PR #12, opened 2026-03-11
Feat/data-connections-command

SailingSF:
Add a slash command to see what data you're connected to and be able to edit it

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -1638,6 +1638,12 @@ async def _handle_data_connections(
                 console.print()
                 continue
 
+            if new_key in all_keys:
+                if not Confirm.ask(f"{new_key} already exists. Overwrite?", default=False, console=console):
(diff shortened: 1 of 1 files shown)


anton/chat.py
```diff
@@ -1456,6 +1456,214 @@ def _minds_test_llm(base_url: str, api_key: str, verify: bool = True) -> bool:
...
+                continue
+
+            scope = Prompt.ask(
+                "Store in",
+                choices=["global", "project"],
+                default="global",
+                console=console,
+            )
+            target_ws = global_ws if scope == "global" else workspace
+            scope_label = "~/.anton/.env" if scope == "global" else f"{workspace.base}/.anton/.env"
```
torrmal: can we make sure the variable does not exist already, or if so if user wants to replace/overwrite?

PR #52, opened 2026-04-01
Update installer 

tino097:
This should be merged just before making the release tag


anton/updater.py
```diff
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import re
```
ZoranPandovski, line 5: Regex is stil used in the script bellow

ZoranPandovski (changes requested):

ZoranPandovski (approved):

PR #93, opened 2026-04-09
Feature/core memory v2

ZoranPandovski:
This PR continues from the #92 , #80, and focuses on the:

Memory: we moved the memory subsystem into anton/core/memory/ so it can be shared with Enterprise (Minds). We have a new base contract that defines the memory protocol. @MinuraPunchihewa, @torrmal, We use Protocol over ABC/inheritance because Enterprise memory replaces mostly all file-based logic. Most of the inline propmpts and constants are moved to settings and core prompts/


anton/core/memory/cortex.py
```diff
@@ -333,6 +340,8 @@ async def compact_all(self) -> None:
             return
 
         for hc in (self.global_hc, self.project_hc):
+            if not isinstance(hc, Hippocampus):
```
ZoranPandovski, line 343: We figure out this on coud with DBHipocampus

PR #102, opened 2026-04-11
warn user scratchpad falls back to synthetic data

Nishu2000-hub:
Added an IMPORTANT instruction to the scratchpad tool description
requiring the LLM to explicitly warn the user with a ⚠️ message
whenever it falls back to synthetic data due to fetch failures.

Impact
Zero risk: prompt-level change only, no logic modified
Protects all users on fresh installs before lessons.md is populated
One line of defence against silent data hallucination


Nishu2000-hub:

ZoranPandovski (changes requested):
Thanks @Nishu2000-hub. I’m just a little confused by the wording around ‘falls back to synthetic data’ since I don’t think that flow exists in our code today. Would you mind clarifying whether this is meant as prompt guidance only? If yes, maybe we can word it that way explicitly.

PR #89, opened 2026-04-08
Release 2.0.0

MinuraPunchihewa:
This release will include the core refactor to support the enterprise (or pro) version of Anton.

PR #107, opened 2026-04-14
Feature/data v2

ZoranPandovski:
This PR splits monolith ds to separate commands and introuduces a DataVault protocol so we can switch between Local and Cloud implementations.
Also in this PR we fix all failing tests for #89


anton/core/datasources/data_vault.py
```diff
@@ -24,8 +25,47 @@ def _slug_env_prefix(engine: str, name: str) -> str:
...
+    def clear_ds_env(self) -> None:
+        """Remove all DS_* variables from os.environ."""
+        ...
+
+    def next_connection_number(self, engine: str) -> int:
+        """Return the next auto-increment number for an engine (1-based)."""
+        ...
+
+
+class LocalDataVault:
```
tino097, line 67: this could implement the new interface

anton/commands/datasource/manage.py
```diff
@@ -0,0 +1,120 @@
...
+from anton.utils.datasources import parse_connection_slug, remove_engine_block, restore_namespaced_env
+from anton.utils.prompt import prompt_or_cancel
+
+if TYPE_CHECKING:
+    from rich.console import Console
+
+
+def handle_list_data_sources(console: "Console", vault: DataVault | None = None) -> None:
+    """Print all saved Local Vault connections in a table with status."""
+    from rich.table import Table
```
tino097, line 18: move this on top, with Console

tino097 (changes requested):

anton/core/datasources/data_vault.py
MinuraPunchihewa: Since it is a `Protocol`, there is no requirement to inherit from it.

MinuraPunchihewa (approved):

tino097 (approved):

PR #125, opened 2026-04-21
Fix: don't skip user messages to extract identity 

ea-rus:
`self._cortex.maybe_update_identity` is called on every 5th message. and only the last message is sent to it
Fix: accumulate previous messages and send all of them on 5th step

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -105,6 +105,7 @@ class ChatSession:
             list(config.initial_history) if config.initial_history else []
         )
         self._pending_memory_confirmations: list = []
+        self._identity_buffer: list[str] = []
         self._turn_count = (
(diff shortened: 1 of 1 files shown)


tino097 (approved):

tino097 (approved):

alecantu7 (approved):

alecantu7:
Nice catch on the missing messages. One concern on the approach: `_identity_buffer` lives only on the live session, so if Anton restarts mid-window the buffered turns are lost and never scanned. Reading the last 5 user messages from `self._history` at flush time would be restart-safe and avoid the extra state.

Minor: this diff references `user_msg_str`, which I don't see defined in `main` — looks like it depends on a sibling change that hasn't landed yet.

PR #127, opened 2026-04-22
Auto complete

ea-rus:
Added auto-completion 

Fixes: https://linear.app/mindsdb/issue/FQE-2220/anton-commands-auto-completion

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -35,7 +35,9 @@ from anton.commands.setup import (
     handle_setup,
     handle_setup_models,
 )
-from anton.commands.ui import handle_explain, handle_theme, print_slash_help, SKILLS_COMMANDS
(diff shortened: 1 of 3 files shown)


anton/chat.py
```diff
@@ -80,8 +80,9 @@
 from prompt_toolkit import PromptSession
 from prompt_toolkit.formatted_text import HTML
 from prompt_toolkit.styles import Style as PTStyle
+from anton.commands.ui import make_completer, THEME_COMMANDS, COMMANDS
```
tino097: can we move on line 38 as there is same from import

tino097 (approved):

anton/chat.py
ea-rus: done

PR #132, opened 2026-04-23
Minds with byok

tino097:
Allow users to use BYOK when querying minds

image


anton/chat.py
```diff
@@ -262,27 +262,61 @@ async def _handle_connect(
...
-            "[anton.success]LLM endpoints available — using Minds server as LLM provider.[/]"
-        )
+    def _has_configured_byok() -> bool:
+        if settings.planning_provider == "anthropic":
+            return bool(
+                settings.anthropic_api_key
+                or os.environ.get("ANTHROPIC_API_KEY")
+                or os.environ.get("ANTON_ANTHROPIC_API_KEY")
+            )
+        if settings.planning_provider == "openai":
```
ZoranPandovski, line 273: Should we do the same check as anthropic here, in casee provider is openai but there is no key set?

anton/cli.py
```diff
@@ -759,7 +759,7 @@ def _setup_other_provider(settings, ws) -> None:
         )
         _setup_anthropic(settings, ws)
 
-    settings.minds_url = None
+    settings.minds_url = "https://mdb.ai"
```
ZoranPandovski, line 762: Bellow we set ANTON_MINDS_URL as none not the above. Maybe update that too

ZoranPandovski (changes requested):
I've just added two minor comments

PR #135, opened 2026-04-24
Dashboard generation improvement

StpMax:
- Templated HTML generation. Dashboards are now produced by filling markers in a template html (GitHub-dark theme, pre-registered ECharts theme and initChart helper) instead of hand-rolling CSS/JS from scratch in every build.

 - Reworked generation prompt. New instruction is short is:
  (1) insights checklist
  (2) a spec YAML plan of the page
  (3) implementation of template markers and DashSpec blocks
  (4) context eraising

 - Main HTML generation prompt is now loaded as a skill when `proactive_dashboards=False`.

 - Context cleanup after build. A new tool `erase_scratchpad_history` wipes the code of every scratchpad cell the agent marked as `# DELETABLE: <desc>` — both in the live pads and in the assistant message history. Once the HTML is on disk, the intermediate generation code no longer occupies context.


StpMax:

torrmal:
i tried this but I think it reduces the flexibility that anton has about what it generates, I got the exeprience that the dashboards it generates using this branch are a bit more limited, lets put this one on hold, and we can discuss a plan on how to improve this part

PR #138, opened 2026-04-27
bump version

tino097:


lucas-koontz (approved):

anton/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "2.0.2"
+__version__ = "2.0.4"
```
martyna-mindsdb, line 1: shouldn't this be 2.0.3?

lucas-koontz: No. The current version was 2.0.3 and we are moving it up

PR #147, opened 2026-05-05
Drag-and-drop image to chat

StpMax:
This PR includes two main changes:

#### Linux clipboard support for image insertion
Added support for the system clipboard on Linux via wl-paste (Wayland) and xclip (X11). Functionality depends on the terminal being used and available tools — if they are missing, a relevant message is displayed to the user.

#### Replacing paths with image placeholders
When dragging or pasting images (on all platforms), the path is automatically replaced with a placeholder of the form [Image #X], similar to Claude Code. Placeholders are highlighted for visibility.

https://www.loom.com/share/ca69b4f8b4644f6fbd028044ab88bc02

#### Added max image size limit 5Mb
Since the maximum image size in claude is set to 5MB, and this is min size across llms, we also set it o 5Mb.
Error before:
image
Error after:
image


ZoranPandovski:
@StpMax what should we do for other platforms is this Linux only?

anton/utils/clipboard.py
```diff
@@ -152,3 +164,206 @@ async def ensure_clipboard(console: Console) -> bool:
...
+#   2. ``ImageRefLexer`` — colorizes [Image #N] in the input line.
+#   3. ``build_image_ref_message`` — at submit time, expands refs into a list
+#      of multimodal content blocks for the LLM.
+
+
+def _media_type_for(fmt: str) -> str:
+    f = (fmt or "PNG").lower()
+    if f in ("jpg", "jpeg"):
+        return "image/jpeg"
+    if f in ("png", "gif", "webp", "bmp"):
```
tino097, line 210: `bmp` format is not supported by OpenAI https://community.openai.com/t/how-to-use-bmp-images-with-api/653232

Also as stated, maybe we should add check for the size

tino097 (changes requested):

anton/utils/clipboard.py
StpMax: Added BMP to PNG conversion.
Regarding the size check — yes, it would be nice to add it. When I'm back, I'll clarify what we're going to do when the limit is exceeded — return an error or silently convert the file. This can be added in a separate PR.

StpMax:
@ZoranPandovski The ability to use the clipboard for Windows and Mac existed before this PR. This PR adds clipboard support for Linux.

StpMax:
@tino097 added max img size limit (screenshots in the description)

tino097 (approved):

PR #153, opened 2026-05-11
fix: enable vision support for OpenAI-compatible client

hazemahmedx0:
## Summary
- Sets `supports_vision=True` in the OpenAI-compatible client config (`client.py:235`)
- Enables vision-capable models routed through this client to process image inputs

## Test plan
- [ ] Verify image inputs work end-to-end with an OpenAI-compatible vision model
- [ ] Confirm non-vision model paths are unaffected


hazemahmedx0:

StpMax (approved):

PR #159, opened 2026-05-13
Fix datasource list status for known engines

robertmonka:
When an engine definition is available, validate required fields before treating an empty credentials map as incomplete. This lets known engines without required fields appear as saved instead of incomplete.

### DIFF
--- a/anton/commands/datasource/manage.py
+++ b/anton/commands/datasource/manage.py
@@ -38,12 +38,12 @@ def handle_list_data_sources(console: "Console", vault: DataVault | None = None)
         source = engine_def.display_name if engine_def else c["engine"]
         fields = vault.load(c["engine"], c["name"]) or {}
 
-        if not fields:
-            status = "[yellow]incomplete[/]"
-        elif engine_def and engine_def.auth_method != "choice":
+        if engine_def and engine_def.auth_method != "choice":
             required = [f.name for f in engine_def.fields if f.required]
             missing = [name for name in required if name not in fields]
             status = "[yellow]incomplete[/]" if missing else "[green]saved[/]"
+        elif not fields:
+            status = "[yellow]incomplete[/]"
         else:
             status = "[green]saved[/]"
 


robertmonka:
Additional context for reviewers:

This is a small display/status fix in `handle_list_data_sources`. The previous ordering treated an empty credentials map as `incomplete` before considering the datasource engine definition. That made the generic empty-fields fallback win even when the engine definition was available and could provide a more precise answer.

What changed:
- For known non-`choice` engines, required fields are evaluated first.
- Missing required fields still show `incomplete`.
- Engines with no missing required fields can show `saved`, even when the stored fields map is empty.
- Unknown engines and `choice` auth engines keep the existing empty-fields fallback behavior.

This is intentionally limited to the status calculation in `anton/commands/datasource/manage.py`; it does not change vault storage, connection setup, or credential loading.

robertmonka:

robertmonka:
recheck

tests/test_datasource.py
```diff
@@ -514,6 +514,23 @@ def test_reload_picks_up_new_engine(self, tmp_path):
         assert reg.get("mysql") is not None
         assert reg.get("mysql").display_name == "MySQL"
 
+    def test_builtin_optima_api_template(self):
```
martyna-mindsdb: The existing tests are testing how the registry parses templates, not whether a specific template has the right values. But the `test_builtin_optima_api_template` test is just checking that the YAML was typed correctly, which isn't something we test for other datasources. Rather, if every new datasource added a test like this, the file would grow a lot without real value, because the file already gets validated by `test_validate_file_returns_engines` and `test_all_engines_sorted`.

Could you remove this test?

robertmonka: Done, removed `test_builtin_optima_api_template`. The generic datasource registry/parser tests remain in place.

martyna-mindsdb: Thanks. I see that you removed the datasource template from the `datasources.md` file as well, although I suggested removing just from the tests.

robertmonka: You are right, I removed it from `datasources.md` as well to keep this PR focused only on the status handling fix.

For context, that datasource is for connecting to an API I wrote for Comarch Optima, which is a popular ERP system in Poland. If you think this datasource belongs in the built-in sources, I can add it back in a separate, dedicated PR with the relevant context.

martyna-mindsdb: That sounds good. Sure, please feel free to create a new PR adding the template for this datasource.

martyna-mindsdb (approved):

PR #160, opened 2026-05-13
Use runtime provider config for CLI scratchpads

robertmonka:
Build datasource CLI scratchpads through the same provider export path used by chat sessions. This preserves runtime factory, base URL, workspace path, and provider connection details for connect, edit, and test commands.

### DIFF
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -24,6 +24,7 @@ from anton.utils.prompt import prompt_or_cancel
 from anton.core.llm.openai import build_chat_completion_kwargs, _is_azure_endpoint
 
 from anton.chat import ChatSession
+from anton.chat_session import get_runtime_factory
 from anton.core.session import ChatSessionConfig
 from anton.core.llm.client import LLMClient
 from anton.core.backends.manager import ScratchpadManager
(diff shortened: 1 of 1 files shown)


robertmonka:
Additional context for reviewers:

This fixes the datasource CLI commands so their scratchpad setup follows the same provider/runtime path used by regular chat sessions. Before this change, `connect`, `edit`, and `test` constructed `ScratchpadManager` inline and only passed provider/model/API key values. That path missed runtime details that are already handled elsewhere in the app.

What changed:
- Added `_build_scratchpad_manager()` as a shared helper for datasource CLI commands.
- Reuses `LLMClient.from_settings()` when needed.
- Gets provider connection details through `coding_provider.export_connection_info()`.
- Passes `runtime_factory`, `coding_base_url`, and `workspace_path` into `ScratchpadManager`.
- Removes the duplicated `anthropic` vs `openai` API-key branching from the CLI command handlers.

The intent is not to create a new configuration path, but to align these CLI-only datasource flows with the existing `ChatSession` scratchpad construction model.

robertmonka:

MinuraPunchihewa (approved):

MinuraPunchihewa:
This looks good to me! Thanks, @robertmonka.

PR #161, opened 2026-05-14
self healing orchestrator

torrmal:
The idea here is to enable better learning from issues capabilities in anton by implementing how the brain does this:
Anterior Cingulate Cortex (ACC) — pattern-level error detection.

Brain analogue
==============

The anterior cingulate cortex is the brain's error-detector. It fires
the *error-related negativity* (ERN) ~80 ms after the brain notices
that an actual outcome diverged from an expected one. The signal it
emits flows downstream to the dopaminergic midbrain (computing the
*reward prediction error*) and onward to the striatum (updating
action policies) and the dorsolateral prefrontal cortex (adjusting
strategy on the next attempt).

### DIFF
--- a/anton/README.md
+++ b/anton/README.md
@@ -601,7 +601,21 @@ Anton's ACC watches a turn unfold, classifies events as they arrive, and at end-
 
 ### Status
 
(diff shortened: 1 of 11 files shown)


ZoranPandovski (approved):

ZoranPandovski:
@torrmal Amazing. Something we can discuss for later is to finda a way to move conflict detection pre-dispatch. Like can we inspect the proposed tool_calls list before running them ?  This way we can figure out how to  catch name-splits, oscillation etc at decision time instead of post run.

PR #175, opened 2026-05-27
Feat/better-tracing

SailingSF:
Add per turn trace id to allow for better eval runs


torrmal (approved):
looks good to me @sejubar can you test that this works?

PR #184, opened 2026-06-17
enable reasoning effort as a param

SailingSF:
Enable reasoning effort for openai, anthropic, and through minds cloud inference

### DIFF
--- a/anton/config/settings.py
+++ b/anton/config/settings.py
@@ -34,6 +34,12 @@ class AntonSettings(CoreSettings):
     # "max"), forwarded to the provider in its native shape when set. None means
     # the provider's own default. The value is validated by the upstream
     # router/provider, not here.
+    #
+    # NOTE: effort only applies to reasoning-capable models. Anthropic returns a
+    # 400 for `effort` on Haiku 4.5 / Sonnet 4.5 — including the default
+    # `coding_model` above — so it needs an Opus-4.5+/Sonnet-4.6-class model; the
+    # OpenAI `reasoning_effort` likewise needs a reasoning model (o-series / GPT-5
+    # class). Set these only when the corresponding model supports it.
     planning_reasoning_effort: str | None = None
     coding_reasoning_effort: str | None = None
 
--- a/anton/config/settings.py
+++ b/anton/config/settings.py
@@ -30,6 +30,13 @@ class AntonSettings(CoreSettings):
     coding_provider: str = "anthropic"
     coding_model: str = "claude-haiku-4-5-20251001"
 
+    # Opaque reasoning-effort level (e.g. "low" | "medium" | "high" | "xhigh" |
+    # "max"), forwarded to the provider in its native shape when set. None means
+    # the provider's own default. The value is validated by the upstream
+    # router/provider, not here.
+    planning_reasoning_effort: str | None = None
+    coding_reasoning_effort: str | None = None
+
     max_tokens: int = 8192  # max output tokens per LLM call
 
     anthropic_api_key: str | None = None
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -66,7 +66,11 @@ class AnthropicProvider(LLMProvider):
     ) -> None:
         self._api_key = api_key
         # Opaque effort level forwarded as ``output_config={"effort": ...}`` on
-        # every call when set (None = the model's default).
+        # every call when set (None = the model's default). Sent via ``extra_body``
+        # rather than a typed kwarg so it doesn't depend on the resolved
+        # ``anthropic`` SDK version exposing ``output_config`` as a parameter
+        # (the Stainless client rejects unknown kwargs, which would otherwise be a
+        # hard TypeError on older SDKs the moment an effort is set).
         self._reasoning_effort = reasoning_effort
         kwargs = {}
         if api_key:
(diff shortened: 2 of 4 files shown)


anton/core/llm/anthropic.py
```diff
@@ -95,6 +100,8 @@ async def complete(
             kwargs["tools"] = merged_tools
         if tool_choice:
             kwargs["tool_choice"] = tool_choice
+        if self._reasoning_effort:
+            kwargs["output_config"] = {"effort": self._reasoning_effort}
```
alecantu7: Shape is correct (`output_config.effort` is the right Anthropic surface), but flagging a dependency-floor risk: `pyproject.toml` declares `anthropic>=0.42.0`, which is older than the SDK version that added `output_config`. The Stainless-generated client does **not** accept arbitrary kwargs, so on an environment that resolves to an old `anthropic`, this line raises `TypeError: unexpected keyword argument 'output_config'` the moment a reasoning effort is set (it's latent today since the default is `None`).

Could you confirm the actually-locked `anthropic` version supports `output_config`, and bump the floor in `pyproject.toml` to match? If you'd rather be resilient to version drift, `extra_body={"output_config": {"effort": self._reasoning_effort}}` achieves the same without depending on the typed param. Same applies to the `stream` path below.

anton/config/settings.py
```diff
@@ -30,6 +30,13 @@ class AntonSettings(CoreSettings):
     coding_provider: str = "anthropic"
     coding_model: str = "claude-haiku-4-5-20251001"
 
+    # Opaque reasoning-effort level (e.g. "low" | "medium" | "high" | "xhigh" |
+    # "max"), forwarded to the provider in its native shape when set. None means
+    # the provider's own default. The value is validated by the upstream
+    # router/provider, not here.
+    planning_reasoning_effort: str | None = None
+    coding_reasoning_effort: str | None = None
```
alecantu7, line 44: Foot-gun worth a one-line note here: `coding_model` defaults to `claude-haiku-4-5-20251001`, and Anthropic returns a 400 for `effort` on Haiku 4.5 (and Sonnet 4.5). So setting `coding_reasoning_effort` while on the default coding model fails at call time rather than at config load. The "validated upstream" design is fine, but a comment noting that effort requires an Opus 4.5+/Sonnet 4.6 class model (and that OpenAI `reasoning_effort` requires a reasoning model) would save the next person a confusing 400.

alecantu7 (commented):
Reviewed the full diff plus both provider implementations and the factory wiring. The change is correct and low-risk: the three provider shapes are all right (`output_config.effort` for Anthropic, top-level `reasoning_effort` for OpenAI chat.completions, `reasoning.effort` for the OpenAI Responses path — added once in `_build_responses_kwargs` so it covers both complete and stream), the `FLAVOR_OPENAI` early-return means the effort is never double-applied, and threading `planning_reasoning_effort`/`coding_reasoning_effort` separately through the factory lambdas is the right call (those lambdas have no other callers, so the signature change is safe). It's fully opt-in (defaults `None`), so existing behavior is unchanged.

Two things worth verifying before merge (left as inline comments), plus a non-blocking note:

- **`anthropic` dependency floor.** `pyproject.toml` pins `anthropic>=0.42.0`, which predates `output_config`. The Stainless SDK rejects unknown kwargs, so on an old resolved version `messages.create(output_config=...)` is a hard `TypeError` whenever effort is set. Confirm the locked version supports it and bump the floor (or route via `extra_body`).
- **Effort vs. the default coding model.** `coding_model` defaults to `claude-haiku-4-5`, and Anthropic 400s on `effort` for Haiku 4.5 / Sonnet 4.5. Setting `coding_reasoning_effort` on the default model fails at call time — worth a docstring note.
- **Non-blocking:** no tests added; `tests/test_client.py` and `tests/test_openai_provider.py` are easy to extend with "forwarded in the right shape per provider / omitted when None" asserts.

Nice, well-commented change overall.

anton/core/llm/anthropic.py
alecantu7: Resolved in fc68052de — switched to `extra_body={"output_config": {"effort": ...}}` on both the complete and stream paths, so it no longer depends on the resolved `anthropic` SDK exposing `output_config` as a typed kwarg. Verified it's the only `extra_body` usage in the file, so no collision. 👍

anton/config/settings.py
alecantu7: Resolved in fc68052de — the docstring note now spells out that effort needs an Opus-4.5+/Sonnet-4.6-class model (Haiku/Sonnet-4.5 400, including the default coding_model) and that OpenAI needs a reasoning model. Exactly what I was after.

alecantu7 (approved):
Approving — both review items are addressed in fc68052de:

- **Dependency-floor risk** → effort now forwarded via `extra_body` on both the complete and stream paths, so it's independent of the resolved `anthropic` SDK version. Confirmed it's the only `extra_body` usage in `anthropic.py` (no collision).
- **Model-requirements foot-gun** → documented in `settings.py` (Opus-4.5+/Sonnet-4.6 for Anthropic, reasoning model for OpenAI).

CI is green. Shape is correct across all three provider paths, fully opt-in (defaults `None`), so existing behavior is unchanged. The only outstanding item is the non-blocking "no tests" note — fine to land without it given the opt-in default. Nice change.

PR #186, opened 2026-06-17
Convert skill storage to Agentskills standard

ea-rus:
Updated Anton's skills storage to match [agentskills specification](https://agentskills.io/specification)

Updates: 

- Format:
  - skill name can't have underscores, they replaced with hyphen
  - meta.json + declarative.md  are replaced with  SKILL.md file
- skill_format.py: a new module that responsible to parse and dump skills. It is used by cowork-server too (to parse and save skills)
- `when_to_use` field was removed (moved to description)
-  added `skills_root` setting: to specify a folder to store skills (default ~/.anton/skills)

- Migration details:
  - if skill folder doesn't have 'SKILL.md' file - make migration
  - meta.json + declarative.md are moved to SKILL.md 
  - stats.json - is untouched 

Not covered:
- scripts/references/assets folders in skill folder. Anton and cowork don't create them.


anton/core/memory/skills.py
```diff
@@ -2,67 +2,52 @@
...
 import shutil
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
 
+from anton.core.tools.skill_format import AgentSkill, dump_skill, parse_skill_dir, normalize_name, DESC_MAX
+
+logger = logging.getLogger(__name__)
 
 _DEFAULT_SKILLS_ROOT = Path("~/.anton/skills").expanduser()
```
MinuraPunchihewa, line 52: nitpik: Should we move this over to where the new `skills_root` default has been defined?

MinuraPunchihewa (approved):

anton/core/memory/skills.py
ea-rus: yes, it think it would be better to move the default value there

PR #187, opened 2026-06-17
Fix web_search crash on None input_tokens from Responses API (ENG-349)

alecantu7:
Fixes ENG-349 — https://linear.app/mindsdb/issue/ENG-349

## What

`web_search()` (the scratchpad helper) crashes on **every call** on the hub agents with:

```
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
  File "anton/core/llm/provider.py", line 291, in compute_context_pressure
    return min(input_tokens / window, 1.0)
```

## Why

The scratchpad `web_search()` helper routes through `/v1/responses`. The MindsHub passthrough returns `usage.input_tokens = None` on web-search (tool-augmented) responses. `_parse_response_object` (`openai.py:1136`) read it with `getattr(usage, "input_tokens", 0)` — but a `getattr` default only fires when the attribute is **absent**, not when it's present-but-`None`. So `None` flowed into `compute_context_pressure`, which did `None / window` → `TypeError`.

The **streaming** Responses path already guards this exact case (`openai.py:1064`, `getattr(...) or 0`); the **non-streaming** `_parse_response_object` never got the same treatment. This is a one-line oversight.

Discovered via an Anton self-generated post-mortem (Nate Herk / MindsHub dashboard session) — `web_search` failed on every call, forcing the agent onto slower direct-HTTP fetching.

## Impact

`web_search` is effectively **100% non-functional on the hub agents** today (Responses path). Research tasks silently degrade to "fetch URLs I can already guess" — no discovery.

## Not a billing issue

`compute_context_pressure` is a **client-side** context-fullness gauge (`input_tokens / context_window`) used only to decide when to compact history on the instance. It is **not** part of any billing path — billing is metered server-side in mindshub-inference from the provider's real reported usage (`usage_box.value` ← Anthropic's actual `usage.input_tokens`), independent of what the client receives. So this fix has **no billing implications**.

## Fix (defense in depth)

- `compute_context_pressure` — treat `None`/`0` input tokens as `0.0` pressure instead of crashing. Guarding the shared helper covers **every** call path (streaming, non-streaming, Anthropic, OpenAI).
- `_parse_response_object` — coerce `None → 0` to match the already-guarded streaming path.
- Tests — `TestComputeContextPressure` covers the `None` regression case, zero, a normal ratio, and the clamp-at-1.0 boundary.

## Testing

`tests/test_provider.py` (15 passed locally, incl. the 4 new cases). `test_openai_provider.py` couldn't run in my local venv due to an unrelated missing-dep (`pydantic_settings`) — CI covers it.

### DIFF
--- a/tests/test_provider.py
+++ b/tests/test_provider.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
 from anton.core.llm.anthropic import AnthropicProvider
+from anton.core.llm.openai import _parse_response_object
 from anton.core.llm.provider import LLMResponse, ToolCall, compute_context_pressure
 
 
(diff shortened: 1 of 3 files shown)


tino097 (approved):
LGTM

tests/test_provider.py
```diff
@@ -3,7 +3,25 @@
 from unittest.mock import AsyncMock, MagicMock, patch
 
 from anton.core.llm.anthropic import AnthropicProvider
-from anton.core.llm.provider import LLMResponse, ToolCall
+from anton.core.llm.provider import LLMResponse, ToolCall, compute_context_pressure
+
+
+class TestComputeContextPressure:
```
tino097, line 11: lest try to add a test with mocking to set the values for input or output token to be `None`

alecantu7: Done — added `test_parse_response_object_coerces_none_usage_tokens`: mocks the Responses object with `usage.input_tokens=output_tokens=None` and asserts the parser coerces both to 0, sets context_pressure to 0.0, and doesn't raise. Also added a sanity case with real counts (preserved + correct 0.5 pressure). 17/17 green locally.

PR #189, opened 2026-06-18
[ENG-344] Error message for tokens exceeded

tino097:
Error message for tokens exceeded

https://linear.app/mindsdb/issue/ENG-344/no-clear-error-when-a-user-hits-the-5m-token-limit-anton-appears-dead


alecantu7 (approved):
Approving — this is exactly the Path-A half of the ENG-344 fix. `_safe_error_message` now surfaces `TokenLimitExceeded` verbatim (no `[agent error]` prefix), so on the dispatch path a spent allowance reads as a quota message rather than a crash; other errors keep the prefix + redaction. The mdb.ai → console.mindshub.ai update is applied consistently across all five 429 sites, version bumped to 2.26.6.18.1, and the new test covers both the no-prefix (quota) and prefix-kept (generic) branches.

Two non-blocking notes: (1) pre-existing wording drift not introduced here — two of the five strings read "upgrade or top up" vs the others' "upgrade or to top up"; (2) CI only shows CLAssistant — the new test doesn't appear to be executed by CI, so worth confirming it runs. Nice, focused change.

PR #190, opened 2026-06-18
ENG-350: scratchpad guardrails — stop the large-output retry loop

alecantu7:
Fixes the wasteful retry loop behind [ENG-350](https://linear.app/mindsdb/issue/ENG-350) — Anton spending 10–15 min and ~360K wasted tokens building large outputs (dashboards) before succeeding. **Enforcement-first:** the prompt already tells the model the right things (one scratchpad, chunk, break down); the model ignores them, so these are runtime guardrails that enforce what's already asked.

This is the **anton** half. Two follow-up PRs land in `cowork-server` (turn on `ANTON_ACC_MODE=active` + emit a killed/timeout status event) and `cowork` (render killed vs running vs done).

### Changes (one commit each)
1. **Cap the inactivity window** (`utils.py`, `settings.py`) — `inactivity = min(max(est×0.5,30), cell_inactivity_max=60)`. A *silent* cell dies in ≤60s regardless of its estimate (the real "running forever" fix). The **total is left scaling** so legit long-but-active cells (a batch loop pinging `progress()`, e.g. sending thousands of emails) run to completion — `progress()` resets inactivity but not total, so a hard total cap would have killed them.
2. **Self-correcting empty-code failure + name-agnostic kill-loop** (`scratchpad.py`, `acc.py`) — the empty-`code` drop (oversized arg truncated in transit) now returns actionable recovery guidance phrased as a failure, so `_apply_error_tracking` counts it toward the circuit breaker instead of silently resetting. `detect_kill_loop` now fires on ≥2 kills across the turn regardless of name (renaming no longer hides the loop).
3. **Failure-type-aware nudge** (`prompts.py`, `session.py`) — scratchpad failures get size/timeout-specific guidance instead of the generic scrape advice ("try archive.org / a different data source") that pushed the model toward rename-and-retry churn.
4. **Single-scratchpad guard** (`tool_handlers.py`, `tool_defs.py`) — challenges an exec on a *new* scratchpad name when the agent already has one in use this task; `confirm_new_scratchpad=true` bypasses for genuine isolation. System-created pads (artifact backend launcher) never count against the agent.

### Not included (deliberate)
- No auto-heartbeat that resets the inactivity timer (would mask genuinely hung cells).
- No hard total-timeout cap (would kill legit long batch loops).
- `est>90` / oversized code is a soft nudge, not a hard reject (atomic long loops are legitimate).

### Tests
`189 passed` across the scratchpad/ACC/session suites. New: `test_resilience_nudge.py`, single-scratchpad-guard tests, inactivity-cap tests; updated the kill-loop and compute-timeouts tests to the new behavior.

### DIFF
--- a/anton/core/backends/utils.py
+++ b/anton/core/backends/utils.py
@@ -8,16 +8,23 @@ def compute_timeouts(estimated_seconds: int) -> tuple[float, float]:
     """
     s = CoreSettings()
     if estimated_seconds <= 0:
-        return float(s.cell_timeout_default), float(s.cell_inactivity_timeout)
-    total = max(estimated_seconds * 2, estimated_seconds + 30)
-    inactivity = max(estimated_seconds * 0.5, 30)
+        total = float(s.cell_timeout_default)
+        inactivity = float(s.cell_inactivity_timeout)
+    else:
+        total = float(max(estimated_seconds * 2, estimated_seconds + 30))
+        inactivity = float(max(estimated_seconds * 0.5, 30))
     # Clamp the silence window: a large estimate must not buy minutes of
     # undetected silence (an est=600 cell would otherwise allow 300s of no
     # output before being killed). A cell quiet for cell_inactivity_max
     # seconds is killed regardless of its estimate. stdout/progress() reset
     # this window, so legitimate long-but-active cells — e.g. a batch loop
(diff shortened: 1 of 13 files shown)


torrmal (approved):

torrmal:
this looks good to me

PR #191, opened 2026-06-18
fix/ Reconcile artifact files[] against disk on read (ENG-372)

alecantu7:
Fixes ENG-372 — https://linear.app/mindsdb/issue/ENG-372

## What

When Anton creates an artifact and then writes its files via scratchpad `open()`, the artifact's `metadata.json` `files[]` was **never updated** — it stayed frozen at the empty state `create_artifact` wrote. So `open_artifact()` / `list_artifacts()` reported `file_count: 0` for artifacts that were fully written on disk. The agent read that as "the file is missing" and burned turns in a search/recovery loop (32 cells in the reported session) even though the file was at the correct path the whole time.

Surfaced by an Anton self-generated post-mortem (artifact `mooresville-home-finder-enhanced`: `dashboard.html` 278 KB on disk, registry `files: []`).

## Root cause

- Scratchpad `open(path, 'w')` writes straight into the artifact folder, **bypassing the store** — there's no write→store notification hook.
- The store already had a reconciliation primitive, `ArtifactStore.rescan_files()`, **but it had zero callers** (dead code — only a docstring reference).
- `store.open()` / `store.list()` read `files[]` straight from the frozen metadata with no disk reconciliation.

(Related dead code, left as a follow-up: `record_turn()` — the provenance hook — also has zero callers, so `provenance[]` is always `[]`. Out of scope here.)

## Fix

Reconcile `files[]` against disk on read, using the primitive that already existed:
- Wire a new `_reconcile_files()` into `open()` and `list()` so both agent tools self-heal (and existing stale artifacts heal on next access).
- **Save-on-change only** — compares the on-disk file set to `files[]` and only rewrites `metadata.json` / bumps `updatedAt` when it actually changed, so calling it on every read is idempotent and cheap (no churn, no spurious `updatedAt` bumps).
- Add `.published.json` to the housekeeping skip set (was `metadata.json` / `README.md` only) to match cowork-server's artifacts service — otherwise publish-state would be miscounted as an artifact file.

## Scope

**Agent-facing only.** cowork-server's artifact service (`_user_files`) already scans the folder from disk directly, so the **user-facing UI/preview was never affected** — this fixes the agent's view (the recovery loops and wrong "file missing" messages).

## Testing

`tests/test_artifacts.py` — 40 passing, incl. 4 new: `open()`/`list()` reflect a scratchpad-written file, `.published.json` is excluded, and reconcile-on-read is idempotent (no re-save when nothing changed).

### DIFF
--- a/anton/core/artifacts/models.py
+++ b/anton/core/artifacts/models.py
@@ -56,9 +56,10 @@ ARTIFACT_TYPES: tuple[str, ...] = (
 class FileEntry(BaseModel):
     """One file inside the artifact folder.
 
-    Re-derived from disk on every save (`ArtifactStore.rescan_files`)
-    rather than mutated in place. The agent never writes here
-    directly — it just creates the files; the server records them.
+    Re-derived from disk on read (`ArtifactStore._reconcile_files`, called
+    by `open()` / `list()`) rather than mutated in place. The agent never
+    populates this directly — it writes the files into the folder via the
+    scratchpad, and the store reconciles `files[]` against disk on access.
     """
 
     path: str  # relative to the artifact folder (e.g. "dashboard.html", "data/prices.csv")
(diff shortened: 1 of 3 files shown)


SailingSF (approved):
LGTM, hopefully this will help token burn quite a lot

PR #194, opened 2026-06-18
prompt: cache-stable assembly (2a) — task-anchored date + memory at tail

torrmal:
## Summary (Phase 2a)
Make Anton's system-prompt **prefix byte-stable across a task's turns** so providers can prefix-cache it and behavior is deterministic.

## Changes
- **Task-anchored, date-only stamp.** `ChatSessionConfig.clock` (e.g. the conversation's `created_at`) replaces the per-turn `datetime.now()` minute clock in `_build_system_prompt`. Same task → identical date every turn → the prefix stops changing each minute. Falls back to today's date for the CLI. The agent fetches precise time via a tool when it needs it.
- **Memory snapshot moved to the volatile tail.** The cortex memory block is relevance-filtered per user message (changes every turn), so it now sits at the very end of the system prompt — everything above it (identity, discipline, tools, project/datasource/skills, suffix) stays cacheable.

## Notes
- This is the low-risk half of Phase 2: byte-stable content + determinism. It does **not** add `cache_control` markers — Anthropic still needs those to actually cache (Phase 2b); OpenAI-compatible providers auto-cache identical prefixes and benefit immediately.
- Behavioral change to validate: memory is now at the end of the system prompt (recency position) rather than mid-prompt.
- **Stacked on** `feat/act-and-surface-assumptions` (#193) — merge that first; this PR's base will retarget to `staging` after.
- Pairs with cowork-server PR (harness passes `clock=conversation.created_at`).

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -60,6 +60,11 @@ from anton.utils.datasources import (
 from anton.core.settings import CoreSettings
 
 
+# Sentinel prefixing a compacted-history summary so later compactions can
+# recognize and update it in place rather than summarize a summary.
+_COMPACTED_MARKER = "[COMPACTED CONTEXT — REFERENCE ONLY]"
+
+
 if TYPE_CHECKING:
     from rich.console import Console
     from anton.context.self_awareness import SelfAwarenessContext
@@ -782,12 +787,18 @@ class ChatSession:
         old_turns = self._history[:split]
         recent_turns = self._history[split:]
 
-        # Serialize old turns into text for summarization
+        # Serialize old turns. Pull out any prior compacted summary so we
+        # UPDATE it in place rather than summarize a summary (which compounds
+        # loss every compaction).
+        prior_summary = ""
         lines: list[str] = []
         for msg in old_turns:
             role = msg.get("role", "unknown")
             content = msg.get("content", "")
             if isinstance(content, str):
+                if content.lstrip().startswith(_COMPACTED_MARKER):
+                    prior_summary = content
+                    continue
                 lines.append(f"[{role}]: {content[:2000]}")
             elif isinstance(content, list):
                 for block in content:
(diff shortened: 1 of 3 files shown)


anton/core/session.py
```diff
@@ -541,8 +549,13 @@ def _record_cell_explainability(
...
         import datetime as _dt
 
-        _now = _dt.datetime.now()
-        _current_datetime = _now.strftime("%A, %B %d, %Y at %I:%M %p")
+        # Task-anchored, date-only stamp. Using the task's anchor (self._clock,
+        # e.g. the conversation's created_at) keeps this byte-identical across
+        # every turn so the system-prompt prefix stays cache-stable; a per-turn
+        # wall clock with minute precision would invalidate the cache each turn.
+        # The agent fetches precise time via a tool when it actually needs it.
+        _now = self._clock or _dt.datetime.now()
```
MinuraPunchihewa: @torrmal I understand the need for prefix caching, but with Cowork, we will be passing the date/time when the task (conversation) created here, but according to the system prompt template that we have, this will be [indicated to the agent as the current date and time](https://github.com/mindsdb/anton/blob/d9a154264f8bb6610f141bc40e8d6a338925f95d/anton/core/llm/prompts.py#L10). Do we need to change this template somehow? Otherwise, say that a user came back to a conversation days or even weeks later, that will still be treated as the 'current' date and time, yes?

torrmal: good point, yes, you are right let me fix that

torrmal:
Good catch — addressed. The date is no longer anchored-as-"current".

**What changed:**
- The cache-stable prefix now carries only a **fixed** `Conversation started: {date}` line (from `created_at`) — it never changes across turns, so it doesn't bust the prefix cache.
- The **real wall clock** is emitted in the **volatile tail** (`Current date and time: …`), recomputed every turn — so a conversation resumed days/weeks later reports the *actual* now, not the creation date.
- `ChatSessionConfig.clock` → `started_at` to match the new semantics.

**Plus, per your "what time was something said" point:** the cowork-server harness now embeds each message's `created_at` into history as `[YYYY-MM-DD HH:MM]` (absolute → fixed per message → still cache-safe), and the compaction summarizer is told to preserve key event dates so the timeline survives compaction. (cowork-server PR #71.)

So: prefix = *started*, history = *per-message stamps*, tail = *live now* — full temporal grounding, cache intact.

PR #197, opened 2026-06-21
feat(select_path): inline file/folder picker for ambiguous paths

hazemahmedx0:
https://linear.app/mindsdb/issue/ENG-325/cowork-file-access

When the agent can't tell which file or folder the user meant, it had two bad options: guess and risk acting on the wrong path, or stop and ask in plain text and break its own flow. This adds a third. The agent calls a `select_path` tool, the user picks, and the same turn keeps going. No extra user message, no restart. It only fires when the path is genuinely ambiguous, so normal requests are untouched.

## Before
image

## After 
Screenshot 2026-06-22 at 2 16 27 AM

https://www.loom.com/share/5495d1aff1a7488db532fbdbd0f55926

What's in here:
- `select_path` tool definition and handler with two modes. Pick mode disambiguates between concrete candidates that already exist in the project. Browse mode is for when there's nothing concrete to offer, so the host opens a picker rooted at a sensible directory.
- A small `SelectionElicitor` protocol (`core/interaction/selection.py`) so the core stays transport agnostic. It only knows "ask for a selection, get a path back" and does not care whether that's a terminal prompt or a GUI.
- A CLI elicitor for the terminal, plus a console fallback and a graceful no-op when running headless.
- Hosts inject their elicitor through `ChatSessionConfig`, and browse picks that come back relative are resolved against the request root.

This is the bottom of a three-repo stack. The picker is useless without a host to drive it, so the other two build on this:
- cowork-server streams the request to the client and feeds the pick back into the paused turn.
- cowork renders the card and opens the native OS picker.

Tested both modes locally through the full cowork stack.

---

Part of the inline path picker stack. Merge bottom up:
1. anton: https://github.com/mindsdb/anton/pull/197 (this PR, the select_path tool and elicitor protocol)
2. cowork-server: https://github.com/mindsdb/cowork-server/pull/77 (streams the request, resolves the pick)
3. cowork: https://github.com/mindsdb/cowork/pull/212 (the native OS picker UI)

Related but independent: mindshub_inference #302 (https://github.com/mindsdb/mindshub_inference/pull/302) is the image side of the same multimodal and file effort. It translates image_url blocks for every provider and transcodes unsupported image types to PNG. It does not depend on this stack and this stack does not depend on it.

### DIFF
--- a/anton/core/interaction/selection.py
+++ b/anton/core/interaction/selection.py
@@ -41,11 +41,22 @@ class SelectionOption:
 
 @dataclass(frozen=True, slots=True)
 class SelectionRequest:
-    """A request for the user to disambiguate between candidate paths."""
+    """A request for the user to choose a file or folder.
+
+    Two modes:
+
+    * ``"pick"`` — disambiguate between the concrete ``options`` (e.g. several
+      files share a name). This is the default.
+    * ``"browse"`` — the target is unknown/unspecified; the host shows a
+      navigable browser rooted at ``root`` and the user locates the path
+      themselves. ``options`` is empty in this mode.
+    """
 
     prompt: str
-    options: tuple[SelectionOption, ...]
+    options: tuple[SelectionOption, ...] = ()
     kind: str = "any"  # "file" | "folder" | "any" — what is being chosen
+    mode: str = "pick"  # "pick" | "browse"
+    root: str = ""  # browse-mode start directory (absolute)
 
 
 @runtime_checkable
--- a/anton/core/interaction/selection.py
+++ b/anton/core/interaction/selection.py
@@ -19,7 +19,7 @@ transport — so the same disambiguation logic serves every host.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Protocol, runtime_checkable
+from typing import Protocol
 
 __all__ = ["SelectionOption", "SelectionRequest", "SelectionElicitor"]
 
@@ -59,9 +59,13 @@ class SelectionRequest:
     root: str = ""  # browse-mode start directory (absolute)
 
 
-@runtime_checkable
 class SelectionElicitor(Protocol):
-    """Strategy for surfacing a :class:`SelectionRequest` and awaiting a choice."""
+    """Strategy for surfacing a :class:`SelectionRequest` and awaiting a choice.
+
+    A Protocol (not an ABC) on purpose: hosts satisfy it by shape, so the
+    cowork-server streaming picker implements ``elicit`` without importing or
+    inheriting this class. The tool depends only on the shape.
+    """
 
     async def elicit(self, request: SelectionRequest) -> str | None:
         """Present *request*, block until the user responds.
--- a/anton/core/interaction/selection.py
+++ b/anton/core/interaction/selection.py
@@ -1,0 +1,61 @@
+"""File/folder disambiguation: value types and the elicitor strategy.
+
+When the agent cannot tell which file or folder the user means, the
+``select_path`` tool offers the candidates and asks the user to pick one
+*within the current turn* — the choice comes back as the tool result, so the
+agent simply keeps going (no separate "I picked X" user message).
+
+How the prompt is surfaced is a host concern, isolated behind
+:class:`SelectionElicitor`:
+
+* standalone CLI injects a terminal picker;
+* the cowork-server harness injects a streaming picker that emits a GUI event
+  and awaits the reply.
+
+The tool and the agent loop depend only on the protocol — never on a concrete
+transport — so the same disambiguation logic serves every host.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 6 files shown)


anton/core/tools/tool_handlers.py
```diff
@@ -578,3 +578,237 @@ async def handle_read_image(
...
+    if chosen is None:
+        return json.dumps({
+            "status": "cancelled",
+            "message": "The user dismissed the picker without choosing. Ask how they would like to proceed.",
+        })
+    path = Path(chosen).expanduser()
+    if not path.is_absolute():
+        path = root / path
+    if not path.exists():
+        return json.dumps({"status": "invalid", "message": "The selected path no longer exists."})
```
ZoranPandovski: Can we move all of this in a util to simplify as:
```
def _status(status: str, message: str = "", **extra) -> str:
    return json.dumps({"status": status, **({"message": message} if message else {}), **extra})

```

anton/core/tools/tool_handlers.py
```diff
@@ -578,3 +578,237 @@ async def handle_read_image(
...
+    """True when *candidate* is inside *root* — the path-traversal guard."""
+    try:
+        candidate.relative_to(root)
+        return True
+    except ValueError:
+        return False
+
+
+def _collect_selection_candidates(
+    session: "ChatSession", tc_input: dict, root: "Path", kind: str
```
ZoranPandovski: Is session used? if not we need to remov it

anton/core/tools/tool_handlers.py
```diff
@@ -578,3 +578,237 @@ async def handle_read_image(
...
+    start = Path(raw).expanduser()
+    if not start.is_absolute():
+        start = root / start
+    start = start.resolve()
+    return start if start.is_dir() else root
+
+
+async def _run_elicitor(elicitor, request):
+    """Run the elicitor, returning (chosen, error_json). error_json is None on success."""
+    import json
```
ZoranPandovski: Add top level imports, we have this in each function that uses it

anton/core/tools/tool_handlers.py
```diff
@@ -578,3 +578,237 @@ async def handle_read_image(
...
+    * **pick** — ``candidates`` or ``pattern`` given: disambiguate concrete
+      matches within the project. Auto-resolves a single match and reports
+      "no matches" for none, so the picker appears only for a genuine (≥2)
+      ambiguity.
+
+    The result is fed back as the tool result, so the agent continues without a
+    separate user message.
+    """
+    import json
+    from pathlib import Path
```
ZoranPandovski: Same here, top levele import

anton/core/interaction/selection.py
```diff
@@ -0,0 +1,72 @@
...
+
+    prompt: str
+    options: tuple[SelectionOption, ...] = ()
+    kind: str = "any"  # "file" | "folder" | "any" — what is being chosen
+    mode: str = "pick"  # "pick" | "browse"
+    root: str = ""  # browse-mode start directory (absolute)
+
+
+@runtime_checkable
+class SelectionElicitor(Protocol):
```
ZoranPandovski, line 62: Why is this a protocol? I think we don;'t need the runtime_checkable, we can simplify this otherwise it will be confusing

ZoranPandovski (changes requested):
Hey @hazemahmedx0 I've added few comments

anton/core/tools/tool_handlers.py
hazemahmedx0: Done. Added `_status(status, message="", **extra)` and routed every select_path result through it, so the repeated json.dumps blocks are gone.

anton/core/tools/tool_handlers.py
hazemahmedx0: Right, it was unused. Dropped the `session` arg and updated the call site.

anton/core/tools/tool_handlers.py
hazemahmedx0: Hoisted `json` and `Path` to the module top and removed the local imports across the select_path functions.

anton/core/tools/tool_handlers.py
hazemahmedx0: Covered by the module-level `json`/`Path` imports now, this local one is gone too.

anton/core/interaction/selection.py
hazemahmedx0: Dropped `@runtime_checkable`, nothing does an isinstance on it so it was dead. I kept it a Protocol on purpose though: cowork-server's StreamingSelectionElicitor satisfies it by shape without importing or inheriting this class, same seam as HippocampusProtocol. Making it an ABC would force a cross-repo inheritance dependency. Added a docstring note explaining that.

ZoranPandovski (approved):

PR #205, opened 2026-06-22
Report format: deliver default, offer alternatives, no persistence

torrmal:
## What

When a user asks for a **report / memo / document**, Anton now:

1. **Produces the default without asking first** — markdown (`report.md`) for written analyses, an HTML app for visual reports/dashboards.
2. The **first time** a document is produced in a conversation, closes with a brief, non-blocking invitation: *"Let me know if you'd like this in a different format (PDF, Word, …) or any other changes."* — offered **once per conversation**, not on every artifact.
3. If the user **names a format** (now, or earlier in the conversation), honors it for the **rest of that conversation** — regenerates and uses it for further documents.

## Deliberately no persistence

The format choice is **conversation context only** — we do **not** `memorize` it as a standing/global preference. This keeps behavior predictable (no cross-conversation surprises, no scope/consent/nag management). To make sure the choice survives a long conversation, the compaction **STATE RECORD** now explicitly preserves any output-format choice under **Constraints**.

## Why this shape

This was the outcome of a design discussion weighing three options (silent inference → global profile; ask-to-remember; deliver-default-and-invite). The deliver-default-and-invite model is act-first compatible (the closing line is an invitation, not a blocking question) and needs no memory infrastructure.

## Files
- `anton/core/llm/prompts.py` — OUTPUT FORMAT guidance in the artifacts section
- `anton/core/session.py` — compaction preserves the format choice under Constraints


torrmal:
Closing — Anton already makes the right default format choices without extra prompt guidance. The only piece we need for this is the deterministic Export control on artifact cards (cowork-server#81 + cowork#218), so this prompt change is unnecessary.

PR #216, opened 2026-06-25
fix(scratchpad): correct Gemini vision format (gate anthropic on host, not openai-compatible)

alecantu7:
Part of ENG-466. Surfaced in the cowork-server #111 review: presenting Gemini to the scratchpad as `openai-compatible` exposed a pre-existing assumption in `scratchpad_boot.py`.

## Bug
`scratchpad_boot.py` forced `vision_format="anthropic"` for **any** `openai-compatible` provider. That's only correct for Minds/MindsHub/MDB.AI (which proxy Anthropic over the OpenAI HTTP envelope). **Gemini** is presented as `openai-compatible` (it speaks OpenAI's protocol at Google's `generativelanguage.googleapis.com` endpoint), so it was being handed **Anthropic-shaped image blocks** — malformed for Google, which expects standard OpenAI `image_url`. → image/vision input through the Gemini scratchpad was broken (text was fine).

## Fix
Gate the anthropic override on the existing `_anthropic_proxy` **host** check (`mdb.ai` / `mindshub.ai`) instead of on the provider string. `OpenAIProvider` already defaults to `supports_vision=True` / `vision_format="openai"`, so:
- Minds/MindsHub → Anthropic format (unchanged) ✅
- Gemini + generic OpenAI-compatible → OpenAI `image_url` (correct) ✅
- direct OpenAI → unchanged ✅

One-line condition change (`if _scratchpad_provider_name == "openai-compatible" or _anthropic_proxy:` → `if _anthropic_proxy:`) + comment.

## Testing
`scratchpad_boot.py` is exec'd in the scratchpad subprocess (not importable), so no unit test; verified by `ast.parse` and by reasoning against `OpenAIProvider`'s vision defaults. 

## Separate observation (not fixed here)
The **main-agent** path (cowork-server `_make_provider`) sets no `vision_format`, so it always uses the `openai` default — meaning **Minds** vision in the main agent likely gets OpenAI `image_url` instead of Anthropic blocks. Pre-existing, separate from this scratchpad fix; flagging for a follow-up.

### DIFF
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -80,23 +80,26 @@ if _scratchpad_model:
                 "ANTON_OPENAI_API_KEY"
             )
             _llm_api_version = os.environ.get("ANTON_OPENAI_API_VERSION") or None
-            # Mirror LLMClient.from_settings: openai-compatible endpoints
-            # (Minds/MindsHub) expect image blocks in Anthropic native format,
-            # not OpenAI image_url data-URLs.
             _llm_provider_kwargs: dict = {
                 "api_key": _llm_api_key or None,
                 "base_url": _llm_base_url or None,
                 "ssl_verify": _llm_ssl_verify,
                 "api_version": _llm_api_version,
             }
-            # Endpoints that proxy to Minds / MindsHub / MDB.AI
-            # speak Anthropic content format over the OpenAI HTTP envelope,
-            # so image blocks must be Anthropic-shaped, not OpenAI image_url.
+            # Only Minds / MindsHub / MDB.AI proxy Anthropic over the OpenAI HTTP
+            # envelope, so ONLY they need image blocks in Anthropic native format.
+            # Gate on the host, NOT on "openai-compatible" generally: other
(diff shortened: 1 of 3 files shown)


alecantu7:
Folded in the main-agent counterpart (`e3670a2`): `LLMClient.from_settings`'s `openai-compatible` factory had the identical unconditional `vision_format="anthropic"`, so a **standalone anton** instance with openai-compatible pointed at Gemini / a generic proxy hit the same image-mangling bug. Applied the same `_anthropic_proxy`-style host gate (mdb.ai / mindshub.ai → anthropic; everything else → openai image_url).

Note: this does **not** affect the cowork product — cowork-server's `build_llm_client._make_provider` constructs `OpenAIProvider` directly with the default `vision_format="openai"`, bypassing `from_settings`. So it's the anton-standalone path only (low urgency), but cheap to fix alongside the scratchpad gate. Added `from_settings` tests (anthropic for mindshub/mdb hosts; openai for Gemini/generic).

#216 now covers both anton vision paths: scratchpad (`scratchpad_boot.py`) and main agent (`from_settings`).

SailingSF (approved):
LGTM

PR #217, opened 2026-06-25
ENG-385: tag CI traffic with is_ci

tino097:
## ENG-385 — Tag CI traffic with `is_ci` (anton)

Adds an env-derived `is_ci` flag to every `send_event` so CI/automation traffic can be filtered out of the product funnel in PostHog (Anton project 355390).

- `_is_ci()` checks common CI env vars (`CI`, `GITHUB_ACTIONS`, `GITLAB_CI`, …), cached per process.
- Stamped as `is_ci=true|false` on every event.
- **No email/staff signal** is added here — this path is anonymous by design (keyed on the hashed `aid`, not a user account), so only CI detection is possible.

### Checks
- `uv run pytest tests/test_analytics.py` → 4 passed (new file)

Pairs with the cowork PR (renderer `is_ci` + `is_internal` + the new funnel events).

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -52,30 +52,32 @@ _TIMEOUT = 3  # seconds
 # a process, so computing it once is sufficient.
 _cached_aid: str | None = None
 
-# Cached CI detection. Env-derived (no PII), so it stays consistent with this
-# module's anonymous design while letting the funnel filter out CI/automation
-# traffic via the ``is_ci`` flag (ENG-385). There is no email/staff signal on
-# this path — events are keyed on the anonymous ``aid``, not a user account.
+# Cached CI detection. Env-derived (no PII), consistent with this module's
+# anonymous design. CI/automation traffic is dropped entirely (see send_event)
+# rather than tagged, so it can't pollute the product funnel. Driven by an
+# explicit Anton-owned signal (ANTON_IS_CI) with known provider markers as a
+# convenience fallback; the bare ``CI`` var is intentionally not consulted —
+# it's frequently set to "false" or leaks into local dev shells (ENG-385).
 _cached_is_ci: bool | None = None
 
-# CI markers across common providers. ``CI`` is set by virtually all of them;
-# the rest catch providers that don't, or set it inconsistently.
-_CI_ENV_VARS = (
-    "CI",
-    "GITHUB_ACTIONS",
-    "GITLAB_CI",
-    "BUILDKITE",
-    "CIRCLECI",
-    "JENKINS_URL",
-    "TF_BUILD",
-)
+
+def _env_true(name: str) -> bool:
+    return os.environ.get(name, "").strip().lower() in {"1", "true", "yes", "on"}
 
 
 def _is_ci() -> bool:
-    """Return True when running under CI/automation (cached, env-only)."""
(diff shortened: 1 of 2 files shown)


tino097:
https://linear.app/mindsdb/issue/ENG-385/free-tier-funnel-instrumentation-in-app-events

anton/analytics.py
```diff
@@ -108,6 +138,7 @@ def send_event(settings: "AntonSettings", action: str, **extra: str) -> None:
         params: dict[str, str] = {
             "action": action,
             "aid": get_installation_id(),
+            "is_ci": "true" if _is_ci() else "false",
```
pnewsam: Instead of adding this property, should we just not send the event?

I don't think there would be any need to send to Posthog.

anton/analytics.py
```diff
@@ -51,6 +52,32 @@
...
+    "JENKINS_URL",
+    "TF_BUILD",
+)
+
+
+def _is_ci() -> bool:
+    """Return True when running under CI/automation (cached, env-only)."""
+    global _cached_is_ci
+    if _cached_is_ci is None:
+        _cached_is_ci = any(os.environ.get(v) for v in _CI_ENV_VARS)
```
pnewsam: **Minor:** I think this is a little brittle as the contract for a product analytics flag. Right now `is_ci` means "any common CI-looking env var is present", but the thing we really want to know seems closer to "this is Anton automation/CI traffic that should be filtered from the funnel."

For GitHub Actions there is a standard marker: GitHub sets `GITHUB_ACTIONS=true` for workflow runs, and also sets `CI=true` (`CI` can currently be overwritten). I would make an Anton-owned variable the primary control, for example `ANTON_IS_CI=true`, and either set that explicitly in our workflows or use `GITHUB_ACTIONS == "true"` as a convenience fallback. That keeps the analytics contract explicit and avoids classifying local/dev usage based on broad ambient CI variables like `CI=false` or a leaked `CI` env var.

A concrete replacement could look like:

```python
def _env_true(name: str) -> bool:
    return os.environ.get(name, "").strip().lower() in {"1", "true", "yes", "on"}


def _is_ci() -> bool:
    return (
        _env_true("ANTON_IS_CI")
        or _env_true("GITHUB_ACTIONS")
        or _env_true("GITLAB_CI")
        or _env_true("BUILDKITE")
        or _env_true("CIRCLECI")
        or _env_true("TF_BUILD")
        or bool(os.environ.get("JENKINS_URL"))
    )
```

pnewsam (commented):
## Code Review

**Verdict: COMMENT**

I think the implementation would be more robust if the product analytics contract were driven by an Anton-owned environment variable, with GitHub Actions' standard variable as an optional fallback. The current broad provider sniffing works, but it makes the meaning of `is_ci` depend on ambient environment details rather than an explicit Anton signal.

### Summary of findings

| Severity | File | Issue |
|----------|------|-------|
| Minor | `anton/analytics.py:78` | Broad CI auto-detection may be the wrong contract for a product analytics flag. |

### What's working well

The change is nicely scoped, uses GitHub's standard `GITHUB_ACTIONS` marker, and includes focused tests around the emitted query parameter.

---
_Reviewed by Codex · tino097/eng-385-free-tier-funnel-instrumentation-in-app-events → staging · 2026-06-25_

PR #220, opened 2026-06-26
Add weekly release cycle workflows (freeze, release PR, unfreeze)

pnewsam:
## Summary

Adds the weekly `staging → main` release cycle to this repo as **thin callers** of the shared reusable workflows in `mindsdb/github-actions` (see **mindsdb/github-actions#32**). Same pattern as `stale-deploy-label.yml`.

| Caller | Name | Trigger |
|---|---|---|
| `staging-freeze.yml` | `Staging Freeze` | Weekly cron + manual dispatch |
| `weekly-merge-staging.yml` | `Create staging to main release PR` | After freeze (`workflow_run`) + dispatch |
| `staging-unfreeze.yml` | `Staging Unfreeze` | Release PR merged + dispatch |

Each caller is ~15 lines — triggers, guards, and a least-privilege `permissions` block. All logic lives once in `github-actions`. The workflow **names are kept identical** so the `workflow_run` chain still fires.

This supersedes the earlier self-contained version of this PR (full workflows plus a branch-lock helper script), incorporating the review feedback: shared reusable workflows, GitHub App auth in place of a PAT, and a ruleset-based freeze.

## Merge order

Merge **mindsdb/github-actions#32 first** — the callers reference it `@main`.


lucas-koontz (changes requested):
Requesting changes as part of a coordinated review of all 7 release-train PRs (byte-identical: mindshub_inference#319, mindshub_frontend#1337, mindshub_services#140, auth#132, anton#220, cowork#285, cowork-server#179). The design is right and we're keeping it; what changes before merge is where the code lives and how it authenticates. Full step-by-step is coming to you on Slack. Summary:

**Blocking (set-wide; this PR is a copy of the shared implementation)**

1. The sync-back push works on this repo (staging has no required-review rule here) but fails on the four internal repos, where staging requires a review with `enforce_admins` on and empty bypass lists. Since all 7 PRs are identical copies, the shared implementation changes and this PR becomes a thin caller too.
2. **Swap the fine-grained PAT for the org GitHub App `mindsdb-release-train`** (already registered; creds stored as org var `RELEASE_APP_CLIENT_ID` + org secret `RELEASE_APP_PRIVATE_KEY`). This matters most on this repo because it's public: a long-lived admin-capable org secret here would be readable by every workflow and every third-party action in the repo, and the org currently allows all actions unpinned. Per-job 1-hour app tokens via `actions/create-github-app-token@v3` shrink that exposure. `RELEASE_AUTOMATION_TOKEN` will not be provisioned.
3. **Single source of truth.** Move the three workflows into `mindsdb/github-actions` as reusable workflows with ~25-line per-repo callers (cron/dispatch/pull_request triggers stay in the callers; keep the workflow names identical so the `workflow_run` chain still fires). Same pattern we already use for `stale-deploy-label.yml`.
4. **Freeze mechanism.** Replace the classic branch-protection round-trip (`set-branch-lock.js`) with a pre-provisioned `staging-freeze` ruleset toggled between `enforcement: active/disabled`. Rulesets layer on top of existing protection, so a bug in the toggle can't clobber the protection config, and the JS helper goes away entirely. I'll pre-create the rulesets with the app as bypass actor.

**Public-repo hardening to fold in (load-bearing here)**

- Unfreeze guard: add `github.event.pull_request.head.repo.full_name == github.repository` alongside the `head.ref` check (a fork can name its branch `staging`; today that only produces a red run since forks never get secrets, but the guard makes the trust boundary explicit and protects against future trigger drift).
- SHA-pin `actions/checkout`, `actions/github-script`, `create-github-app-token` (full commit SHA + version comment); `persist-credentials: false` on the checkouts that never push.
- Move the cron off the top of the hour (e.g. `47 13 * * 5`): xx:00 is GitHub's documented high-load slot, and the "6am PST" comment is only true in winter. Also note: on public repos GitHub auto-disables scheduled workflows after 60 days of repo inactivity; the freeze's skip logic makes that mostly harmless, but worth knowing.

**Rollout once reworked:** central `github-actions` PR first, canary caller on cowork-server with one manual full cycle (freeze, release PR, merge, unfreeze, sync-back), then the remaining six.

lucas-koontz (approved):
Approving. Verified: mindsdb/github-actions#32 merged and its merge commit (75df118f) is exactly the ref pinned in all three callers; caller files are byte-identical across the 7 repos (blob-SHA check); CI green. Rollout per the pilot plan: cowork-server merges first and validates via manual dispatch, the rest follow.

PR #226, opened 2026-06-30
fix(scrub): redact provider API keys from tool output before it reaches the LLM (ENG-463)

tino097:
## Summary

Fixes [ENG-463](https://linear.app/mindsdb/issue/ENG-463) — a raw provider API key reached the LLM's context (the agent was caught writing a live `mdb_` key into generated scratchpad code). *A model can only emit a secret it can see.*

**Root cause:** `scrub_credentials()` (anton) is already applied to every tool result before it's appended to history (`session.py:1472` for strings, `:1464` for multimodal text blocks) — but it only redacted `DS_*` datasource secrets. Provider keys (`mdb_`/`sk-`/`AIza`) never start with `DS_`, so a key surfaced via a tool result, a scratchpad traceback (env values appear in frame locals), or a settings echo passed straight through into model context — and was re-sent to the provider as prompt text on every subsequent turn.

## Change (anton, `anton/utils/datasources.py`)
Extend `scrub_credentials` to also redact, after the existing `DS_*` logic:
1. **By value** — the values of `_PROVIDER_SECRET_VARS` (`OPENAI/ANTHROPIC/GEMINI/ANTON_*_API_KEY`, `ANTON_MINDS_API_KEY`) present in `os.environ`, with an informative `[VAR]` label (mirrors the `DS_*` secret handling). Precise, no false positives.
2. **By shape** — `_SECRET_KEY_PATTERN` (`mdb_…`, `sk-…`/`sk-proj-`/`sk-ant-`, `AIza…`, with length floors) → `[REDACTED_API_KEY]`, so a key that didn't originate from one of our env vars (e.g. one the model already emitted) is still caught.

No change to `scratchpad_boot.py` needed: the scratchpad result/traceback is returned as the tool-result string, which already flows through `scrub_credentials`. Base URLs and short `sk-`/`mdb_` fragments are deliberately left readable to avoid over-redaction.

## Testing
`uv run pytest tests/test_scrubbing.py` → **10 passed** (5 new: value-redaction with label for minds/openai keys; pattern-redaction of `mdb_`/`sk-ant-`/`AIza` keys with no env var set; base-URL + short-`sk-` left intact). Existing `DS_*` scrub tests in `test_datasource.py` still green (23 passed).

## Scope / follow-up (NOT in this PR)
This closes the **model-visible** path (the core of ENG-463 — what the model sees and re-emits). The issue also notes the key lands in the **SSE `/responses` stream, `cowork.db`, and the HTTP cache**. Those persistence paths are emitted in `cowork-server` (`harnesses/anton_harness/stream_formatter.py` `StreamToolResult`) independently of anton's history scrub, and overlap ENG-462 (cache mitigation already shipped). A defense-in-depth follow-up could redact at that stream layer too (pattern-based, since cowork-server's process may not hold the provider env values). Flagging rather than bundling — happy to do it next if you want full coverage.

### DIFF
--- a/anton/utils/datasources.py
+++ b/anton/utils/datasources.py
@@ -20,6 +20,31 @@ _DS_SECRET_VARS: set[str] = set()
 # DS_* var names for **ALL** fields of registered engines.
 _DS_KNOWN_VARS: set[str] = set()
 
+# Provider credential env vars injected into the scratchpad execution env for
+# the code to USE. Their values must never reach the LLM (ENG-463): a model can
+# only emit a secret it can see, and the agent was caught writing a raw `mdb_`
+# key into generated code. Scrubbed by exact value (labeled) from anything that
+# re-enters model context.
+_PROVIDER_SECRET_VARS: tuple[str, ...] = (
+    "OPENAI_API_KEY",
+    "ANTHROPIC_API_KEY",
+    "GEMINI_API_KEY",
+    "ANTON_OPENAI_API_KEY",
+    "ANTON_ANTHROPIC_API_KEY",
+    "ANTON_GEMINI_API_KEY",
+    "ANTON_MINDS_API_KEY",
+)
+
+# Well-known provider key formats — scrubbed by shape so a key is redacted even
+# when it didn't come from one of our env vars (e.g. the model already emitted
+# it, or it arrived via an unexpected path). Length floors keep this from
+# matching short incidental `sk-`/`mdb_` strings.
+_SECRET_KEY_PATTERN = re.compile(
+    r"mdb_[A-Za-z0-9._-]{10,}"   # MindsHub
+    r"|sk-[A-Za-z0-9_-]{20,}"    # OpenAI / Anthropic (sk-, sk-proj-, sk-ant-)
+    r"|AIza[A-Za-z0-9_-]{30,}"   # Google / Gemini
+)
+
 
 def _reset_registered_ds_vars() -> None:
     """Clear the DS_* var registries so they can be rebuilt from current vault state."""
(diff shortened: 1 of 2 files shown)


torrmal (approved):
looks good to me, @sejubar?

alecantu7 (approved):
**Approving** — correct root-cause fix, well-tested.

The key thing is *where* this scrubs: at the inbound-to-context boundary. I confirmed `scrub_credentials` runs at all four `session.py` tool-result sites (the streaming and non-streaming paths, for both string and multimodal-text results) plus `publisher.py` — so extending it from `DS_*` to provider keys covers every path a tool result re-enters model context. A model can only emit a secret it can see; cut it off at ingest and the agent can't write a live `mdb_` key into generated code.

Two composing layers are the right design:
- **By value** — exact env values of `_PROVIDER_SECRET_VARS`, labeled. Precise, no false positives, `len>8` guard.
- **By shape** — `_SECRET_KEY_PATTERN` (`mdb_`/`sk-`/`AIza`, length floors) backstops anything not sourced from a known env var. Since all tracked providers (OpenAI/Anthropic/Gemini/Minds) have shapes the pattern covers, nothing realistic escapes both.

Tests lock the load-bearing behavior, including the negative cases (short `sk-`/base URL left readable). LGTM.

Two non-blocking notes (no change required here):
1. **Scope boundary — worth a follow-up.** This closes the **model-visible / re-emission** path. A key the model *does* emit still flows **outbound** to the SSE `/responses` stream + `cowork.db` messages unredacted — those are emitted in cowork-server's `stream_formatter`, independent of anton's history scrub. ENG-462 handles the cache; a defense-in-depth pass to redact at that stream layer would close the persistence side. Flagging so it's tracked, not bundled.
2. **Maintenance coupling.** Adding a new BYOK provider means updating `_PROVIDER_SECRET_VARS` (and `_SECRET_KEY_PATTERN` if its key shape differs), or a new key could slip through — worth a one-line comment by the lists.

PR #144, opened 2026-04-30
ok, inline terms

torrmal:
This resolves 
https://github.com/mindsdb/anton/issues/112
and it's a cleaner experience

PR #227, opened 2026-06-30
feat(datasources): surface non-secret connection identity in the agent prompt (ENG-508)

alecantu7:
## What

`build_datasource_context` (the "## Connected Data Sources" system-prompt section) listed only `DS_*` **var names**, so the agent could see *that* a Gmail was connected but not **which account** — and with several connected it couldn't pick the right one ("send from my support email"). Asked the address directly, it can't read it back either: the scrubber redacts any `DS_*` value > 8 chars for unregistered connectors.

This adds a short **non-secret identity** to each line — the account `email`, or the DB `host` (+ `database`):

```
## Connected Data Sources
- `gmail-support` (gmail) — support@acme.com → DS_GMAIL_SUPPORT__EMAIL, DS_GMAIL_SUPPORT__APP_PASSWORD
- `postgres-prod` (postgres) — db.acme.com/sales → DS_POSTGRES_PROD__HOST, DS_POSTGRES_PROD__PASSWORD, ...
```

The identity comes from the **prompt** (read from the vault), which isn't scrubbed — so the agent can finally tell connections apart and select among them.

## Scope / safety

- **Scoped, not a dump.** Only `email` / `host` / `database` are shown — inherently non-secret, account-identifying fields. It deliberately does **not** print every non-secret field (which would surface opaque `client_id` or config like `ssl_mode`), and never a secret.
- **Helps existing connections too.** Because it reads the stored `fields`, it labels connections saved under an old random slug (`gmail-548bdb`), not just newly-named ones.
- The scrubber is unchanged; this routes identity via the prompt rather than via scratchpad output.

## Tests

`tests/test_datasource_context_identity.py` — identity selection (email/host/host+db), and that secrets and opaque field *values* never reach the prompt. `uv run --with pytest python -m pytest tests/test_data_vault.py tests/test_datasource.py tests/test_datasource_context_identity.py` → **221 passed**.

## Coordination

Pairs with the cowork-server side ([cowork-server#144](https://github.com/mindsdb/cowork-server/pull/144), ENG-508): readable identity-derived slugs + non-destructive save + secure_keys. This anton change reaches a deployed cowork-server via the usual PyPI/`main` path (release + `anton-agent` pin), so it lands after merge → release. Part of ENG-508.

### DIFF
--- a/anton/utils/datasources.py
+++ b/anton/utils/datasources.py
@@ -148,7 +148,11 @@ def build_datasource_context(vault: DataVault, active_only: str | None = None) -
             fields = vault.load(c["engine"], c["name"]) or {}
             secure_keys = None
         prefix = _slug_env_prefix(c["engine"], c["name"])
-        var_names = ", ".join(f"{prefix}__{k.upper()}" for k in fields)
+        # Skip `_`-prefixed bookkeeping (`_connector_id`, `_method`, `_label`) —
+        # they're not credential env vars the agent should reference.
+        var_names = ", ".join(
+            f"{prefix}__{k.upper()}" for k in fields if not k.startswith("_")
+        )
         # Prefer a user/agent-assigned label ("Support"); otherwise the derived
         # non-secret identity (email / host).
         identity = str(fields.get("_label", "")).strip() or _connection_identity(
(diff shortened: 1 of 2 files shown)


alecantu7:
## Self-review

Fixed in `01f9a6c`: `_connection_identity` now **respects `secure_keys`** — it skips any field a record explicitly marks secret (read together with the fields via one `read_record` call), as defensive depth even though email/host are never normally secret.

Otherwise the change held up: identity is scoped to email/host (not a dump), no secret reaches the prompt (asserted by test), and it labels pre-existing random-slug connections too. Paired with cowork-server #144. Suite: 222 passed.

SailingSF (approved):
👍

PR #228, opened 2026-06-30
ENG-513: validate stale training-data assumptions before blocking tasks

torrmal:
## Summary

- Adds a new conversation-discipline rule to Anton's system prompt: before using a training-data fact as a reason to refuse or block a task, verify it online first
- Treats the user's question itself as evidence the fact may have changed (e.g., asking about SpaceX stock → SpaceX may now be public)
- Anton states what it's checking and why: *"My training says X, but that could be outdated — let me verify."*

## Motivation

Anton was asserting stale facts from training (e.g. "SpaceX is not a publicly traded company") as blocking conditions, preventing queries that would otherwise succeed. The fix is a prompt-level rule rather than code since this is a reasoning/judgment issue.

## Test plan

- [ ] Ask Anton: *"Give me the performance of SpaceX stocks over the past 3 months"* — it should verify public status online instead of refusing
- [ ] Ask about a company whose status is definitively unchanged — Anton should still proceed normally (validate is fast, not harmful)
- [ ] Ask about a genuinely private company — Anton validates, confirms it's still private, then explains cleanly

Fixes ENG-513

### DIFF
--- a/anton/core/llm/prompts.py
+++ b/anton/core/llm/prompts.py
@@ -223,6 +223,13 @@ Phrase it like "Assuming you mean X (the common case), so I'll…" or "Going wit
 granularity since you didn't specify." Surface each assumption as it happens so the user \
 can redirect mid-flight instead of being blocked up front. Acting silently is wrong; \
 acting out loud with your assumptions visible is right.
+- NEVER let a training-data fact BLOCK a task without first verifying it is still current. \
+Facts like company public/private status, leadership, product availability, regulatory \
+status, or market listings can change after your training cutoff. If something you learned \
+during training would prevent you from completing a request (e.g., "that company isn't \
+publicly traded so I can't fetch stock data"), treat the user's question itself as evidence \
+the fact may have changed — validate it online FIRST, then proceed. State what you're \
+checking and why: "My training says X, but that could be outdated — let me verify."
 - Only STOP and ASK when acting on a guess would be costly to undo or is genuinely \
 unknowable: destructive or irreversible actions (deleting data, spending money, sending \
 messages on the user's behalf), credentials or access you can't obtain, or a fork where \


alecantu7 (approved):
**Approving** — now a clean, focused change (base `staging`, +7/−0, one bullet in the conversation-discipline list). The rule is the right call: it's a reasoning/judgment issue, so a prompt rule is the correct lever, not code.

On the two consequences worth watching:
- **Prompt bloat:** ~120 tokens appended to an existing list — negligible and in-style.
- **Latency / over-validation:** bounded, because the rule is scoped to the **refusal path** ("if something you learned during training would *prevent* you from completing a request"). It doesn't fire on normal factual answers — only when Anton was about to refuse, which is rare and exactly where a web check pays for itself (those queries were previously broken).

Two **non-blocking** wording suggestions (take or leave):
1. **Web-unavailable fallback.** Some harnesses run with web disabled; there "validate online FIRST" has no tool. A clause like *"if you can't verify, flag the uncertainty and proceed rather than block"* keeps it from stranding the model when there's no web tool.
2. **Make it explicitly refusal-only** — one word ("ONLY when you would otherwise refuse/block") removes any risk the model generalizes it to double-checking every time-sensitive mention (the main over-validation/latency tail).

**Testing note (can't be exercised in the dev env here):** verify both branches of the behavior, and mind the ground-truth durability lesson —
- **Changed fact → correct & proceed:** *SpaceX* now works (it IPO'd 2026-06-12, Nasdaq: SPCX), so it exercises "training said private → verify → it's public → proceed."
- **Unchanged fact → confirm & explain (don't fabricate):** *Anthropic* is a better probe here — still private as of July 2026 (confidential S-1 filed 2026-06-01; listing expected ~Oct 2026). Ask "Anthropic's stock performance" → the rule should validate, confirm it's still private, and explain cleanly rather than inventing a price.
- ⚠️ Anthropic is itself time-sensitive (expected to IPO ~Oct 2026), so anchor/re-verify that case then — same trap that flipped SpaceX. ENG-381's `grounding-spacex-ipo-01` already covers the changed-fact branch and its `llm_calls` metric can catch over-validation.

Net: mergeable as-is; the two wording tweaks would tighten the latency tail.

PR #238, opened 2026-07-07
feat(ENG-644): prompt caching via cache_control on the stable prefix and history

hazemahmedx0:
## Summary

- **Turn on Anthropic prompt caching with zero behavior change.** The prompt was already assembled stable-first / volatile-last by design, but no cache markers were ever emitted, so the ~16k-token fixed prefix (system prompt + tool schemas) was billed at full price on every call, up to 7 times per user turn. This PR places two `cache_control` markers: one on the stable system block (which also caches the tool definitions that precede it in the request) and one on the last history message (so the conversation caches incrementally and only the newest exchange is fresh each round). Expected effect: 60 to 80 percent cheaper input on multi-call turns plus lower time to first token, with the model seeing byte-identical content.
- **`SystemPrompt` is a `str` subclass, not a new interface.** It IS the full assembled prompt string, so every existing consumer that measures, substring-checks, or serializes the system prompt keeps working untouched (all 1152 existing tests pass without modification). Cache-aware providers additionally read `.stable` / `.volatile` to place markers. Passing a plain `str` anywhere keeps the exact pre-caching behavior.
- **Compaction stays correct under caching.** The API's `input_tokens` excludes cached tokens, so `context_pressure` is now computed on the total (fresh + cache read + cache write). Without this, a fully-cached long conversation would report near-zero pressure and history compaction would never fire. `Usage` gains `cache_read_input_tokens` / `cache_creation_input_tokens`, which downstream telemetry (cowork-server PR #172) already has fields waiting for.
- **Fixed a real leak: the scratchpad tool description grew for the life of the process.** `_build_core_tools` appended the package list and memory wisdom to the shared module-level `ToolDef` instance on every session build; each new session inherited all previous sessions' appends. Now each session gets its own copy (`dataclasses.replace`), which also keeps the tool schema byte-stable for the cache.
- **Kill switch:** `ANTON_PROMPT_CACHING=false` falls back to plain strings end to end, for gateways that reject or strip cache markers. OpenAI-family paths coerce to plain text (their prefix caching is automatic; the stable-first ordering already helps them).

## Testing

1. `pytest tests/` — 1152 passed, 17 skipped (13 new tests in `tests/test_prompt_caching.py`).
2. Key invariants tested: `build_parts()` output is byte-identical to `build()`; plain-str system produces the exact old request shape; the session's live history list is never mutated by marker placement; pressure math uses the total context; the kill switch returns a plain str.
3. Real-world verification happens behind the telemetry from ENG-642: after deploy, `cache_read_input_tokens` must be greater than zero from the second call of a turn onward, and billed input cost on multi-round eval turns should drop at least 50 percent.

Refs ENG-644 (Linear). Draft until before/after numbers from the ENG-642 baseline prove the win.

---

## Glossary

| Term | Definition |
|------|------------|
| Prompt caching | Anthropic API feature: mark the stable request prefix; later calls read it from cache at 10 percent of input price with lower latency (5 minute TTL) |
| cache_control | The marker object (`{"type": "ephemeral"}`) placed on a content block or tool to declare a cache boundary |
| Stable / volatile split | Everything per-session-constant (identity, tool guidance, project context) vs what changes per turn (live clock, memory snapshot); the split point already existed in the builder, this PR just exposes it |
| Context pressure | Total context size divided by the model window; anton compacts history when it crosses 0.7 |
| Incremental history caching | Each call marks its own last message; the provider still hits entries created at earlier boundaries, so only the newest exchange is uncached |

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -17,6 +17,7 @@ from .provider import (
     StreamToolUseDelta,
     StreamToolUseEnd,
     StreamToolUseStart,
+    SystemPrompt,
     ToolCall,
     Usage,
     compute_context_pressure,
@@ -53,6 +54,79 @@ def _build_native_web_tools(
     return entries, beta
 
 
+_CACHE_MARKER = {"type": "ephemeral"}
+# Block types cache_control may legally sit on in a message's content array.
+_CACHEABLE_BLOCK_TYPES = {"text", "tool_result", "image"}
+
+
+def _system_param(system: str | SystemPrompt) -> str | list[dict]:
+    """SystemPrompt → content blocks with a cache marker on the stable prefix.
+
+    The marker caches everything before it in the request (tools + the stable
+    system block); the volatile tail rides after it, uncached, so the live
+    clock and memory snapshot never invalidate the prefix. Plain strings pass
+    through unchanged (uncached, pre-existing behavior).
+    """
+    if not isinstance(system, SystemPrompt):
+        return system
+    blocks = [{"type": "text", "text": system.stable, "cache_control": _CACHE_MARKER}]
+    if system.volatile:
+        blocks.append({"type": "text", "text": system.volatile})
+    return blocks
+
+
+def _mark_history_for_cache(messages: list[dict]) -> list[dict]:
+    """Cache marker on the final content block of the last message.
+
+    Each call marks its own last message; Anthropic still hits the entries
+    created at earlier boundaries, so history caches incrementally — only the
+    newest exchange is fresh each round. Copy-on-write: the caller's message
+    list (the session's live history) is never mutated. Anything unexpected →
+    return the input unchanged; caching is best-effort.
+    """
+    if not messages:
+        return messages
+    last = messages[-1]
+    content = last.get("content") if isinstance(last, dict) else None
+    marked_content = None
+    if isinstance(content, str) and content:
+        marked_content = [
+            {"type": "text", "text": content, "cache_control": _CACHE_MARKER}
+        ]
+    elif isinstance(content, list) and content and isinstance(content[-1], dict):
+        if content[-1].get("type") in _CACHEABLE_BLOCK_TYPES:
+            marked_content = content[:-1] + [
+                {**content[-1], "cache_control": _CACHE_MARKER}
+            ]
+    if marked_content is None:
+        return messages
+    return messages[:-1] + [{**last, "content": marked_content}]
+
+
+def _usage_from(model: str, api_usage, input_tokens: int, output_tokens: int) -> Usage:
+    """Build a Usage with cache stats; pressure uses the TOTAL context size.
+
+    With caching active the API's ``input_tokens`` excludes cached tokens, so
+    compaction decisions must add the cache read/write counts back in. Cache
+    fields are coerced through ``isinstance(int)`` — gateways may omit them,
+    return null, or (in tests) hand back mock objects.
+    """
+    def _int_or_zero(value) -> int:
+        return value if isinstance(value, int) else 0
+
+    cache_read = _int_or_zero(getattr(api_usage, "cache_read_input_tokens", 0))
+    cache_creation = _int_or_zero(getattr(api_usage, "cache_creation_input_tokens", 0))
+    return Usage(
+        input_tokens=input_tokens,
+        output_tokens=output_tokens,
+        context_pressure=compute_context_pressure(
+            model, (input_tokens or 0) + cache_read + cache_creation
+        ),
+        cache_read_input_tokens=cache_read,
+        cache_creation_input_tokens=cache_creation,
+    )
+
+
 class AnthropicProvider(LLMProvider):
     name: str = "anthropic"
 
@@ -84,7 +158,7 @@ class AnthropicProvider(LLMProvider):
         self,
         *,
         model: str,
-        system: str,
+        system: str | SystemPrompt,
         messages: list[dict],
         tools: list[dict] | None = None,
         tool_choice: dict | None = None,
@@ -94,11 +168,12 @@ class AnthropicProvider(LLMProvider):
         web_entries, beta_headers = _build_native_web_tools(native_web_tools)
         merged_tools = list(tools or []) + web_entries
 
+        use_cache = isinstance(system, SystemPrompt)
         kwargs: dict = {
             "model": model,
             "max_tokens": max_tokens,
-            "system": system,
-            "messages": messages,
+            "system": _system_param(system),
+            "messages": _mark_history_for_cache(messages) if use_cache else messages,
         }
         if merged_tools:
             kwargs["tools"] = merged_tools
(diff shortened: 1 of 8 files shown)


hazemahmedx0 (commented):
## Code Review: prompt caching

**Reviewed commit:** `e3fb6ec2ad075a4a2ee82282cd1b4b4433145e16`

Reviewed against the invariants that matter for this change: byte-identical prompt content with caching on/off, no mutation of the live history list, pressure math on the true context size, `str`-subclass safety, and Anthropic `cache_control` placement rules. Also ran the full suite: **1152 passed, 17 skipped** (2m08s).

### Verdict: APPROVED

Clean, well-scoped, and the risky parts are the ones with tests. Details below.

### What I verified

**1. Byte-identity (caching on == off).** `build()` is now `build_parts()` joined with `stable + volatile`, and the volatile tail (`Current date and time…` + memory snapshot) is character-for-character what the old `build()` appended — so the model sees identical text either way. `_system_param` and `_mark_history_for_cache` only re-shape (str → content-block list, add a `cache_control` key); no text is edited, sliced, or reordered. `test_stable_plus_volatile_is_byte_identical_to_build` and `test_split_point_is_the_volatile_tail` lock this in.

**2. Live history is never mutated.** `_mark_history_for_cache` is correct copy-on-write throughout: `messages[:-1] + [{**last, "content": marked_content}]` rebuilds the outer list and the last message dict, and the last content block is copied via `{**content[-1], "cache_control": …}` rather than mutated in place. Earlier blocks (`content[:-1]`) are shared by reference but never touched. `test_original_list_never_mutated` and `test_block_content_marks_last_block_only` cover both the str and nested-list shapes.

**3. Pressure on total context.** With caching, Anthropic's `input_tokens` excludes cached tokens, so `_usage_from` adds `cache_read + cache_creation` back before computing pressure — otherwise a fully-cached long conversation would report ~0 pressure and compaction would never fire. Good catch. The OpenAI paths are correctly left alone: OpenAI's `prompt_tokens` already includes cached tokens, so pressure there is already computed on the full context — consistent with the Anthropic path, no double-count. `complete()` and `stream()` both route through `_usage_from` (stream captures cache fields from the `message_start` usage), so the two paths agree.

**4. `SystemPrompt(str)` safety.** Equality/hash/serialization all inherit from `str` and key on the full assembled string, so existing consumers that measure/substring/serialize the prompt keep working. Confirmed no code path anywhere relies on `type(x) is str` (only `isinstance` checks, which a subclass satisfies). It flows untouched from `_build_system_prompt` → `plan_with_recovery` → client → provider; nothing concatenates or coerces it on the main turn path.

**5. Per-session scratchpad copy.** `dataclasses.replace(SCRATCHPAD_TOOL)` fixes a genuine process-lifetime leak (the module constant's `description` was being appended to on every session build, so schemas accumulated across sessions). Shallow copy is sufficient here because the only mutated field is `description`, an immutable str that's reassigned — the shared `input_schema`/`handler` are read-only in this path. Per-session appends (packages, wisdom) mutate the copy before `register_tool`, so they still reach the registered tool. Full suite passing confirms the handler/schema still resolve.

**6. Anthropic placement.** Two breakpoints (stable system block + last history message) — within the 4-breakpoint limit; the system marker implicitly caches the preceding tool defs. `_CACHEABLE_BLOCK_TYPES = {text, tool_result, image}` and the "unexpected shape → return unchanged" fallback keep marking best-effort and safe.

### Suggestions (non-blocking)

- `_mark_history_for_cache` skips a `tool_use`-terminated last message (not in `_CACHEABLE_BLOCK_TYPES`). The API does accept `cache_control` on `tool_use` blocks, so this leaves a small incremental-cache win on the table in the rare case the last message at call time ends in a tool call. Safe as-is; worth a follow-up only if telemetry shows those rounds are common.
- `plan_with_recovery` / `plan_stream_with_recovery` still annotate `system: str`. Harmless (a `SystemPrompt` is a `str`), but widening to `str | SystemPrompt` would match the provider/client signatures and document the intent.
- The compaction (`code(...)`) and verifier planning calls pass plain-str systems, so they stay uncached. Fine given they're short one-offs — noting it only so it's a conscious choice.

Nice work on keeping the assembled text byte-stable and putting the invariants under test rather than in the description.

hazemahmedx0 (commented):
## Follow-up review: OpenAI-envelope cache stats (`0cfa01cc`)

Re-reviewed the delta over `e3fb6ec2` and re-ran the suite: **1156 passed, 17 skipped** (1m49s).

### Verdict: APPROVED

The new commit surfaces prompt-cache stats from OpenAI-shaped usage envelopes (OpenAI automatic caching + gateways proxying Anthropic) into the same `Usage` fields the Anthropic path uses. Checked the three things that could go wrong:

**1. Fresh-vs-total semantics consistent across all four sites.** Both chat-completions paths (`complete`, `stream`) now go through `_usage_from_openai`, and both Responses-API paths (streaming + `_parse_response_object`) do the inline equivalent. Every site lands on the same contract: `input_tokens` = fresh portion, `context_pressure` computed on the total. The two Responses sites correctly compute pressure from the *pre-subtraction* total local (the `max(... - cache_read, 0)` is field-local and doesn't clobber the var passed to `compute_context_pressure`) — easy place to get an off-by-cache-read bug, and it's right. This also matches the Anthropic provider, where the API's `input_tokens` is already the fresh (neither-read-nor-written) portion. The chat path subtracts both `cache_read` and `cache_creation`; the Responses paths subtract only `cache_read` — correct, since OpenAI-native Responses has no cache-write concept and doesn't emit that field.

**2. Mock/None coercion is safe.** `_usage_from_openai`'s `_int()` guards every numeric pull with `isinstance(int)`, so a `MagicMock` attribute or a gateway-returned `null` coerces to `0` rather than propagating a mock/None into arithmetic; `usage_obj is None` short-circuits to all-zeroes. The inline Responses sites apply the same `isinstance(int)` guard to `cached_tokens`. `test_mock_or_null_fields_coerce_to_zero` and `test_missing_usage_is_zeroes` cover both.

**3. No behavior change when cache fields are absent.** With no `prompt_tokens_details` and no `cache_creation_input_tokens`, `cache_read`/`cache_creation` are `0`, so `input_tokens = max(prompt_total - 0 - 0, 0) = prompt_total` and pressure is unchanged from the old `prompt_tokens`-based math — i.e. identical to pre-caching behavior for plain OpenAI responses. `test_plain_usage_without_cache_fields` locks this in. The `max(…, 0)` floor is a nice guard against a gateway reporting cached+creation > prompt_total.

### One note (non-blocking)

The chat-completions mapping assumes the gateway reports `prompt_tokens` as the *grand total* (fresh + read + creation) — the test encodes exactly that (140k = 8k + 130k + 2k). That's the right reading of OpenAI's semantics and matches the named gateway's contract; just flagging it as the one external assumption the fresh-portion math rests on, in case a future gateway reports `prompt_tokens` excluding cache-creation.

Field semantics are now uniform across Anthropic, chat-completions, and Responses paths. Good consolidation.

hazemahmedx0:
Closing: superseded by ENG-648 Phase 1 (prompt caching pending the gateway design decision). Keeping the branch as reference.

PR #241, opened 2026-07-08
gdrive connector with google picker api

martyna-mindsdb:
Linear ticket: https://linear.app/mindsdb/issue/ENG-687/google-drive-google-picker-api
___

Demo video: https://www.loom.com/share/6c689e3708664779aa13900ae20e4e49
___

Adds Google Drive Picker guidance to `build_chat_session`—the function backing Anton's own dispatch/CLI runtime—so that files a user explicitly granted access to via the Picker are surfaced to the agent by name. A plain `files.list()`/`files.search()` call won't return them because the `drive.file` scope only covers files created by the app itself.

Since this logic runs on every chat turn and touches every connection in the vault, it also hardens the surrounding loop against malformed records and partial failures so one bad connection can't prevent guidance from being generated for every other connection.

| File | What & why |
|------|------------|
| `anton/core/runtime.py` | Added picked-files extraction and an **IMPORTANT** guidance block telling the agent which files it can access directly via `files.get()` (mirrors `cowork-server`'s `harness.py`, used by a different, non-`cowork-server` call path). Moved the per-connection `try/except` inside the loop so one bad record doesn't abort processing for subsequent connections. Initialized `engine` and `name` before the `try` block so the exception handler's own logging can't fail on malformed connection records. Validated picked-file data after JSON parsing by dropping non-dictionary entries and entries missing an `id`, preventing malformed data from crashing `f.get(...)` or embedding a literal `"None"` file ID in the prompt. Decoupled the guidance block from the `auth_type == "oauth"` heuristic, since gating the entire block behind that fragile check could silently suppress the guidance. Reads the renamed `_picked_files` field while maintaining backward compatibility with the legacy `picked_files` key. |
| `uv.lock` | Lockfile bump. |

### DIFF
--- a/anton/core/runtime.py
+++ b/anton/core/runtime.py
@@ -173,7 +173,10 @@ async def build_chat_session(
         episodic=episodic,
         system_prompt_context=SystemPromptContext(
             runtime_context=build_runtime_context(settings),
-            suffix=system_prompt_suffix,
+            # SystemPromptContext.suffix is typed str (default ""), not
+            # Optional — pass "" rather than None when no suffix was given,
+            # or ChatSystemPromptBuilder.build()'s suffix.strip() crashes.
+            suffix=system_prompt_suffix or "",
         ),
         output_dir=str(output_dir),
         workspace=workspace,
--- a/anton/core/runtime.py
+++ b/anton/core/runtime.py
@@ -14,6 +14,7 @@ Public API:
 """
 from __future__ import annotations
 
+import json
 import logging
 import os
 from pathlib import Path
@@ -142,6 +143,13 @@ async def build_chat_session(
 
     data_vault = LocalDataVault() if LocalDataVault is not None else None
     google_drive_oauth_connected = False
+    # {connection_name: [{"id": ..., "name": ...}, ...]} — files the user
+    # explicitly granted access to via the Google Picker. drive.file scope
+    # only covers files the app created itself, so without calling these
+    # out by name the agent has no way to know they're reachable at all;
+    # it isn't enough to just leave PICKED_FILES sitting in the env like
+    # any other DS_ credential field.
+    google_drive_picked_files: dict[str, list[dict]] = {}
     if data_vault is not None:
         try:
             for conn in data_vault.list_connections():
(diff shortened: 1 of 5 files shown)


anton/core/runtime.py
```diff
@@ -154,16 +176,74 @@ async def build_chat_session(
                     fields = data_vault.load(engine, name) or {}
                     if fields.get("auth_type") == "oauth":
                         google_drive_oauth_connected = True
-        except Exception:
-            logger.debug("Could not inject Anton data vault env", exc_info=True)
+                    # Fall back to the pre-rename key so a vault record
+                    # written before the _picked_files rename doesn't
+                    # silently lose its guidance.
+                    raw_picked = fields.get("_picked_files") or fields.get("picked_files")
```
ZoranPandovski: This renaming happend in this PR so better to keep one name

anton/core/runtime.py
```diff
@@ -154,16 +176,74 @@ async def build_chat_session(
...
+                            google_drive_picked_files[name] = files
+            except Exception:
+                logger.debug(
+                    "Could not inject Anton data vault env for %s/%s",
+                    engine, name, exc_info=True,
+                )
 
     integration_guidance = ""
-    if google_drive_oauth_connected:
+    # Also shown whenever picked_files exist even if google_drive_oauth_connected
```
ZoranPandovski: Maybe avoid a lot of comments since they are from the llm reviewer.

anton/core/runtime.py
```diff
@@ -154,16 +176,74 @@ async def build_chat_session(
...
+            f"- {f.get('name', 'untitled')} (id: {f.get('id')}, connection: {conn_name})"
+            for conn_name, files in google_drive_picked_files.items()
+            for f in files
+        ]
+        # Deliberately imperative and structurally separate from the
+        # paragraph above — a soft prose mention was tested and the
+        # agent silently dropped these files whenever a user asked to
+        # "list" Drive files, because it just reported files.list()'s
+        # output verbatim without cross-referencing earlier context.
+        integration_guidance += (
```
ZoranPandovski: When Anton starts a chat, it builds a system prompt that tells the model things like "here are the data sources you're connected to." The codebase already has one function whose job is exactly that: build_datasource_context() in [datasources.py](vscode-webview://1838v6npjth1igil5advm58si0ia3h3v453v35l8svpsk75lc5mu/anton/utils/datasources.py). I guess we need to move this there

ZoranPandovski (changes requested):
I've added  few minor comments

ZoranPandovski (approved):

PR #243, opened 2026-07-09
[ENG-688] Scheduled task fixes

tino097:
https://linear.app/mindsdb/issue/ENG-688/scheduled-task-slack-spam-incident-four-small-verified-fixes 

https://github.com/mindsdb/cowork/pull/378
https://github.com/mindsdb/cowork-server/pull/181

### DIFF
--- a/anton/utils/datasources.py
+++ b/anton/utils/datasources.py
@@ -7,7 +7,12 @@ import yaml
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from anton.core.datasources.data_vault import DataVault, LocalDataVault, _slug_env_prefix
+from anton.core.datasources.data_vault import (
+    DataVault,
+    LocalDataVault,
+    _slug_env_prefix,
+    is_secret_key,
+)
 from anton.core.datasources.datasource_registry import DatasourceRegistry, _YAML_BLOCK_RE
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


pnewsam:
Reviewed with cowork-server #181 and cowork #378. This one is clean — `_register_unregistered_connection_vars` correctly closes the gap where custom/connector-spec engines fell into the conservative unknown-`DS_*` scrub and leaked non-secret values like `base_url` as `[DS_...]` markers. Registering known-but-not-secret field names so `scrub_credentials` leaves them readable is the right fix, and pairing it with `restore_namespaced_env` in the harness (which also injects env + clears stale vars) is a good consolidation. Tests cover both the `secure_keys` and name-heuristic classification paths.

A couple of minor notes (non-blocking):

- **`secure_keys` casing.** `is_secret = field_name in secure_keys` is an exact match, while the env key uses `field_name.upper()`. If a stored record's `secure_keys` ever differs in case from the field name (e.g. `["Token"]` vs field `token`), the field would be classified non-secret and its value would stay readable. Probably can't happen given how records are written, but a case-insensitive membership check would be defensive.
- **Empty `secure_keys`.** `secure_keys is not None` means an explicit `[]` disables the `looks_secret` fallback entirely — every field is treated non-secret. That's correct if `[]` truly means "no secrets", but worth confirming a connector-spec save never stores `[]` to mean "unclassified".
- The `rich >=13.0,<15` pin in `uv.lock` — assume this is a known incompatibility with rich 15; a one-line note on why would help future bumps.

pnewsam (approved):
Approving after minor comment

PR #249, opened 2026-07-14
ENG-664: expose compaction result for host persistence + skip low-yield re-summarization

ea-rus:
## Summary
- Adds `ChatSession.last_compaction` — a read-only result of the most recent  history compaction this session ran (`{"summary": str, "covered_through": int}`   or `None`). `summary` is the exact compacted message content (including the `_COMPACTED_MARKER` prefix) so a host can seed it back verbatim next turn and  have `_summarize_history` recognize + extend it in place. `covered_through` is how many of the session's seed (`initial_history`) messages are now folded into it, letting a host map that count onto its own message list to
  find the cutoff.
- `_summarize_history` now skips the LLM round-trip when there's little new material to fold in (excluding any already-carried-forward prior summary from the measurement) — avoids paying for a summarization call that would barely extend an already-large summary.
- No change to the compaction algorithm itself, and no new required constructor parameters — existing hosts are unaffected.

This is the anton-side half of ENG-664 (cowork-server persists `last_compaction` across turns instead of resending full history every turn). The once-per-turn compaction guard (`_compacted_this_turn`) already existed and needed no change.

## Test plan
- [x] New `tests/test_history_compaction.py`: `last_compaction` is `None` before any compaction, populated correctly after one, and the low-yield-skip guard (including the "prior summary doesn't count as new material" case).

Fixes https://linear.app/mindsdb/issue/ENG-664/conversations-resend-the-full-history-every-turn-and-throw-away-the

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1258,6 +1258,8 @@ class ChatSession:
                 max_tokens=2048,
             )
             summary = summary_response.content or "(summary unavailable)"
+            # Record the compaction ONLY on success.
+            self._last_compacted_count = compacted_count
         except Exception:
             # If summarization fails, just do a simple truncation
             summary = f"(Earlier conversation with {len(old_turns)} turns — summarization failed)"
@@ -1275,7 +1277,6 @@ class ChatSession:
             f"{summary}"
         )
         summary_msg = {"role": "user", "content": summary_body}
-        self._last_compacted_count = compacted_count
 
         # If the recent portion starts with a user message, insert a minimal
         # assistant separator to avoid consecutive user messages (API error).
@@ -1476,6 +1477,12 @@ class ChatSession:
             self._history = [placeholder, *tail]
         else:
             self._history = [placeholder, separator, *tail]
+
+        # A hard truncate discards the compacted summary — history[0] is now
+        # the truncation placeholder, not the summary. Clear the compaction
+        # record so `last_compaction` reports None (keep-full-history) rather
+        # than letting a host persist the placeholder as the durable summary.
+        self._last_compacted_count = None
 
     async def plan_with_recovery(
         self,
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -353,6 +353,11 @@ class ChatSession:
         self._history: list[dict] = (
             list(config.initial_history) if config.initial_history else []
         )
+        # Seed length + how many leading history messages the last compaction
+        # folded into its summary — lets a host map the compaction back onto
+        # its own message list (see `last_compaction`).
+        self._seed_len = len(self._history)
+        self._last_compacted_count: int | None = None
         self._pending_memory_confirmations: list = []
         self._turn_count = (
             sum(1 for m in self._history if is_user_turn(m))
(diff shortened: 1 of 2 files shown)


Dav1dF:
Read this alongside the prompt-caching work (the gateway's cache alignment from mindsdb/mindshub_inference — ENG-601/ENG-586) to check for a conflict, and wanted to note the opposite on the record: this design is exactly what prefix caching needs.

- Replaying the persisted summary **verbatim** keeps the seed byte-stable between compactions, and the tail is append-only — so a cached prefix survives across turns.
- The low-yield re-summarization skip is also the cache-friendly behavior: every re-summarization rewrites the front of history and resets the cached prefix, so avoiding marginal ones saves twice.

One expectation to write down somewhere (here is fine): when cache alignment turns on in prod, cache-hit dips will correlate with compaction events. That is inherent to any history trimming, and `last_compaction` / the compaction event give us the attribution — no change needed in this PR, just flagging so nobody reads the dips as a caching bug later.

Linear: [ENG-664](https://linear.app/mindsdb/issue/ENG-664); caching context on [ENG-586](https://linear.app/mindsdb/issue/ENG-586).

anton/core/session.py
```diff
@@ -1225,6 +1275,7 @@ async def _summarize_history(self) -> None:
             f"{summary}"
         )
         summary_msg = {"role": "user", "content": summary_body}
+        self._last_compacted_count = compacted_count
```
tino097: with claude help:
 This line runs even when llm.summarize failed: the except above `substitutes summary = "(Earlier conversation with N turns — summarization failed)"` and falls through, so `last_compaction` ends up reporting a marker-wrapped, fact-free placeholder as a valid compaction (covered_through = N).

Pre-PR that failure mode was self-healing — the placeholder died with the per-turn session and the next turn rebuilt full history. But the whole point of this property is that the host persists it: paired with `cowork-server #192`, a single transient summarize error (rate limit, network blip) permanently replaces the covered messages with the placeholder in every future turn. It can't recover, either — the placeholder carries `_COMPACTED_MARKER`, so the next compaction "updates it in place" instead of starting over.
```

Repro:
llm.summarize = AsyncMock(side_effect=RuntimeError("blip"))
session = ChatSession(ChatSessionConfig(llm_client=llm, initial_history=ten_messages))
await session._summarize_history()
session.last_compaction
# {'summary': '[COMPACTED CONTEXT — REFERENCE ONLY]\n…(summarization failed)', 'covered_through': 6}

```
Fix: record the compaction only on success — move this assignment into the try, right after summary = summary_response.content or .... The in-memory degraded compaction can stay as-is; the host then just falls back to full history next turn. Suggested regression test: summarize raises → last_compaction is None.

anton/core/session.py
```diff
@@ -602,6 +607,23 @@ def _acc_has_similar(rule: str) -> bool:
     def history(self) -> list[dict]:
         return self._history
 
+    @property
+    def last_compaction(self) -> dict | None:
```
tino097, line 641: "summary": `self._history[0]["content"]` is read lazily at access time, but by then `history[0]` isn't guaranteed to still be the summary: `hard_truncate_history` replaces it with `"[Earlier conversation was truncated due to persistent context overflow.]"`, and the recovery ladder hits exactly compact-then-truncate in one turn on a double overflow `(plan_stream_with_recovery: _summarize_history() at L1574, then hard_truncate_history() at L1587)`. `_last_compacted_count` is still set from the earlier compaction, so a host reading this at turn end persists the truncation placeholder as the conversation's durable summary, with covered_through messages dropped from all future replays. This placeholder doesn't even carry `_COMPACTED_MARKER`, so later compactions treat it as ordinary user text.

```
Repro:
await session._summarize_history()   # real summary, count set
session.hard_truncate_history()
session.last_compaction
# {'summary': '[Earlier conversation was truncated due to persistent context overflow.]', 'covered_through': 6}
```
Fix: reset `self._last_compacted_count = None` inside `hard_truncate_history` — after a hard truncate the compaction result no longer describes `history[0]`, and `None` is already the host's keep-full-history path. 
Optionally the property could also verify `history[0] is a str` starting with `_COMPACTED_MARKER` before reporting (defense against any future history rewrite) — but note that wouldn't catch the failed-summarize case in my other comment, since that placeholder doescarry the marker. Suggested regression test: `compact → hard_truncate_history() → last_compaction is None`.

tino097 (changes requested):
Claude finds issues

ea-rus:
@tino097 thank you! fixed

tino097 (approved):

tino097:
Thanks @ea-rus !!

PR #255, opened 2026-07-16
ENG-716: add WAITING verdict to the completion verifier

alecantu7:
## What & why

The completion-verifier loop (`anton/core/session.py`) classified **every tool-using turn that ended by asking the user a question** as `INCOMPLETE` and force-continued it — so the agent answered its own questions (or fabricated) instead of waiting. Root cause: the verdict vocabulary had no "waiting for the user" state, so the verifier overrode the agent's correct instinct to stop and ask.

## Evidence (three converging confirmations)

1. **The original ticket screenshot, traced to its prod Langfuse session (Jul 10):** verifier `INCOMPLETE` on the question-ending turn → forced continuation → the exact screenshot text (*"The verifier note is noted … I'll make sensible default assumptions and draft the skill now"*). A later verdict in the same session read verbatim: **"The assistant is *waiting for user feedback* … has not been completed."** — the verifier naming the missing state itself.
2. **Reproduced on a fresh, memory-empty staging instance** (verified clean via Langfuse) — so it's structural, not a memory artifact.
3. **Cost, confirmed in prod:** the verifier fired a full-conversation call on the expensive planning model on every tool-using turn.

## What changed

- **`WAITING` verdict** — a turn ending in a genuine question to the user (or a reasoned refusal) is a valid stop, not unfinished work.
- **Cheap model + structured output** — the verdict now runs via `generate_object_code` (coding model) with a typed `_VerifierVerdict` over a compact **request + reply + per-tool-outcome** view, instead of a full-transcript call on the planning model parsed from free-text `STATUS:` lines. Cheaper and structurally reliable; the tool-outcome log preserves the verifier's ability to catch "claimed done but a tool errored."
- **`verify_min_tool_rounds`** (default `1` = current behavior) to later skip trivial single-tool-round turns once verdict logs confirm they're rarely `INCOMPLETE`.
- **Leak fix** — internal `SYSTEM:` continuation/diagnosis injections now carry a non-disclosure clause so they aren't echoed to the user (*"the verifier note is noted"*).
- **Verdict logging** for measurability.

## Why it's low-risk

Genuine multi-step work completes inside the **inner** tool loop, where the verifier returns `COMPLETE` once with no continuation (verified). The outer continuation loop — the only thing `WAITING` touches — fires only *after* a text-only response, so legitimate completion barely uses it. On a verifier error we fail-safe to `COMPLETE` (failing to `INCOMPLETE` would risk re-introducing the over-continuation this fixes).

## Testing

- Full suite: **1194 passed**, 17 skipped; only pre-existing scratchpad-exec env failures remain (identical on clean `staging`).
- **New** `test_waiting_verdict_stops_without_continuation` guards the fix (WAITING stops, no continuation injection, tool-outcome log reaches the verifier).
- Broader testing caught and fixed a regression: the E2E stub spoke the old free-text verifier protocol; updated it to emit structured `_VerifierVerdict` tool calls (see commit `dbdcdf4`).

## Self-review trail

Ran an adversarial self-review; findings + resolutions are posted as a comment below (2 fixed, 1 kept-intentional, 1 low/noted). The `AttributeError`-on-init risk was checked and cleared (`AntonSettings` inherits `CoreSettings`, so the new setting flows to the cowork path).

## Security

Reviewed the diff: the reduced view sends **less** data to the LLM than before; the verdict is a constrained enum (no injected control flow); the log line carries only status/counters/reason (no credentials or reply content). No new endpoints, auth, or deserialization.

## Merge order (not stacked)

anton merges **first**; a follow-up cowork-server `anton-agent` bump ships it to the desktop path. Stated here so the two don't get stacked.

Closes ENG-716.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1823,11 +1823,17 @@ class ChatSession:
         # task isn't actually done yet.
         continuation = 0
         _max_rounds_hit = False
+        import logging as _logging
+        _verifier_log = _logging.getLogger(__name__)
 
         while True:  # Completion verification loop
             tool_round = 0
             error_streak: dict[str, int] = {}
             resilience_nudged: set[str] = set()
+            # Compact per-tool outcome log for this iteration, so the completion
+            # verifier can cross-check the assistant's claims against what tools
+            # actually did without receiving the full transcript (ENG-716).
+            tool_outcomes: list[str] = []
 
             while llm_response.tool_calls:
                 tool_round += 1
@@ -2090,6 +2096,7 @@ class ChatSession:
                             severity=1,
                             round_idx=tool_round,
                         )
+                        tool_outcomes.append(f"{tc.name}=ok")
                         tool_results.append(
                             {
                                 "type": "tool_result",
@@ -2132,6 +2139,7 @@ class ChatSession:
                         severity=5 if _failed else 1,
                         round_idx=tool_round,
                     )
+                    tool_outcomes.append(f"{tc.name}={'error' if _failed else 'ok'}")
                     tool_results.append(
                         {
                             "type": "tool_result",
@@ -2263,6 +2271,7 @@ class ChatSession:
             # needs for the verdict, keeps the call cheap, and — being plain text
             # — avoids any tool_use/tool_result pairing constraints (ENG-716).
             request_text = (user_message or "").strip() or "(see the assistant's message)"
+            tools_summary = ", ".join(tool_outcomes) if tool_outcomes else "none"
             verify_messages = [
                 {"role": "user", "content": f"USER'S REQUEST:\n{request_text}"},
                 {
@@ -2272,8 +2281,11 @@ class ChatSession:
                 {
                     "role": "user",
                     "content": (
-                        "Based on the assistant's latest message above, which status "
-                        "applies to the user's request?"
+                        f"TOOLS RUN THIS TURN (name=outcome): {tools_summary}\n\n"
+                        "Based on the assistant's latest message and the tool outcomes "
+                        "above, which status applies to the user's request? If a tool the "
+                        "task depended on returned an error but the assistant implied "
+                        "success, that is INCOMPLETE or STUCK — not COMPLETE."
                     ),
                 },
             ]
@@ -2297,8 +2309,7 @@ class ChatSession:
                 # than forcing a continuation the user never asked for.
                 status, reason = "COMPLETE", "verifier unavailable"
 
-            import logging as _logging
-            _logging.getLogger(__name__).info(
+            _verifier_log.info(
                 "completion-verifier verdict=%s continuation=%d/%d tool_rounds=%d reason=%s",
                 status, continuation, self._max_continuations, tool_round, reason,
             )
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -145,47 +145,83 @@ class _VerifierVerdict(BaseModel):
     reason: str = Field(description="One brief sentence explaining the verdict.")
 
 
+def _render_tool_result_content(content, cap: int) -> str:
+    """Render a tool_result's content as bounded plain text.
+
+    Never serializes raw payloads: a multimodal result (e.g. read_image) can
+    carry megabytes of base64, so we keep only text blocks and mark images with
+    a placeholder rather than ``json.dumps``-ing the whole thing (ENG-716).
+    """
+    if isinstance(content, str):
+        return content[:cap] or "(empty result)"
+    if isinstance(content, list):
+        parts: list[str] = []
+        for block in content:
+            if not isinstance(block, dict):
+                continue
+            if block.get("type") == "text":
+                parts.append((block.get("text") or "").strip())
+            elif block.get("type") in ("image", "image_url"):
+                parts.append("[image]")
+        return (" ".join(p for p in parts if p)[:cap]) or "[non-text result]"
+    return str(content)[:cap]
+
+
 def _render_verify_transcript(
     history: list[dict],
     *,
-    max_messages: int = 16,
+    max_convo: int = 10,
+    max_tool: int = 12,
     tool_cap: int = 400,
     text_cap: int = 2000,
 ) -> str:
     """Render a compact, text-only view of the recent conversation for the
     completion verifier.
 
-    Gives the verifier enough to judge reliably — prior turns (so referential
-    follow-ups keep their task context) and truncated tool-result *content* (so
-    a claimed-successful tool can be cross-checked, not just its ok/error flag)
-    — without shipping the full raw transcript or tripping the provider's
-    tool_use/tool_result pairing constraints (ENG-716). Internal ``SYSTEM:``
-    injections are omitted; they are directives, not conversation.
+    Budgets the conversational thread and the tool activity *separately* so a
+    voluminous tool loop can't crowd out the user/assistant turns the verifier
+    needs to resolve referential requests ("do the same for the other file"):
+    the most recent ``max_convo`` user/assistant text turns plus the most recent
+    ``max_tool`` tool events, merged back into chronological order. Speaker is
+    taken from the message role (list-based *user* content — images/files — is
+    labelled USER, not ASSISTANT); multimodal blocks are rendered block-by-block
+    (text kept, images as a placeholder, never raw base64); internal ``SYSTEM:``
+    injections are dropped before budgeting so they don't consume slots. Keeps
+    the call cheap and free of tool_use/tool_result pairing constraints
+    (ENG-716).
     """
-    lines: list[str] = []
-    for msg in history[-max_messages:]:
+    convo: list[tuple[int, str]] = []   # (orig_index, line) — user/assistant text
+    tools: list[tuple[int, str]] = []   # (orig_index, line) — tool activity
+
+    for i, msg in enumerate(history):
         role = msg.get("role")
         content = msg.get("content")
+        speaker = "USER" if role == "user" else "ASSISTANT"
         if isinstance(content, str):
             text = content.strip()
             if not text or (role == "user" and text.startswith("SYSTEM:")):
                 continue
-            speaker = "USER" if role == "user" else "ASSISTANT"
-            lines.append(f"{speaker}: {text[:text_cap]}")
+            convo.append((i, f"{speaker}: {text[:text_cap]}"))
         elif isinstance(content, list):
             for block in content:
                 if not isinstance(block, dict):
                     continue
                 btype = block.get("type")
-                if btype == "text" and (block.get("text") or "").strip():
-                    lines.append(f"ASSISTANT: {block['text'][:text_cap]}")
+                if btype == "text":
+                    text = (block.get("text") or "").strip()
+                    if text:
+                        convo.append((i, f"{speaker}: {text[:text_cap]}"))
+                elif btype in ("image", "image_url"):
+                    convo.append((i, f"{speaker}: [image]"))
                 elif btype == "tool_use":
-                    lines.append(f"ASSISTANT called tool: {block.get('name')}")
+                    tools.append((i, f"ASSISTANT called tool: {block.get('name')}"))
                 elif btype == "tool_result":
-                    c = block.get("content")
-                    text = c if isinstance(c, str) else json.dumps(c)
-                    lines.append(f"TOOL RESULT: {text[:tool_cap]}")
-    return "\n".join(lines) or "(no conversation)"
+                    rendered = _render_tool_result_content(block.get("content"), tool_cap)
+                    tools.append((i, f"TOOL RESULT: {rendered}"))
+
+    kept = convo[-max_convo:] + tools[-max_tool:]
+    kept.sort(key=lambda entry: entry[0])
+    return "\n".join(line for _, line in kept) or "(no conversation)"
 
 
 @dataclass
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2309,13 +2309,18 @@ class ChatSession:
             # and free of tool_use/tool_result pairing constraints (ENG-716). The
             # assistant's latest reply is already in history (appended above).
             transcript = _render_verify_transcript(self._history)
+            # Always state the current request explicitly: a long tool-heavy turn
+            # can push the turn's opening user message out of the transcript window,
+            # and the request is the anchor for the whole judgment (ENG-716).
+            request = (user_message or "").strip()
+            request_header = f"USER'S CURRENT REQUEST: {request}\n\n" if request else ""
             verify_messages = [
                 {
                     "role": "user",
                     "content": (
                         "Assess the conversation below (tool results are truncated) and "
                         "decide the status of the USER's most recent request.\n\n"
-                        f"{transcript}\n\n"
+                        f"{request_header}{transcript}\n\n"
                         "Which status applies? A tool the task depended on that returned "
                         "an error — or returned no usable data while the assistant implied "
                         "success — means INCOMPLETE or STUCK, not COMPLETE."
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -6,8 +6,10 @@ from dataclasses import asdict, dataclass, field, replace
 from datetime import datetime
 import json
 import re
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Literal
 import os
+
+from pydantic import BaseModel, Field
 
 from anton.core.backends.base import Cell, ScratchpadRuntimeFactory
 from anton.core.backends.local import local_scratchpad_runtime_factory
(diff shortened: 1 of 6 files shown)


alecantu7:
## Self-review (adversarial pass before requesting review)

Ran an 8-angle review over the diff + enclosing code. Findings and resolutions:

| # | Finding | Severity | Resolution |
|---|---|---|---|
| 1 | Reduced verify view was **tool-blind** — a turn where a tool errored but the assistant claimed success would be judged `COMPLETE`, defeating the `STUCK` backstop | High | ✅ **Fixed** (`92f47b1`) — track a compact per-tool outcome log (`name=ok/error`) and feed it to the verifier with explicit guidance |
| 2 | On verifier error we fail-safe to `COMPLETE`, losing the `plan_with_recovery` retry — a transient model blip ends a genuinely-unfinished turn | Medium | ⚖️ **Kept intentional** — failing to `INCOMPLETE` would risk re-introducing the over-continuation this PR removes; a transient stop is user-recoverable |
| 3 | `import logging` ran inside the verification loop | Low | ✅ **Fixed** (`92f47b1`) — hoisted to one logger per turn |
| 4 | Tool-driving tests reach the verifier via auto-mock verdict | Low | 📝 **Noted** — non-blocking; they pass and don't assert verifier behavior |

Also verified (not a bug): the new `verify_min_tool_rounds` setting flows to the cowork path because `AntonSettings` inherits `CoreSettings`, so `s.verify_min_tool_rounds` resolves whether `config.settings` is `CoreSettings` or `AntonSettings`.

anton/core/session.py
```diff
@@ -2230,55 +2265,72 @@ async def _stream_and_handle_tools(
...
-                            "has been fully completed based on the conversation above."
-                        ),
-                    }
-                ]
+            # Ask the cheap coding model to self-assess completion, using a
+            # compact text-only view (the user's request + the assistant's latest
+            # message) rather than the full transcript. That is all the verifier
+            # needs for the verdict, keeps the call cheap, and — being plain text
+            # — avoids any tool_use/tool_result pairing constraints (ENG-716).
+            request_text = (user_message or "").strip() or "(see the assistant's message)"
```
pnewsam: **Major:** Preserve the task's conversational context here.

`user_message` is only the latest raw turn input. For common follow-ups such as "yes, do that" or "use the second option", the verifier receives no antecedent but is still asked whether the request is complete. The previous full-history verifier could resolve those references; this compact view can produce false `COMPLETE` or `WAITING` verdicts.

Please include a bounded recent conversation or a compact resolved-task summary rather than only the current message, and add coverage for a referential follow-up.

anton/core/session.py
```diff
@@ -2230,55 +2265,72 @@ async def _stream_and_handle_tools(
...
+            tools_summary = ", ".join(tool_outcomes) if tool_outcomes else "none"
```
pnewsam: **Major:** `name=ok/error` is not enough evidence to cross-check the assistant's completion claim.

A successful call may still return no rows, an empty artifact, or a result that does not satisfy the request; all of those become `tool=ok`, so the verifier has no way to distinguish them from a correct result. This removes a capability the prior full-transcript verifier had.

Please include a scrubbed, bounded result summary (or another structured semantic outcome) for each tool. The E2E assertion should also check the actual outcome/summary rather than only the `TOOLS RUN THIS TURN` marker.

pnewsam (changes requested):
## Code Review

**Verdict: REQUEST_CHANGES**

The `WAITING` state and structured verdict are sensible improvements, but the reduced verification view removes evidence required to assess completion reliably: prior conversational context and successful tool-result content. This can allow unfinished or incorrectly completed tasks to be classified as complete.

### Summary of findings

| Severity | File | Issue |
|----------|------|-------|
| Major | `anton/core/session.py:2273` | The verifier receives only the latest raw user message, so referential follow-ups lose the task context. |
| Major | `anton/core/session.py:2274` | `name=ok/error` does not contain enough result evidence to cross-check successful tool claims. |

### What's working well

The explicit `WAITING` state directly addresses the forced-continuation bug, the structured verdict removes fragile text parsing, and the new E2E coverage verifies that a `WAITING` verdict stops the loop. CI and CodeQL are green.

---
_Reviewed by Codex · `alejandrocantu/eng-716-completion-verifier-waiting-verdict` → `staging` · 2026-07-16_

alecantu7:
## Addressed both Major findings (`2e44a3f`)

Thanks — both were right; the reduced view was over-reduced. Replaced it with an **evidence-complete but bounded** view via a new `_render_verify_transcript`, instead of the request+reply+`ok/error` summary.

**Finding 1 — referential context (`session.py:2273`).** The verifier now receives a compact text render of the **recent conversation** (prior turns included, last ~16 messages), so follow-ups like *"now do the same for the other file"* keep their task context.

**Finding 2 — tool evidence (`session.py:2274`).** Dropped the coarse `name=ok/error` flag; the view now carries **truncated tool-result content** (each result capped at ~400 chars). A tool that returned 200-but-empty, or errored while the assistant implied success, is now visible to the verifier for cross-checking.

It stays cheap: history is text-rendered (no full raw transcript), tool output is truncated, messages are capped, and it sidesteps `tool_use`/`tool_result` pairing constraints. Internal `SYSTEM:` injections are omitted (they're directives, not conversation).

**Tests added/updated:**
- `tests/test_verify_transcript.py` — context retention, tool-result evidence, `SYSTEM:` omission, truncation, empty-history safety.
- e2e `test_waiting_verdict_stops_without_continuation` now asserts truncated tool-result evidence reaches the verifier.
- Full suite: 1200 passed (only pre-existing scratchpad-exec env failures remain).

Ready for another look.

anton/core/session.py
```diff
@@ -120,6 +122,72 @@ def _scrub_user_input(user_input: str | list[dict]) -> str | list[dict]:
...
+
+    Gives the verifier enough to judge reliably — prior turns (so referential
+    follow-ups keep their task context) and truncated tool-result *content* (so
+    a claimed-successful tool can be cross-checked, not just its ok/error flag)
+    — without shipping the full raw transcript or tripping the provider's
+    tool_use/tool_result pairing constraints (ENG-716). Internal ``SYSTEM:``
+    injections are omitted; they are directives, not conversation.
+    """
+    lines: list[str] = []
+    for msg in history[-max_messages:]:
```
pnewsam: **Major:** Preserve the initiating user request independently of this tail window.

A single turn with eight tool rounds already creates enough assistant tool-use and user tool-result messages for `history[-16:]` to drop the original request; the configured limit permits up to 25 rounds. In that case the rendered transcript contains tool activity and the final assistant reply but no `USER:` request, so the verifier cannot assess completion. Filtering after slicing also lets omitted internal messages consume the budget.

Please anchor the latest non-`SYSTEM:` textual user request (and any context needed to resolve it), then fill the remaining budget from the tail. Add a regression test whose current turn exceeds eight tool rounds.

anton/core/session.py
```diff
@@ -120,6 +122,72 @@ def _scrub_user_input(user_input: str | list[dict]) -> str | list[dict]:
...
+                continue
+            speaker = "USER" if role == "user" else "ASSISTANT"
+            lines.append(f"{speaker}: {text[:text_cap]}")
+        elif isinstance(content, list):
+            for block in content:
+                if not isinstance(block, dict):
+                    continue
+                btype = block.get("type")
+                if btype == "text" and (block.get("text") or "").strip():
+                    lines.append(f"ASSISTANT: {block['text'][:text_cap]}")
```
pnewsam: **Major:** Derive the speaker from the parent message role here.

`turn_stream` supports list-based user content for image/file inputs. Their text blocks currently render as `ASSISTANT`, so a multimodal request such as "describe this image and save a report" yields two assistant lines and no current `USER` request. Mixed user content retained by history truncation has the same problem.

Use `USER` when `role == "user"` and add a multimodal-user regression test.

anton/core/session.py
```diff
@@ -120,6 +122,72 @@ def _scrub_user_input(user_input: str | list[dict]) -> str | list[dict]:
...
+                elif btype == "tool_use":
+                    lines.append(f"ASSISTANT called tool: {block.get('name')}")
+                elif btype == "tool_result":
+                    c = block.get("content")
+                    text = c if isinstance(c, str) else json.dumps(c)
```
pnewsam: **Minor:** Avoid serializing the entire multimodal payload before truncating it.

`read_image` returns an image block containing up to roughly 5 MB of base64 followed by a useful text summary. `json.dumps(c)` materializes the whole payload, then the first 400 characters retain only JSON/base64 and discard that summary.

Render multimodal content block-by-block: keep bounded text blocks and a compact image placeholder, without copying encoded image data into the verifier transcript.

pnewsam (changes requested):
## Code Review — Follow-up

**Verdict: REQUEST_CHANGES**

The original review findings are addressed for short, string-only histories, but the new transcript renderer still loses the active request in two supported paths: longer tool loops and multimodal user messages. These can leave the verifier without any correctly labeled current user request. There is also a smaller performance/evidence issue when rendering multimodal tool results.

### Summary of findings

| Severity | File | Issue |
|----------|------|-------|
| Major | `anton/core/session.py:166` | The fixed 16-message tail can discard the initiating request during a supported multi-round tool turn. |
| Major | `anton/core/session.py:181` | Text blocks from multimodal user messages are incorrectly labeled as assistant messages. |
| Minor | `anton/core/session.py:186` | Multimodal results serialize megabytes of base64 before truncation and omit the useful text summary. |

### What's working well

The follow-up does restore prior conversational context and real tool-result evidence for the common string-only path, removes the weaker ok/error heuristic, and adds focused unit coverage. All current CI and CodeQL checks pass.

---
_Reviewed by Codex · `alejandrocantu/eng-716-completion-verifier-waiting-verdict` → `staging` · 2026-07-16_

alecantu7:
## Follow-up self-review of the new view (`4905a26`)

Re-reviewed the evidence-complete transcript view and caught one edge case I introduced: the transcript is the last ~16 messages, so a **long tool-heavy turn** (~7+ tool rounds, 2 messages each) can evict the turn's **opening user request** from the window — leaving the verifier to judge completion without seeing what was asked. Fixed by always prepending the current request (`USER'S CURRENT REQUEST: …`) to the view, independent of the window; the e2e WAITING test now asserts that header reaches the verifier.

Full suite still green bar the pre-existing scratchpad-exec env failures.

anton/core/session.py
```diff
@@ -2230,55 +2302,73 @@ async def _stream_and_handle_tools(
...
+            # referential follow-ups plus truncated tool-result evidence to
+            # cross-check success claims, but far smaller than the raw transcript
+            # and free of tool_use/tool_result pairing constraints (ENG-716). The
+            # assistant's latest reply is already in history (appended above).
+            transcript = _render_verify_transcript(self._history)
+            # Always state the current request explicitly: a long tool-heavy turn
+            # can push the turn's opening user message out of the transcript window,
+            # and the request is the anchor for the whole judgment (ENG-716).
+            request = (user_message or "").strip()
+            request_header = f"USER'S CURRENT REQUEST: {request}\n\n" if request else ""
```
pnewsam, line 2361: **Major:** The literal current-request header does not preserve the antecedent of a referential request.

For example, after a prior turn establishes how `report_q1.csv` was cleaned, the current request may be "now do the same for the other file". Eight tool rounds can evict that prior turn from the 16-message transcript, leaving the verifier with this header but no explanation of what "the same" means. The new E2E assertion checks only that the marker exists, so it does not cover this case.

Please retain a bounded set of relevant preceding user/assistant turns independently of the tool-activity tail, and add a long-turn regression test with a referential current request.

anton/core/session.py
pnewsam: **Major:** This still misattributes text blocks from list-based user messages as assistant output.

The new request header mitigates the latest multimodal request, but prior multimodal user context remains incorrectly labeled. A later request such as "use the same style for the next image" therefore sees the style-setting user instruction as an `ASSISTANT` line, which can corrupt the context the verifier uses to resolve the reference.

Derive the speaker from the parent `role` and add coverage for a referential follow-up to a multimodal user turn.

anton/core/session.py
pnewsam: **Minor:** This still serializes the entire multimodal payload before truncating it.

`read_image` can return roughly 5 MB of base64 followed by a useful text summary. `json.dumps(c)` materializes the full payload, while the retained 400-character prefix contains encoded image data and drops the summary.

Render multimodal content block-by-block: retain bounded text blocks and an image placeholder without copying encoded data into the verifier transcript.

pnewsam (changes requested):
## Code Review — Follow-up 2

**Verdict: REQUEST_CHANGES**

The explicit current-request header fixes long, self-contained requests, but it does not retain the antecedent needed by referential requests after a long tool loop. The multimodal role-labeling and base64-rendering findings also remain unchanged. These leave two supported conversation paths where the verifier receives incorrect or insufficient task context.

### Summary of findings

| Severity | File | Issue |
|----------|------|-------|
| Major | `anton/core/session.py:2316` | A literal current-request header does not preserve prior context needed to resolve referential requests. |
| Major | `anton/core/session.py:181` | Text blocks from multimodal user messages are still rendered as assistant messages. |
| Minor | `anton/core/session.py:186` | Multimodal results still serialize megabytes of base64 before truncation and omit the useful text summary. |

### What's working well

The new header does ensure that self-contained current requests survive long tool-heavy turns, and all current CI and CodeQL checks pass.

---
_Reviewed by Codex · `alejandrocantu/eng-716-completion-verifier-waiting-verdict` → `staging` · 2026-07-16_

alecantu7:
## Addressed the follow-up findings (`959c76a`)

All three are legit — the renderer was too naïve. Reworked `_render_verify_transcript`:

**Major — multimodal user text mislabeled as ASSISTANT (`:181`).** Speaker is now taken from the message `role`, so list-based *user* content (image/file inputs) is labelled `USER`. Covered by `test_multimodal_user_text_labeled_user_not_assistant`.

**Major — referential antecedent lost after long turns (`:166` / `:2316`).** Replaced the flat last-16 window with **separate budgets** for the conversation thread (`max_convo`) and tool activity (`max_tool`), merged back in chronological order. Tool volume can no longer evict prior user/assistant turns, so referential requests keep their antecedent. `SYSTEM:` injections are filtered **before** budgeting so they don't consume slots. Covered by `test_long_tool_turn_keeps_request_and_antecedent` (10 tool rounds + a referential request — both the antecedent and the request survive) and `test_system_injections_do_not_consume_budget`. The explicit current-request header stays as a belt-and-suspenders for self-contained asks.

**Minor — multimodal tool_result base64 (`:186`).** New `_render_tool_result_content` renders block-by-block: bounded text kept, images as an `[image]` placeholder, raw base64 never serialized. Covered by `test_multimodal_tool_result_keeps_summary_drops_base64`.

Full suite: 1204 passed (only the pre-existing scratchpad-exec env failures). Ready for another look.

anton/core/session.py
```diff
@@ -120,6 +122,108 @@ def _scrub_user_input(user_input: str | list[dict]) -> str | list[dict]:
...
+                        convo.append((i, f"{speaker}: {text[:text_cap]}"))
+                elif btype in ("image", "image_url"):
+                    convo.append((i, f"{speaker}: [image]"))
+                elif btype == "tool_use":
+                    tools.append((i, f"ASSISTANT called tool: {block.get('name')}"))
+                elif btype == "tool_result":
+                    rendered = _render_tool_result_content(block.get("content"), tool_cap)
+                    tools.append((i, f"TOOL RESULT: {rendered}"))
+
+    kept = convo[-max_convo:] + tools[-max_tool:]
```
pnewsam, line 231: **Major:** Tool-call preamble text still allows tool volume to evict the conversational thread from this budget.

When the model emits a text block such as "Processing step 4" alongside each `tool_use` (a normal response shape), those blocks are appended to `convo`. Ten tool rounds then fill `max_convo=10`, dropping both the referential request and its antecedent. The explicit request header preserves only "now do the same for the other file", so the verifier still cannot determine what "the same" means.

The new long-turn test does not reproduce this because each assistant tool-call message contains only a `tool_use` block. Please classify text from messages containing `tool_use` as tool activity (or independently reserve actual user turns and their conversational responses), and add preamble text to that regression test.

pnewsam (changes requested):
## Code Review — Follow-up 3

**Verdict: REQUEST_CHANGES**

The multimodal role-labeling and base64-rendering findings are resolved. One Major context-budget issue remains: assistant preamble text emitted alongside tool calls is counted as conversation, so a realistic long tool loop can still evict both a referential request and its antecedent from the transcript.

### Summary of findings

| Severity | File | Issue |
|----------|------|-------|
| Major | `anton/core/session.py:222` | Tool-call preamble text consumes the conversation budget and can evict the conversational thread. |

### What's working well

Multimodal user content now retains the correct role, image payloads are represented without base64 serialization, useful tool-result summaries are preserved, and all current CI and CodeQL checks pass.

---
_Reviewed by Codex · `alejandrocantu/eng-716-completion-verifier-waiting-verdict` → `staging` · 2026-07-16_

alecantu7:
## Addressed the preamble finding (`bcd091e`)

Good catch — this one was real and my earlier long-turn test missed it (its tool-call messages had only `tool_use` blocks, no preamble).

Assistant text emitted **alongside a `tool_use`** is now routed to the **tool** budget, not `convo`. The conversation budget is reserved for real user turns and standalone replies, so a long tool loop's preambles ("Processing step N") can no longer evict the referential request or its antecedent.

The regression test now emits preamble text on every round. I verified it's a genuine guard: simulating the old behavior (preamble → `convo`), both the request *and* the antecedent are evicted after 10 rounds; with the fix, both survive (and the antecedent's answer too).

Full suite: 1204 passed (only the pre-existing scratchpad-exec env failures).

pnewsam (approved):
## Code Review — Final

**Verdict: APPROVE**

All previously identified issues are resolved. The verifier now preserves referential context through long tool-heavy turns, routes tool-call preamble text outside the conversational budget, labels multimodal user content correctly, and renders multimodal tool evidence without serializing base64 payloads.

No actionable issues found. Code looks good to merge.

### Validation

- All 9 focused transcript tests pass.
- The previously failing long-turn reproduction now preserves the request and antecedent.
- CI and CodeQL are green.
- `git diff --check` passes.

---
_Reviewed by Codex · `alejandrocantu/eng-716-completion-verifier-waiting-verdict` → `staging` · 2026-07-16_

PR #257, opened 2026-07-17
Surface provider 401s as typed ProviderAuthError with provider-aware invalid-key copy

ianu82:
## What

Provider HTTP 401s (invalid/missing API key) are now surfaced as a typed `ProviderAuthError` instead of being swallowed into assistant prose, and the 401 message is provider-aware instead of always blaming OpenAI.

## Why

User-reported (2026-07-17): with a missing/invalid MindsHub key, a chat turn rendered this as ordinary assistant text:

> "An unexpected error occurred: Invalid API key — check your OpenAI API key configuration.. Please try again or rephrase your request."

Two defects:

1. **Swallowed auth errors.** The `turn_stream` retry loop in `anton/core/session.py` re-raises `TokenLimitExceeded` / `ModelUnavailableError` so the server can map them to actionable error cards, but a provider-auth `ConnectionError` was auto-retried twice (pointless — a 401 is deterministic) and then wrapped into the generic "An unexpected error occurred: … Please try again or rephrase" fallback prose. cowork-server's `turn_errors.is_auth_error` (which maps auth failures to a reconnect/update-key card) could therefore never fire for streaming turns.
2. **Misleading copy.** `_raise_for_status_error` in `anton/core/llm/openai.py` hardcoded "check your OpenAI API key configuration" even when `OpenAIProvider` was serving MindsHub (`api.mindshub.ai` / `mdb.ai`) or another OpenAI-compatible endpoint.

## How

- New `ProviderAuthError(ConnectionError)` in `anton/core/llm/provider.py`, beside `TokenLimitExceeded` / `ModelUnavailableError`. Subclasses `ConnectionError` so legacy call sites keep working. Its docstring pins the copy contract: every message must contain "Invalid API key" (cowork-server matches the lowercase substring `"invalid api key"`).
- Raised at every provider 401 site:
  - `anthropic.py` (both complete/stream): message unchanged — `Invalid API key — check your ANTHROPIC_API_KEY environment variable.`
  - `openai.py` `_raise_for_status_error`: new optional `auth_hint` parameter (default `None` → existing OpenAI copy, so external callers stay source-compatible). `OpenAIProvider.__init__` computes the hint once from `base_url` and all four call paths (chat/stream × completions/responses) pass it:
    - `mindshub.ai` / `mdb.ai` → `Invalid API key — check your MindsHub API key in Settings.`
    - other non-empty base URL → `Invalid API key — check the API key configured for your OpenAI-compatible endpoint.`
    - no base URL (direct OpenAI) → `Invalid API key — check your OpenAI API key configuration.` (unchanged)
- `session.py`: `ProviderAuthError` added to both curated escape points — the no-retry isinstance check and the exhausted-retries re-raise clause. A 401 now propagates immediately (no retries, no fallback prose), letting the server map it to the reconnect/update-key card.

Deliberately **not** widened to plain `ConnectionError`: the generic "temporarily unavailable" mapping also raises bare `ConnectionError` for transient failures, which must keep auto-retrying. A regression test pins this.

## Testing

- `uv run --extra dev pytest tests/ -q` — **1208 passed, 13 skipped** (full suite green; baseline 1200).
- New/updated tests:
  - `tests/test_status_error_mapper.py`: 401 → `ProviderAuthError` (and still a `ConnectionError`), HTML-body 401, hint override vs. default copy, and a parametrized `test_auth_hint_per_base_url` covering None / `mindshub.ai` / `mdb.ai` / generic base URLs — each case also asserts the `"invalid api key"` substring contract.
  - `tests/test_chat.py`: streaming turn re-raises `ProviderAuthError` after exactly 1 LLM call with no fallback prose; a plain transient `ConnectionError` still retries (4 calls) and falls back — proving the escape is scoped to the typed subclass.
- Cross-checked read-only against cowork-server `cowork/handlers/turn_errors.py::is_auth_error`: all four message variants contain the matched substring.
- Live repro context: the original failure was reproduced in the Cowork desktop app (MindsHub provider, invalid key) and the pipeline was verified working end-to-end with a valid key — the only broken part was this error surfacing.

## Notes for reviewers

- Small, standalone, no behavior change for non-401 paths. Targets `staging`; please review and merge manually (not part of the Browser Control M1 PR stack).
- The `auth_hint` parameter default keeps `_raise_for_status_error` source-compatible for any external caller.

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -10,6 +10,7 @@ from .provider import (
     ContextOverflowError,
     LLMProvider,
     LLMResponse,
+    ProviderAuthError,
     ProviderConnectionInfo,
     StreamComplete,
     StreamEvent,
@@ -120,7 +121,7 @@ class AnthropicProvider(LLMProvider):
         except anthropic.APIStatusError as exc:
             if exc.status_code == 401:
                 msg = "Invalid API key — check your ANTHROPIC_API_KEY environment variable."
-                raise ConnectionError(msg) from exc
+                raise ProviderAuthError(msg) from exc
             elif (
                 exc.status_code == 429
                 and isinstance(exc.body, dict)
@@ -267,7 +268,7 @@ class AnthropicProvider(LLMProvider):
         except anthropic.APIStatusError as exc:
             if exc.status_code == 401:
                 msg = "Invalid API key — check your ANTHROPIC_API_KEY environment variable."
-                raise ConnectionError(msg) from exc
+                raise ProviderAuthError(msg) from exc
             elif (
                 exc.status_code == 429
                 and isinstance(exc.body, dict)
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -43,9 +43,6 @@ def _raise_for_status_error(
         lets the caller supply provider-aware wording (MindsHub vs. a
         generic OpenAI-compatible endpoint vs. direct OpenAI); None keeps
         the direct-OpenAI copy so external callers stay source-compatible.
-        Whatever the hint, the message must contain "Invalid API key" —
-        cowork-server's provider_auth detection matches the lowercase
-        substring "invalid api key".
       - 403 with a structured gateway code (``model_access_denied`` /
         ``model_disabled``) → ModelUnavailableError carrying the code +
         model, with actionable copy. Detection is code-exact on purpose:
@@ -633,6 +630,9 @@ class OpenAIProvider(LLMProvider):
         # chat.completions path and as ``reasoning={"effort": ...}`` on the
         # Responses API path. None = the model's default.
         self._reasoning_effort = reasoning_effort
+        is_minds_endpoint = bool(base_url) and (
+            "mindshub.ai" in base_url or "mdb.ai" in base_url
+        )
         # Whether to attach langfuse-style headers (Langfuse-Session-Id,
         # Langfuse-Tags, Langfuse-Metadata) to outbound requests. Default-on
         # only for the MindsHub-backed deployment, which is the curated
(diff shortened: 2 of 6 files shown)


ianu82:
Scope note on the two Codex findings above: both point at `anton/utils/datasources.py`, which this PR does not modify. This PR's diff against `staging` touches only `anton/core/llm/{provider,anthropic,openai}.py`, `anton/core/session.py`, and two test files.

The review appears to have been computed while the PR was briefly based on `main` (the base was corrected to `staging` shortly after opening), so the diff included staging-only commits — the flagged code was introduced by #241 / #243, which are already merged into `staging`.

That said, the findings themselves look valid for that pre-existing staging code:

- **P1 (unknown custom secrets)**: `_register_unregistered_connection_vars` registers every field of a legacy record (no `secure_keys`) into `_DS_KNOWN_VARS`, while secret classification falls back to the `_SECRET_KEY_TOKENS` name heuristic (`password`, `passphrase`, `secret`, `token`, `key`, `credential`, `auth`). A field named e.g. `bearer`, `jwt`, or `session` matches none of these, so its value is no longer caught by the conservative unknown-`DS_*` redaction in `scrub_credentials()` — a regression relative to the pre-#241 behavior. (The shape-based `_SECRET_KEY_PATTERN` only covers `mdb_`/`sk-`/`AIza` prefixes, so it won't catch a bearer/JWT value.)
- **P2 (Drive resourceKey)**: the picker guidance in `build_datasource_context` formats picked files as `name (id, connection)` and instructs `files.get(fileId=...)` only; entries carrying a `resourceKey` lose it, and link-shared files that require it will fail to open.

Since neither is introduced by this PR, they should be fixed in a separate change against `staging` rather than folded into this focused auth-surfacing fix.

PR #258, opened 2026-07-17
ENG-847: Fix scratchpad web_search() on the minds-cloud gateway

ea-rus:
## Problem
Scratchpad `web_search()` returns an empty string under the minds-cloud provider (the
standard config), so the agent silently loses web access and falls back to scraping.
Affects real cowork/anton usage, not just evals.

Cause: the scratchpad picks its web "flavor" by provider name, but `OpenAIProvider.name`
is always `"openai"` even on the minds gateway — so it chose the OpenAI Responses API and
dispatched web_search to an endpoint the minds gateway doesn't implement (minds serves it
over the chat.completions passthrough) → empty. The gateway itself is fine (a direct
chat.completions call with `tools=[{"type":"web_search"}]` returns real results).

## Fix
Resolve the flavor by HOST, not provider name: minds/mdb.ai base URL → passthrough.
Extracted as `OpenAIProvider.resolve_web_flavor()`, called from `scratchpad_boot.py`.
All prior cases unchanged; only the broken `openai`-name + minds-base case now works.

## Repro
cowork-server on `minds_cloud`, then in the scratchpad:
`print(repr(web_search("current US CPI year-over-year rate")))`
Before: `''`. After: a non-empty answer with sources.

## Tests
`tests/test_scratchpad_flavor.py` (new, 7 cases); `test_scratchpad.py` + `test_web_tools.py`
still green (107 passed).

Fixes https://linear.app/mindsdb/issue/ENG-847/scratchpad-web-search-silently-returns-empty-on-the-minds-cloud

### DIFF
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -104,28 +104,11 @@ if _scratchpad_model:
                 _llm_provider_kwargs["vision_format"] = "anthropic"
             # Resolve the OpenAI "flavor" so the injected web_search() helper can
             # route through whatever native web tooling the endpoint exposes.
-            # Mirrors LLMClient._resolve_openai_compatible_flavor: direct OpenAI
-            # BYOK uses the Responses API (web_search); an openai-compatible base
-            # URL pointing at Minds/mdb.ai uses the chat.completions passthrough;
-            # any other openai-compatible endpoint is generic (no native search).
-            if _scratchpad_provider_name == "openai":
-                _llm_provider_kwargs["flavor"] = _ProviderClass.FLAVOR_OPENAI
-            else:
-                _minds_url_norm = (
-                    os.environ.get("ANTON_MINDS_URL", "").rstrip("/").lower()
-                )
-                _base_norm = (_llm_base_url or "").rstrip("/").lower()
-                if _minds_url_norm and _base_norm in (
-                    _minds_url_norm,
-                    f"{_minds_url_norm}/api/v1",
-                ):
-                    _llm_provider_kwargs["flavor"] = (
-                        _ProviderClass.FLAVOR_MINDS_PASSTHROUGH
-                    )
-                else:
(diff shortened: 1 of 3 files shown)


StpMax (approved):

alecantu7:
Late review note (PR's already merged — flagging for follow-up, not blocking): nice fix. Resolving the flavor by host is the right layer, and because it sets the provider *flavor* rather than just patching `web_search()`, it also fixes `get_llm().complete()` and `generate_object()` — they route through the same `complete()` dispatch, so the whole scratchpad blast radius is covered by this one change.

**One residual gap worth a small follow-up (hardening, not a regression in this PR):** the non-streaming `complete()` path still returns an empty result **silently** on an empty 200. This PR guarantees MindsHub no longer takes the Responses path, so *this* bug is fully fixed — but the underlying silent-empty failure mode remains. If any endpoint ever returns a 200 with no content (a different provider, a future regression, a truncated body), `_parse_response_object` (`anton/core/llm/openai.py:1326-1372`) via `_complete_via_responses` (`:1127-1167`) will still hand back `LLMResponse(content='', usage=Usage(0,0,0.0), stop_reason=None)` with no exception raised. That's exactly the "agent silently gets nothing and misdiagnoses it as a backend outage" mode that made this bug so hard to triage in the first place.

The streaming chat path already guards this at `openai.py:1056-1073` (raises `TransientProviderError(code="truncated_stream")` on empty content + `stop_reason is None`). Mirroring that onto the non-streaming `complete()` path — and the Anthropic non-streaming equivalent at `anton/core/llm/anthropic.py:181-202` — would give defense-in-depth so the next empty-200 fails loud instead of silent. Low priority; just capturing it so it isn't lost.

(Context: surfaced during a duplicate investigation, ENG-1004, now closed as a dup of ENG-847.)

ea-rus:
@alecantu7 fixed here https://github.com/mindsdb/anton/pull/297

PR #259, opened 2026-07-17
feat: anton_state SDK + artifact lifecycle integration for stateful artifacts

StpMax:
## What

Adds durable state support for `fullstack-stateful-app` artifacts on the anton side
([ENG-704](https://linear.app/mindsdb/issue/ENG-704)): a `STATE` SDK with two
interchangeable drivers, and its integration into the artifact lifecycle
(backend template, generation rules, publish bundle, local runs).

### 1. `anton_state` — the STATE SDK (new top-level package)

One async `Store` API (get / put / delete / query, TTL, conditional writes /
optimistic locking) over two drivers behind a common `Driver` protocol:

- **SQLite driver** (local dev): stdlib `sqlite3`, WAL mode, items stored as JSON.
- **HTTP driver** (published runtime): stdlib `urllib`, a thin synchronous RPC to
  a trusted state broker (`mindshub_services`) that owns the actual DynamoDB
  table; the namespace is derived server-side from the signed token, never sent
  by the artifact. Reads are retried on transient failure; mutations are not
  (their outcome would be unknown). No new runtime deps — `anton_state` itself
  never imports `boto3`.

**Parity by construction** — the API is the *intersection* of both backends, and
known divergences are eliminated in driver code rather than left to generation
prompts:

- strict cloud-grade validation on both sides (key types, empty keys, 400 KB item
  cap, TTL must be an epoch-seconds number);
- `ConsistentRead=True` by default on base-table reads → read-after-write behaves
  the same locally and in the cloud;
- lazy TTL on both sides: expired items are filtered on read; conditional writes
  see expired "zombies" identically;
- monotonic version (`_v`) assignment is identical on both sides, preventing an
  ABA lost-update on optimistic-lock writes;
- query is bounded identically on both sides (default/max row caps) so a
  collection that exceeds them doesn't silently return "everything" locally and
  truncate in prod;
- secondary indexes (GSIs) are out of scope for v1 and rejected by the manifest
  schema — not partially supported on one side only.

**Schema is a single source of truth**: `state_manifest.json` (declarative,
pydantic-validated) is read both by backend code (`open_store()` /
`StateSchema.from_manifest`) and by the publish/provisioning pipeline — code and
provisioned table cannot drift by construction. Migrations are out of scope for
v1 (key-schema must match on republish; enforced by the provisioning gate in
mindshub_services).

Driver selection mirrors `SECRETS`: `backend.STATE is None` → SQLite;
`{url, token}` dict (overlaid by the cloud runner with a per-invocation signed
token for the broker) → HTTP driver.

### 2. Lifecycle integration

- **Publisher**: vendors the `anton_state` package into every fullstack bundle
  (version-locked, no package index needed at runtime); ships
  `state_manifest.json` in the bundle and duplicates it in the publish payload
  for the provisioning pipeline; defensively excludes `.anton_state.db`
  (+ `-wal`/`-shm`) from bundles.
- **Backend launcher**: local backends run in a scratchpad venv where
  `anton_state` isn't installed — the launcher injects an isolated PYTHONPATH
  dir exposing *only* the `anton_state` package (symlink, copy fallback), so
  anton's own dependencies never shadow the backend venv's. Covers
  cowork-server too (it calls `launch_artifact_backend`).
- **Backend template & generation rules** (`build-fullstack-backend` builtin
  skill): module-level `STATE = None` slot symmetric to `SECRETS`, a
  `get_store()` pattern with `__file__`-relative paths; `STATE` is positioned
  for *light* state (counters, settings, sessions, simple keyed documents);
  heavy/relational needs go to an external connected database — not forced
  into KV.

### Packaging note

`anton_state` is added to `[tool.hatch.build.targets.wheel].packages` — not to
publish it standalone, but because anton itself imports it at runtime
(`publisher._vendor_anton_state`, `backend_launcher._anton_state_pythonpath_dir`):
the wheel is the source both vendoring and local injection copy from. No new
runtime or dev dependencies are added; the HTTP driver is stdlib-only.

## Out of scope (follow-ups in other repos)

- Cloud data-plane — the state broker (`functions/artifact_state`), the shared
  DynamoDB table it owns, per-invocation token minting, and table provisioning
  at publish: `mindshub_services` + `terraform` PRs.

Until those land, published stateful artifacts get `STATE = None` in the cloud;
local behavior is fully functional.

## Testing

`pytest tests/test_anton_state/ tests/test_publisher_state.py
tests/test_backend_launcher_env.py tests/test_prompts_state.py`

- driver parity suites (SQLite + HTTP driver against a fake broker): roundtrip,
  conditional writes, optimistic locking, TTL zombies, number round-trips;
- single-source manifest seam test (publish side and backend side read the same
  schema);
- vendoring/bundle content, isolated PYTHONPATH injection, skill/rules smoke
  checks;
- lazy-import guard: the SQLite path never imports the HTTP driver.


ea-rus (commented):
LLM findings 

## Blocking

**1. PR description is out of sync with the implementation.** The description details a DynamoDB driver on boto3 + per-invocation STS creds, selection on `{table, region, credentials}`, `float↔Decimal` conversion, and "new dev deps (boto3, moto)". None of that is in the diff: the cloud driver is `http_driver.py` (broker RPC), selected on `{url, token}`; `pyproject.toml` only changes the wheel `packages` line — no boto3/moto added; the skill tests assert the opposite (`"table, region" not in T`, `"url, token" in T`). Please rewrite the description to match the broker design. Side effect: `tests/test_anton_state/test_lazy_dynamo_import.py` asserts `anton_state.dynamo_driver` / `boto3` are not imported — neither exists in the package, so the test is vacuous; drop it or repoint it at `http_driver` not being imported on the SQLite path.

**2. `publish()` does not validate the manifest.** `_read_state_manifest` only does `json.loads` — no `StateSchema` validation. A malformed/hand-written `state_manifest.json` (wrong shape, `gsis`, non-string key type) is shipped to the provisioning pipeline as-is and only fails later at runtime. This contradicts the SKILL claim "this makes a malformed manifest impossible". Run `StateSchema.from_manifest(path)` in `publish()` to fail fast before upload.

## Should fix

**3. `update`/`increment` bypass all validation (parity gap).** `SQLiteDriver._rmw` deliberately skips `validate_item`, so locally these succeed where the cloud broker/DynamoDB would reject: no `MAX_ITEM_BYTES` (400 KB) check → oversized items; no value-type check on `set_fields` → unsupported types stored locally; key/reserved attrs overwritable, e.g. `set_fields={"pk": "other"}` diverges the stored body from the `pk` column. At minimum apply the size + value-type checks to `set_fields`/`add_fields`.

**4. Items created via `increment`/`update` lack logical keys (`pk`/`sk`/`_key`).** `_rmw` starts from `{}` and adds only the mutated field + `_v`, unlike `put`. Through `Collection` this is a footgun:
```python
c = Collection(get_store(), "counters")
await c.increment("visits", field="n")   # body: {"n": 1, "_v": 1} — no _key/pk/sk
for i in await c.list():
    i["_key"]                            # KeyError on mutation-born items
Not covered by tests (test_increment_atomic only checks the field value). Either synthesize the logical keys in _rmw or document the asymmetry loudly in the SKILL.

5. Raw sqlite3.OperationalError escapes as a non-StateError. The async facade runs the sync driver in a thread pool; concurrent writes serialize on BEGIN IMMEDIATE with busy_timeout=5000. Beyond that the raw sqlite3.OperationalError: database is locked propagates (re-raised in except Exception: rollback; raise) instead of StateThrottled/StateUnavailable — another local≠cloud divergence and an unwrapped error. Wrap it in a StateError.

StpMax:
Thanks for the thorough review, went through the blocking + should-fix items:

- **Description out of sync (#1)**: updated — now describes the HTTP driver → state broker (`mindshub_services`) design with `{url, token}` selection instead of DynamoDB/boto3/STS, dropped the stale GSI mention (v1 rejects GSIs in the manifest schema), and fixed the file reference for the backend template (now the `build-fullstack-backend` builtin skill, not `prompts.py`). Left the now-vacuous `test_lazy_dynamo_import.py` untouched for now — happy to drop/repoint it in a follow-up if useful.
- **`publish()` doesn't validate the manifest**: not doing this one. The manifest isn't hand-written or edited by users — it's generated once, automatically — so there's no realistic path for a malformed one to reach `publish()`.
- **`update`/`increment` bypass validation (#3)**: fixed. Both now type-check `set_fields`/`add_fields` up front (same value-type checks `put` already applies) and the mutated item is size-capped (400 KB) before being persisted, via shared `check_value`/`check_number`/`check_size` helpers in `validation.py`. Added tests in `test_sqlite_driver.py` covering non-numeric `by`/`add_fields`, unsupported `set_fields` value types, oversized items, and that a rejected mutation doesn't partially apply.
- **Items from `increment`/`update` lack `pk`/`sk`/`_key` (#4)**: leaving as-is. It's symmetric on both drivers by design (the broker doesn't synthesize logical keys either), so it's not a local/cloud parity gap — just an ODM-level footgun. Since backend code using `Collection` is agent-generated rather than hand-written, we can steer generation away from the pattern instead of changing the driver/ODM contract.
- **Raw `sqlite3.OperationalError` escapes (#5)**: fixed. `put`/`delete`/`_rmw` now catch `sqlite3.OperationalError` (lock contention past `busy_timeout`) and re-raise as `StateThrottled`, matching how the HTTP driver maps broker errors. Added tests exercising this via a flaky-connection proxy.

`pytest tests/test_anton_state/` (79) and the lifecycle suites are green.

ea-rus (approved):

PR #263, opened 2026-07-21
ENG-940: explicit encoding="utf-8" on scratchpad/chat reads (launcher-independent UTF-8)

alecantu7:
## What

Add explicit `encoding="utf-8"` to every text **read and write** in anton's scratchpad/chat path, so they decode UTF-8 regardless of host locale or whether `PYTHONUTF8` is set. Completes **ENG-940** (the "suspenders" half of ENG-824's fix).

## Why

ENG-824 (GBK/CJK Windows scratchpad crash — `'gbk' codec can't decode byte …`) shipped the **belt** only: `PYTHONUTF8=1` on the scratchpad subprocess (anton#253) and on the cowork-server sidecar that hosts in-process anton (cowork#420). But ENG-824's "Done when" also required explicit `encoding="utf-8"` on the reads, and that was left undone — so the reads relied *entirely* on a launcher setting the env. Any launch path that doesn't (bare `uv run` / CLI, a provisioned/Docker cowork instance, OpenClaw's launcher) re-crashes on a GBK host at the root-caused site `code = script_path.read_text()`. Explicit encoding makes these sites **launcher-independent** (belt *and* suspenders, as specced).

## How

Every text read/write in the scratchpad/chat path now passes `encoding="utf-8"`:

**`anton/chat.py`**
- `script_path.read_text(...)` — the ENG-824 root crash site (scratchpad script; commented).
- `env_path.read_text(...)` + `env_path.open("a", ...)` — `.env` read/append.
- `published_json` / `legacy_json` / `pub_file` `.read_text(...)` + paired `.write_text(...)` — publish-map JSON (round-trip safe).

**`anton/core/backends/local.py`**
- `.python_version` and `requirements.txt` read + write.

(`console.file.write(...)` stdout calls are untouched — that's stdio, covered by UTF-8 mode's `surrogateescape`, not file I/O.)

## Tests (`tests/test_scratchpad_utf8.py`)

- `test_scratchpad_script_read_is_locale_independent` — a real non-ASCII fixture: the explicit UTF-8 read round-trips verbatim, while a host-locale (GBK) read **crashes-or-corrupts** (payload-independent — we don't rely on which failure mode a given byte sequence hits). Proves the fix stands on its own, **independent of `PYTHONUTF8`**.
- `test_chat_module_has_no_bare_read_text` — regression guard: fails the instant any `.read_text(...)` in `anton/chat.py` drops `encoding=`.

Local: `tests/test_scratchpad_utf8.py` 7/7 green; 108 pass across the chat/publish/scratchpad suites. (The 2 `test_chat_scratchpad` failures are a pre-existing sandbox issue — the scratchpad subprocess can't spawn here — verified identical on `origin/staging` without this change.)

## Not included (tracked on ENG-940)

The optional launcher-agnostic `cowork.cli:main` UTF-8 re-exec guard — a separate `cowork-server` change; left as optional hardening.

## Security

Encoding-only correction on existing read/write sites; no new inputs, endpoints, or secret handling.

Closes the explicit-encoding acceptance item of ENG-824.

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -640,7 +640,7 @@ async def _handle_publish(
     published_map = {}
     try:
         if published_json.is_file():
-            published_map = json.loads(published_json.read_text())
+            published_map = json.loads(published_json.read_text(encoding="utf-8"))
     except Exception:
         pass
     prev = published_map.get(published_key)
@@ -650,7 +650,7 @@ async def _handle_publish(
         legacy_json = artifacts_root / ".published.json"
         try:
             if legacy_json.is_file():
-                legacy_map = json.loads(legacy_json.read_text())
+                legacy_map = json.loads(legacy_json.read_text(encoding="utf-8"))
                 legacy_entry = legacy_map.get(file_key)
                 if isinstance(legacy_entry, dict) and legacy_entry.get("report_id"):
                     prev = legacy_entry  # carry over report_id/url/last_md5/access
@@ -751,7 +751,7 @@ async def _handle_publish(
         })
         published_map[published_key] = entry
         try:
-            published_json.write_text(json.dumps(published_map, indent=2))
+            published_json.write_text(json.dumps(published_map, indent=2), encoding="utf-8")
         except Exception:
             pass
 
@@ -880,7 +880,7 @@ async def _handle_unpublish(
     artifacts_root = _Path(settings.artifacts_dir)
     for pub_file in artifacts_root.rglob(".published.json"):
         try:
-            data = _json.loads(pub_file.read_text())
+            data = _json.loads(pub_file.read_text(encoding="utf-8"))
         except Exception:
             continue
         if not isinstance(data, dict):
@@ -900,7 +900,7 @@ async def _handle_unpublish(
                 changed = True
         if changed:
             try:
-                pub_file.write_text(_json.dumps(data, indent=2))
+                pub_file.write_text(_json.dumps(data, indent=2), encoding="utf-8")
             except Exception:
                 pass
 
@@ -1021,7 +1021,12 @@ async def _agent_zero(console: Console, session: "ChatSession", settings) -> str
         description="Demo dashboard comparing NVDA stock and BTC prices over time.",
         type="html-app",
     )
-    code = script_path.read_text()
+    # Read as UTF-8 explicitly, never the host-locale default: on a non-UTF-8
+    # Windows code page (GBK/cp936) a bare read_text() crashes decoding UTF-8
+    # script content (ENG-824's root-caused crash site — 'gbk' codec can't
+    # decode …). Explicit encoding is launcher-independent, unlike the
+    # PYTHONUTF8 belt (ENG-940).
+    code = script_path.read_text(encoding="utf-8")
     output_dir = str(_store.folder_for(_demo_artifact.slug))
     output_html = str(Path(output_dir) / "dashboard.html")
     code = (
@@ -1146,9 +1151,9 @@ def _persist_first_run_done(settings) -> None:
 
     env_path = Path.home() / ".anton" / ".env"
     env_path.parent.mkdir(parents=True, exist_ok=True)
-    existing = env_path.read_text() if env_path.is_file() else ""
+    existing = env_path.read_text(encoding="utf-8") if env_path.is_file() else ""
     if "ANTON_FIRST_RUN_DONE" not in existing:
-        with env_path.open("a") as f:
+        with env_path.open("a", encoding="utf-8") as f:
             if existing and not existing.endswith("\n"):
                 f.write("\n")
             f.write("ANTON_FIRST_RUN_DONE=true\n")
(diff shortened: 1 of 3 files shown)


alecantu7:
### Self-review

- **Scope check (why only `chat.py` + `local.py`):** ENG-824 root-caused the bare reads in `chat.py`; I extended to `core/backends/local.py` (the scratchpad backend: `.python_version` / `requirements.txt`). The other scratchpad-path readers the ticket mentioned — `artifacts/store.py`, `publisher.py`, and `chat.py:353` — already pass `encoding="utf-8"`, so they're not touched. Swept both files for any remaining bare `read_text()` / text `open(...)`; none left.
- **JSON reads were not the active crash vector** (anton writes those with `json.dumps`, i.e. `ensure_ascii=True` → ASCII on disk). I still pinned them for defense-in-depth + round-trip safety, but the real vector is the **script read** (arbitrary model-written UTF-8) and `.env` — which the test models directly.
- **Test is `PYTHONUTF8`-independent by construction** (it asserts on explicit-encoding reads + a crash-or-corrupt comparison of the raw bytes), so it proves the suspenders, not the belt — and passes on a UTF-8 CI box.
- **Deferred:** the optional `cowork.cli:main` re-exec guard (covers CLI/Docker launchers) — separate `cowork-server` change, noted on ENG-940.

alecantu7:
### Addressed the encode-side sibling (de6d6fb) — from ENG-940's new evidence

The Langfuse churn note (sabrina/eddie/janis) surfaced the **encode-side** sibling of ENG-824: a non-ASCII Windows path (pt-BR `Área de Trabalho`, emoji filename) gets surrogate-escaped into lone surrogates (`\udcXX`) on a non-UTF-8 host, and a **strict** UTF-8 encode then raises `surrogates not allowed` and kills the session. Explicit `encoding="utf-8"` (this PR's original scope) does **not** fix that — strict UTF-8 forbids lone surrogates. Two changes:

1. **The one host-side strict encode of generated content** is the cell payload sent to the subprocess (`local.py` `payload.encode("utf-8")`). Extracted `_encode_cell_payload()` using `errors="surrogatepass"` so it can't crash when the host isn't in UTF-8 mode; the subprocess (always UTF-8 mode via `_utf8_env`) decodes it fine.
2. **The two `chat.py` JSON writes need no change** — `json.dumps` defaults to `ensure_ascii=True`, so surrogates become ASCII `\uXXXX` escapes and never reach the encoder (verified). Flagging so it's a considered decision, not an omission.

**Tests:** pinned the surrogate-safe encode (strict **raises**, surrogatepass round-trips) + a broadened accented-Latin + emoji case that must pass through unmangled. Confirmed the surrogate test **fails** if the helper reverts to strict. 9/9 green.

Scope note: kept in this PR since it's the same launcher-independence goal for the same file. The generated-file *content* writes happen inside the scratchpad subprocess, which is already belt-covered (UTF-8 mode on every launcher), so they're out of scope here.

alecantu7:
### Adversarial self-review of the encode-side fix — one real bug caught (5aa166d)

**A — Wrong `errors=` mode (correctness, fixed).** My first pass used `errors="surrogatepass"`. Verified empirically that this does **not** survive the round-trip: the scratchpad subprocess always runs in UTF-8 mode, so its stdin decodes with `surrogateescape` — and `surrogatepass` emits the 3-byte CESU form that `surrogateescape` then re-mangles (`\udc81` → three surrogates). Switched to `errors="surrogateescape"`, which is correct on both counts: it's the inverse of the `os.fsdecode` that *created* these lone surrogates (U+DC80..U+DCFF), so it restores the original path bytes, and it matches the subprocess's `surrogateescape` stdin decode → the path arrives verbatim.

**B — The test asserted a false symmetry (fixed).** It decoded with `surrogatepass` (mirroring the buggy encode), so it would have passed for *either* mode — it didn't actually pin anything. Now it decodes with `surrogateescape` (what the subprocess does) and genuinely discriminates: **fails** if the helper reverts to `surrogatepass`, verified.

**C — Honest limitation (not fixed, flagging).** I hardened the one host-side strict encode of generated content (the cell payload), and the codec is now correct — but I have **not** trace-confirmed that this exact frame is where sabrina/eddie/janis crashed (their sessions predate the belt, and the surrogate path could alternatively enter via the spawn `cwd=`, which the OS fsencode handles without crashing). So this stands as **defense-in-depth for non-belt launchers** with a verified-correct codec, not a trace-confirmed reproduction of those three frames. If we want certainty, the next step is a Langfuse trace showing the crashing frame.

9/9 tests green.

ZoranPandovski (approved):

alecantu7:
### Correction: this PR does NOT fix the encode-side crash (tracked in ENG-981)

Verified the real raiser is `scratchpad_boot.py:757 compile(code, ...)` on stdin data — a different file/channel than this PR touches. `compile()` is always strict UTF-8, and the ENG-824 belt's lenient `surrogateescape` stdin is what *creates* the lone surrogate in the first place, so neither the belt nor this PR's `encoding="utf-8"` changes fix it.

My `_encode_cell_payload(surrogateescape)` change only **relocates** the failure (parent encode → child `compile()`), incidentally healing only when the surrogates are a full byte-escaped multibyte pair — input-dependent, not a real fix. The commit message + the `test_cell_payload_encode_*` tests overclaim; I'll correct them in the ENG-981 follow-up.

**What stands:** the read-side `encoding="utf-8"` additions (launcher-independent decode). **What's deferred to ENG-981:** the deterministic surrogate heal before `compile()`.

PR #239, opened 2026-07-07
ENG-648: Reduce token usage — prompt caching, prompts→skills, cheap router tier

torrmal:
Linear: [ENG-648](https://linear.app/mindsdb/issue/ENG-648/reduce-anton-cowork-server-token-usage-promptskills-migration-and)

Full analysis + plan in `plans/eng-648-token-efficiency.md`. **Phases 2 and 3 are implemented in this PR**; Phase 1 (prompt caching) is pending a design decision with the `mindshub_inference` gateway.

## Problem

Every turn — even a trivial "answer from context" one — paid for a **~11.4K-token system prompt**, **~4.2K tokens of tool schemas**, the **full conversation history**, and the frontier planning model, multiplied by up to **25 tool rounds per turn**, with **no prompt caching at all**.

### Phase 3 — Cheap front-model router (respond-or-delegate) · `anton/core/llm/router.py`

With `ANTON_ROUTER_ENABLED=true`, a cost-effective router model (new `router` role on `LLMClient`, defaults to the coding role, i.e. Haiku) runs first on every text turn with a **~420-token prompt and no tool schemas**, over a condensed text-only view of history, and either:

- **answers directly** when the reply is fully derivable from the conversation + stable general knowledge (planning model never invoked — the turn costs a single small Haiku call), or
- **delegates** via a forced `delegate` tool call, optionally naming skills that get **preloaded into history as a synthetic `recall_skill` exchange** — payload-identical to what the planning model would have fetched itself, minus the full-context planning round.

Conservative by construction: image turns, truncated router answers, empty responses, unknown skill labels, and router exceptions all fall through to the planning model. Off by default until eval'd.

### Phase 2 — Prompt tutorials → built-in skills · `anton/core/llm/builtin_skills.py`

~57% of the system prompt was two always-on tutorials + a URL catalog. They now ship as built-in skills — `html-dashboards` (~2.7K tok), `backend-apps` (~3.9K tok), `public-data-sources` (~0.7K tok) — listed as one-line entries in the procedural-memory section and loaded on demand via the existing `recall_skill` tool or the router's preload path. Built-ins resolve before the on-disk `SkillStore` and shadow same-labelled user skills. Prompt content is unchanged; it just moved.

**Measured: base system prompt 45,511 → 18,847 chars (~11,377 → ~4,711 tokens, −59%).**

### Test coverage

[...]

PR #266, opened 2026-07-22
ENG-648: cheap front-model thalamus gate (respond-or-delegate), off by default

ea-rus:
Moved from https://github.com/mindsdb/anton/pull/239

Adds the **thalamus** — a cheap front-model that every text turn hits first
when `ANTON_ROUTER_ENABLED=true`. It either answers trivial/from-context turns
directly (skipping the full prompt + tool schemas + planning model) or delegates
to the planning model, optionally preloading skills so the cortex doesn't spend
a round on `recall_skill`. **Off by default** until eval'd.

## What
- `anton/core/llm/thalamus.py` — the gate: `condense_history` (text-only view),
  `gate_turn` (decision parsing), `DELEGATE_TOOL`, `ThalamicDecision`. Fails
  *open*: any doubt/error/truncation/empty answer → delegate.
- `LLMClient.gate()` — one cheap gating call on the router role (no
  `native_web_tools`).
- `CoreSettings`: `router_enabled` (default False), `router_max_tokens` (1024).
- `ChatSession`: `_gate_turn` + `_inject_recalled_skills`, wired into both
  `turn()` and `turn_stream()`. Image turns skip the gate. Preloaded skills
  resolve through the same `SkillStore` `recall_skill` uses (built-ins included
  via #244's mechanism) — payload byte-identical to a real recall.

## Naming
Mechanism = **thalamus** (brain-inspired, matches anton's architecture);
config surface stays **`router_*`** (consistent with cowork-server / PR C).

### DIFF
--- a/anton/core/llm/client.py
+++ b/anton/core/llm/client.py
@@ -157,6 +157,29 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
         )
 
+    async def gate(
+        self,
+        *,
+        system: str,
+        messages: list[dict],
+        tools: list[dict] | None = None,
+        tool_choice: dict | None = None,
+        max_tokens: int | None = None,
+    ) -> LLMResponse:
+        """One cheap gating call on the router role — see `anton.core.llm.thalamus`.
+
+        No ``native_web_tools``: the thalamus must never do work itself,
+        only answer from context or delegate.
+        """
+        return await self._router_provider.complete(
+            model=self._router_model,
+            system=system,
+            messages=messages,
+            tools=tools,
+            tool_choice=tool_choice,
+            max_tokens=max_tokens or self._max_tokens,
+        )
+
     async def _generate_object_with(
         self,
         schema_class,
--- a/anton/core/llm/thalamus.py
+++ b/anton/core/llm/thalamus.py
@@ -1,0 +1,332 @@
+"""Thalamus — the gate between an incoming turn and the cortex.
+
+Brain analogue
+==============
+
+The thalamus is the brain's central relay. Almost every sensory and
+motor signal passes through it on the way to the cortex — it is the
+first thing a signal hits, and it decides what happens next. Four of
+its properties are worth stealing:
+
+1. **It gates; it does not compute.** The thalamus relays signals and
+   modulates their gain. The heavy interpretation happens in the cortex.
+   A thalamus that started "thinking" would be both slow and a liability.
+
+2. **Its default posture is inhibitory.** The thalamic reticular
+   nucleus (TRN) is a shell of inhibitory neurons that suppresses relay
+   by default and selectively *dis*inhibits the one channel that earns
+   passage — Crick's "searchlight." Nothing gets through on a maybe.
+
+3. **It has two firing modes.** *Tonic* mode faithfully relays a simple,
+   attended signal. *Burst* mode is an alerting spike that says "cortex,
+   wake up, this one needs you." Every signal resolves to one or the
+   other.
+
+4. **The cortex talks back — a lot.** Corticothalamic fibers outnumber
+   the feed-forward ones roughly ten to one: the cortex constantly
+   biases what the thalamus lets through, based on context and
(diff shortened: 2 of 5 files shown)


StpMax (changes requested):
## Minor

### 1. Synthetic preload skips the `_already_in_history` check — diverges from real `recall_skill`

`handle_recall_skill` in `anton/core/tools/recall_skill.py` returns a short stub (not the full body) when a skill was already recalled earlier in the conversation, to avoid re-sending the procedure. `_inject_recalled_skills` in `anton/core/session.py` always injects the full body via `_format_skill_response(skill)`. If the thalamus names an already-preloaded skill in a longer conversation, its body gets duplicated in context (wasted tokens), whereas a real `recall_skill` call would have returned the stub. The docstring claim "payload byte-identical to a real recall" holds only for the happy path. Either route the injection through the same `_already_in_history` guard, or soften the wording in the docstring / PR description.

### 2. Gate-call cost is invisible for delegated turns

On `ACTION_DELEGATE`, `decision.response` (the gate call's usage) is never surfaced — only direct answers surface it via `StreamComplete`. Since the router hits *every* text turn when enabled, the entire gating traffic will be absent from telemetry / token accounting. Not a bug, but before eval this is worth seeing — otherwise the "cheapness" assessment is incomplete. At minimum, log the gate usage.

### 3. `condense_history`: truncation by `max_chars` happens before merge, not after

In `anton/core/llm/thalamus.py`, when consecutive same-role messages are merged, their content is concatenated without re-truncating, so the resulting merged block can significantly exceed `max_chars`. Not a problem for the defaults (1500 chars, 16 messages), but the cap is documented as "per message" and isn't actually guaranteed. A comment or a post-merge truncation would fix the mismatch.

## Discussion (non-blocking)

### 4. Thalamus prompt carries no date/context but allows "stable general knowledge"

`_build_system_prompt` (with the date, etc.) is not invoked on the direct-answer path. The "information that could be stale → delegate" wording partly covers this, but for borderline queries ("what's today's date", "how old are you") the model may confidently answer wrong. Consider explicitly listing date/time as "always delegate", or leave it to eval.

### 5. Latency

When the flag is on, every delegated turn (the common case) pays an extra full round-trip before real work begins. This is an inherent, acknowledged trade-off ("off until eval'd") — just worth keeping as an eval metric alongside the direct-answer hit rate, not only the hit rate itself.

## Nits

- The module docstring (~70 lines of neuro-analogy) is nice but is a noticeable fraction of the file — author's call.
- The `LLMResponse(content=decision.text)` fallback in `turn_stream` is a dead branch (for `ACTION_RESPOND`, `gate_turn` always sets `response`) — harmless.

ea-rus:
Fixes: 

1. Synthetic preload no longer duplicates an already-recalled skill.
_inject_recalled_skills now skips any label whose full body is already in history (via _already_in_history), mirroring handle_recall_skill's stub path — so a later delegate naming the same skill can't re-inject the procedure. The counter is not bumped on the skip either, matching the real recall. The "byte-identical to a real recall" claim now holds for the repeat case too. Added test_preload_skips_skill_already_in_history (a second preload of the same label is a no-op).

2. Gate-call cost is now surfaced on delegate.
Instead of just logging it, the gate call's usage is emitted as its own StreamComplete(response=decision.response) on the delegate path — exactly like a planning round — so any consumer that sums StreamComplete usage (e.g. the CLI's X in / Y out counter) now counts the gate on every outcome, not only on direct answers. The cowork/dispatch consumers pass on StreamComplete, so this is a no-op there (their accounting is server-side). Added test_gate_usage_counted_on_delegate_streaming (two StreamCompletes billed: gate + planning).

3. condense_history now truncates after merge.
Extracted _truncate_middle (hard cap — the marker counts against the budget) and moved truncation to run after the merge and the max_messages window, per surviving entry. A block merged from several near-cap same-role messages is now capped, so the whole view is bounded by max_messages * max_chars. Docstring updated to state that bound. Added test_merged_block_respects_max_chars_cap.

StpMax (approved):

PR #276, opened 2026-07-27
fix(session): completion verifier fails toward a real message, not silence (ENG-1079)

torrmal:
## Summary
- The completion-verifier's `generate_object_code` call can throw (provider hiccup, malformed structured output, etc.). Before this fix, that exception was silently mapped to a fake `COMPLETE` verdict, so the task loop just broke with no message at all — the agent looked like it had died on the first hurdle it hit, even in a fresh session (ENG-1079).
- On a verifier-call exception, this now routes through the same honest-diagnosis pattern already used for `STUCK`: append a `SYSTEM` note admitting the check itself failed, and let the model generate a real message — summarize progress, be honest about the internal failure, and ask the user how they'd like to proceed. Functionally a WAITING-style stop, but with a real, model-generated message instead of silence.
- Also persists that generated message to history directly, rather than relying on the shared post-loop fallback (which re-appends the stale pre-verification `reply` and would otherwise leave history out of sync with what the user actually saw).
- **Follow-up**: none of the three "hand control back to the user" paths (STUCK, budget-exhausted, and the new verifier-failure path) asked the model for a solvability self-assessment — just "what I tried" + "next steps," which can read as "here's what happened, good luck" instead of an actual recommendation. Added a shared `_SOLVABILITY_CLAUSE` ("is this still solvable on your own with a different approach, or does it genuinely need the user's input/decision/credentials") wired into all three.

## Context
Traced from a regression report that Anton seems to get stuck/"die" on the first hurdle even after restarting from scratch. Root-caused to the ENG-716 verifier rework (`db10fdcf`/`92f47b14`/etc.), which moved the verdict to a `generate_object_code` call on the cheap coding model over a compact transcript — a less forgiving code path than before, sitting behind a silent fail-safe that had never existed pre-ENG-716.

## Test plan
- [x] `tests/test_verifier_failsafe.py` — verifier exception yields a real streamed message (not silence), makes exactly one honest-diagnosis LLM call, does NOT inject a "Continue working" continuation, persists the real message to history, and asks for the solvability self-assessment.
- [x] `uv run pytest tests/ -k "session or verifier or loop_safety" --ignore=tests/e2e` — 89 passed, 2 skipped.
- [x] `uv run pytest tests/e2e/scenarios/test_loop_safety.py` — passes in an isolated worktree; occasional single-test timeouts under load are pre-existing environmental flakiness (real subprocess CLI spawn, confirmed to pass in isolation both before and after this change).

## QA Instructions

**Where to test:** a build that actually contains this change — desktop on the **git update channel** (lands minutes after merge), or hosted after the manual snapshot bake + SAM redeploy. PyPI-channel desktop builds will NOT receive an anton-only change (ENG-1094).

**Precondition for every scenario:** the turn must use at least one tool round (a scratchpad task, not pure Q&A) — the completion verifier is skipped otherwise.

1. **Happy-path regression (sonnet):** give a small multi-step task ("fetch example.com and save a summary to a file"). It should complete normally — no check-in message, no behavior change.
2. **The fix, via the real user path:** in the model picker select **`kimi`** (Kimi K3 — its verdict call 400s deterministically until ENG-1095 lands, which makes it a natural live trigger, no test rigging needed). Give the same multi-step task.
   - **Before this PR:** the agent does the work, then the turn just ends — no final message (silent fake-COMPLETE).
   - **After:** the progress line briefly shows *"Something went wrong — checking in with you..."*, then the agent posts a real message that (a) summarizes what it did, (b) is honest that an internal check failed, (c) states whether it thinks the task is still solvable, and (d) asks how to proceed. It must NOT mention "the verifier".
3. **History integrity:** reply "continue" after the check-in — the agent's next answer must be consistent with the check-in message it just showed (the diagnosis is persisted, not lost).
4. **Logs:** backend log shows `completion-verifier verdict=ERROR … failing toward an honest diagnosis, not a silent COMPLETE`, and per-attempt lines with exception type only (no conversation content).
5. **STUCK-path regression:** a genuinely blocked task (e.g. ask it to read a file that doesn't exist and insist) still produces the stuck diagnosis as before, now ending with a solvability self-assessment.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -234,6 +234,18 @@ def _safe_error_detail(exc: BaseException) -> str:
     return name
 
 
+# Shared closing instruction for every path that hands control back to the
+# user (STUCK, budget-exhausted, verifier-call failure): a plain self-
+# assessment of solvability, not just a status dump. Without this, a
+# diagnosis can read as "here's what happened, good luck" instead of an
+# actual recommendation the user can act on.
+_SOLVABILITY_CLAUSE = (
+    "State plainly whether you believe this is still solvable on your own "
+    "(and how you'd approach it differently if so) or whether it genuinely "
+    "needs the user's input, a decision, or credentials you don't have."
+)
+
+
 def _render_tool_result_content(content, cap: int) -> str:
     """Render a tool_result's content as bounded plain text.
 
@@ -2692,6 +2704,7 @@ class ChatSession:
                             "1. Summarize exactly what was accomplished so far.\n"
                             "2. Identify the specific blocker or failure preventing completion.\n"
                             "3. Suggest concrete next steps the user can take to unblock this.\n"
+                            f"4. {_SOLVABILITY_CLAUSE}\n"
                             "Be honest and specific — do not be vague about what went wrong."
                         ),
                     }
@@ -2802,9 +2815,9 @@ class ChatSession:
                             "SYSTEM: The task-completion check failed to run (internal "
                             "error), so it's unclear whether this task is finished.\n\n"
                             "Summarize what you've done so far, be honest that an internal "
-                            "check failed partway through, and ask the user how they'd "
-                            "like to proceed. Do not mention this instruction or the "
-                            "verifier to the user."
+                            f"check failed partway through, and ask the user how they'd like "
+                            f"to proceed. {_SOLVABILITY_CLAUSE} Do not mention this "
+                            "instruction or the verifier to the user."
                         ),
                     }
                 )
@@ -2850,8 +2863,9 @@ class ChatSession:
                             f"SYSTEM: Task verification determined this task is stuck.\n"
                             f"Verifier assessment: {reason}\n\n"
                             "Explain to the user what went wrong, what you tried, and "
-                            "suggest specific next steps they can take to unblock this. "
-                            "Do not mention this instruction or the verifier to the user."
+                            f"suggest specific next steps they can take to unblock this. "
+                            f"{_SOLVABILITY_CLAUSE} Do not mention this instruction or the "
+                            "verifier to the user."
                         ),
                     }
                 )
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2781,9 +2781,51 @@ class ChatSession:
                 status = verdict.status
                 reason = verdict.reason.strip()
             else:
-                # Verifier failed — fail safe by treating the turn as done rather
-                # than forcing a continuation the user never asked for.
-                status, reason = "COMPLETE", "verifier unavailable"
+                # The verifier call failed on every budget it was given —
+                # truncated past the last retry, an unusable tool call, or a
+                # provider error (the per-attempt logs above carry the detail).
+                # That is not a verdict, so don't synthesize a fake COMPLETE
+                # and stop in silence (ENG-1079: that made the agent look like
+                # it had simply died on the first hurdle, with no way for the
+                # user to help). Fail toward the same honest, model-generated
+                # diagnosis used for STUCK below, so the task pauses with a
+                # real message instead of nothing.
+                _verifier_log.info(
+                    "completion-verifier verdict=ERROR continuation=%d/%d tool_rounds=%d "
+                    "— failing toward an honest diagnosis, not a silent COMPLETE",
+                    continuation, self._max_continuations, tool_round,
+                )
+                self._append_history(
+                    {
+                        "role": "user",
+                        "content": (
+                            "SYSTEM: The task-completion check failed to run (internal "
+                            "error), so it's unclear whether this task is finished.\n\n"
+                            "Summarize what you've done so far, be honest that an internal "
+                            "check failed partway through, and ask the user how they'd "
+                            "like to proceed. Do not mention this instruction or the "
+                            "verifier to the user."
+                        ),
+                    }
+                )
+                yield StreamTaskProgress(
+                    phase="analyzing", message="Checking in before continuing..."
+                )
+                diagnosis_response = None
+                async for event in self.plan_stream_with_recovery(system=system):
+                    yield event
+                    if isinstance(event, StreamComplete):
+                        diagnosis_response = event
+                # Persist the actual message the user just saw — not the stale
+                # pre-verification `reply` the post-loop fallback below would
+                # otherwise re-append, which would leave history (and thus the
+                # model's own memory of what it just told the user) out of sync
+                # with what was streamed.
+                if diagnosis_response is not None:
+                    self._append_history(
+                        {"role": "assistant", "content": diagnosis_response.response.content or ""}
+                    )
+                break
 
             _verifier_log.info(
                 # No `reason` — it's the model's free-text justification, derived
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2815,14 +2815,15 @@ class ChatSession:
                             "SYSTEM: The task-completion check failed to run (internal "
                             "error), so it's unclear whether this task is finished.\n\n"
                             "Summarize what you've done so far, be honest that an internal "
-                            f"check failed partway through, and ask the user how they'd like "
+                            "check failed partway through, and ask the user how they'd like "
                             f"to proceed. {_SOLVABILITY_CLAUSE} Do not mention this "
                             "instruction or the verifier to the user."
                         ),
                     }
                 )
                 yield StreamTaskProgress(
-                    phase="analyzing", message="Checking in before continuing..."
+                    phase="analyzing",
+                    message="Something went wrong — checking in with you...",
                 )
                 diagnosis_response = None
                 async for event in self.plan_stream_with_recovery(system=system):
@@ -2834,9 +2835,23 @@ class ChatSession:
                 # otherwise re-append, which would leave history (and thus the
                 # model's own memory of what it just told the user) out of sync
                 # with what was streamed.
-                if diagnosis_response is not None:
+                diagnosis_text = (
+                    (diagnosis_response.response.content or "").strip()
+                    if diagnosis_response is not None
+                    else ""
+                )
+                if diagnosis_text:
                     self._append_history(
-                        {"role": "assistant", "content": diagnosis_response.response.content or ""}
+                        {"role": "assistant", "content": diagnosis_text}
+                    )
+                else:
+                    # An empty diagnosis would silently recreate the exact
+                    # out-of-sync history this path exists to fix (the
+                    # post-loop fallback re-appends the stale reply). Make it
+                    # visible rather than circular.
+                    _verifier_log.warning(
+                        "verifier-failure diagnosis returned no content — "
+                        "history falls back to the pre-verification reply"
                     )
                 break
 
@@ -2863,7 +2878,7 @@ class ChatSession:
                             f"SYSTEM: Task verification determined this task is stuck.\n"
                             f"Verifier assessment: {reason}\n\n"
                             "Explain to the user what went wrong, what you tried, and "
-                            f"suggest specific next steps they can take to unblock this. "
+                            "suggest specific next steps they can take to unblock this. "
                             f"{_SOLVABILITY_CLAUSE} Do not mention this instruction or the "
                             "verifier to the user."
                         ),
--- a/tests/test_verifier_failsafe.py
+++ b/tests/test_verifier_failsafe.py
@@ -1,0 +1,131 @@
+"""Completion-verifier fail-safe (ENG-1079).
+
+Before this fix, an exception from the verifier's ``generate_object_code``
+call (provider hiccup, malformed structured output, etc.) was silently
+treated as a COMPLETE verdict — the task loop just broke with no message,
+making the agent look like it had died on the first hurdle. It should
+instead behave like the STUCK path: append an honest SYSTEM note and let the
+model generate a real message summarizing progress and asking the user how
+to proceed.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from tests.conftest import make_mock_llm
+
+from anton.core.session import ChatSession, ChatSessionConfig
+from anton.core.llm.provider import (
+    LLMResponse,
+    StreamComplete,
+    StreamTaskProgress,
+    ToolCall,
+    Usage,
+)
+
+
+@pytest.fixture()
+def workspace():
+    # Keep scratchpad venvs inside the repo workspace (pytest runs sandboxed and
+    # can't write to the real home directory).
+    base = Path(__file__).resolve().parent / ".pytest-workspace"
+    base.mkdir(parents=True, exist_ok=True)
+    return MagicMock(base=base)
+
+
+def _text_response(text: str) -> LLMResponse:
+    return LLMResponse(
+        content=text,
+        tool_calls=[],
+        usage=Usage(input_tokens=10, output_tokens=20),
+        stop_reason="end_turn",
+    )
+
+
+def _scratchpad_response(text: str, action: str, name: str, code: str = "") -> LLMResponse:
+    tc_input: dict = {"action": action, "name": name}
+    if code:
+        tc_input["code"] = code
+    return LLMResponse(
+        content=text,
+        tool_calls=[ToolCall(id="tc_1", name="scratchpad", input=tc_input)],
+        usage=Usage(input_tokens=10, output_tokens=20),
+        stop_reason="tool_use",
+    )
+
+
+class _FakeAsyncIter:
+    def __init__(self, items):
+        self._items = items
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        if not self._items:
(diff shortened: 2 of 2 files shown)


alecantu7:
Measured the verifier failure rate in prod Langfuse to check the root cause behind this fix. The fail-safe direction here is right, and the data supports landing it — but two findings change the details. Full write-up in ENG-1081 / ENG-1082; the parts that touch this PR:

**1. Merge order matters.** The verifier failure isn't a rare "provider hiccup" — it's **171 of 687** verdict calls since ENG-716 (24.9%; 37.1% for external users), and it's concentrated: **98.6% of `mindshub_air` calls** (145/147). So on its own, this PR would show *"an internal check failed, how would you like to proceed?"* to those users on nearly **every tool-using turn**. Worth landing the token-ceiling fix (ENG-1081) first or together, so this path stays the rare last resort it's designed to be.

**2. The root cause is a token ceiling, not provider flakiness — and it isn't from ENG-716.** Every failure has `completion_tokens == max_tokens` exactly (256/256), with ~1,300 chars of reasoning prose in `content`, cut off before the tool call. Three aliases narrate before acting (`kimi`, `mindshub_air` = kimi-k2p6, `deepseek`); the other nine emit the forced tool call immediately. Ran the identical verifier payload against all 14 aliases: 9/9 succeed at `max_tokens=256` in 43–115 tokens with zero prose, those 3 fail 0/2 at exactly 256.

The old pre-ENG-716 free-text verifier passed `max_tokens=256` too, and **21% (116/554)** of legacy verdicts were unparseable for the same reason. So `db10fdcf` didn't make the verifier less reliable — it changed what truncation *costs*, from "keep working" to "stop silently". Which is exactly the gap this PR closes; just worth correcting in the PR/ticket framing so the fix lands in the token budget too.

**3. Heads-up for any truncation handling here: `finish_reason` is unusable on this gateway.** It returns `"stop"` while handing back exactly `max_tokens` tokens, for 9 of 11 aliases (all Anthropic- and OpenAI-backed ones; only Gemini reports `"length"`). Filed as ENG-1082 with the two-line fix. Until that ships, truncation has to be detected as `completion_tokens >= max_tokens`.

Empirically, both halves of the ceiling fix are needed — `mindshub_air`, 3 runs each: `256` → 0/3, `256` + "no preamble" prompt → 0/3, `1024` alone → 1/3, `2048` + no-preamble → 3/3. Note it used **1,654 output tokens even with** the no-preamble instruction, so 1024 would not have been enough.

One small thing on the diff itself: the new failure path logs `verdict=ERROR` but doesn't distinguish truncation from a genuine provider error. Given 25% of calls hit this branch, splitting the two would make the rate measurable from logs alone rather than needing a Langfuse scan next time.

alecantu7 (commented):
Adversarial review. The direction is right — failing toward a real message is strictly better than the two options ENG-716 chose between, and sharing `_SOLVABILITY_CLAUSE` across all three hand-back paths is the correct instinct. Seven findings below, ranked. **#1 is a confirmed bug in code this PR edits**; #2 and #3 are substantive; the rest are small.

---

### 1. The history-sync bug you fixed still exists in the two other paths this PR touches — confirmed, with repro

The PR body diagnoses this exactly right: the post-loop fallback "re-appends the stale pre-verification `reply`... leaving history out of sync with what the user actually saw." That fix is applied only to the new verifier-error path. **`STUCK` and budget-exhausted have the identical shape and the identical bug** — and both are edited in this PR for `_SOLVABILITY_CLAUSE`.

Both do: append the `SYSTEM` note → stream the diagnosis → `break`, with no assistant append. So `self._history[-1]["role"] == "user"`, the post-loop guard at `session.py:2795` doesn't fire, and it appends `llm_response.content` — the pre-verification reply, which was **already** appended at `session.py:2615`.

Verified by driving a STUCK verdict through `turn_stream` (probe test, not committed):

```
===== HISTORY AFTER A STUCK VERDICT =====
  0. user      load my db
  1. assistant [text "Running." + tool_use scratchpad]
  2. user      [tool_result]
  3. assistant PRE-VERIFY REPLY: all set!
  4. user      SYSTEM: Task verification determined this task is stuck. ...
  5. assistant PRE-VERIFY REPLY: all set!        <-- duplicated

  DIAGNOSIS present in history?   False
  PRE-VERIFY REPLY count:         2
```

So on the STUCK path the model's next turn sees itself cheerfully saying "all set!" **twice** and never sees the honest diagnosis it just streamed to the user. That's worse than the case this PR fixes: the model is now actively misinformed about its own last message. Budget-exhausted is structurally identical (`session.py:2618-2637` — append note, stream, `break`).

Since all three paths are touched here, suggest fixing all three, and extracting the shared shape rather than a third copy of it — something like `_stream_diagnosis(system)` that appends the `SYSTEM` note, yields the progress event, streams, and persists the streamed text. Three hand-back paths that must stay in lockstep have already drifted once; a helper makes the next drift impossible.

### 2. The swallow is still undiagnosable for operators

`except Exception:` binds no exception, and `verdict=ERROR` logs no cause. This PR fixes the silent failure for the *user* but keeps it silent for whoever has to debug it.

That matters concretely: the reason ENG-1079 sat for 10 days is that four different failure modes (truncation, provider 5xx, schema mismatch, disabled-model 403) all logged the identical `reason="verifier unavailable"`. As of today **~25% of verdict calls take this branch** (ENG-1081 — 171/687 measured in prod, 98.6% on `mindshub_air`), and you still can't tell from logs which of the four you're looking at. `except Exception as exc:` plus `type(exc).__name__` in the log line is a one-liner that would have made this whole investigation a grep.

### 3. Cost: this adds a full planning-model turn over the whole history to a quarter of verified turns

`plan_stream_with_recovery(system=system)` runs on the **planning** model against the entire history — the expensive call (cf. ENG-578). At the measured 24.9% failure rate that's an extra full-history planning turn on a quarter of all verified turns; for a `mindshub_air` user (98.6% failure) it's an extra one on nearly *every* turn, on top of the wasted verdict call.

Two implications:
- **Land ENG-1081 (#277) first or together.** It removes ~99% of the triggers, which turns this from a per-turn tax into the rare fallback it's designed to be.
- Consider latching per session: once the diagnosis path has fired and the *next* verdict call fails the same way, stop paying for a fresh full-history diagnosis every turn. (#277 does this for the truncation retry — same reasoning, same shape.)

### 4. An empty diagnosis silently reverts to the old behaviour

`_append_history` no-ops on empty content (`session.py:682-684`), so if the diagnosis returns empty (refusal, provider hiccup), nothing is appended → the post-loop appends the stale reply → you're back to exactly the out-of-sync history this PR fixes, with no trace. The diagnosis call inherits the default `max_tokens` so truncation is unlikely, but `if diagnosis_response is not None` isn't quite the guard you want — `if diagnosis_response and diagnosis_response.response.content` plus a warning log when it's empty makes the failure visible instead of circular.

### 5. The progress copy says the opposite of what's happening

`"Checking in before continuing..."` — the turn is *stopping* to hand control back, not continuing. Users read this text. Something like `"Something went wrong — checking in with you..."` matches the actual behaviour.

### 6. Test gaps that follow from #1

`test_verifier_failsafe.py` covers the new path well (I like that it pins the call count at 3 and asserts no "Continue working" injection). Nothing covers history sync on the STUCK or budget-exhausted paths — the two places the bug still lives — and nothing asserts the stale reply *isn't* duplicated. If #1 is fixed here, those are the two tests that lock it down.

### 7. Nits

- `tests/test_verifier_failsafe.py:35` uses `Path(__file__).resolve().parent / ".pytest-workspace"`; every other test file uses `parents[1]` (repo root). As written it creates a stray `tests/.pytest-workspace`.
- Two `f"..."` prefixes with no placeholders in the new and edited prompt blocks.
- The new path returns before the existing `completion-verifier verdict=%s ... reason=%s` summary line, so log consumers see a different line shape for this branch. Fine by me, but worth knowing if anything parses that line.

---

**Note for whichever of us merges second:** #277 and this PR both rewrite the same `try/except`. #277 keeps the fail-safe in an `else:` branch after a budget-retry loop, so the honest-diagnosis body here drops into that `else:` directly. Deliberately not stacked.

alecantu7:
Rebased onto current `staging` (was CONFLICTING) and finished per the review plan — taking this over from @torrmal per assignment on ENG-1079.

**What the rebase changed (and why the diff moved):**

- ENG-1081 (#278/#279) landed after this branch was cut and rewrote the verifier call into a budget ladder (`_VERIFIER_TOKEN_BUDGETS = (2048, 4096)` with a truncation retry and per-attempt safe logging). The original `except Exception:` this PR patched no longer exists — **but the silent fake-COMPLETE it targeted survived** as the ladder's `verdict is None` fallback, so the fix is still needed and now lives there: the honest-diagnosis block is grafted onto that branch, covering **both** routes into silence — a provider/schema exception *and* truncation that exhausts every budget (a route that didn't exist when this PR was written).
- Dropped the PR's own `verdict=ERROR` log line — the ladder already logs each failed attempt with `_safe_error_detail` (type + status only, never content); kept one summary line with continuation/tool-round context.
- `_SOLVABILITY_CLAUSE` and its wiring into all three hand-back paths (STUCK, budget-exhausted, verifier-failure) survived the rebase unchanged.
- Added `test_truncation_exhausting_every_budget_also_gets_the_diagnosis` — pins the new second route (both budgets truncated → diagnosis, not silence). Jorge's original test covers the exception route and passes unchanged against the ladder.

**Ordering note from the earlier review comment is now satisfied:** the ceiling fix (ENG-1081) is merged, so this fail-safe fires rarely instead of on ~25% of verdicts — landing it no longer spams narrating-model users. It also becomes the user-facing net for ENG-1095's population: fable's empty verdicts and kimi-K3's tool_choice 400s currently end tasks in silence through exactly this fallback.

**Tests:** full suite 1,357 passed / 17 skipped (2 pre-existing environmental failures in `test_chat_scratchpad` excluded — they fail identically on clean `origin/staging` in this sandbox).

**Security pass:** no new inputs cross a trust boundary (the SYSTEM note is a static string; the diagnosis reply was already model content bound for history), and the failure logging deliberately uses the ENG-1081 `_safe_error_detail` path rather than exception messages, so no conversation content lands in app logs.

alecantu7 (commented):
Adversarial self-review of the rebased head (convention #6), before requesting human review. No blocking findings; three notes and one follow-up, with the reasoning recorded.

**Verified correct:**
- Control flow of the graft: the `else` branch `break`s before the `verdict=%s` summary log, so `status`/`reason` are never read undefined; when a real verdict exists the summary line fires exactly as before. Telemetry keyed on `completion-verifier verdict=` still sees failures (the `verdict=ERROR` line in the else branch).
- All three routes to `verdict is None` land in the diagnosis: provider/schema exception, non-truncated `StructuredOutputError` (this is fable's empty-`{}` verdict, ENG-1095), and truncation exhausting both ladder budgets (new test pins it).
- History stays role-alternating via `_append_history` (SYSTEM note merges/appends correctly; empty diagnosis content is a no-op append).
- No content leaks: the new log line carries counters only; per-attempt failure detail continues through `_safe_error_detail`.
- No textual or logical overlap with #286's regions — merge order between them is free.
- **A strength worth naming:** the diagnosis deliberately avoids the mechanism that just failed — it's a *plain-text* call on the *planning* model, while the verdict was a *forced-tool* call on the *coding* model. Verified empirically for fable (ENG-1095): text calls answer perfectly on the same payload where structured calls return `{}`. So for the ENG-1095 population the diagnosis will reliably render.

**Note 1 — diagnosis stream yielding no `StreamComplete` (rare).** If `plan_stream_with_recovery` ends without a completion (streams that die usually *raise* instead, which takes the turn-level retry path), nothing is appended, and the post-loop fallback re-appends the stale pre-verification `reply` — a duplicated assistant message in history. Considered and left as-is: the alternatives are worse (a synthetic fallback message makes history claim something the user never saw; suppressing the fallback leaves history ending on a user message). Duplicate-but-true beats novel-but-false.

**Note 2 — known consequence for ENG-1095 users until that fix lands.** kimi-K3's verifier call 400s deterministically, so K3 users will now see the honest check-in on essentially every multi-step turn — recurring, but honest and actionable, versus today's silent task death. This is strictly an improvement and also a forcing function for ENG-1095's gateway capability fix.

**Note 3 — UI copy nit.** The progress line says "Checking in before continuing..." but the turn stops after the check-in. Kept Jorge's wording (the message the *model* then generates is what the user actually reads); flagging in case the reviewer prefers "Checking in…".

**Follow-up (not this PR):** the diagnosis stream itself — like the pre-existing STUCK and budget-exhausted streams — is not covered by #286's output-truncation recovery, so an ENG-1042-style silent burn *inside the diagnosis* remains theoretically possible (~0.6% band conditions). Once both PRs are merged, wrapping those three auxiliary streams with `_recover_truncated_stream` is a one-liner each; noting it here so the last silent path is on record rather than forgotten.

anton/core/session.py
```diff
@@ -2692,6 +2704,7 @@ async def _stream_and_handle_tools(
                             "1. Summarize exactly what was accomplished so far.\n"
                             "2. Identify the specific blocker or failure preventing completion.\n"
                             "3. Suggest concrete next steps the user can take to unblock this.\n"
+                            f"4. {_SOLVABILITY_CLAUSE}\n"
```
pnewsam, line 2707: **Non-blocking follow-up:** The new verifier-failure branch correctly captures and persists its streamed diagnosis, but the existing budget-exhausted path here—and the `STUCK` path below—still only yield the diagnosis. Because history then ends on the synthetic user note, the post-loop fallback appends the stale pre-verification reply again, so the next turn does not see the diagnosis shown to the user.

This behavior predates the PR and does not block the ENG-1079 fix, but it would be worth reusing the capture-and-persist logic from the new branch for all three hand-back paths, with regression coverage for `STUCK` and budget exhaustion.

pnewsam (commented):
## Code Review

**Verdict: COMMENT (non-blocking)**

The ENG-1079 change is safe to merge. The verifier-failure fallback now produces a real user-facing diagnosis, persists that diagnosis in history, and covers both provider/schema failures and exhausted truncation budgets. The remaining history-sync issue in the older `STUCK` and budget-exhausted paths predates this PR, so I have left it as follow-up feedback rather than a merge blocker.

Validation on the rebased head:
- Focused fail-safe tests: 2 passed
- Broader session/verifier selection: 127 passed, 2 skipped
- `git diff --check` and Python compilation passed

The fallback’s use of the planning model’s plain-text path is also a good separation from the structured coding-model call that failed.

alecantu7:
Review feedback addressed — thanks @pnewsam. Item-by-item disposition (covering both pnewsam's verdict and the 07-27 review items that were still open after the rebase):

**Fixed in the latest commit:**
- **Empty diagnosis silently reverting to the old behavior** — now guarded: an empty diagnosis logs `"verifier-failure diagnosis returned no content"` and lets the post-loop fallback keep history consistent (stale-but-true beats empty). New test pins it.
- **Progress copy** — `"Checking in before continuing..."` → `"Something went wrong — checking in with you..."`; the turn is stopping to hand control back, and the copy now says so.
- **Nits** — fixture path uses the repo-root `.pytest-workspace` like every other test file; the two placeholder-less `f` prefixes dropped.

**Filed as follow-up (ENG-1155), per the "predates this PR" agreement:**
- **History desync on the STUCK and budget-exhausted paths** + the tests that lock it — same capture-persist-guard treatment this PR gave the verifier-failure path; deliberately after #276/#286 merge since all of it lives in `session.py`.
- **Per-session latch for deterministically failing verifiers** — matters most for the ENG-1095 population (kimi-K3 400s on every verdict call) so they get one diagnosis per session, not one per turn.

**Already resolved by events:**
- "Land ENG-1081 first" — merged (#278/#279); this PR now sits on top of it, so the diagnosis is the rare fallback it was designed to be.
- Log-line shape note — the else-branch logs `verdict=ERROR` with the same `completion-verifier` prefix, so consumers keyed on that prefix still see every outcome.

Full suite 1,359 passing on the new head.

anton/core/session.py
alecantu7: Agreed on all points — filed as [ENG-1155](https://linear.app/mindsdb/issue/ENG-1155/stuck-and-budget-exhausted-diagnoses-desync-history-deterministic) with exactly the shape you describe: extract this branch's capture-persist-guard logic into a shared helper, apply it to the STUCK and budget-exhausted paths, with regression coverage per path (streamed text — not the stale reply — is what history must hold). Scheduled after #276/#286 merge since all three touch this file; the ticket also folds in the per-session latch for deterministically failing verifiers (the kimi-K3/ENG-1095 population), which lives in the same block. Resolving as tracked.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

The follow-up commit addresses the actionable feedback without introducing regressions. Empty diagnosis responses are now observable, the progress copy accurately reflects that the turn is stopping, and the test cleanup is consistent with the repository. The pre-existing `STUCK`/budget-exhausted history issue is appropriately tracked in ENG-1155.

Validation on the updated head:
- Focused fail-safe tests: 3 passed
- Broader session/verifier selection: 128 passed, 2 skipped
- `git diff --check` and Python compilation passed

No remaining merge blockers found.

PR #280, opened 2026-07-28
feat(llm): capture Claude/OpenAI reasoning as a real subtype:reasoning channel (ENG-1109)

torrmal:
Also fixes [ENG-1134](https://linear.app/mindsdb/issue/ENG-1134/completion-verifier-forces-unnecessary-continuations-same-answer-re) — commit `8e28138b`, `fix(verifier): judge final outcome, not intermediate tool errors`. Two tickets landed in this PR; adding the reference so the verifier change is discoverable from ENG-1134 and so a future revert of this PR is known to touch both.

---

## Summary
Anton has no way to distinguish the model's own reasoning/narration ("let me verify that first...", "confirmed — this changed, let me re-check...") from the final answer — both travel through `StreamTextDelta`, the only text-carrying event that exists. That means narration gets permanently baked into the persisted answer instead of showing as a live, disappearing "current thought." The frontend already supports this correctly (built for cowork's ENG-1108) via a `thought.progress` + `subtype: reasoning|thinking` event — the Hermes harness already emits it; Anton just never did.

Confirmed via direct code reading, not assumption:
- `StreamTextDelta` is the only text event in `provider.py`'s `StreamEvent` union.
- Anthropic's extended-thinking content blocks were never parsed (misclassified as plain text, or `thinking_delta`/`signature_delta` silently dropped).
- OpenAI's Responses-API reasoning-summary events were never requested (`reasoning.summary` was missing from the request kwargs) and never handled in the streaming loop.
- The MindsHub-managed gateway path (mdb.ai/mindshub.ai) proxies even **Claude** calls over an OpenAI **chat.completions**-shaped envelope (`FLAVOR_MINDS_PASSTHROUGH`), not the Responses API — so a correct fix needs 3 code paths, not just "enable Claude thinking."

### Changes

- **`provider.py`**: new `StreamReasoningDelta` event, added to `StreamEvent`.
- **`anthropic.py`**: `output_config.effort` (verified real — Claude 4.6/4.7's adaptive-thinking parameter, already sent whenever `reasoning_effort` is configured for a compatible model) already triggers server-side thinking; the gap was purely on the parsing side. Added `content_block_start` handling for `thinking`/`redacted_thinking` block types, and `content_block_delta` handling for `thinking_delta` → `StreamReasoningDelta`. `signature_delta` is intentionally a no-op (cryptographic verification signature, not user-facing text). No new request parameter — reuses the existing, safely-gated `reasoning_effort` config.
- **`openai.py` (Responses API / `FLAVOR_OPENAI`)**: added `"summary": "auto"` alongside the existing `effort` in the `reasoning` kwarg (only sent when `reasoning_effort` is already configured, same gating as today), plus a new `response.reasoning_summary_text.delta` branch.
- **`openai.py` (chat.completions / `FLAVOR_MINDS_PASSTHROUGH` + `FLAVOR_OPENAI_COMPATIBLE_GENERIC`)**: reads `delta.reasoning_content` defensively via `getattr` — the de facto convention several OpenAI-compatible reasoning gateways (DeepSeek, vLLM's reasoning parser) use. **This one is best-effort** — flagging clearly for review since I couldn't verify the actual mdb.ai gateway's wire format from this repo alone.
- **`session.py`**: the synthetic "Thinking..." timer (`reasoning_start`/`reasoning_done`) now also closes on the first `StreamReasoningDelta`, matching its existing behavior for text/complete.

Companion PR in cowork-server (maps `StreamReasoningDelta` → the wire `subtype: reasoning` event, matching hermes_harness's existing shape exactly): mindsdb/cowork-server#241 (opening next).

No frontend changes needed — cowork's ENG-1108 already built the client-side handling for this exact `subtype: reasoning|thinking` shape.

Linear: https://linear.app/mindsdb/issue/ENG-1109/anton-capture-claudeopenai-reasoning-as-a-real-subtypereasoning

## Test plan
- [x] New tests: `TestAnthropicProviderReasoningStream` (thinking_delta → StreamReasoningDelta, signature_delta ignored, effort passed via extra_body), `TestResponsesAPIReasoningSummary` (kwargs gating, reasoning_summary_text.delta mapping), `TestChatCompletionsReasoningContent` (reasoning_content read + graceful absence)
- [x] Full suite: 1377 passed, 17 skipped — 2 pre-existing failures in `test_chat_scratchpad.py` (subprocess sandbox execution issue), confirmed identical on unmodified `staging` via `git stash`, unrelated to this change
- [x] `ast.parse` sanity check on all 4 edited modules; `ruff check` clean (one pre-existing, unrelated `E741` elsewhere in `session.py`)

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -16,6 +16,7 @@ from .provider import (
     ProviderConnectionInfo,
     StreamComplete,
     StreamEvent,
+    StreamReasoningDelta,
     StreamTextDelta,
     StreamToolUseDelta,
     StreamToolUseEnd,
@@ -262,6 +263,11 @@ class AnthropicProvider(LLMProvider):
                                 "json_parts": [],
                             }
                             yield StreamToolUseStart(id=block.id, name=block.name)
+                        elif block.type in ("thinking", "redacted_thinking"):
+                            # Adaptive thinking (triggered by output_config.effort,
+                            # set above when self._reasoning_effort is configured)
+                            # — the model's own reasoning, not the final answer.
+                            blocks[idx] = {"type": "thinking"}
                         else:
                             blocks[idx] = {"type": "text"}
 
@@ -278,6 +284,10 @@ class AnthropicProvider(LLMProvider):
                                 yield StreamToolUseDelta(
                                     id=info["id"], json_delta=delta.partial_json
                                 )
+                        elif delta.type == "thinking_delta":
+                            yield StreamReasoningDelta(text=delta.thinking)
+                        # signature_delta carries only a verification signature
+                        # for the thinking block, no user-facing text — ignored.
 
                     elif event.type == "content_block_stop":
                         idx = event.index
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -20,6 +20,7 @@ from .provider import (
     ProviderConnectionInfo,
     StreamComplete,
     StreamEvent,
+    StreamReasoningDelta,
     StreamTextDelta,
     StreamToolUseDelta,
     StreamToolUseEnd,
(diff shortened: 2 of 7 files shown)


torrmal:
Heads-up for reviewers: this branch now also carries a **second, separate fix** (commit `8e28138b`) for **ENG-1134** — folded in here at the author's request to avoid a proliferation of PRs.

**ENG-1134 (completion verifier over-firing):** the verifier flagged a turn INCOMPLETE whenever an errored tool result appeared anywhere in the transcript — even when the model recovered and answered correctly — forcing a continuation that re-ran generation and re-streamed a whole new answer (they concatenate in one SSE response → the same answer shown 2–3×). The fix rewords the verifier's judgment (schema field description + a new `_VERIFIER_JUDGMENT_RUBRIC` constant) to judge the **final delivered outcome**: an errored/empty tool result only means INCOMPLETE/STUCK when the final answer depends on data that never arrived and wasn't recovered. The hallucinated-success safeguard is preserved.

This is unrelated to the ENG-1109 reasoning-channel work in the rest of the PR — please review the two commits independently. Only the root-cause lever is here; the complementary "replace-not-append on continuation" client change stays tracked in ENG-1134.

pnewsam (approved):
The new reasoning event abstraction and provider mappings are a net improvement, with good targeted coverage across Anthropic, OpenAI Responses, and compatible chat-completions paths. The verifier rubric change also preserves the hallucinated-success safeguard while avoiding unnecessary continuations. Targeted review tests passed (74/74).

PR #282, opened 2026-07-28
Feat/cloud turn

ZoranPandovski:
anton.cloud_turn is the entrypoint that runs a single anton turn headlessly inside an isolated cloud sandbox (one turn per request).  This is how anton runs in the multi-tenant cloud product, as opposed to the local desktop CLI.

What this PR does: it makes that entrypoint safe to run in a shared sandbox where the agent can execute arbitrary code. Concretely, when it builds the agent session, it leaves out the desktop-only behavior that would leak data across tenants — it doesn't load the workspace's .env into the process, doesn't touch personal memory on disk, and doesn't inject datasource credentials into the environment; conversation history comes from the request instead of local disk. It also strips secrets (like the model API key) out of the environment the agent's generated code runs in, isolates the output stream at the OS level so stray prints or library output can't corrupt the JSON protocol, returns short scrubbed error messages instead of raw tracebacks, and restricts the pod to a small reviewed set of tools.

Not included yet: the model API key still has to be provided to the pod via environment variables, and model selection is passed through as-is; both are tightened in a later Gateway change(model calls go through a gateway with a short-lived, run-scoped token; )


tino097 (approved):
LGTM

PR #283, opened 2026-07-28
ENG-1124: persist the scratchpad namespace so state survives a turn

alecantu7:
## What

Makes the scratchpad's namespace actually survive the pad process being replaced, so the agent stops rebuilding work every time the user replies.

Fixes https://linear.app/mindsdb/issue/ENG-1124

## Why

The scratchpad is documented to the model as stateful — *"Variables, imports, and data persist across cells — like a notebook you drive programmatically"* (`prompts.py:86`, and in the tool description it re-reads on every call). That holds **within** a turn and was silently false **across** turns: `cowork-server` builds a new `ChatSession` per user message (`harness.py:232`/`:277`, no cache), `ChatSession.__init__` builds its own `ScratchpadManager` (`session.py:456`), so `get_or_create` spawns a brand-new interpreter for the same pad name. Reattachment isn't just absent — IPC is over stdin/stdout pipes, so it's impossible.

The bridge for that already existed **and was switched on**: `cowork-server` sets `ANTON_SCRATCHPAD_PERSIST_SESSION=true` unconditionally (`harness.py:663`). It had never worked. `SESSION_PATH` defaulted to `/anton_scratchpad_session.pkl` — the filesystem root — and nothing anywhere set the override:

- **macOS desktop** — verified on a dev machine: `uid=501`, `os.access('/', W_OK)` → `False`, `open(...,'wb')` → `OSError errno 30 EROFS: Read-only file system`.
- **hosted** — the container runs `USER anton` uid 1000 (`cowork/Dockerfile:155`), so `/` is root-owned → `EACCES`.

And `_dump_namespace`'s error return was discarded at the call site. So it reported as enabled and did nothing from **2026-05-26** (`cowork-server` `cd01a6f`, titled *"fixed session persistence"*) until now.

### What it cost, concretely

Customer session `93aad8e5` (Mynda Treacy, hosted): 26 cells over ~6 minutes built the whole forecast — trend analysis, per-account method classification, all Jul–Dec values, `master_df`, the monthly P&L, and a fully formatted sheet. Next turn:

```
Lost state: name 'master_df' is not defined
```

Then 15 cells re-doing the identical work, and at 05:12 a *third* derivation of the same analysis. ~20 of the session's 55 cells spent rebuilding. That rebuild is also what inflated her context to ~190k, which is what tripped the TPM ceiling into a 429 storm (ENG-1122).

Session `97b4113c` is the clean proof this is the turn boundary and not pad-name churn: it used **one** pad name throughout and still logged *"State was lost between turns"* with `name 'wb' is not defined`.

## How

- **`local.py` composes the snapshot path**, scoped per conversation **and** per pad. Both segments matter: without the pad segment two named scratchpads overwrite each other; without the conversation segment two conversations in one workspace sharing a pad name read each other's variables — a correctness bug and a confidentiality one.
- **`session_id` is plumbed** `ChatSession` → `ScratchpadManager` → runtime, not passed via `os.environ`. An env var would race across concurrent conversations in one server process. Added to the factory protocol with a default so existing hosts, the remote factory and test doubles keep working.
- **No default path.** Unset now disables persistence *loudly* rather than aiming writes somewhere unwritable.
- **Failures are surfaced on `logs`** — deliberately not `error`, which would feed the consecutive-error circuit breaker and the resilience nudge, turning a snapshot problem into an apparent cell failure. The swallowed error is why this survived two months.
- **100 MB cap** (`ANTON_SCRATCHPAD_SESSION_MAX_BYTES`). Writing after every cell is free while it always fails; once it works, a namespace holding large frames would be rewritten each cell. Over the cap we skip and say so on `logs`.
- **Atomic write** — temp file then `os.replace`, temp name suffixed with the pid. The process is killed abruptly at turn end and by the inactivity watchdog, so writing straight to the real path would eventually leave a torn pickle.
- **`reset()` and `cleanup()` discard the snapshot.** Caught in self-review: without this, `reset` — documented as *"clearing all state"* — would reload the snapshot on restart and silently stop resetting, and `remove` would delete the venv but leave the pickle.
- **Injected helpers excluded** (`get_llm`, `agentic_loop`, `web_search`, `query_minds_data`, `progress`, `sample`, `__builtins__`): rebuilt each boot, and some close over live network clients. Agent-authored functions and data are kept — verified dill handles them, including nested references between agent-defined functions.

### Deliberately out of scope

Normalising the pad name for the **venv** path. It's the same one-line temptation, but it changes which pads share an environment and orphans existing venvs (a measured 179 of them / 2.8 GB on one dev machine). That belongs with its migration in **ENG-1133**.

## Testing

8 new cases in `TestSessionPersistence`:

| case | asserts |
|---|---|
| `test_namespace_survives_restart` | variables, imports and agent-defined functions survive a pad restart |
| `test_namespace_isolated_per_pad_name` | a different pad name sees nothing |
| `test_namespace_isolated_per_session_id` | same pad name in two conversations stays separate, and conversation A still has its own state later |
| `test_oversized_namespace_is_skipped_and_reported` | over the cap → skipped, reported on `logs`, `error` stays `None` |
| `test_no_session_path_reports_instead_of_pretending` | unwritable dir → says so rather than silently no-op'ing |
| `test_reset_still_clears_state` | regression guard for the `reset` semantics above |
| `test_cleanup_removes_the_snapshot` | `remove` leaves no pickle |
| `test_snapshot_path_contains_a_traversing_pad_name` | `../../etc/passwd` as a pad name stays inside the root |

**Verified these fail on `origin/staging` without the source change** (3 of the original 4 fail; the isolation one is a no-regression guard, so passing both ways is correct).

End-to-end check against the product shape — real workspace + conversation id:
```
snapshot: <workspace>/.anton/scratchpad-sessions/7593cdaa-…/forecast.pkl   864 bytes
turn 2 → '(3, 1) workbook'   error: None
```

Full suite: **1341 passed, 17 skipped**. The 2 failures in `test_chat_scratchpad.py` are **pre-existing on `origin/staging`** — verified by stashing this change and re-running.

## Security

Reviewed, since this newly writes a pickle to disk and derives a path from model-chosen input:

- **Pickle load is code execution.** The file is written and read only under the agent's own workspace, alongside the venvs it already controls. The scratchpad already executes agent-authored Python as that user, so this is not an escalation — but the path must not be attacker-steerable, hence the next two points.
- **Path containment.** The pad name and session id are both sanitised to `[A-Za-z0-9._-]` with leading dots stripped, and the resolved path is checked with `relative_to` against the snapshot root before use. Covered by a test using `../../etc/passwd`.
- **Cross-conversation confidentiality.** Per-conversation scoping is what stops one conversation loading another's namespace — including data from a datasource enabled only in the other. Tested, not assumed.
- **Data at rest.** Snapshots live inside the project directory, so existing project deletion (`shutil.rmtree(project.path)`) reclaims them. Not world-readable (default umask, inside the user's own tree). Related: ENG-392 — credentials present in a namespace would now be persisted, which is why retention matters (below).

## Follow-ups (not in this PR)

- **`cowork-server`**: prune `<project>/.anton/scratchpad-sessions/<conversation_id>/` on conversation delete, plus an orphan sweep. Today only *project* deletion reclaims anything, so without this the fix adds a slow leak — especially on desktop, where files sit in the user's home and survive uninstall.
- **ENG-1133**: pad-name normalisation for venv paths + the venv orphan sweep.
- **ENG-1124 Fix 5**: why the ENG-350 `confirm_new_scratchpad` guard never fires — still an open investigation, untouched here.

### DIFF
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import json
 import os
+import hashlib
 import re
 import shutil
 import sys
@@ -113,31 +114,31 @@ def snapshot_dir(venvs_base: Path, session_id: str | None) -> Path:
     return venvs_base.parent / "scratchpad-sessions" / (session or "_no-session")
 
 
+def _pad_filename(pad_name: str) -> str:
+    """An injective, length-bounded filename for a pad.
+
+    `_safe_segment` alone is NOT injective — it maps every unsafe character to `_`, so
+    `'my pad'`, `'my_pad'` and `'my/pad'` all collapse to `my_pad` and would share one
+    snapshot, meaning one pad loads another's namespace. Appending a digest of the
+    ORIGINAL name keeps distinct pads distinct. Also truncated, because a pad name is
+    model-chosen and most filesystems cap a path component at 255 bytes — an
+    over-long name would fail the write instead of saving state.
+    """
+    stem = _safe_segment(pad_name, "scratchpad")[:80]
+    digest = hashlib.sha1((pad_name or "").encode("utf-8")).hexdigest()[:8]
+    return f"{stem}-{digest}.pkl"
+
+
 def snapshot_file(venvs_base: Path, session_id: str | None, pad_name: str) -> Path | None:
     """Snapshot path for one pad, or None if it would escape the snapshot root."""
     base = snapshot_dir(venvs_base, session_id)
-    path = base / f"{_safe_segment(pad_name, 'scratchpad')}.pkl"
+    path = base / _pad_filename(pad_name)
     # Belt for the sanitiser: never hand back a path outside the snapshot root.
     try:
         path.resolve().relative_to(base.resolve())
     except (ValueError, OSError):
         return None
     return path
-
-
-def pads_with_snapshots(venvs_base: Path, session_id: str | None) -> set[str]:
-    """Pad names this conversation has a saved namespace for.
-
-    This is what lets the single-scratchpad guard work across turns. The guard used
-    to consult only an in-memory set on `ChatSession`, which is rebuilt every turn —
-    and since the agent switches pad names precisely *at* turn boundaries, that set
-    was always empty when it mattered, so the guard never fired (ENG-1124 Fix 5).
-    The snapshot directory is the same fact, on disk, already maintained.
-    """
-    try:
-        return {p.stem for p in snapshot_dir(venvs_base, session_id).glob("*.pkl")}
-    except OSError:
-        return set()
 
 
 class LocalScratchpadRuntime(ScratchpadRuntime):
@@ -187,7 +188,7 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
             _venvs_base if _venvs_base is not None else default_venvs_base(workspace_path)
         )
 
-    def _session_snapshot_path(self) -> Path | None:
+    def _session_snapshot_path(self, *, create: bool = False) -> Path | None:
         """Where this pad's namespace snapshot lives, or None if it can't be written.
 
         Scoped per conversation *and* per pad. Both matter: without the pad segment two
@@ -201,8 +202,8 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
         ANTON_SCRATCHPAD_SESSION_PATH unset so the failure is reported rather than silent.
         """
         path = snapshot_file(self._venvs_base, self._session_id, self.name)
-        if path is None:
-            return None
+        if path is None or not create:
+            return path
         try:
             path.parent.mkdir(parents=True, exist_ok=True)
         except OSError:
@@ -554,7 +555,7 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
         # Cowork process can write to. Every save failed and every failure was
         # discarded, so state never survived a turn. This is the only place that knows
         # the pad name, so it is where the path is composed.
-        snapshot = self._session_snapshot_path()
+        snapshot = self._session_snapshot_path(create=True)
         if snapshot is not None:
             env["ANTON_SCRATCHPAD_SESSION_PATH"] = str(snapshot)
         else:
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -104,14 +104,32 @@ def _safe_segment(value: str, fallback: str) -> str:
     return re.sub(r"[^A-Za-z0-9._-]", "_", value or "").strip("._") or fallback
 
 
-def snapshot_dir(venvs_base: Path, session_id: str | None) -> Path:
-    """Namespace-snapshot directory for one conversation.
+def snapshot_dir(venvs_base: Path, session_id: str | None) -> Path | None:
+    """Namespace-snapshot directory for one conversation, or None if it must not persist.
 
-    Falls back to a shared `_no-session` bucket when the host supplies no session
-    id (bare CLI use, tests), where there is only one conversation anyway.
+    Requires a session id, and requires that id to be path-safe as supplied. There is
+    deliberately NO shared fallback bucket:
+
+    * A shared bucket is a confidentiality boundary, not a convenience. Cowork's
+      transient `CredentialProbe` builds a `ChatSession` with **no** session id and
+      parses `DS_*` datasource credentials in the scratchpad — and
+      `ANTON_SCRATCHPAD_PERSIST_SESSION` is process-global, so a probe inherits it once
+      any normal chat has switched it on. With a shared bucket those credentials land on
+      disk under a predictable path and a later probe reusing the pad name reloads them.
+    * `_safe_segment` is not injective, so `tenant/a` and `tenant_a` would resolve to one
+      directory. Rather than transform the id (which would break the path cowork-server
+      computes when it prunes), refuse anything that is not already path-safe. A UUID —
+      what every real host passes — is unchanged, so this enforces the cross-repo
+      invariant instead of merely documenting it.
+
+    The cost is that bare CLI use gets no cross-process persistence. That is the right
+    trade: the CLI is one long-lived process, so its namespace lives in memory anyway.
     """
-    session = _safe_segment(session_id or "", "")
-    return venvs_base.parent / "scratchpad-sessions" / (session or "_no-session")
+    if not session_id:
+        return None
+    if _safe_segment(session_id, "") != session_id:
+        return None
+    return venvs_base.parent / "scratchpad-sessions" / session_id
 
 
 def _pad_filename(pad_name: str) -> str:
@@ -130,8 +148,10 @@ def _pad_filename(pad_name: str) -> str:
 
 
 def snapshot_file(venvs_base: Path, session_id: str | None, pad_name: str) -> Path | None:
-    """Snapshot path for one pad, or None if it would escape the snapshot root."""
+    """Snapshot path for one pad, or None if it must not or cannot be persisted."""
     base = snapshot_dir(venvs_base, session_id)
+    if base is None:
+        return None
     path = base / _pad_filename(pad_name)
     # Belt for the sanitiser: never hand back a path outside the snapshot root.
     try:
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -81,6 +81,65 @@ def _utf8_env(base: "os._Environ[str] | dict[str, str]") -> dict[str, str]:
 _MAX_OUTPUT = 10_000
 
 
+# ── Namespace-snapshot layout (ENG-1124) ────────────────────────────────────
+# One derivation, shared by the writer (the runtime) and the single-scratchpad
+# guard (which needs to know which pads a conversation has already used, across
+# turns). Keeping it in one place is the point: if these two disagreed, the guard
+# would silently stop firing again.
+
+def default_venvs_base(workspace_path: Path | None) -> Path:
+    """Where a workspace's scratchpad venvs live."""
+    if workspace_path is not None:
+        return workspace_path / ".anton" / "scratchpad-venvs"
+    return Path("~/.anton/scratchpad-venvs").expanduser()
+
+
+def _safe_segment(value: str, fallback: str) -> str:
+    """A path-safe version of a model-chosen string.
+
+    Deliberately NOT case-folded or separator-collapsed — that would change which
+    pads share state, and belongs with the venv-path normalisation in ENG-1133.
+    """
+    return re.sub(r"[^A-Za-z0-9._-]", "_", value or "").strip("._") or fallback
+
+
+def snapshot_dir(venvs_base: Path, session_id: str | None) -> Path:
+    """Namespace-snapshot directory for one conversation.
+
+    Falls back to a shared `_no-session` bucket when the host supplies no session
+    id (bare CLI use, tests), where there is only one conversation anyway.
+    """
+    session = _safe_segment(session_id or "", "")
+    return venvs_base.parent / "scratchpad-sessions" / (session or "_no-session")
+
+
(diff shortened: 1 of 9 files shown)


alecantu7 (commented):
## Adversarial self-review — 5 findings, 3 of them real bugs

Went looking for ways to break this rather than re-reading it. Findings below; fixes in follow-up commits.

### 🔴 F1 — an agent variable named `sample` is silently replaced by the injected helper

`_SESSION_EXCLUDE` skips names **by string**, so any agent variable sharing a name with an injected helper is dropped from the snapshot and then re-injected as the helper on the next turn. Verified:

```
turn 1:  sample = [1,2,3]
turn 2:  print(sample)  →  <function sample at 0x1030caca0>
```

No error, no warning — the agent's data becomes a function. `sample` is a *very* plausible variable in data work (`sample = df.sample(100)`), and `progress` is plausible too. This is silent cross-turn data corruption, and worse than the loss it replaces.

**Fix:** exclude by **identity**, not by name — skip `k` only when `ns[k] is` the object we injected. If the agent rebinds the name, persist its value. `__builtins__` and `_anton_explainability_queries` stay unconditional.

### 🔴 F2 — two different pad names can share one snapshot file

`_safe_segment` maps every unsafe character to `_`, so distinct pads collide:

```
'my pad'  -> my_pad.pkl
'my_pad'  -> my_pad.pkl
'my/pad'  -> my_pad.pkl
```

One pad then loads another's namespace — exactly the cross-contamination the per-pad scoping is supposed to prevent. I wrote a test for the *cross-conversation* case and missed the *within-conversation* one.

**Fix:** make the filename injective — append a short hash of the original name so sanitisation can't merge two pads.

### 🟠 F3 — the size cap bounds retention, not cost

`_dump_namespace` writes the **entire** pickle to the temp file, *then* stats it and deletes if over cap. So a 2 GB namespace costs a full 2 GB serialise + write on **every cell** before being thrown away. The cap was added specifically to avoid that cost, and it doesn't. The ticket's claim that "over the cap the write is skipped" is misleading — the write happens and is discarded.

**Fix:** wrap the file object and abort the write once it passes the cap, so the cost is bounded too.

### 🟡 F4 — dead code with a misleading docstring

`pads_with_snapshots()` in `local.py` is defined and **never called** — I wrote it, then replaced its use with the explicit `agent_pads()` record. Its docstring still claims it is *"what lets the single-scratchpad guard work across turns"*, which is now false. Actively misleading to the next reader. Remove it.

### 🟡 F5 — `reset` / `cleanup` create a directory in order to delete a file

`_discard_session_snapshot()` calls `_session_snapshot_path()`, which `mkdir`s the snapshot directory. So discarding a snapshot can *create* the folder it is trying to clean. Harmless, but sloppy — split path computation from directory creation.

### Cross-repo invariant worth a reviewer's eye (also noted on cowork-server#250)

anton **sanitises** the session id into the directory name (`_safe_segment`); cowork-server#250 computes the same directory from a raw `str(conversation_id)`. For UUIDs those are identical, so it works — but it is an undocumented invariant across two repos. If a host ever passes a non-UUID `session_id`, the prune silently stops matching and the leak returns. Documented in both PRs; worth deciding whether to share the rule explicitly.

### Considered and deliberately kept

- **Repeated `logs` notes.** If the dump fails on every cell, the agent sees the note every cell. That is the intent — loud beats silent, and the note is actionable ("write large results to disk"). Accepted cost.
- **`record_agent_pad` read-modify-write is not atomic across processes.** Turns are serialised per conversation and each conversation has its own file, so there is no realistic writer race. Losing a write only returns the guard to its previous behaviour.
- **Pickle load is code execution.** The file sits in the agent's own workspace, and the scratchpad already runs agent-authored Python as that user, so no new trust boundary — but it is why F2 matters and why the path is containment-checked.

alecantu7 (commented):
_Review notes moved to internal discussion._

anton/core/backends/local.py
```diff
@@ -80,6 +82,65 @@ def _utf8_env(base: "os._Environ[str] | dict[str, str]") -> dict[str, str]:
...
+
+
+def snapshot_dir(venvs_base: Path, session_id: str | None) -> Path:
+    """Namespace-snapshot directory for one conversation.
+
+    Falls back to a shared `_no-session` bucket when the host supplies no session
+    id (bare CLI use, tests), where there is only one conversation anyway.
+    """
+    session = _safe_segment(session_id or "", "")
+    return venvs_base.parent / "scratchpad-sessions" / (session or "_no-session")
```
pnewsam: **Blocking:** Please do not persist an unscoped runtime into a shared `_no-session` bucket. Cowork's transient `CredentialProbe` constructs `ChatSession` without `session_id`, parses `DS_*` credentials in the scratchpad, and can inherit `ANTON_SCRATCHPAD_PERSIST_SESSION=true` from an earlier normal chat because that flag is process-global. A later probe in the same workspace using the same pad name can reload those values, and cowork-server#250 explicitly never sweeps this bucket. Disable persistence when `session_id` is absent, or give each probe a unique scoped ID and clean it afterward.

Non-blocking hardening while here: `_safe_segment` is not injective for supplied IDs (`tenant/a` and `tenant_a` map to the same directory). Cowork UUIDs are safe, but a digest or validated identifier would protect other hosts.

anton/core/backends/scratchpad_boot.py
```diff
@@ -710,6 +880,10 @@ def emit(self, record):
 _logging.root.addHandler(_cell_log_handler)
 _logging.root.setLevel(_logging.INFO)
 
+# All injections are done. Record the helper objects so `_dump_namespace` can tell
+# "this is still our helper" from "the agent rebound this name to its own data".
+_INJECTED_HELPERS = {k: namespace[k] for k in _INJECTED_HELPER_NAMES if k in namespace}
```
pnewsam: **Non-blocking correctness follow-up:** This records whatever is currently in `namespace`, not necessarily the helper object defined during this boot. I reproduced `sample = [1, 2, 3]` surviving the first restart, then becoming `<function sample ...>` after the second: on restart one, the loaded list is recorded in `_INJECTED_HELPERS`, so the next dump excludes it as though it were the injected helper. Track the freshly defined helper functions separately from the post-`setdefault` namespace values. This is narrow enough not to block the broader persistence improvement.

anton/core/backends/manager.py
```diff
@@ -58,6 +132,7 @@ async def get_or_create(self, name: str) -> ScratchpadRuntime:
                 coding_api_key=self._coding_api_key,
                 coding_base_url=self._coding_base_url,
                 workspace_path=self._workspace_path,
+                session_id=self._session_id,
```
pnewsam: **Non-blocking compatibility follow-up:** Passing this keyword unconditionally breaks existing custom runtime factories that implement the previous protocol signature: Python raises `TypeError: unexpected keyword argument 'session_id'`. Giving the protocol parameter a default does not adapt existing callable implementations. All in-repo factories are compatible, so I would not hold this PR for it, but external embedders may need a compatibility path or a clearly documented breaking change.

pnewsam (changes requested):
Requesting one focused change before merge: persistence must not put runtimes without a session ID into a shared namespace. In the paired cowork-server integration, transient connector probes omit `session_id`, parse credential material, and can inherit the process-wide persistence flag after a normal chat has enabled it. A later probe using the same workspace/pad name can then reload that state, while cowork-server intentionally exempts `_no-session` from sweeping. Disabling persistence when `session_id` is absent—or assigning probes unique scoped IDs with cleanup—closes the boundary. The helper-shadow and custom-factory comments are real but explicitly non-blocking; overall this remains a strong net improvement.

alecantu7:
Thanks @pnewsam — all four addressed, and the blocking one was a boundary I created rather than inherited.

### Blocking: no persistence without a conversation scope — fixed

You're right, and the reasoning is worse for me than the bug: I had **already** gated the `_agent_pads.json` record on `session_id` after noticing test pollution, and did not carry the same reasoning to the namespace snapshot — the more sensitive of the two. The credential path you describe (`CredentialProbe` → no `session_id` → inherits the process-global persist flag → predictable path → later probe reloads it) is exactly right.

Taken your first option: it refuses to persist when `session_id` is absent. No fallback bucket exists any more. Bare CLI loses cross-process persistence, which I think is the right trade — the CLI is one long-lived process, so its namespace lives in memory regardless.

Also took the non-blocking hardening in the same comment, with one deviation worth flagging: rather than adding a digest to the session segment, it now **refuses** an id that isn't path-safe as supplied. A digest would have changed the directory name and broken the path cowork-server computes when it prunes. Refusing keeps the two repos agreeing *and* closes the `tenant/a` vs `tenant_a` collision — so the cross-repo invariant is now enforced on this side rather than merely documented.

### `_INJECTED_HELPERS` — reproduced, and it's a real bug in my own fix

Confirmed precisely as you described:

```
turn 1: sample = [1,2,3]
turn 2: [1, 2, 3]
turn 3: <function sample at 0x108cf7100>
turn 4: <function sample at 0x1054bb100>
```

The capture read `namespace` *after* `setdefault`, so on the first restore it recorded the agent's restored list as though it were our helper, and the next dump excluded it. My test only did one restart, so it passed — the same failure mode as the original bug, one layer out.

Helpers are now recorded at the point of definition by `_inject_helper`, which records and injects in one call so the two cannot drift. Test extended to three restarts.

### Custom factory compatibility — fixed rather than documented

Cheap enough not to defer, and there's precedent: `session_id` is only passed to factories whose signature accepts it (or that take `**kwargs`), using the same probe pattern cowork-server uses to stay compatible with older anton builds. Resolved once at construction, not per call. Verified against a legacy-signature factory, the current one, and a `**kwargs` one.

### Note on the paired PR

Your `_no-session` sweep exemption on #250 is now guarding something anton never writes. I kept it deliberately — an older anton in the field still creates that bucket — and updated the comment to say why. Also took both follow-ups you left there: `delete_turn` now invalidates the snapshot, and the sweep is project-aware so a conversation move doesn't strand a directory.

Full suite: **1354 passed**; the 2 `test_chat_scratchpad.py` failures are pre-existing on `origin/staging`.

anton/core/backends/local.py
```diff
@@ -114,16 +196,56 @@ def __init__(
...
+        )
+
+    def _session_snapshot_path(self, *, create: bool = False) -> Path | None:
+        """Where this pad's namespace snapshot lives, or None if it can't be written.
+
+        Scoped per conversation *and* per pad. Both matter: without the pad segment two
+        named scratchpads would overwrite each other, and without the conversation
+        segment two conversations in the same workspace that happen to use the same pad
+        name would read each other's variables — a correctness bug and a confidentiality
+        one. Falls back to a shared `_no-session` bucket when the host supplies no
```
pnewsam: **Minor:** This docstring still says unscoped runtimes fall back to a shared `_no-session` bucket, but the new security boundary intentionally disables persistence when `session_id` is absent. Please update it to match the implementation; the stale wording is particularly confusing given the confidentiality fix.

pnewsam (approved):
Re-reviewed at `0acad30a`.

The blocking confidentiality boundary is closed: runtimes without a session ID no longer persist into a shared bucket, and non-path-safe session IDs are refused. The repeated-restart helper-shadow regression is fixed, and the factory signature probe preserves compatibility with older runtime factories.

Targeted validation passed (44 tests), and CI is green. One minor documentation mismatch remains inline; it does not block merge.

anton/core/backends/local.py
alecantu7: Fixed in `10542e425a7d73cb200093376415bd83cebccf90` — docstring now says persistence is refused without a path-safe session id, with no fallback bucket, and points at `snapshot_dir` for the reasoning.

Same staleness existed in `ScratchpadManager._agent_pads_file`, which justified returning `None` in terms of the fallback rather than in terms of `snapshot_dir` refusing outright — corrected in the same commit. Docstrings only, no behaviour change; 91 scratchpad tests still pass.

PR #285, opened 2026-07-29
ENG-764: lazy-load artifact tools behind an artifacts skill

ea-rus:
What

The tool schemas sent to the LLM cost ~5k tokens per request. The five artifact tools (create_artifact, list_artifacts, open_artifact, update_artifact, launch_backend) plus the always-on ARTIFACTS_PROMPT accounted for ~1.8k of that, even in sessions that never produce an artifact.

This PR defers them: they're no longer sent by default and are unlocked on demand when the model recalls the artifacts skill.

How

- New builtin skill artifacts — its one-line description acts as the discovery hook in ## Procedural memory; its body holds the full registration contract (moved out of the system prompt).
- Bundle map (SKILL_TOOL_BUNDLES in tool_defs.py) — maps artifacts / build-html-dashboard / build-fullstack-backend to the five artifact tools.
- On-demand registration (_register_tool_bundle) — wired into both recall paths: model self-recall (handle_recall_skill) and router preload (_inject_recalled_skills). Gated on a bound workspace, deduped, sticky for the session.
- Same-turn availability — tools is rebuilt inside both tool-call loops after dispatch, so tools registered by a recall_skill this round are visible to the follow-up request in the same turn.
- Resume support — _replay_tool_bundles_from_history re-unlocks bundles from prior recall_skill calls in the loaded history, so sticky tools survive a session rebuild (e.g. server restart). Runs after extra tools so the tool order stays cache-stable.
- Shrunk ARTIFACTS_PROMPT — from ~766 to ~213 tokens: now a mandatory pointer to recall_skill("artifacts"); the full contract lives in the skill body.

Impact

- ~1.8k tokens saved per request for sessions that don't touch artifacts.
- Sticky registration → at most one prompt-cache invalidation per session (on activation); stable tail ordering avoids repeat busts.

Fixes: https://linear.app/mindsdb/issue/ENG-764/token-usage-by-tools

### DIFF
--- a/anton/core/llm/prompts.py
+++ b/anton/core/llm/prompts.py
@@ -223,56 +223,17 @@ possible question at once — ask 1-3 at a time, then follow up."""
 
 ARTIFACTS_PROMPT = """\
 ARTIFACTS (applies to all user-facing output):
-Any file you create that the user is meant to open, view, download, or run \
-is an ARTIFACT. Artifacts MUST be registered with `create_artifact` BEFORE \
-any file is written. The tool claims a dedicated folder under \
-`<workspace>/artifacts/<slug>/`, writes `metadata.json` + `README.md` for you, \
-and returns the absolute folder path. Write ALL of the artifact's files into \
-that returned path.
-
-WHEN TO REGISTER:
-- HTML dashboards, charts, reports, infographics → `type="html-app"`, \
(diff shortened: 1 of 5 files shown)


ea-rus:
Evals of this feature shows no significant improvements

PR #286, opened 2026-07-29
ENG-1042: revive main-loop truncation recovery — silent turns at the output-token cap

alecantu7:
## What

A completion can burn its entire `max_tokens` on internal reasoning (or narration) and return **nothing** — no text, no tool call, no error. Measured in the 07-22→07-29 Langfuse sweep: **38 fully-silent generations across 12 users**, 22 from one named customer (Mynda Treacy) whose "it simply does nothing" report this explains.

Linear: [ENG-1042](https://linear.app/mindsdb/issue/ENG-1042)

## Root cause chain

anton already had a recovery for exactly this (`session.py`, two gates), but it gated on `stop_reason in ("max_tokens", "length")` — and the MindsHub gateway reports `finish_reason: "stop"` at the cap (ENG-1082, gateway fix in review as mindshub_inference#367). The recovery was **dead code for every hosted user**.

## Changes

1. **Both main-loop gates now use `looks_truncated()`** (ENG-1081's shared classifier): `output_tokens >= budget` is the evidence; an honest `stop_reason` still counts, a dishonest one no longer disables the net. Works today, with or without the gateway fix.
2. **The retry always CHANGES the call** (ticket Done-when: the three prod failures *are* an unchanged retry loop dying identically 14 min apart):
   - output budget doubled for that one retry (`plan_stream_with_recovery(max_tokens=…)`), and
   - a variant-aware nudge: *continue where you left off* when partial text arrived; *answer now, deliberate less* when the whole budget went to reasoning and nothing was emitted (the old "continue" nudge is meaningless there).
3. **Double failure is surfaced, never silent**: if the retry also burns its doubled budget with nothing visible, the user sees an explicit notice (yielded before the withheld `StreamComplete` so it lands inside the message) and it's recorded in history for the next turn.
4. **`CoreSettings.max_tokens`** (default 8192, `ANTON_MAX_TOKENS`) — ticket Fix 4: `LLMClient.from_settings` already read `getattr(settings, "max_tokens", 8192)` but the field never existed, so the fallback always won. Hosts (cowork-server overlay — follow-up) can now raise the budget.

The truncated-**text** variant (partial deliverable presented as complete) is covered by the same gate: the partial answer is preserved in history and the continue-nudge retry finishes it.

## Tests

`tests/test_session_truncation_recovery.py` (10 tests, all passing; full suite 1,345 passed — the 2 `test_chat_scratchpad` failures are pre-existing on clean `origin/staging`, real-subprocess sandbox issue):

- gate fires on the gateway dialect (`stop_reason "stop"` + `output_tokens == budget`) — the dead-code revival itself
- **retry differs from the failed call in both budget and messages** (explicit ticket Done-when requirement)
- silent vs partial nudge variants; partial text preserved in history
- double silent burn yields a visible failure notice, ordered before the final `StreamComplete`, and lands in history
- healthy completions and at-cap-with-tool-call responses are untouched

## Security pass

Reviewed the changed surface: no new inputs cross a trust boundary (nudges/notice are static strings; model content was already appended to history on this path), no credentials/secrets in the new log lines (token counts and stop_reason only), no new endpoints or deserialization. Nothing flagged.

## Out of scope / follow-ups

- Gateway `finish_reason` fix: ENG-1082 (mindshub_inference#367, in review) — independent, complementary.
- Reserved thinking headroom at the gateway (ticket Fix 1): recorded on the ticket — the effort API (`output_config: {effort}`) has no separate thinking-budget knob, so headroom is implemented reactively here (doubled retry budget) instead.
- cowork-server `UserSettings`/overlay for `max_tokens` so hosted deployments can set it per-user.

## QA Instructions

**Where to test:** a build containing this change (desktop git channel after the weekly release; or the sandbox driver on the branch), pointed at a gateway **that has ENG-1082/#367** — staging today, prod after next week's promotion. ⚠️ **Against a pre-#367 gateway this fix is inert on streamed calls** (the gateway sends no streaming usage — ENG-1158 — and mislabels truncation as "stop"), so QA against prod before the gateway promotes will falsely fail.

1. **Deterministic recovery check (validated on staging 2026-07-30):** run anton with `ANTON_MAX_TOKENS=400`, model `sonnet` via MindsHub, prompt "write a 1200-word explanation of X, no tools". Expect the log `Response truncated at the output budget … retrying once with max_tokens=800`, a continuation in the streamed text, and no silent cut-off. (BYO direct Anthropic key also works — real providers report streaming usage, so the token clause fires without #367.)
2. **Silent-variant nudge:** same setup at `ANTON_MAX_TOKENS=200` on a reasoning-heavy question — if a response burns the budget with nothing visible, the retry log shows `silent=True` and the answer still arrives (or the explicit double-burn notice does).
3. **Double-burn notice:** if both attempts die at the cap, the user-visible message "I ran out of output-token budget twice in a row…" must appear — never a silent end.
4. **Healthy-path regression:** normal tasks at the default 8192 behave identically to before (no retries in logs, no nudges in history).
5. **Config knob (Fix 4):** `ANTON_MAX_TOKENS=16384` is honored (visible in the retry log arithmetic: retries at 32768).
6. **Post-prod re-measure (7 days):** baseline to beat — 38 fully-silent generations / 12 users / week; sonnet-5 40–80k band 1.92% (ENG-1042 sweep protocol; verify candidates by payload, not exact-cap counts alone).

### DIFF
--- a/anton/core/llm/client.py
+++ b/anton/core/llm/client.py
@@ -100,6 +100,17 @@ class LLMClient:
         return self._planning_provider
 
     @property
+    def max_tokens(self) -> int:
+        """Default output-token budget for calls that don't pass their own.
+
+        Exposed so the session's truncation recovery can compare a
+        response's ``output_tokens`` against the budget the call actually
+        ran with (ENG-1042) — the gateway's ``finish_reason`` can't be
+        trusted at the cap (ENG-1082).
+        """
+        return self._max_tokens
+
+    @property
     def coding_provider(self) -> LLMProvider:
         """The LLM provider used for coding/skill execution."""
         return self._coding_provider
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -44,6 +44,7 @@ from anton.core.llm.provider import (
     ToolCall,
     TransientProviderError,
 )
+from anton.core.llm.structured import looks_truncated
 from anton.core.llm.thalamus import (
     ACTION_RESPOND,
     ThalamicDecision,
@@ -88,6 +89,36 @@ from anton.core.settings import CoreSettings
 # Sentinel prefixing a compacted-history summary so later compactions can
 # recognize and update it in place rather than summarize a summary.
 _COMPACTED_MARKER = "[COMPACTED CONTEXT — REFERENCE ONLY]"
+
+# Truncation-recovery nudges (ENG-1042). Two variants of the same failure —
+# the response burned its whole output budget without producing a tool call:
+#
+# - Partial text arrived → the answer was cut mid-flight; ask the model to
+#   pick up where it stopped (the pre-existing recovery message).
+# - Nothing visible arrived → the whole budget went to internal reasoning
+#   (reasoning models share one max_tokens between thinking and answer);
+#   "continue where you left off" is meaningless when nothing was emitted,
+#   so ask for the answer up front instead.
+_TRUNCATED_CONTINUE_NUDGE = (
+    "SYSTEM: Your response was truncated because it exceeded the output token limit. "
+    "Continue exactly where you left off. If you were about to call a tool, "
+    "call it now. If the code you were writing was too long, split it into smaller parts."
+)
+_TRUNCATED_SILENT_NUDGE = (
+    "SYSTEM: Your previous response spent its entire output-token budget before "
+    "producing any visible text or tool call — the user saw nothing. Respond again, "
+    "and lead with the answer or the tool call immediately; keep deliberation brief. "
+    "If the output you are building is large, produce it in smaller parts."
+)
+# Shown to the user when the retry ALSO burns its (doubled) budget with
+# nothing visible. The one outcome this ticket forbids is the turn ending
+# silently.
+_TRUNCATION_FAILURE_NOTICE = (
+    "I ran out of output-token budget twice in a row before completing a "
+    "response, so I could not finish this step. Try splitting the request "
+    "into smaller steps; if this keeps happening, lowering the reasoning "
+    "effort in Settings also helps."
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


alecantu7 (commented):
Adversarial self-review (convention #6). Findings:

1. **Test used private `session._history`** → switched to the public `history` property. Fixed in the follow-up commit.

2. **The gate is still blind on surfaces that report no streaming usage.** `looks_truncated` needs `output_tokens`; the ticket's sweep found 13,546 gateway-handler observations with zero recorded output tokens. On such a surface the token-count clause can't fire — an *honest* `stop_reason` ("length"/"max_tokens") still trips the gate, a dishonest one doesn't. No regression vs. the old gate (which was fully dead there), but full coverage of that surface needs the gateway usage gap fixed (tracked on ENG-1042) and/or ENG-1082 (mindshub_inference#367) so `finish_reason` becomes trustworthy. Deliberately not hand-rolled around here.

3. **Double failure inside the tool loop can be followed by verifier continuations.** After the failure notice, the completion verifier may still judge the turn INCOMPLETE and restart the tool loop (bounded by `max_continuations`). Deliberate: the verifier's continuation injects new context and may genuinely rescue the task; the notice wording says "could not finish this step", not "the task is dead", so it stays honest either way.

4. **The retry raises the budget but does not lower reasoning effort.** Effort is baked into the provider instance at construction (`planning_factory(planning_reasoning_effort)`); per-call effort would mean threading a new parameter through every provider signature. The ticket allows any of {lower effort, raise budget, answer-now nudge} — this PR does the latter two. If doubling proves insufficient on real traffic (re-measure step), per-call effort is the next lever.

5. **`budget * 2` is not clamped to a per-model output ceiling.** A host that sets `ANTON_MAX_TOKENS=32768` gets a 65,536-token retry, which some models reject with a 400. That failure is *visible* (propagates through the existing turn error handling), which beats a silent clamp that quietly reintroduces the cliff; per-model ceilings live in the gateway, not anton. Left as-is knowingly.

alecantu7:
## Live QA finding — one claim in this PR needs correcting, and it changes the rollout ordering

Ran the real `ChatSession` (this branch) against prod MindsHub in an isolated sandbox, with `ANTON_MAX_TOKENS=400` as a deterministic truncation trigger. Result: **a genuinely truncated streamed response (cut mid-sentence at the cap) passed through the new gate undetected** — `StreamComplete` arrived with `usage.output_tokens = 0` and `stop_reason: "stop"`.

Root cause, verified separately: **the gateway never emits usage on streaming responses**, even when `stream_options: {"include_usage": true}` is requested (anton requests it — `openai.py:621`). Confirmed zero usage chunks on prod *and* on the #367-carrying PR env. Filed as **ENG-1158**.

**What this corrects:** the PR body says the token-count clause "works today, with or without the gateway fix." That's true for **non-streaming** calls (the verifier/memory ladders — usage is present there) and for direct-provider users, but **anton's main loop streams**, so on the hosted path the token clause is blind and the recovery rests entirely on the honest `finish_reason` label. Concretely:

- **ENG-1082 / mindshub_inference#367 (merged, staging) is a hard prerequisite for this PR's primary case on hosted, not just a complement.** Post-#367 the label clause fires (`looks_truncated` already honors `"length"`) and the recovery works — but **prod promotion of the gateway must precede or accompany this PR's rollout** for Mynda-class users to benefit.
- **ENG-1158** (gateway honoring `include_usage`) restores the token clause on streams as defence in depth and is the durable fix for client-side accounting.
- The unit tests stand (they exercise both clauses); it's the *field conditions on hosted streams* that were narrower than claimed.

Also verified live in the same sandbox session, for the record: the same driver on the #276 branch produced the full ENG-1079 chain against real kimi — verifier 400 → ladder logs → honest diagnosis streamed ("Something went wrong — checking in with you...", 606-token check-in). That composition works today because the verifier path is non-streaming.

Sandbox notes: fake `$HOME` (no real `~/.anton`/`~/.cowork` read or written), memory off, analytics off, traces tagged `eng-qa-local`, spend ≈ a few thousand tokens on paid aliases.

pnewsam (approved):
Reviewed the complete diff and the truncation-recovery state transitions. The retry changes both the prompt and output budget, preserves partial output in history, and surfaces a second silent burn before the terminal stream event. The new targeted tests and the broader session tests pass locally; no actionable code defects found.

Rollout note: hosted streaming currently reports zero output usage, so production effectiveness still depends on deploying the ENG-1082 gateway `finish_reason` fix first or alongside this change, as documented in the live-QA follow-up.

alecantu7:
## Composed proof on the STAGING gateway (has #367) — the recovery fires where prod couldn't

Follow-up to yesterday's finding. Ran the real `ChatSession` (this branch) with `ANTON_MAX_TOKENS=400` against **`api.staging.mindshub.ai`** (which carries the merged ENG-1082/#367), using a funded staging identity. The exact log line:

```
WARNING anton.core.session: Response truncated at the output budget with no tool call
(output_tokens=0, budget=400, stop_reason=length, silent=False) — retrying once with max_tokens=800
```

**Why this is the key result:** `output_tokens=0` (ENG-1158 — the gateway still sends no streaming usage) means the token-count clause of `looks_truncated()` is blind. The recovery fired **entirely on `stop_reason=length`** — the honest label #367 added. Against **prod** yesterday, the identical run returned `stop_reason=stop` + `output_tokens=0`, so both clauses were dead and the truncation passed through undetected.

So the dependency is now demonstrated, not just reasoned: **#286's hosted-streaming efficacy requires #367 in the same environment.** Confirmed corollaries on staging in the same session:
- ENG-1082 before/after (paid models): staging `finish_reason: "length"` vs prod `"stop"` — sonnet, opus, gpt; non-streaming, streaming final chunk, and `/v1/messages` → `stop_reason: "max_tokens"`. All ✅.
- ENG-1158 reproduces on staging (0 usage chunks with `include_usage`) — the token clause stays blind on streams until that lands.

Rollout consequence unchanged: **promote #367 to prod before/with this PR**, or #286 is inert on the hosted streaming path for exactly the users it targets.

Sandbox: fake `$HOME`, memory off, analytics off, staging key used only against `*.staging.mindshub.ai`, traces tagged `eng-qa-local`.

PR #287, opened 2026-07-29
ENG-1084: memory & extraction subsystems silently did nothing on narrating models — shared truncation ladder for every forced-schema call

alecantu7:
## What

[ENG-1084](https://linear.app/mindsdb/issue/ENG-1084): ENG-1081 fixed the completion verifier's token ceiling; the identical defect lived at **eight** more `generate_object*` call sites, and every failure was swallowed — five of them without a log line. The user-visible result: **memory silently stops working** (no identity facts, no lessons, no consolidation) for anyone on a narrating model, with nothing anywhere saying so.

Confirmed live in prod before building this (see ticket comment): identity extraction dying at exactly its 512 cap and consolidation at exactly its 2048 cap on `mindshub_air`, output pure narration, no tool call, silently swallowed.

## Changes

**Shared ladder** (`anton/core/llm/structured.py`):
- `generate_with_truncation_retry` — the ENG-1081 budget ladder (default 2048→4096), generalized. A retry is bought **only** for `StructuredOutputError.truncated`; a genuine failure (fable's `{}` verdicts at 8 tokens — ENG-1095) is re-raised immediately since a bigger budget can't fix it.
- `no_preamble_instruction(schema)` — the verifier's "call the tool first" suffix, tool name derived from the schema class.

**Eight call sites** route through it (old → new budgets):
| site | old | new |
|---|---|---|
| cortex identity extraction | 512 | 2048→4096 |
| connect_collector credential extraction | 512 | 2048→4096 |
| cerebellum diff pass | 600 | 2048→4096 |
| datasource custom spec | 1024 | 2048→4096 |
| skills draft | 1500 | 2048→8192 |
| consolidator lessons | 2048 | 2048→4096 |
| cortex compaction | 4096 | 4096→8192 |
| share.py session meta | **300** (ticket said "default 8192" — it was 300, tighter than the verifier's infamous 256) | 2048→4096 |

**Logging, content-safe:** every swallow now logs the exception **type name only** — never the message, because a `ValidationError` quotes the rejected model output, and at the credential site that's the user's pasted secrets. Also fixed the two pre-existing cerebellum logs that interpolated `%s exc`.

**Deliberately not touched:** `session.py` — the verifier already has this ladder (ENG-1081), and #276/#286 own that file; zero conflict surface among the three open PRs. Also out of scope, noted on the ticket: `cortex._filter_rules_by_relevance` (plain-text call, fails open — different failure semantics).

## Why fixed raises alone would not have worked

Narration scales with input — consolidation was observed filling 2,048 *exactly* (above ENG-1081's measured 1,654 worst case) because it feeds in a whole scratchpad session. The retry is the load-bearing part; per-site budgets are sized to the site's input.

## Tests

`tests/test_memory_truncation_retry.py` (10 tests; full suite 1,366 passed, the 2 pre-existing environmental `test_chat_scratchpad` failures excluded):
- **Per-subsystem prose-at-the-cap through the real `LLMClient`** (Done-when requirement): identity extraction, cerebellum flush, consolidation, credential extraction — each rescued by the retry and producing its result.
- Ladder semantics: retry raises the budget (never an unchanged re-issue), genuine failures not retried, exhaustion raises rather than returning None.
- **Content-safety asserted against a planted secret**: failure logs carry the type name and must not contain the exception message.
- No-preamble instruction verified in the system prompt the provider actually received.

## Security pass

The credential-extraction site was the sharpest edge: its exception messages can quote pasted secrets — logging is type-name-only there and everywhere else; the two pre-existing content-interpolating log lines were fixed. No new inputs cross a trust boundary; the no-preamble suffix is static text.

## QA Instructions

**Where to test:** a build containing this change — desktop on the **git update channel**, or hosted after the snapshot bake. The affected population is narrating models, so QA runs on **`mindshub_air`** (the free/trial tier default — the real user path; sonnet users were never bitten).

1. **Identity extraction (the confirmed prod failure):** fresh session on `mindshub_air`, send: *"Hi, I'm Alex — I'm a data analyst at Acme and I prefer metric units."* Continue 2–3 more turns (extraction fires in the background). Then verify the facts landed: desktop — the identity section of Global memory (`~/.cowork` memory files) mentions Alex/Acme; or in Langfuse, the generation whose system prompt starts *"Extract identity facts"* now returns a **tool call** (before: 512 tokens of narration, no tool call, exactly at the cap).
2. **Lessons/consolidation:** run a scratchpad task that fails then succeeds (e.g. "read data.csv" when it doesn't exist, then "create it and retry"). After the turn, the consolidation generation in Langfuse (system prompt *"You are a memory consolidation system…"*) returns a tool call within ≤4096 tokens; a later session's context includes the extracted lesson.
3. **Credential extraction:** connect a datasource and paste all credentials as one blob (`host=… user=… password=…`). On `mindshub_air` this previously fell back to field-by-field prompting; now the blob parses in one shot.
4. **Failure visibility (not rescue):** switch to **`kimi`** (K3 — structured calls 400 until ENG-1095): memory features still degrade gracefully, but the backend log now shows `identity-extraction failed (…)` / `consolidation failed (…)` warnings naming the subsystem — exception type only, never message content (verify no user text appears in the log lines).
5. **Regression (sonnet/haiku):** memory features behave exactly as before — these models answer the forced calls in tens of tokens and never touch the retry.

**Langfuse re-measure recipe** (1 week post-deploy): count `mindshub_air` generations at exactly 2048/4096 output with no tool call and a subsystem prompt in the input — baseline was ~95/week at 2048; expect near-zero. Per the ENG-1084 ticket comment: verify candidates by payload, never by exact-cap token count alone (false positives are real).

### DIFF
--- a/anton/commands/datasource/custom.py
+++ b/anton/commands/datasource/custom.py
@@ -9,6 +9,10 @@ from pydantic import BaseModel, Field
 
 from anton.connect_collector import extract_variables
 from anton.commands.datasource.helpers import show_credential_help
+from anton.core.llm.structured import (
+    generate_with_truncation_retry,
+    no_preamble_instruction,
+)
 from anton.core.datasources.datasource_registry import DatasourceEngine, DatasourceField
 from anton.utils.datasources import persist_custom_engine
 from anton.utils.prompt import prompt_or_cancel
@@ -252,11 +256,6 @@ async def handle_add_custom_datasource(
     )
 
     try:
-        from anton.core.llm.structured import (
-            generate_with_truncation_retry,
-            no_preamble_instruction,
-        )
-
         # 1024 sat inside the measured narration range (245–1,654+) — the
         # shared ladder gives narrating models room to reach the forced call
         # (ENG-1084).
--- a/anton/commands/datasource/custom.py
+++ b/anton/commands/datasource/custom.py
@@ -252,11 +252,21 @@ async def handle_add_custom_datasource(
     )
 
     try:
-        spec: _CustomDatasourceSpec = await session._llm.generate_object(
+        from anton.core.llm.structured import (
+            generate_with_truncation_retry,
+            no_preamble_instruction,
+        )
+
+        # 1024 sat inside the measured narration range (245–1,654+) — the
+        # shared ladder gives narrating models room to reach the forced call
+        # (ENG-1084).
+        spec: _CustomDatasourceSpec = await generate_with_truncation_retry(
+            session._llm.generate_object,
             _CustomDatasourceSpec,
-            system="You are a data source connection expert.",
+            system="You are a data source connection expert."
+            + no_preamble_instruction(_CustomDatasourceSpec),
             messages=[{"role": "user", "content": llm_prompt}],
-            max_tokens=1024,
+            subsystem="custom-datasource-spec",
         )
     except Exception:
         console.print(
--- a/anton/commands/share.py
+++ b/anton/commands/share.py
@@ -15,6 +15,11 @@ from typing import TYPE_CHECKING
 from pydantic import BaseModel, Field
 from rich.console import Console
 
+from anton.core.llm.structured import (
+    generate_with_truncation_retry,
+    no_preamble_instruction,
+)
+
 if TYPE_CHECKING:
     from anton.config.settings import AntonSettings
     from anton.core.llm.client import LLMClient
@@ -60,11 +65,6 @@ async def _generate_meta(
     session_id: str,
 ) -> tuple[str, str]:
     try:
-        from anton.core.llm.structured import (
-            generate_with_truncation_retry,
-            no_preamble_instruction,
-        )
-
         conversation_text = _format_history_for_llm(history)
         # 300 was tighter than the verifier's infamous 256 (ENG-1081), for a
         # call asked to produce a whole-session distillation — narrating
--- a/anton/commands/share.py
+++ b/anton/commands/share.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import getpass
 import json
+import logging
 import os
 import re
 import tempfile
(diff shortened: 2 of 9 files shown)


alecantu7 (commented):
Adversarial self-review (convention #6):

1. **Fixed — inconsistent import placement.** The three command sites used function-level imports of the shared helpers while the memory modules imported at top; no cycle risk exists (`structured.py` imports only `provider`), so all are top-level now.

2. **Accepted cost — background budgets rise.** Identity extraction (fires every ~5 turns) can now spend up to 4,096 output tokens instead of 512. This only pays out when a model actually narrates — non-narrating models answer these calls in tens of tokens and `max_tokens` costs nothing unused — and a truncated-then-swallowed call at 512 was *already* paying 512 for zero value on every pass. Net cost on narrating models buys a working memory subsystem.

3. **Known limitation — the ladder can't rescue what isn't truncation.** fable's `{}` verdicts (ENG-1095) and kimi-K3's tool_choice 400s still fail these subsystems — now with a WARNING log naming the subsystem instead of silence, which is this ticket's contract ("a silent memory subsystem shows up in logs"). Actually *working* on those models is ENG-1095's gateway fix.

4. **Deliberate non-migration — the verifier's own inline ladder stays.** `session.py`'s verifier duplicates this pattern (it's where the pattern came from, ENG-1081). Migrating it onto the shared helper belongs in a follow-up after #276 and #286 merge — three open PRs editing the same block is how conflicts happen. Noting per convention #7: the shared helper is the canonical pattern going forward; the verifier is the one known remaining inline copy.

5. **Ticket-table correction recorded**: `share.py`'s budget was **300** (not "default 8192" as the ticket table said) — tighter than the verifier's infamous 256, for a whole-session distillation call.

6. **Verified against regression**: 1,366 passing including 10 new tests; the content-safety tests assert a planted secret in exception messages never reaches `caplog`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the shared truncation-retry helper and all eight call-site migrations. Retries remain limited to confirmed truncation, non-truncation failures preserve the existing fallback behavior with content-safe logging, and the per-site budget overrides match their workloads. No actionable findings.

### Validation

- 237 targeted regression and surrounding tests passed.
- `git diff --check` and Python compilation passed.

The new tests cover successful retry, non-truncated failure, retry exhaustion, prompt propagation, subsystem recovery, and secret-safe logging.

PR #288, opened 2026-07-30
fix(llm): unwrap array-shaped provider errors; 404 → model-unavailable (ENG-1145)

pnewsam:
## Problem

A Gemini chat turn on a dead model surfaced as *"Server returned 404 — the LLM endpoint may be temporarily unavailable. Try again in a moment"* ([ENG-1145](https://linear.app/mindsdb/issue/ENG-1145)) — misleading on two counts: the failure is permanent (the model isn't served), not a transient outage, and the provider's real message was thrown away.

Root cause: Google's Gemini OpenAI-compat endpoint wraps chat errors in a single-element **array** (`[{"error":{…}}]`) while OpenAI uses a bare object. The SDK stores whatever it parsed, so `exc.body` is a `list` — and every structured branch in `_raise_for_status_error` (auth / quota / model) reads it as a dict, misses, and falls through to the generic message. (Pinned by a new test: `assert isinstance(exc.body, list)`.)

## Changes

- Unwrap a single-element list at the top of `_raise_for_status_error`, so all downstream checks see the error object regardless of provider body shape.
- Map a **404** to `ModelUnavailableError` (permanent — "switch models in Settings") instead of the transient "try again" `ConnectionError`, surfacing the provider's own message (e.g. Google's "no longer available to new users").

## Test plan

- `tests/test_status_error_mapper.py` (+2, real SDK via `httpx.MockTransport`): object-body 404 → `ModelUnavailableError` (not transient, no "temporarily unavailable"); Gemini array-body 404 unwrapped and message surfaced.
- `uv run pytest -k "openai or status or provider or llm or transient"` → 240 passed, 4 skipped.

## Related / rollout

Part of a 3-repo fix (cowork-server defaults + live overlay, cowork onboarding errors). This change reaches the desktop app only once anton is repinned in cowork-server — folds into the anton staging promotion tracked in ENG-1159.

### DIFF
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -14,6 +14,7 @@ from anton.utils.datasources import scrub_credentials
 from .provider import safe_parse_tool_input
 from .provider import (
     ContextOverflowError,
+    EndpointConfigurationError,
     LLMProvider,
     LLMResponse,
     ModelUnavailableError,
@@ -155,18 +156,30 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
                 or "does not exist" in msg_l
             )
         )
-        detail = f" {provider_msg.strip()}" if isinstance(provider_msg, str) and provider_msg.strip() else ""
+        # Provider message as a leading-space fragment with a normalized
+        # terminator, so the appended copy reads cleanly whether or not the
+        # provider punctuated its own message (Gemini's ends in a period; a raw
+        # proxy/FastAPI detail may not). Named `suffix`, not `detail`: the `detail`
+        # bound above is the FastAPI 429 detail, and reusing the name would leak
+        # this 404-local value into any branch added after this block.
+        clean = provider_msg.strip() if isinstance(provider_msg, str) else ""
+        if clean and clean[-1] not in ".!?":
+            clean += "."
+        suffix = f" {clean}" if clean else ""
         if model_specific:
+            reason = f":{suffix}" if suffix else "."
             raise ModelUnavailableError(
-                f"The model '{model}' isn't available:{detail} Switch models in Settings.",
+                f"The model '{model}' isn't available{reason} Switch models in Settings.",
                 code="model_not_found", model=model,
             ) from exc
-        # Not model-specific → almost always a misrouted/misconfigured endpoint.
-        # Permanent for this request (retrying won't help), but the remedy is the
-        # endpoint config, not the model — so don't send the user to switch models.
-        raise ConnectionError(
+        # Not model-specific → almost always a misrouted/misconfigured endpoint
+        # (bad base URL, missing /v1, proxy route). Permanent for this request,
+        # but the remedy is the endpoint config, not the model — a distinct type
+        # so the CLI defaults it to `setup` (fix provider/endpoint), not `retry`,
+        # and never to "switch models" (ENG-1145 review).
+        raise EndpointConfigurationError(
             f"The model endpoint returned 404 — check the endpoint URL and model "
-            f"configuration.{detail}"
+            f"configuration.{suffix}"
         ) from exc
 
     # Retryable provider/infra failures — overload/api_error (incl. the mid-stream
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -131,25 +131,43 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
             )
         raise ModelUnavailableError(msg, code=code, model=model) from exc
 
-    # A 404 from an OpenAI-compatible endpoint means the model id isn't served
-    # here — retired, not enabled for this key, or never existed — which is
-    # permanent for this request, NOT the transient outage the generic branch
-    # below implies ("try again in a moment"). Surface the provider's own copy
-    # so the user switches models instead of retrying; Google's is e.g.
-    # "models/gemini-2.5-flash is no longer available to new users." (ENG-1145)
+    # A 404 needs care: this mapper is shared by direct OpenAI, Gemini, MindsHub,
+    # Azure, and arbitrary OpenAI-compatible endpoints, so a bare 404 does NOT by
+    # itself mean the model is missing — a wrong base URL, a missing ``/v1``, a
+    # reverse-proxy route, or an unsupported API path all 404 too, and there
+    # "switch models" is the wrong remedy. Only treat it as model-not-found when
+    # the structured body actually points at the model: OpenAI's
+    # ``code="model_not_found"``, or a model-oriented message (Gemini's
+    # ``status="NOT_FOUND"`` with "models/<id> is not found / no longer
+    # available"). Everything else is surfaced as an endpoint/configuration
+    # failure carrying the provider's own words. (ENG-1145 review)
     if exc.status_code == 404:
         provider_msg = envelope.get("message") or body.get("message")
-        if isinstance(provider_msg, str) and provider_msg.strip():
-            msg = (
-                f"The model '{model}' isn't available: {provider_msg.strip()} "
-                "Switch models in Settings."
+        msg_l = provider_msg.lower() if isinstance(provider_msg, str) else ""
+        status_str = str(body.get("status") or envelope.get("status") or "").upper()
+        model_specific = code == "model_not_found" or (
+            "model" in msg_l
+            and (
+                status_str == "NOT_FOUND"
+                or "not found" in msg_l
+                or "not available" in msg_l
+                or "no longer available" in msg_l
+                or "does not exist" in msg_l
             )
-        else:
-            msg = (
-                f"The model '{model}' isn't available at this endpoint. "
-                "Switch to a supported model in Settings."
-            )
-        raise ModelUnavailableError(msg, code="model_not_found", model=model) from exc
+        )
+        detail = f" {provider_msg.strip()}" if isinstance(provider_msg, str) and provider_msg.strip() else ""
+        if model_specific:
+            raise ModelUnavailableError(
+                f"The model '{model}' isn't available:{detail} Switch models in Settings.",
+                code="model_not_found", model=model,
+            ) from exc
+        # Not model-specific → almost always a misrouted/misconfigured endpoint.
+        # Permanent for this request (retrying won't help), but the remedy is the
+        # endpoint config, not the model — so don't send the user to switch models.
+        raise ConnectionError(
+            f"The model endpoint returned 404 — check the endpoint URL and model "
+            f"configuration.{detail}"
+        ) from exc
 
     # Retryable provider/infra failures — overload/api_error (incl. the mid-stream
     # HTTP-200 case), 5xx, or a plain 429 — get backed off and retried by the
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -83,7 +83,16 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
             "Invalid API key — check your OpenAI API key configuration."
         ) from exc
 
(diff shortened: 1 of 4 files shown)


anton/core/llm/openai.py
```diff
@@ -122,6 +131,26 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
             )
         raise ModelUnavailableError(msg, code=code, model=model) from exc
 
+    # A 404 from an OpenAI-compatible endpoint means the model id isn't served
+    # here — retired, not enabled for this key, or never existed — which is
+    # permanent for this request, NOT the transient outage the generic branch
+    # below implies ("try again in a moment"). Surface the provider's own copy
+    # so the user switches models instead of retrying; Google's is e.g.
+    # "models/gemini-2.5-flash is no longer available to new users." (ENG-1145)
+    if exc.status_code == 404:
```
pnewsam, line 145: **Minor:** This classifies every 404 from every OpenAI-compatible backend as a missing model. A bad base URL, missing `/v1`, reverse-proxy route, or unsupported API path can also return 404; in those cases `ModelUnavailableError` and “Switch models” give the wrong remedy. Could we require model-specific evidence from the structured body/message (for example Gemini’s `NOT_FOUND` plus a model-oriented message), and otherwise surface the provider message as a generic endpoint/configuration failure? A generic endpoint-404 test would pin the distinction.

pnewsam (commented):
The array-body normalization is the right fix and the new tests pin the Gemini SDK behavior well. I found one narrower classification issue to address or consciously accept: this mapper is shared by direct OpenAI, Gemini, MindsHub, Azure, and arbitrary OpenAI-compatible endpoints, so a bare 404 does not always establish that the model is missing.

anton/core/llm/openai.py
pnewsam: Good catch — fixed in 4742022a. A 404 now maps to ModelUnavailableError only with model-specific evidence: OpenAI's `code=model_not_found`, or a model-oriented message (Gemini `status=NOT_FOUND` / "models/<id> is not found | no longer available | does not exist"). Everything else surfaces as an endpoint/configuration failure carrying the provider's own message. Added generic-endpoint-404 tests (FastAPI `{detail}` + HTML body) and an OpenAI model_not_found-code test.

pnewsam:
Addressed the review in 4742022a: a 404 now maps to `ModelUnavailableError` ("switch models") only with model-specific evidence — OpenAI's `code=model_not_found`, or a model-oriented message (Gemini `status=NOT_FOUND` / "is not found | no longer available | does not exist"). Anything else (bad base URL, missing `/v1`, proxy route) surfaces as an endpoint/configuration failure carrying the provider's own message. Added generic-endpoint-404 tests (FastAPI `{detail}` + HTML body) and an OpenAI model_not_found test; 24 mapper tests pass.

alecantu7 (commented):
Reviewed as one of the three ENG-1145 PRs (with cowork-server#256 and cowork#531). Verified at `4742022`: `tests/test_status_error_mapper.py` → 24 passed.

**The root-cause diagnosis is right and the array unwrap is in the correct place** — ahead of the `envelope`/`detail` extraction, so all three downstream checks (auth / quota / model) see the unwrapped dict. Pinning it with `assert isinstance(exc.body, list)` on an SDK-constructed error rather than a hand-built `APIStatusError` is the right instinct; hand-built fixtures are how ENG-747's card sat dead in production.

**The 404 handling is the best part of this PR and I want to be explicit about it.** A bare 404 → "switch models" was exactly the objection I'd have raised — this mapper is shared by direct OpenAI, Gemini, MindsHub, Azure and arbitrary compat endpoints, and a missing `/v1`, a reverse-proxy route, or a typo'd base URL all 404 too. Requiring model-specific evidence (`code == "model_not_found"`, or a model-oriented message with `NOT_FOUND`/"not found"/"no longer available") before sending the user to Settings pre-empts it. Two notes on it inline.

**One cross-repo gap, and it's the thing I'd most want addressed** — not necessarily here, but decided somewhere.

The other two PRs in this set both establish the rule that **Gemini returns a bad key as 400, not 401/403**, and both act on it: cowork#531's `classifyOpenAICompatibleResult` and cowork-server#256's `validate_openai_compatible` label an auth-shaped message at any status as "Invalid API key". anton has no 400 branch at all — `_raise_for_status_error` handles 401 and then falls through.

So the coverage is asymmetric in the worst direction: **onboarding and Settings now name a bad Gemini key correctly, but a key that is revoked or rotated mid-conversation still surfaces generically** — and that's the path where the user loses in-flight work rather than just failing a validation click. Same underlying quirk, same provider, three layers, two of them fixed.

If you'd rather not widen this PR, that's reasonable — but it's worth a line on ENG-1145 saying the runtime path is knowingly left, because "we fixed the Gemini 400" will read as complete otherwise.

anton/core/llm/openai.py
```diff
@@ -122,6 +131,44 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
...
+        detail = f" {provider_msg.strip()}" if isinstance(provider_msg, str) and provider_msg.strip() else ""
+        if model_specific:
+            raise ModelUnavailableError(
+                f"The model '{model}' isn't available:{detail} Switch models in Settings.",
+                code="model_not_found", model=model,
+            ) from exc
+        # Not model-specific → almost always a misrouted/misconfigured endpoint.
+        # Permanent for this request (retrying won't help), but the remedy is the
+        # endpoint config, not the model — so don't send the user to switch models.
+        raise ConnectionError(
```
alecantu7: The comment says "Permanent for this request (retrying won't help)" — but a bare `ConnectionError` is precisely the class the CLI treats as transient. `chat.py:1980`:

```python
default=(
    "setup" if isinstance(exc, ModelUnavailableError)
    else ("retry" if isinstance(exc, ConnectionError) else "setup")
)
```

So a misrouted-endpoint 404 lands on **`retry` as the default choice** — the one thing the comment says won't work. And it's the case where `setup` is actually the right destination: `handle_setup_models` is the provider/key/endpoint flow, which is where a wrong base URL gets fixed.

I checked how far this reaches before flagging it: the session loop's automatic backoff keys on `TransientProviderError` (`session.py:2098`, gated on `session_backoff`), **not** on bare `ConnectionError`, so this does *not* silently auto-retry a doomed request. The consequence is limited to the interactive prompt's default, which makes it small — but it's small and wrong rather than small and right, and the fix is a line.

Options, in preference order: raise a distinct subclass (e.g. an `EndpointConfigurationError(ConnectionError)`) and default it to `setup`; or add it to the `setup` side of that ternary. The concern in your comment — that `setup` reads as "switch models" — is a copy problem in the prompt string, not a routing one, and worth solving there rather than by routing a permanent failure to retry.

anton/core/llm/openai.py
```diff
@@ -122,6 +131,44 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
             )
```
alecantu7, line 132: Two small things in this block, neither blocking.

**`detail` is reused for a different meaning.** It's bound at the top of the function as `body.get("detail") or envelope.get("detail")`, then rebound here as a pre-formatted `" <provider message>"` fragment. Both 404 branches raise, so nothing downstream sees the clobbered value and it's correct today — but a future `elif` added after this block would inherit the shadowed one. A separate name (`suffix`, `provider_detail`) costs nothing.

**The message spacing depends on `detail` carrying its own leading space.** `f"...isn't available:{detail} Switch models in Settings."` reads fine both ways, but with a provider message that already ends in a period you get `"… new users. Switch models in Settings."` and without one `"… available to new users Switch models"`. Worth normalising the terminator rather than relying on Google's punctuation.

anton/core/llm/openai.py
pnewsam: Good call, fixed in 34b98c78. That non-model 404 now raises a new EndpointConfigurationError(ConnectionError), and chat.py defaults it to setup rather than retry, ie. a wrong base URL lands on handle_setup_models instead of re-sending a doomed request. It stays a ConnectionError subclass so existing except sites are unaffected, and since the session backoff keys on TransientProviderError it still won't auto-retry. Went with the distinct-subclass option you listed first; the endpoint-404 tests now assert the type.

anton/core/llm/openai.py
pnewsam: Both fixed in 34b98c78. Renamed the 404-local fragment to suffix so it no longer shadows the detail bound at the top of the function. And normalized the terminator, ie. the provider message gets a period if it lacks one, so 'no longer available' no longer runs into 'Switch models'. Added a no-terminator test case.

pnewsam:
On the cross-repo gap you flagged, ie. anton's runtime path has no 400 branch so a Gemini key revoked or rotated mid-conversation still surfaces generically and defaults to retry: agreed, and you're right it would read as complete otherwise. I've left it out of this PR to avoid widening the coordinated release, and written it up on ENG-1145 as a follow-up, ie. a 400/401 auth-shaped branch that surfaces the invalid-key copy and defaults to setup (the 401 routes to retry today for the same reason, so both move together). The two inline notes on the 404 handling are addressed in 34b98c78.

alecantu7 (commented):
Verified at `34b98c7` — `test_status_error_mapper.py` → **25 passed**. Both points closed, and the first one better than I asked for.

`EndpointConfigurationError(ConnectionError)` is the right shape: the CLI now defaults a misrouted 404 to `setup`, existing `except ConnectionError` sites are untouched because it subclasses, and the session backoff still can't auto-retry it since that keys on `TransientProviderError`. The tests assert the *type*, not just the message, so the routing can't silently regress. And the `detail` → `suffix` rename plus the terminator normalization — with a no-punctuation test case — closes the shadowing and the run-on sentence.

**Your 401 observation is correct, and it's worth stating plainly on the ticket because it's the same bug sitting next to the one we just fixed.** I checked: `openai.py:82` raises a bare `ConnectionError` for a 401, and `chat.py`'s ternary defaults any bare `ConnectionError` to `retry` — so *today*, an invalid API key tells the user their best option is to retry, which cannot possibly help. That's pre-existing and not this PR's doing, but it's now the only remaining case where the default contradicts the failure's own permanence. Agreed both should move together in the follow-up.

Also agreed on deferring the 400 auth-shaped branch: widening a three-repo coordinated release for it would be the wrong trade, and it's written up. My only ask was that it not read as complete, and the ENG-1145 note does that.

One thing I'd flag for whoever merges the set: the follow-up is now load-bearing for a real user path. Onboarding and Settings name a bad Gemini key correctly as of #256/#531, so a user whose key is fine at setup and revoked later gets a *worse* experience than one who typed it wrong — generic copy plus a retry suggestion. Worth not letting that follow-up drift.

alecantu7 (approved):
Approving at `34b98c7`. `test_status_error_mapper.py` → 25 passed.

Both review points closed, and the 404 handling is the part I'd highlight for anyone reading this later: requiring model-specific evidence before saying "switch models" is what keeps a wrong base URL or a missing `/v1` from being blamed on the model, and `EndpointConfigurationError` makes that distinction typed rather than message-shaped.

Two things deliberately **not** in this PR, both agreed and tracked on ENG-1145 — recording them here so the merge doesn't read as closing the class:

1. **No 400 branch.** Gemini answers a bad key with 400, and cowork#531 / cowork-server#256 both act on that. anton doesn't, so a key revoked or rotated *mid-conversation* still surfaces generically. Net effect after this set ships: a mistyped key is handled better than a revoked one.
2. **The 401 default.** `openai.py:82` raises a bare `ConnectionError`, which the CLI ternary defaults to `retry` — so an invalid key currently suggests the one action that cannot help. Pre-existing, and the same shape as the endpoint-404 default this PR just fixed.

Neither blocks the merge and widening a three-repo coordinated release for them would be the wrong trade. But (1) is a live user path, so the follow-up shouldn't drift.

PR #290, opened 2026-07-30
ENG-1166: keep publish/datasource tools on resumed + rebuilt sessions

alecantu7:
## What

`rebuild_session` (`anton/chat_session.py`) — used by **`/resume`** (`commands/session.py` `restore_session`) **and, per its docstring, after any settings/model change** — built its `ChatSessionConfig` with **no `tools=` argument**. Since `config.tools` feeds `ChatSession._extra_tools` (registered in `_build_tools`), a rebuilt session registered only *core* tools and silently dropped **`publish_or_preview`** and **`connect_new_datasource`**.

Symptom: publishing from a resumed session fails with `Tool 'publish_or_preview' failed: Tool publish_or_preview not found`, while a **fresh** session (`chat.py:1360`, which *did* pass `tools=[CONNECT_DATASOURCE_TOOL, PUBLISH_TOOL]`) works. Reported in ENG-798 (misdiagnosed by the agent as a "platform outage" — see ENG-1165).

## Root cause = drift

Two session builders, one maintained tool list:
- `chat.py` (fresh) → had the tools.
- `chat_session.py::rebuild_session` (resume / settings-change) → didn't.

## Fix

Extract the extra-tool list to a single source — `DEFAULT_SESSION_TOOLS` in `anton/tools.py` — and use it in **both** builders (`list(...)` so each session gets its own copy). Now they can't diverge again. Adds `tests/test_session_tools.py`: asserts the list contains both tools and that **both** builders reference the shared constant (drift guard).

## Verification

- All three files compile; `DEFAULT_SESSION_TOOLS` → `['connect_new_datasource', 'publish_or_preview']`.
- New regression tests pass (2 passed).
- Confirmed live during triage: publish backend was up the whole time (`view.mindshub.ai/list`→200, `auth`→valid) — the failure was purely the missing tool registration on resume.

## Scope / not covered

- **Desktop (`cowork-server`) is unaffected** — it never calls `rebuild_session`, and its harness (`anton_harness/harness.py`) always registers a (cowork-flavored) publish tool. Verified.
- **Heads-up, separate audit:** there's a *third* builder, `runtime.py::build_chat_session`, which defaults `tools=[]` unless a caller passes `extra_tools`. Its callers (cron/headless/scheduled?) should be audited to confirm they pass the tool list — out of scope here, worth a follow-up.
- A full `rebuild_session` integration test would need to mock its network call (`refresh_knowledge`); the included source-level drift guard covers the actual failure mode cheaply. Follow-up welcome.

## Repro

1. Fresh `anton` session → publish → works.
2. `anton` → `/resume` any session → publish → `Tool publish_or_preview not found` (before this fix).
3. Likely also: fresh session → change model mid-session → publish → same (same `rebuild_session` path).

Fixes ENG-1166. Origin ENG-798. Companion (agent fabrication/diagnosability) ENG-1165.

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -68,7 +68,7 @@ from anton.commands.skills import (
     handle_skills_list,
 )
 from anton.commands.share import handle_share_export, handle_share_import, handle_share_status, handle_share_history
-from anton.tools import CONNECT_DATASOURCE_TOOL, PUBLISH_TOOL
+from anton.tools import DEFAULT_SESSION_TOOLS
 from anton.utils.prompt import (
     prompt_or_cancel,
     prompt_minds_api_key,
@@ -1357,7 +1357,7 @@ async def _chat_loop(
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
         output_dir=settings.artifacts_dir,
-        tools=[CONNECT_DATASOURCE_TOOL, PUBLISH_TOOL],
+        tools=list(DEFAULT_SESSION_TOOLS),
         web_search_enabled=settings.web_search_enabled,
         web_fetch_enabled=settings.web_fetch_enabled,
     ))
--- a/anton/chat_session.py
+++ b/anton/chat_session.py
@@ -89,6 +89,7 @@ def rebuild_session(
     from anton.core.llm.client import LLMClient
     from anton.chat import ChatSession
     from anton.core.session import ChatSessionConfig
+    from anton.tools import DEFAULT_SESSION_TOOLS
 
     state["llm_client"] = LLMClient.from_settings(settings)
 
@@ -118,6 +119,10 @@ def rebuild_session(
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
         output_dir=settings.artifacts_dir,
+        # ENG-1166: without this, resumed / post-settings-change sessions
+        # register only core tools and silently lose publish_or_preview +
+        # connect_new_datasource. Mirror the fresh-session builder (chat.py).
+        tools=list(DEFAULT_SESSION_TOOLS),
         web_search_enabled=settings.web_search_enabled,
         web_fetch_enabled=settings.web_fetch_enabled,
     ))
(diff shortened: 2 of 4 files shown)


alecantu7:
## Adversarial self-review — safe

Attacked the ways this change (add a shared constant + swap two imports + add `tools=` to `rebuild_session`) could go wrong:

- **Circular import?** No. `anton/tools.py` imports only `ToolDef` (from `anton.core.tools.tool_defs`) — not `chat`/`chat_session` — so `DEFAULT_SESSION_TOOLS` in `tools.py` + the new imports create no cycle. `chat_session.py`'s import is function-local in `rebuild_session` anyway (deferred).
- **Import swap safety?** `chat.py`'s `CONNECT_DATASOURCE_TOOL`/`PUBLISH_TOOL` were used only at the (now-changed) line 1360, and **nothing imports those names *from `anton.chat`***, so swapping its import to `DEFAULT_SESSION_TOOLS` breaks no other module.
- **Regression?** `test_session_tools.py`, `test_share.py` (exercises `restore_session`/`rebuild_session` — the changed path), `test_publish_tool_access.py`, `test_chat_context.py` → **24 passed**.
- **New object-sharing risk?** None. Callers pass `list(DEFAULT_SESSION_TOOLS)` — a fresh list of the same module-level `ToolDef` singletons the fresh path (`chat.py`) *already* passed directly. Object sharing is identical to before; no new per-session mutation surface.
- **Parity?** `DEFAULT_SESSION_TOOLS` == exactly what the fresh builder passed (`[CONNECT_DATASOURCE_TOOL, PUBLISH_TOOL]`), so `rebuild_session` now matches fresh precisely — no more, no less.

**Scope note (not a defect):** `runtime.py::build_chat_session` is a *separate* builder that defaults `tools=[]`; it is **not** on the `/resume` or settings-change path (those go through `rebuild_session`), so it's out of scope here. Flagged in the PR description for a separate audit of its headless/cron callers.

Verdict: safe. No changes needed.

MinuraPunchihewa (approved):

PR #291, opened 2026-07-31
ENG-1169: Map M3 wallet denials (402/429) to TokenLimitExceeded so the out-of-credits card renders

alecantu7:
## What

The M3 authorization gate (ENG-608, live ~Jul 20–27) denies out-of-credits as **402 `wallet_empty`** and a spent free allowance as **429 `included_allowance_exhausted`** (no FastAPI `detail`). anton's status-error mapper had no branch for either, so:

- the **402** fell to the generic "temporarily unavailable" ConnectionError, got auto-retried twice by `turn_stream` (with an "adjust your approach" note injected into history), and was finally **rendered as assistant prose** — the turn *succeeded*, `response.failed` was never emitted, and the ENG-562 out-of-credits card never showed. The user hit a hard paywall with no path to refill (live repro attached on the ticket).
- the **allowance 429** misclassified as a retryable plain rate limit → backoff → wrong `provider_overloaded` card.

## How

- **`wallet_denial_code()` helper in `provider.py`** — reads the code from both body dialects (SDK-unwrapped top level per ENG-747, and the wire envelope). Code-exact on purpose: the gate's velocity 429 (`rate_limited`, ENG-878) stays transient, and BYOK 402s (e.g. OpenRouter's insufficient-credits dialect) stay generic — mirroring cowork-server's `test_byok_402_stays_generic` decision.
- **Both mapper twins** (`openai.py` + `anthropic.py`): 402/429 + wallet code (body) or `X-MindsHub-Reason` header → `TokenLimitExceeded`, which `session.py` already re-raises unretried and cowork-server (since its Jul 27 release) already maps to the `token_limit` card.
- **Mid-stream defensive branch** on both openai stream paths: a wallet denial arriving inside an open stream fails fast onto the credits card instead of entering the 30s `stream_error` backoff (the gate denies pre-stream today; this is insurance).
- **Tests**: real-SDK fixtures only (`httpx.MockTransport`, the ENG-747 lesson — this exact class of card died in prod once from hand-built fixtures). 12 new tests: both dialects, header-only fallback (code-stripped anthropic lane), BYOK-402-stays-generic, velocity-429-stays-transient, and the `wallet_denial_code` unit matrix.

## Verification

- `pytest tests/ --ignore=tests/e2e`: **1406 passed**; the 2 `test_chat_scratchpad` failures are pre-existing on clean `origin/staging` (local sandbox blocks exec), verified by stash-and-rerun.
- Message copy keeps the `402`+`credit` phrasing so cowork-server's string heuristic also matches as a belt on version-skewed installs (isinstance on `TokenLimitExceeded` is the primary path).

## Security pass

Reviewed the changed surface: no new inputs are trusted — the wallet code is compared against a frozen allowlist (never interpolated into copy), provider body content is not echoed into user-facing messages (curated copy only), and the existing `scrub_credentials` logging path is unchanged. No secrets, no new endpoints.

## Delivery note

anton-only fix: reaches git-channel desktops on merge-to-main; PyPI-channel installs need a cowork-server bump (ENG-1094). Companion UI PR (single-CTA card) lands separately in cowork.

Ticket: ENG-1169

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -27,6 +27,7 @@ from .provider import (
     Usage,
     classify_transient,
     compute_context_pressure,
+    wallet_denial_code,
 )
 
 logger = logging.getLogger(__name__)
@@ -44,6 +45,9 @@ def _raise_for_status_error(
 
     - 401 → ConnectionError (invalid-key copy; cowork-server keys on this phrase).
     - 429 WITH a quota ``detail`` → TokenLimitExceeded (keeps its own card).
+    - 402/429 with an M3 gate wallet code (``wallet_empty`` /
+      ``included_allowance_exhausted``, body or X-MindsHub-Reason header)
+      → TokenLimitExceeded (ENG-1169). Code-exact: BYOK 402s stay generic.
     - overloaded/api_error (incl. the mid-stream HTTP-200 case), 5xx, plain 429
       → TransientProviderError (retryable — see ENG-673).
     - anything else → the generic "temporarily unavailable" ConnectionError.
@@ -58,6 +62,30 @@ def _raise_for_status_error(
         msg = f"Server returned 429 — {body['detail']}"
         msg += " Visit https://console.mindshub.ai to upgrade or to top up your tokens."
         raise TokenLimitExceeded(msg) from exc
+
+    # MindsHub M3 wallet taxonomy (ENG-1169) — same branch as the openai twin
+    # so the mapping can't drift. Today the gateway's anthropic-dialect lane
+    # strips both the code and the X-MindsHub-* headers (tracked separately),
+    # so this fires only against a fixed gateway or a proxy that preserves
+    # them — but a wallet denial that DOES arrive here must hit the credits
+    # card, not the generic copy. Code/reason-exact: a BYOK Anthropic billing
+    # error carries neither and stays generic.
+    gate_reason = ""
+    if getattr(exc, "response", None) is not None:
+        gate_reason = exc.response.headers.get("x-mindshub-reason", "")
+    wallet_code = wallet_denial_code(body) or (
+        gate_reason if gate_reason in ("wallet_empty", "included_allowance_exhausted") else None
+    )
+    if exc.status_code in (402, 429) and wallet_code:
+        what = (
+            "your MindsHub credits are used up."
+            if wallet_code == "wallet_empty"
+            else "your included token allowance is exhausted."
+        )
+        raise TokenLimitExceeded(
+            f"Server returned {exc.status_code} — {what} Add credits at "
+            "https://console.mindshub.ai/settings/organization/billing to continue."
+        ) from exc
 
     transient = classify_transient(exc.status_code, body, provider=provider, model=model)
     if transient is not None:
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -1131,6 +1131,24 @@ class OpenAIProvider(LLMProvider):
             # must classify it here or it surfaces as an opaque generic error on
             # the OpenAI/MindsHub path (ENG-673, Sam's review). Body type sits at
             # the top level (SDK-unwrapped); classify_transient handles that shape.
+            # Out-of-credits smuggled into an open stream (defensive — the M3
+            # gate denies pre-stream today): permanent, so it must fail fast
+            # onto the credits card, not enter the 30s stream_error backoff
+            # and surface as a misleading "provider overloaded" (ENG-1169).
+            # Checked BEFORE the transient classifier — permanent-first, the
+            # same policy as the request-time mapper — so a wallet denial that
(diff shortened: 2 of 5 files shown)


alecantu7:
**Self-review findings (adversarial pass):**

1. **[fixed in 738314a]** `classify_transient` gap: a wallet-coded 429 reaching a *direct* classifier call (the mid-stream paths pass status through) would classify as retryable `rate_limited` and enter the backoff loop — the exact failure mode this PR fixes at the mapper layer. The `insufficient_quota` guard already existed for this precedent; wallet codes now get the same code-exact `return None` + a regression test.

2. **[accepted as-is]** The 429-allowance message copy ("your included token allowance is exhausted") does not match cowork-server's string-heuristic belt (`is_token_limit_error` needs `402`+credit/quota/token or 'limit exceeded for tokens'). Primary detection is `isinstance(exc, TokenLimitExceeded)`, which always works in-process; the string belt only matters for a version-skewed cowork-server that can't import anton's type, and that skew combination (new anton + very old cowork-server) can't occur on managed installs (updater moves them together).

3. **[accepted as-is]** The mid-stream wallet guard sits *after* `classify_transient` — a body carrying both a transient type AND a wallet code would retry. No real dialect emits that shape; ordering matches the mapper's 'permanent-first within each arm' policy closely enough, and putting wallet first would shadow genuine overload events if the gate ever reused the codes.

4. Precedence audit: 429+`detail`+wallet-code → legacy detail branch wins (unchanged behavior, same card). 402 with HTML body + reason header → header fallback maps it. Gemini array-unwrap happens before the wallet read (`raw_body`), so the branch sees the unwrapped dict.

alecantu7:
**Adversarial review round 2 (agent-assisted, mutation-tested) — 3 findings fixed in 12d25e8, 2 accepted with rationale:**

**Fixed:**
1. **TypeError on unhashable wire code (minor/major):** `wallet_denial_code` did `code in _WALLET_DENIAL_CODES` — frozenset membership *hashes*, so a buggy/hostile OpenAI-compatible endpoint sending `"code": ["wallet_empty"]` crashed the classifier with `TypeError` instead of mapping (verified by probe; blast radius included the mapper, classify_transient's plain-429 branch, and both mid-stream handlers). Now isinstance-guarded, matching the tuple-membership convention every other wire-value check in these files uses. Regression test added.
2. **Mid-stream guards were mutation-dead:** neutering both guards left the suite green — violating this suite's own stated bar (untested mapping wiring "ships dead", per the file docstring). Added real-SDK e2e stream tests (`httpx.MockTransport` SSE): wallet denial mid-stream → `TokenLimitExceeded`, the allowance variant, and the ordering case below. All three fail if the guards are removed.
3. **Guard ordering contradicted the permanent-first policy:** the mid-stream wallet check ran *after* `classify_transient`, so a wallet denial that also carried a transient-looking `type` (e.g. `api_error` + `code: wallet_empty`) would enter the 30s backoff and surface as `provider_overloaded`. Reordered to permanent-first (same policy the request-time mapper documents); mid-stream copy now also distinguishes `wallet_empty` from `included_allowance_exhausted`.

**Accepted as-is (documented, no code change):**
4. **Interior swallow points cost one extra doomed request each** before the card renders: `_gate_turn` (router thalamus call, `except Exception → None`), tool-internal LLM calls (`Tool 'x' failed:` string), and the completion verifier (`except Exception → break`). In all three the next planning call re-raises and the card still renders — wallet-empty is account-global, so the extra request is denied cheaply pre-dispatch at the gate. Tightening those three interceptors is out of scope here (they guard against unrelated failure classes by design — "routing must never break a turn").
5. **Copy divergence** between the request-time branch, the legacy 429-detail branch (still says "upgrade", pre-M3 language), and the mid-stream branch: cosmetic — cowork-server replaces all of them with its curated card copy; only CLI users see these strings. The legacy branch's copy predates this PR; happy to sweep it in a follow-up if wanted.

Review also confirmed (attacks that held): all four openai call paths + both anthropic paths route through the mapper with no bypass; a request-time 402 on the planning call escapes `turn_stream` unretried (re-raise precedes both retry mechanisms); Gemini list-wrapped wallet bodies still map; the header check is case-insensitive and value-exact; no known provider dialect (OpenAI/Anthropic/Gemini/OpenRouter/Cloudflare-HTML) false-positives; every other new branch is mutation-caught by the suite.

pnewsam (approved):
Reviewed the complete five-file diff at `12d25e8e`. No actionable findings.

The wallet-denial mapping remains code/reason-exact, preserves velocity rate limits as transient and BYOK 402 responses as generic, and covers both request-time and defensive mid-stream paths.

Validation:
- `pytest tests/ --ignore=tests/e2e -q`: 1,413 passed, 17 skipped
- Focused real-SDK mapper/stream suite: 44 passed
- GitHub CI: green

I did not exercise the live MindsHub gateway; validation used its checked-in contract plus real-SDK mock transports.

PR #295, opened 2026-07-31
ENG-1092:  move current date/time out of the cached system prefix

ea-rus:
'Current date and time' info was move from system prompt to the last user message. 
It will increase usage of cached tokens (instead of not cached)

BEFORE:
```
system:  ...<stable prefix>...
         Current date and time: Monday, August 03, 2026 at 05:14 PM   -- volatile (local TZ) 
         <memory snapshot>
messages:
  ...
  assistant: "[2026-08-03 15:14] Got it — I'll analyze..."            
  user     : "yes"                                                    -- current turn: NOT stamped
```

AFTER
```
system:  ...<stable prefix>...
         <memory snapshot>                                       -- removed TZ
messages:
  ...
  assistant: "[2026-08-03 15:14] Got it — I'll analyze..."     
  user     : "[2026-08-03 15:15] yes"                             -- current turn now stamped (UTC)
```

Fixes https://linear.app/mindsdb/issue/ENG-1092/improve-prompt-cache-reuse-move-volatile-parts-out-of-the-system


StpMax (approved):
### Findings

**1. `anton/core/session.py:171-188`, `anton/core/llm/prompt_builder.py` (build()) — Multimodal/attachment turns lose all temporal context (regression).**
`_stamp_user_content()` only stamps `user_input` when it is a plain `str`; when it's a `list[dict]` (image or file attachment, with or without a caption), it is returned completely unchanged — no timestamp is added anywhere, not even to a text block inside the list. Since the system prompt's dynamic clock (`current_datetime`) was deleted outright and replaced with a static, time-free note, there is now **no fallback source of "now" at all** for such a turn.

*Failure scenario:* A user sends an image with a caption (a real, reachable path — confirmed via `cowork-server`'s `anton_harness/harness.py` `_to_anton_input()`, which only collapses a *single* text block to a plain string; any attachment produces a `list[dict]` that reaches `session.turn_stream()` unchanged). Before this PR, that turn still carried `Current date and time: ...` in the system prompt. After this PR, the model has zero temporal signal for that turn — a genuine behavior regression versus the pre-PR baseline, not just a cosmetic gap. This matters for anything time-sensitive (e.g. "what's due today" alongside a screenshot, or scheduling from an attached file).

*Suggested fix:* Either stamp the first text block inside a multi-block message too, or keep a minimal fallback (e.g. stamp based on `now` even for list content, attached as a synthetic leading text block) so every turn always carries a time signal.

**2. `anton/core/session.py:1849-1852` and `2032-2036` — Stamping logic duplicated across `turn()` and `turn_stream()` instead of centralized.**
The `_scrub_user_input` → `_stamp_user_content(user_input, datetime.now(timezone.utc))` → `_append_history(...)` sequence is repeated verbatim in both entry points. The codebase already centralizes history-append invariants in `_append_history` specifically so they can't be violated accidentally at a call site (see the comment at `session.py:825-827`). Stamping is exactly that kind of invariant — a future third call site (or a contributor editing only one of the two methods) could easily forget the stamp step and silently reintroduce the cache-busting bug this PR is fixing. Moving the stamp step into `_append_history` (guarded on `role == "user"`) would remove the duplication and make the invariant impossible to skip, consistent with the existing pattern in that method.

ea-rus:
1. cowork-server's history stamper (anton_harness/harness.py _stamped) also only stamps str content — it leaves list[dict] (attachment) messages unstamped on replay. So if we stamp list content only in anton's _stamp_user_content, an attachment turn would read [ts] caption live but come back unstamped when replayed from persisted history — the temporal signal would vanish after the turn, and the live-vs-replay divergence would also cost cross-turn cache on that message. 
The symmetrical stamp of mixed content for cowork/anton can be done in separated PRs
3.  I'd recommend against centralizing. role == "user" would over-stamp — tool_result messages and synthetic inserts (recovery note, "Cancelled by user.") are also role user but aren't turns. Doing it safely needs an explicit "user turn" flag threaded into _append_history — more complexity in the one invariant-critical method, not less. The duplication is just the two turn entry points (which already share _scrub_user_input). Added comments why it's not in _append_history.

PR #297, opened 2026-07-31
ENG-847:  fail loud on empty 200 in non-streaming complete (follow up of #258)

ea-rus:
## Summary

Based on https://github.com/mindsdb/anton/pull/258#issuecomment-5063836574

Fail loud on an empty HTTP 200 in the non-streaming `complete()` paths. Previously they silently returned an empty `LLMResponse` (`content=''`, zero usage, `stop_reason=None`), which the agent misdiagnoses as a backend outage — the exact failure mode that made ENG-847 hard to triage.

This mirrors the guard the streaming path already has (`openai.py`, `code="truncated_stream"`) onto the three non-streaming return sites, as defense-in-depth follow-up flagged in review on #258.

## Changes

- **`provider.py`** — add shared helper `raise_on_empty_response(...)`. Raises `TransientProviderError(code="empty_response", session_backoff=False)` only when content, tool calls, and `stop_reason` are all empty. One helper so the mapping can't drift between providers.
- Wire it into the three non-streaming paths: OpenAI chat-completions `complete()`, OpenAI Responses path (`_parse_response_object`), and Anthropic `complete()`.

`session_backoff=False` matches the streaming carve-out: an empty-from-start 200 is a strong broken/misconfigured-endpoint signal, so it fails fast instead of looping the retry budget. `stop_reason` is the discriminator — a real 200 always carries one, so the guard only fires on a genuinely empty/truncated response, not a legitimately empty turn.

## Tests

- New `TestRaiseOnEmptyResponse`: raise on empty 200, passthrough for content / tool calls / real stop_reason, and end-to-end via `_parse_response_object`.
- Updated two existing `_parse_response_object` fixtures to include a realistic `status="completed"` (they previously omitted `status`, which is unrealistic for a completed Responses object).

All provider + transient-retry suites pass (112 tests).

Fixes https://linear.app/mindsdb/issue/ENG-847/scratchpad-web-search-silently-returns-empty-on-the-minds-cloud

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -27,6 +27,7 @@ from .provider import (
     Usage,
     classify_transient,
     compute_context_pressure,
+    raise_on_empty_response,
 )
 
 logger = logging.getLogger(__name__)
@@ -189,6 +190,11 @@ class AnthropicProvider(LLMProvider):
                 tool_calls.append(
                     ToolCall(id=block.id, name=block.name, input=block.input)
                 )
+
+        raise_on_empty_response(
+            content=content_text, tool_calls=tool_calls,
+            stop_reason=response.stop_reason, provider="Anthropic", model=model,
+        )
 
         input_tokens = response.usage.input_tokens
         return LLMResponse(
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -32,6 +32,7 @@ from .provider import (
     Usage,
     classify_transient,
     compute_context_pressure,
+    raise_on_empty_response,
 )
 
 logger = logging.getLogger(__name__)
@@ -933,6 +934,11 @@ class OpenAIProvider(LLMProvider):
                     )
                 )
 
+        raise_on_empty_response(
+            content=content_text, tool_calls=tool_calls,
+            stop_reason=choice.finish_reason, model=model,
+        )
+
         usage_obj = response.usage
         input_tokens = usage_obj.prompt_tokens if usage_obj else 0
         return LLMResponse(
@@ -1458,6 +1464,11 @@ def _parse_response_object(response, model: str) -> LLMResponse:
     input_tokens = (getattr(usage, "input_tokens", 0) or 0) if usage else 0
     output_tokens = (getattr(usage, "output_tokens", 0) or 0) if usage else 0
 
+    status = getattr(response, "status", None)
+    raise_on_empty_response(
+        content=content_text, tool_calls=tool_calls, stop_reason=status, model=model,
+    )
+
     return LLMResponse(
         content=content_text,
         tool_calls=tool_calls,
@@ -1466,5 +1477,5 @@ def _parse_response_object(response, model: str) -> LLMResponse:
             output_tokens=output_tokens,
             context_pressure=compute_context_pressure(model, input_tokens),
         ),
-        stop_reason=getattr(response, "status", None),
+        stop_reason=status,
     )
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -471,6 +471,32 @@ def classify_transient(
     return None
 
 
+def raise_on_empty_response(
+    *, content: str, tool_calls: list, stop_reason: str | None,
+    provider: str = "", model: str = "",
+) -> None:
+    """Fail loud on an empty 200: no content, no tool calls, no stop reason.
+
+    The non-streaming mirror of the streaming truncated-response guard (ENG-673).
+    An empty-from-start 200 is:
+
+    - a weak incident signal (real mid-incident silence surfaces as a dropped
+      connection / read timeout, which back off above), and
+    - a strong broken/misconfigured-endpoint signal.
+
+    So it fails fast (``session_backoff=False``) rather than looping the retry
+    budget — and, critically, raises instead of handing back an empty
+    ``LLMResponse`` the agent would misdiagnose as a backend outage.
+    """
+    if content or tool_calls or stop_reason is not None:
+        return
+    raise TransientProviderError(
+        f"{provider or 'The model provider'} returned an empty response — try again in a moment.",
+        provider=provider or "The model provider", code="empty_response",
+        session_backoff=False, model=model,
+    )
+
+
 class ModelUnavailableError(ConnectionError):
     """Raised when the gateway rejects the requested model with a structured 403.
 
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -518,7 +518,9 @@ def raise_on_empty_response(
     budget — and, critically, raises instead of handing back an empty
     ``LLMResponse`` the agent would misdiagnose as a backend outage.
     """
-    if content or tool_calls or stop_reason is not None:
+    # Empty-string stop_reason ("" — no real provider sends it) is treated as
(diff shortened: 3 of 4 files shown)


alecantu7 (commented):
Reviewed with particular attention to interactions with this week's truncation-family merges (#278/#286/#287), since "empty response" is exactly the territory those classify. Verdict: clean — the all-three-empty predicate threads correctly past every case in the family. Walked each one:

- **Capped-but-labeled response** (the ENG-1042 silent burn, non-streaming): `stop_reason` present (`length`/`max_tokens`/even a lying `stop`) → guard passes → flows to `looks_truncated` → the budget ladders/recovery own it. No double-handling.
- **fable's `{}` verdicts** (ENG-1095): `tool_calls` present → passes → `raise_unusable_tool_call` classifies as non-truncated, correctly not retried.
- **fable's auto-mode near-empties**: `stop_reason: end_turn/stop` present → passes — correctly NOT misdiagnosed as a broken endpoint.
- **The genuinely dead 200** (ENG-847/1004's Responses shape-mismatch parse): all three empty → raises loud. The fix.

Also verified, not assumed:

1. **ENG-847's original surface is covered**: the scratchpad subprocess constructs the same `OpenAIProvider`/`AnthropicProvider` classes (`scratchpad_boot.py:252-299`), so `web_search()`/`get_llm()` route through the guarded `complete()`/`_parse_response_object`. In the cell, the raise surfaces as a visible exception the model can adapt to — strictly better than the silent `''` it gets today.
2. **Downstream handling of the new raise is right in every merged handler**: verifier ladder → generic except → content-safe log → `verdict is None` → the #276 honest diagnosis; memory sites → the #287 subsystem-named warnings; turn-level paths → the ENG-673 count-based retry (`session_backoff=False` matches the streaming carve-out's reasoning exactly).
3. Ran the provider/transient suites plus all four of this week's family test files on this branch: 106/106.
4. The `stop_reason=status` consolidation in `_parse_response_object` and the fixture-realism fix (`status="completed"`) are both right.

One nit, note-level: `stop_reason is not None` treats an empty-string `stop_reason` as "present" — no real provider sends `""`, so this is theoretical; `if content or tool_calls or stop_reason:` would be marginally stricter. Not worth a push on its own.

Freeze note: anton staging is post-cut, so this rides next week's train — which is the right speed for it (the empty-200 has been silently wrong forever; a week of staging soak alongside #382 is the correct path, no exception warranted).

ea-rus:
fixed pr conflicts, and this:

> One nit, note-level: stop_reason is not None treats an empty-string stop_reason as "present" — no real provider sends "", so this is theoretical; if content or tool_calls or stop_reason: would be marginally stricter. Not worth a push on its own.

alecantu7 (commented):
Re-reviewed the two new commits at `f9f72b4`. `test_provider.py` + `test_status_error_mapper.py` → **61 passed**, now 0 behind `staging`.

**The nit fix is exactly right**, and I probed the predicate directly rather than reading it:

```
RAISES   all empty / None            (code=empty_response, backoff=False)
RAISES   stop_reason=""  ← the nit
passes   stop_reason="end_turn"      ← legitimately empty turn, correctly NOT flagged
passes   content present
passes   tool_calls present
```

That third row is the one that mattered to get right: switching `stop_reason is not None` → truthiness makes the guard *stricter*, so the question was whether it could now fire on a real turn that legitimately produced no text. It can't — a genuine 200 carries a stop reason, and any non-empty one is truthy. The change only bites in the case we want (`""` with nothing else), so it's strictly better with no new false-positive surface.

**I also checked the merge, since it pulled in #291 (ENG-1169 wallet denials) which edits the same provider files.** No interaction: #291 maps 402/429 inside `_raise_for_status_error` — the *status-error* path — while this guard sits on the *success* path after a 200 parses. A wallet denial never reaches it. And all three call sites survived the merge intact (`anthropic.py:222`, `openai.py:971`, `openai.py:1538`).

**One thing worth adding before this merges: the new commit is untested.**

`TestRaiseOnEmptyResponse` covers `stop_reason=None`, `"stop"`, and `"tool_use"` — but nothing passes `""`, which is the only behaviour `f9f72b4` changed. I confirmed that rather than eyeballing it: reverting the predicate back to `stop_reason is not None` leaves **all 61 tests passing**.

So the fix is real but unpinned — the next person tidying that condition back to an explicit `is not None` (which reads more "correct" to a Python reviewer, and was the original form) gets a green suite. Two lines closes it:

```python
def test_empty_string_stop_reason_counts_as_absent(self):
    with pytest.raises(TransientProviderError):
        raise_on_empty_response(content="", tool_calls=[], stop_reason="")
```

Not worth blocking on by itself — it's a theoretical shape no provider sends today, which is exactly why I raised it as note-level the first time. But since the code now carries a comment asserting the behaviour, a test asserting it costs nothing and stops the comment from outliving the code.

Nothing else from me. The freeze note from my earlier review still applies: anton `staging` is post-cut, so this rides next week's train, which remains the right speed for a guard that changes how an empty 200 surfaces.

alecantu7:
@ea-rus one clarification, because I under-sold this in my last review and don't want it to stall on a mixed signal: I called the missing test "not worth blocking on by itself", and we've since decided to **hold the approval for it**. So it's the one thing outstanding — sorry for the moving target.

The reason it changed weight: I mutation-tested it rather than eyeballing it. Reverting your predicate back to the original form —

```python
if content or tool_calls or stop_reason is not None:
```

— leaves **all 61 tests in `test_provider.py` + `test_status_error_mapper.py` passing.** So `f9f72b4`'s behaviour is currently unpinned, and the specific way it regresses is plausible rather than exotic: `is not None` is the more idiomatic-looking form and *was* the original, so a future tidy-up reverts it and CI stays green. The code now carries a comment asserting the empty-string case, and a comment that outlives its behaviour is worse than no comment.

Two lines in `TestRaiseOnEmptyResponse`:

```python
def test_empty_string_stop_reason_counts_as_absent(self):
    with pytest.raises(TransientProviderError):
        raise_on_empty_response(content="", tool_calls=[], stop_reason="")
```

Everything else on the PR is verified and good — I probed the predicate directly (`""` and `None` raise; `"end_turn"`, content, and tool_calls all pass, so a legitimately-empty turn is correctly not flagged), and I checked the staging merge against #291's wallet-denial mapping: no interaction, since that lives on the status-error path and this guard is on the success path after a 200 parses. All three call sites survived intact.

Ping me when it's pushed and I'll approve straight away.

alecantu7 (approved):
Approving — `6c29c641` closes the last item, and it's mutation-verified rather than eyeballed.

As pushed: **62 passed**. Reverting the predicate back to the original `stop_reason is not None`:

```
FAILED tests/test_provider.py::TestRaiseOnEmptyResponse::test_empty_string_stop_reason_counts_as_absent
1 failed, 61 passed
```

That's the gap closed exactly — the same revert left all 61 green before, and now fails precisely this test and nothing else. The comment on the predicate no longer outlives its coverage, which was the actual worry.

The comment you wrote is better than what I asked for: it states *why* `""` counts as absent **and** says out loud that it pins against the revert, so the next person tidying that line reads the reason before changing it.

Rest of the state, re-checked on `6c29c641`: all 9 checks green (`run-tests` 3m15s), `MERGEABLE / CLEAN`, and **0 behind `origin/staging`** — your 08-03 merge is still current, so no rebase before merge. Final diff 4 files, +104/-4.

Thanks for the patience on the moving target — the weight change was mine, not a new requirement.

PR #289, opened 2026-07-30
ci: publish Anton staging release candidates (ENG-1159)

pnewsam:
## Problem

cowork-server staging depends on Anton changes before they reach `anton/main`. Anton currently publishes only stable packages from `main`, so downstream staging has had to use a mutable Git branch or hand-pin a commit. That makes the cross-repository release train difficult to reproduce and promote.

## Design

Give Anton separate, branch-owned package publishing workflows, consistent with the other release-train repositories:

- `release.yml` remains production-only: pushes to `main` validate, create a stable CalVer release, publish it to PyPI, and run live release e2e;
- the new `publish-staging.yml` is staging-only: pushes to `staging` validate, create a PEP 440 `rc` release, and publish it to PyPI;
- `tests.yml` remains the pull-request validation workflow; branch-push validation belongs to the corresponding publishing workflow, avoiding duplicate disconnected run trees;
- both publishing streams run the full unit + stub-e2e suite before creating a tag or publishing;
- live API e2e remains production-only so staging commits do not consume external API capacity;
- permissions and failure/recovery notifications remain scoped to each release stream.

This gives downstream repositories immutable Anton artifacts while preserving the intuitive branch relationship: Anton staging produces RCs for cowork-server staging, and Anton main produces stable packages for cowork-server main.

## One-time PyPI setup required

Before merging, add an additional Trusted Publisher to the PyPI `anton-agent` project with:

- owner: `mindsdb`
- repository: `anton`
- workflow: `publish-staging.yml`
- environment: `pypi`

PyPI binds trusted publishing to the workflow filename. The existing `release.yml` publisher remains unchanged for production; the staging workflow needs its own publisher identity. No API token or other long-lived credential is introduced.

## Downstream and merge order

The coordinated [cowork-server #255](https://github.com/mindsdb/cowork-server/pull/255) resolves the published Anton package whose release tag points to the exact corresponding branch HEAD. Its tests, images, and RC wheel then consume that one immutable Anton artifact.

1. Register the `publish-staging.yml` Trusted Publisher.
2. Merge this PR into Anton staging and wait for its RC to publish.
3. Merge cowork-server #255 into cowork-server staging.

## Validation

- `actionlint` 1.7.12 passes for all changed workflows.
- YAML parsing and whitespace checks pass.
- Workflow graph validation passes across all 10 workflows: each event produces exactly one run tree.
- Anton test suite: 1,427 passed, 17 skipped.
[...]

PR #298, opened 2026-07-31
ci: publish Anton staging release candidates (ENG-1159)

pnewsam:
## Problem

cowork-server staging needs Anton changes before they reach `anton/main`, but Anton only publishes stable packages from `main` (`release.yml`). Downstream staging has had to install Anton from a mutable git branch or a hand-pinned commit, so tests, images, wheels, and lockfiles can observe different Anton snapshots. This is the producer half of **[ENG-1159](https://linear.app/mindsdb/issue/ENG-1159)**.

The consumer half already shipped in cowork-server **[#225](https://github.com/mindsdb/cowork-server/pull/225)** (ENG-958): its [`publish-staging.yml`](https://github.com/mindsdb/cowork-server/blob/main/.github/workflows/publish-staging.yml) ("Pin anton-agent to its latest staging rc" step) pins `anton-agent==<latest rc>` and falls back to a loose constraint otherwise. It hits that fallback on every run today because Anton publishes no rc's. This PR closes the loop.

## Design — mirrors the cowork-server pattern

- **`publish-staging.yml` (new)** — staging only: push to `staging` → validate → cut a PEP 440 pre-release (`2.YY.M.DD.SEQrcN`) via the shared `calver-release.yml` reusable → publish to PyPI via Trusted Publishing → notify.
- **`release.yml`** — same intent, restructured to match cowork-server's `publish.yml`: gate on the reusable tests, scope `permissions` per job (only the release job gets `contents: write`), guard on `refs/heads/main`, and fold the test job into `notify`.
- **`tests.yml`** — now a `workflow_call` reusable invoked by both publishers; drops its standalone `push: [main, staging]` run and push-only `notify`. Branch-push validation now happens inside the release run. PR gating is unchanged.

**Simpler than cowork-server's:** Anton doesn't depend on itself, so there's no self-pin step (cowork-server's "pin anton-agent to its latest rc") and no `build-deploy` job (Anton ships no image) — the build derives the rc version from the git tag via `hatch-vcs`. It does set `SETUPTOOLS_SCM_PRETEND_VERSION` on the build step, but only to pin the exact version the release job minted so a re-run/re-dispatch on an already-tagged head can't `git describe` a stale duplicate — not as a self-pin substitute. Same guard in `release.yml`.

**Prod safety:** resolvers ignore pre-releases unless a specifier names one, and PyPI's `info.version` (read by the prod desktop updater) excludes them — so prod can never pick up an rc.

## PyPI setup — done ✅

The **second Trusted Publisher** for `publish-staging.yml` (environment `pypi`) is registered on the [`anton-agent` project](https://pypi.org/manage/project/anton-agent/settings/publishing/), done during review — so the staging `publish` job's OIDC auth is ready and this is no longer a merge blocker.

## Notes

- Supersedes the self-closed [anton#289](https://github.com/mindsdb/anton/pull/289) / cowork-server#255. The consumer side ([cowork-server#225](https://github.com/mindsdb/cowork-server/pull/225)) is already merged on `main`, so only this producer workflow is needed.
- Follow-up (not in this PR): cowork-server pins "latest by upload time, only if rc", so once Anton publishes both streams a newer stable can win the race and drop it back to the loose constraint. Tightening to "latest rc specifically" would make the pairing strict.

### DIFF
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -1,14 +1,21 @@
-name: Auto-release and publish to PyPI
+name: Staging pre-release and publish to PyPI
 
-# Stable release stream: on every push to main, validate, cut a CalVer release,
-# publish it to PyPI, and run live release e2e. Staging pre-releases (rc) live in
-# publish-staging.yml so each stream keeps its own run tree, permissions,
-# notifications, and publishing identity.
+# rc pre-release stream: on every push to staging, validate, cut a PEP 440
+# pre-release (2.YY.M.DD.SEQrcN), and publish it to PyPI so cowork-server staging
+# consumes an immutable, versioned Anton artifact instead of a mutable git branch
+# or a hand-pinned commit (ENG-1159). Resolvers ignore pre-releases unless a
+# specifier names one, and PyPI's `info.version` (which the prod desktop updater
+# reads) excludes them, so prod installs can never pick these up.
 #
 # The publish job MUST live in this top-level workflow — PyPI Trusted Publishing
 # does not support reusable workflows and matches the OIDC claim on THIS
-# filename. Register release.yml as a trusted publisher (environment `pypi`) at
-# https://pypi.org/manage/project/anton-agent/settings/publishing/
+# filename. Register publish-staging.yml as a trusted publisher (environment
+# `pypi`) at the anton-agent PyPI project's publishing settings.
+#
+# Anton has nothing to self-pin (it does not depend on itself), so — unlike
+# cowork-server's staging publisher — this builds from a clean checkout of the
+# tag and needs no version-pretend override: hatch-vcs derives the exact rc
+# version straight from the tag, identical to the stable path in release.yml.
 #
 # No workflow-level `permissions:` block: only the release job needs
 # `contents: write`; every other job declares its own, so the repo default
@@ -18,11 +25,11 @@ name: Auto-release and publish to PyPI
 
 on:
   push:
-    branches: [main]
+    branches: [staging]
   workflow_dispatch:
 
 concurrency:
-  group: auto-release-${{ github.ref }}
+  group: staging-prerelease-${{ github.ref }}
   cancel-in-progress: false
 
 jobs:
@@ -31,20 +38,21 @@ jobs:
       contents: read
     uses: ./.github/workflows/tests.yml
 
-  auto-release:
+  release:
     needs: unit-tests
-    # workflow_dispatch can be pointed at any ref; only main cuts a stable release.
-    if: github.ref == 'refs/heads/main'
+    # workflow_dispatch can be pointed at any ref; only staging mints rc tags.
+    if: github.ref == 'refs/heads/staging'
     permissions:
       contents: write  # tag push + release creation
     uses: mindsdb/github-actions/.github/workflows/calver-release.yml@main
     with:
       calver-major: "2"
+      prerelease: true
       runs-on: ubuntu-latest
 
   publish:
-    name: Publish to PyPI
-    needs: auto-release
+    name: Build and publish pre-release to PyPI
+    needs: release
     runs-on: ubuntu-latest
     environment: pypi
     permissions:
@@ -53,7 +61,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ needs.auto-release.outputs.tag }}
+          ref: ${{ needs.release.outputs.tag }}
           fetch-depth: 0  # hatch-vcs needs tags to derive version
 
       - name: Setup uv
@@ -66,32 +74,23 @@ jobs:
 
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
-        # Trusted publisher (OIDC) — release.yml must be registered at the
-        # anton-agent PyPI project's publishing settings (environment: pypi).
-
-  e2e:
-    needs: auto-release
-    permissions:
-      contents: read
-    uses: ./.github/workflows/tests_e2e_release.yml
-    with:
-      tag: ${{ needs.auto-release.outputs.tag }}
-    secrets: inherit
+        # Trusted publisher (OIDC) — publish-staging.yml must be registered at
+        # the anton-agent PyPI project's publishing settings (environment: pypi).
 
   notify:
-    # Alert the eng channel if ANY job in the release pipeline fails (tests, tag,
-    # PyPI publish, or release e2e).
+    # Alert the eng channel if ANY job in the staging release/publish pipeline
+    # fails.
     # One job covers both outcomes: a `uses:` job cannot branch on status, so
     # the aggregate result picks the failed or recovered message, and a
     # cancelled run stays silent.
-    needs: [unit-tests, auto-release, publish, e2e]
-    if: ${{ github.ref == 'refs/heads/main' && !cancelled() && !contains(needs.*.result, 'cancelled') }}
+    needs: [unit-tests, release, publish]
+    if: ${{ github.ref == 'refs/heads/staging' && !cancelled() && !contains(needs.*.result, 'cancelled') }}
     permissions:
       contents: read
       actions: read  # the prior-run lookup behind the recovery message
     uses: mindsdb/github-actions/.github/workflows/notify-main-failure.yml@main
     with:
-      env-name: "release"
+      env-name: "staging prerelease"
       status: ${{ contains(needs.*.result, 'failure') && 'failed' || 'recovered' }}
       runs-on: ubuntu-latest
     secrets: inherit
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -92,5 +92,12 @@ jobs:
     with:
       env-name: "staging prerelease"
       status: ${{ contains(needs.*.result, 'failure') && 'failed' || 'recovered' }}
+      # This publisher only ever runs on staging, so the alert is always
+      # freeze-scoped: reserve release-blocking escalation for the freeze
+      # window rather than paging the eng channel on every midweek failure.
+      # Preserves the policy the shared CI notify applied via
+      # `freeze-scoped: ${{ github.ref_name == 'staging' }}` before this
+      # pipeline took ownership of the alert.
+      freeze-scoped: true
       runs-on: ubuntu-latest
     secrets: inherit
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -12,10 +12,13 @@ name: Staging pre-release and publish to PyPI
 # filename. Register publish-staging.yml as a trusted publisher (environment
 # `pypi`) at the anton-agent PyPI project's publishing settings.
 #
-# Anton has nothing to self-pin (it does not depend on itself), so — unlike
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-staging.yml
```diff
@@ -0,0 +1,96 @@
...
+    # cancelled run stays silent.
+    needs: [unit-tests, release, publish]
+    if: ${{ github.ref == 'refs/heads/staging' && !cancelled() && !contains(needs.*.result, 'cancelled') }}
+    permissions:
+      contents: read
+      actions: read  # the prior-run lookup behind the recovery message
+    uses: mindsdb/github-actions/.github/workflows/notify-main-failure.yml@main
+    with:
+      env-name: "staging prerelease"
+      status: ${{ contains(needs.*.result, 'failure') && 'failed' || 'recovered' }}
```
pnewsam, line 102: **Minor:** This move drops the freeze-aware alert policy currently used for staging CI. Without `freeze-scoped: true`, every midweek staging pipeline failure escalates to the engineering channel instead of reserving release-blocking alerts for the freeze window. Pass `freeze-scoped: true` in this `with:` block.

pnewsam (commented):
## Code Review

**Verdict: REQUEST CHANGES**

The core producer design is sound, but this head is not merge-ready. It is 36 commits behind `staging`, GitHub reports it as non-mergeable, and the conflict in `.github/workflows/tests.yml` overlaps meaningful changes: current `staging` added `uv lock --check` and moved development dependencies from `--extra dev` to `--group dev`. Rebase and preserve those current test steps while adding `workflow_call` and removing the standalone push notification.

The PyPI Trusted Publisher for `publish-staging.yml` with environment `pypi` also needs to be configured before merge, because the merge push immediately enters the tag-and-publish pipeline.

Non-blocking integration caveat: cowork-server still selects the most recently uploaded Anton version and pins it only when that version is an RC. If ENG-1159 requires strict RC pairing, update the consumer to select the latest RC specifically rather than deferring that change.

### Validation

- Inspected all three workflow patches at head `1a9301b` and the current `staging` versions.
- Confirmed the PR CI run passed unit and stub E2E tests.
- Confirmed the branch is 36 commits behind `staging` and conflicts in `tests.yml`.
- Did not exercise PyPI OIDC publishing or verify the Trusted Publisher configuration.

pnewsam:
Thanks for the review. Addressed the blocking item and confirmed the rest.

**Rebased onto current `staging`** (was 36 behind, non-mergeable) — now 0 commits behind at `549cf3e4`. The only conflict was `.github/workflows/tests.yml`; `release.yml` (untouched by staging) and the new `publish-staging.yml` applied cleanly.

**`tests.yml` conflict resolved to preserve current staging test steps** while keeping this PR's reusable-workflow change:
- Kept staging's `Check uv.lock matches pyproject.toml` (`uv lock --check`) step.
- Kept staging's `--group dev` on both the unit and stub-e2e pytest steps (this PR had branched from the older `--extra dev`).
- Applied `workflow_call:` and removed the standalone `push:` trigger.
- Removed the in-file `notify` job — failure notification is now carried by the calling publishers (`release.yml` needs `[unit-tests, auto-release, publish, e2e]`; `publish-staging.yml` needs `[unit-tests, release, publish]`), so direct-push alerting isn't lost, just relocated per the reusable pattern.

All three workflows pass YAML validation, and `release.yml`'s reference to `tests_e2e_release.yml` resolves against staging.

**PyPI Trusted Publisher** for `publish-staging.yml` (environment `pypi`): still needs to be registered on the `anton-agent` project before merge — flagging so it's done ahead of the merge push entering the tag/publish pipeline.

**Strict RC pairing (non-blocking):** agreed this is a cowork-server consumer change ("latest by upload time, pin only if RC" → "latest RC specifically"). Tracking it as a follow-up rather than folding it into this producer PR.

pnewsam:
Addressed the freeze-scoped comment in `4cdd4f64`.

`publish-staging.yml`'s `notify` job now passes `freeze-scoped: true`. Since this publisher only runs on staging (`if: github.ref == 'refs/heads/staging'`), that hardcodes the same policy the shared CI notify previously applied via `freeze-scoped: ${{ github.ref_name == 'staging' }}` — release-blocking escalation stays reserved for the freeze window instead of paging on every midweek staging failure. `release.yml` (main) intentionally omits the flag, since main failures always block.

lucas-koontz:
> Thanks for the review. Addressed the blocking item and confirmed the rest.
> 
> **Rebased onto current `staging`** (was 36 behind, non-mergeable) — now 0 commits behind at `549cf3e4`. The only conflict was `.github/workflows/tests.yml`; `release.yml` (untouched by staging) and the new `publish-staging.yml` applied cleanly.
> 
> **`tests.yml` conflict resolved to preserve current staging test steps** while keeping this PR's reusable-workflow change:
> 
>     * Kept staging's `Check uv.lock matches pyproject.toml` (`uv lock --check`) step.
> 
>     * Kept staging's `--group dev` on both the unit and stub-e2e pytest steps (this PR had branched from the older `--extra dev`).
> 
>     * Applied `workflow_call:` and removed the standalone `push:` trigger.
> 
>     * Removed the in-file `notify` job — failure notification is now carried by the calling publishers (`release.yml` needs `[unit-tests, auto-release, publish, e2e]`; `publish-staging.yml` needs `[unit-tests, release, publish]`), so direct-push alerting isn't lost, just relocated per the reusable pattern.
> 
> 
> All three workflows pass YAML validation, and `release.yml`'s reference to `tests_e2e_release.yml` resolves against staging.
> 
> **PyPI Trusted Publisher** for `publish-staging.yml` (environment `pypi`): still needs to be registered on the `anton-agent` project before merge — flagging so it's done ahead of the merge push entering the tag/publish pipeline.
> 
> **Strict RC pairing (non-blocking):** agreed this is a cowork-server consumer change ("latest by upload time, pin only if RC" → "latest RC specifically"). Tracking it as a follow-up rather than folding it into this producer PR.

image

.github/workflows/publish-staging.yml
```diff
@@ -0,0 +1,103 @@
...
+          ref: ${{ needs.release.outputs.tag }}
+          fetch-depth: 0  # hatch-vcs needs tags to derive version
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+
+      - name: Build package
+        run: uv build
```
lucas-koontz, line 81: issue: If someone clicks 'Re-run all jobs', or dispatches this workflow on an unchanged staging head, calver-release mints a second rc tag on a commit that already has one. The build step asks git which version to build, and `git describe` answers with the older tag: the run that minted rc2 builds rc1, and PyPI rejects the upload as a duplicate. Pinning the build to the version the release job just minted removes the guesswork (the lines 15-18 header comment would need updating to match). Same edge on release.yml re-runs.

```suggestion
      - name: Build package
        env:
          # A re-run or re-dispatch on an already-tagged head leaves two CalVer
          # tags on one commit and `git describe` resolves the older one; build
          # exactly the version the release job minted.
          SETUPTOOLS_SCM_PRETEND_VERSION: ${{ needs.release.outputs.version }}
        run: uv build
```

.github/workflows/publish-staging.yml
```diff
@@ -0,0 +1,103 @@
+name: Staging pre-release and publish to PyPI
+
+# rc pre-release stream: on every push to staging, validate, cut a PEP 440
+# pre-release (2.YY.M.DD.SEQrcN), and publish it to PyPI so cowork-server staging
+# consumes an immutable, versioned Anton artifact instead of a mutable git branch
+# or a hand-pinned commit (ENG-1159). Resolvers ignore pre-releases unless a
+# specifier names one, and PyPI's `info.version` (which the prod desktop updater
+# reads) excludes them, so prod installs can never pick these up.
```
lucas-koontz, line 8: issue: README's 'Releasing' section and docs/docs/developer/release-and-versioning.md still describe push-to-main as the only publish path and say only the release workflow creates v* tags. The first staging push after merge contradicts both: rc tag, GitHub pre-release, PyPI upload. Worth a short staging-rc paragraph in both while this is fresh.

lucas-koontz (approved):
Read the whole diff at 4cdd4f6, plus calver-release.yml and notify-main-failure.yml against it. The permission grants match the called workflows' declarations exactly, the notify aggregation handles skipped needs correctly, and the recovered lookup keys on this file's path, so the first green run stays quiet. That all checks out.

Nothing blocks. Two notes inline: a version-skew edge when one commit carries two rc tags (with a one-env-var fix), and the two release docs that still describe main as the only publish stream.

The PyPI side is ready: I registered publish-staging.yml (environment pypi) on anton-agent, so the pre-merge gate in the body is cleared. Workflow naming across repos (release.yml here vs publish.yml in cowork-server) is a follow-up I'll take; the rename has to move the trusted-publisher registration in lockstep anyway.

.github/workflows/publish-staging.yml
```diff
@@ -0,0 +1,103 @@
...
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      contents: read
+      id-token: write  # required for trusted publisher (OIDC)
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.release.outputs.tag }}
+          fetch-depth: 0  # hatch-vcs needs tags to derive version
```
lucas-koontz, line 68: fyi: While you're in the release docs for the rc stream: README "Versioning" and docs/docs/developer/release-and-versioning.md both say the version source of truth is `__version__` in anton/__init__.py. It's been hatch-vcs tag derivation since before this PR (pyproject's `[tool.hatch.version] source = "vcs"`; this checkout comment is the tell). Predates this PR, fix whenever those sections get touched.

.github/workflows/publish-staging.yml
pnewsam: Fixed in 60b9e6d0. Pinned the build to `needs.<release>.outputs.version` (`SETUPTOOLS_SCM_PRETEND_VERSION`) on both `publish-staging.yml` (`needs.release`) and `release.yml` (`needs.auto-release`), and rewrote the 15-18 header comment that claimed no pretend-override was needed. Confirmed `calver-release.yml` exposes `version` (= tag minus the leading `v`), so the pin resolves.

.github/workflows/publish-staging.yml
pnewsam: Also fixed in 66ae9c14. Corrected the source-of-truth claim in both docs to tag-derived via hatch-vcs (`[tool.hatch.version] source = "vcs"`); the dev doc's "hatch reads it via regex" line was flatly wrong. Removed the now-moot "bump `__version__`" instructions too.

One pre-existing mismatch I left alone as a separate cleanup: the CalVer table says `PATCH` is `0` for scheduled releases, but `calver-release.yml` computes the day sequence starting at `1` — docs vs tooling disagree and it predates this PR.

.github/workflows/publish-staging.yml
pnewsam: Done in 66ae9c14. Added a "Staging release candidates" section to both the README "Releasing" section and `docs/docs/developer/release-and-versioning.md` (rc tag → GitHub pre-release → PyPI, plus the prod-never-sees-rc note), and dropped the "push-to-main is the only publish path" framing.

PR #299, opened 2026-07-31
ENG-1155: hand-back diagnoses persist as streamed; latch dead verifiers

alecantu7:
## What

Both ENG-1155 items, plus **a fourth hand-back site the ticket didn't name** and **two stale-base artifacts** that were sitting on `staging`.

**Six commits, read in order** — the first is the failing baseline, so the diff shows the defect caught and then closed:

| commit | contents |
|---|---|
| `fce54ea` | tests only; 3 deliberately failing. Establishes the bug. |
| `ae23b31` | the fix for both ENG-1155 items. |
| `3b7e2a2` | two defects found reviewing `ae23b31`, and their fixes. |
| `2044d8b` | third self-review finding: transient provider errors must not latch. |
| `d4b7785` | pins the failed and truncating re-probe paths; 3 comment/log accuracy fixes. |
| `1b19a1a` | addresses @pnewsam's review — widens the transient carve-out, corrects 2 comments. |

## 1. History desync (ticket bug #1)

`STUCK`, budget-exhausted **and the tool-round circuit breaker** each streamed a model-generated diagnosis and never captured it. The post-loop fallback guarded on `history[-1]["role"] != "assistant"` — after the SYSTEM injection the last entry is a *user* message, so it fired and re-appended the reply the verifier had just rejected.

History after a `STUCK` turn, before this PR:

```
[3] assistant  STALE_REPLY
[4] user       SYSTEM: Task verification determined this task is stuck…
[5] assistant  STALE_REPLY          ← appended a second time
```

So the message the user actually read was **absent entirely** (not substituted, as the ticket describes) and the rejected reply was present **twice**. The model's memory of "what I just told the user" was wrong on exactly the turns where the user is being asked to decide.

- New `_stream_handback_diagnosis()` — #276's capture-persist-guard block, now shared by all four paths instead of copied.
- `_reply_persisted` replaces the role-based fallback guard, which is what allowed the double-append. **Per verification-loop iteration, not per turn** — see the self-review note below; getting this wrong drops continuation answers.

### The fourth site

The **25-round tool cap** has the identical defect and is not in the ticket. It is also the **highest-traffic of the four** — the cap fires on real sessions far more often than `STUCK` does (`STUCK` is ~1.9% of verdicts). Fixed here rather than left as the one path the shared helper doesn't cover.

## 2. Latch (ticket item #2)

kimi-K3 rejects the forced `tool_choice` with a 400 on *every* verdict call (ENG-1095) until the gateway fix lands, so the ENG-1079 diagnosis path fired on every multi-step turn — a full-history planning call plus a "checking in" message each time.

Behaviour now:

- First hard failure → diagnose as before.
- Second consecutive hard failure → **latch without a second diagnosis**. One diagnosis per session, matching the ticket's "Done when".
- While latched → skip the verdict call, one log line per skipped verification.
- **Every `_VERIFIER_LATCH_REPROBE_TURNS` (10) skips → spend one verdict call.** Success clears the latch; failure stays latched and does not re-diagnose.

The re-probe is not in the ticket and is a **deliberate deviation**: the ticket asks for "reset on a successful verdict", which is unreachable as written, because a latched session makes no verdict calls. Without the re-probe the latch is permanent for the session — a user who switches off the broken model gets no verification for the rest of it.

**Truncation deliberately does not latch** — a blown budget is one model being verbose on one transcript, and it keeps its ENG-1081 retry ladder. Only hard failures count. (The `_VERIFIER_TOKEN_BUDGETS` comment asserting "there is no latch" was corrected — it's true of truncation, false in general, and a comment contradicting the code is how the watchdog print-vs-progress confusion started.)

## 3. Stale-base artifacts removed (unplanned)

PR #293 — a Dependabot **postcss bump in `/docs`** — was squash-merged from a stale base and carried two unrelated reverts into `staging` that are **not in `main`**:

- `status, reason = "COMPLETE", "verifier unavailable"` — the silent fail-safe **ENG-1079 deleted**, re-added at the top of the else-branch, directly above the comment explaining why it must not exist.
- a duplicate `from typing import Any, NoReturn` in `llm/structured.py`.

The first was **dead code** — the branch still diagnoses and `break`s, and `test_verifier_exception_yields_real_message_not_silent_stop` passes either way, so there was **no behaviour regression in prod**. It never reached `main`; the 2026-08-02 release (`v2.26.8.2.1`) shipped correct code. Removed here since it's in the block this PR rewrites. The tell that #293 was stale-based: its squash body contains ENG-1081's commit messages.

Worth a separate decision on how a `/docs` dependency bump reached `anton/core/session.py` at all.

## Self-review found two defects in this PR

Full write-up in [this comment](https://github.com/mindsdb/anton/pull/299#issuecomment-5173829295). Both were introduced by `ae23b31`, both fixed in `3b7e2a2`:

1. **`_reply_persisted` was turn-scoped** → on `INCOMPLETE` whose continuation finishes with no tool calls, the loop breaks before the append and the stale flag suppressed the fallback, so **the answer the user just read vanished from history**. Same class of bug as this PR fixes, re-introduced on the continuation path. The old role-based guard handled that case correctly (verified by reverting it).
2. **The latch's reset was unreachable** → dead code, latch outlived its cause. Fixed by the re-probe above.

Both existed because every test drove the verifier to a *terminal* verdict; nothing exercised `INCOMPLETE` → continuation → text-only finish. The existing e2e continuation test queues a tool call every round, so it misses it too. That gap is closed.

## Tests

**10 new** (8 in `test_verifier_failsafe.py`, 2 in `test_loop_safety.py`). **Every one verified to fail without its fix**, including reverting individual hunks for the max-rounds, per-iteration-flag and re-probe cases.

| | |
|---|---|
| unit | 1419 pass |
| e2e | 39 pass |
| pre-existing failures | 2 in `test_chat_scratchpad` — scratchpad subprocess under sandboxed pytest, **identical on clean `staging`**, verified by stash rather than assumed |

## Security

Reviewed per convention. No new endpoints, no deserialization, no path handling. The helper logs a label and a fixed string — never diagnosis content, history, or an exception message, keeping the ENG-1081 property that verifier logs carry no conversation content (`_safe_error_detail`). Latch state is per-session in memory, not persisted, and holds no user data. The removed silent-COMPLETE line was dead, so removing it cannot change what users see.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -211,11 +211,21 @@ class _VerifierVerdict(BaseModel):
 #
 # 2048 is sized from a measured distribution, not one sample: 16 identical calls
 # spanned 245–1654 output tokens (median ~290). That 6.7x per-call spread is also
-# why there is no "this model can't do it" latch — one truncation is a tail
-# sample, not proof about the next turn — and why the 4096 retry exists rather
-# than a single bigger budget. 1024 was measurably too small. Nothing pays for
-# headroom it doesn't use; first-party models answer in 43–115 tokens either way.
+# why *truncation* never latches — one truncation is a tail sample, not proof
+# about the next turn — and why the 4096 retry exists rather than a single bigger
+# budget. 1024 was measurably too small. Nothing pays for headroom it doesn't
+# use; first-party models answer in 43–115 tokens either way.
+#
+# A *hard* failure does latch, which is a different claim: a 400 rejecting the
+# forced `tool_choice` is a statement about the model, not about this transcript
+# (ENG-1095/ENG-1155). See `_verifier_latched`.
 _VERIFIER_TOKEN_BUDGETS = (2048, 4096)
+
+# Turns a latched session skips before spending one verdict call to see whether
+# the cause has gone away (user switched model, gateway fix shipped). Without a
+# re-probe the latch is permanent for the session and "reset on a successful
+# verdict" can never fire, since a latched session makes no verdict calls.
+_VERIFIER_LATCH_REPROBE_TURNS = 10
 
 # Appended to the verifier system prompt. Shortens the preamble on narrating
 # models but is not sufficient alone (0/3 at 256 with it), so it pairs with the
@@ -474,6 +484,7 @@ class ChatSession:
         # failure twice in a row latches, a successful verdict clears it.
         self._verifier_hard_failures = 0
         self._verifier_latched = False
+        self._verifier_latch_skips = 0
         self._context_pressure_threshold = s.context_pressure_threshold
         self._max_consecutive_errors = s.max_consecutive_errors
         self._resilience_nudge_at = s.resilience_nudge_at
@@ -2470,14 +2481,22 @@ class ChatSession:
         # task isn't actually done yet.
         continuation = 0
         _max_rounds_hit = False
-        # True once the verification block has put the assistant reply in
-        # history, so the post-loop fallback knows not to append it a second
-        # time (ENG-1155 double-append).
+        # Set per verification-loop iteration (see below): tells the post-loop
+        # fallback the current reply is already in history, so it must not append
+        # it a second time (ENG-1155 double-append).
         _reply_persisted = False
         import logging as _logging
         _verifier_log = _logging.getLogger(__name__)
 
         while True:  # Completion verification loop
+            # Per-iteration, NOT per-turn: the flag means "the reply currently in
+            # `llm_response` is already in history". A continuation replaces
+            # `llm_response`, so a stale True would make the post-loop fallback
+            # skip a reply that was never persisted — dropping the answer the
+            # user just read on any INCOMPLETE turn whose continuation needs no
+            # tools (tool_round stays 0, so the loop breaks at the gate below
+            # before the append).
+            _reply_persisted = False
             tool_round = 0
             error_streak: dict[str, int] = {}
             resilience_nudged: set[str] = set()
@@ -2923,14 +2942,31 @@ class ChatSession:
             # (ENG-1155). The turn ends on the model's own answer, which is
             # already in history — unverified, but that beats a per-turn
             # interruption we know the cause of.
+            # Re-probe occasionally so the latch can't outlive its cause: the
+            # user may switch off the broken model mid-session, or the gateway
+            # fix may land. Without this, "reset on a successful verdict" is
+            # unreachable — a latched session skips every verdict call, so there
+            # is never another success to reset on. A failed re-probe stays
+            # latched and does NOT re-diagnose (the >=2 branch below breaks
+            # before the diagnosis), so the cost is one verdict call per
+            # _VERIFIER_LATCH_REPROBE_TURNS turns.
             if self._verifier_latched:
+                self._verifier_latch_skips += 1
+                if self._verifier_latch_skips < _VERIFIER_LATCH_REPROBE_TURNS:
+                    _verifier_log.info(
+                        "completion-verifier skipped — latched after %d consecutive "
+                        "hard failures (skip %d/%d before re-probe); "
+                        "continuation=%d/%d tool_rounds=%d",
+                        self._verifier_hard_failures, self._verifier_latch_skips,
+                        _VERIFIER_LATCH_REPROBE_TURNS, continuation,
+                        self._max_continuations, tool_round,
+                    )
+                    break
                 _verifier_log.info(
-                    "completion-verifier skipped — latched after %d consecutive "
-                    "hard failures; continuation=%d/%d tool_rounds=%d",
-                    self._verifier_hard_failures, continuation,
-                    self._max_continuations, tool_round,
+                    "completion-verifier re-probing after %d skipped verifications",
+                    self._verifier_latch_skips,
                 )
-                break
+                self._verifier_latch_skips = 0
 
             # Ask the cheap coding model to self-assess completion over a compact,
             # text-rendered view of the recent conversation: enough context for
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2985,8 +2985,9 @@ class ChatSession:
                 self._verifier_latch_skips += 1
                 if self._verifier_latch_skips < _VERIFIER_LATCH_REPROBE_TURNS:
                     _verifier_log.info(
-                        "completion-verifier skipped — latched after %d consecutive "
-                        "hard failures (skip %d/%d before re-probe); "
+                        "completion-verifier skipped — latched after %d hard "
+                        "failures with no successful verdict between them "
+                        "(skip %d/%d before re-probe); "
                         "continuation=%d/%d tool_rounds=%d",
                         self._verifier_hard_failures, self._verifier_latch_skips,
                         _VERIFIER_LATCH_REPROBE_TURNS, continuation,
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import httpx
 import random
 from collections.abc import AsyncIterator, Callable
 from dataclasses import asdict, dataclass, field, replace
@@ -220,6 +221,25 @@ class _VerifierVerdict(BaseModel):
 # forced `tool_choice` is a statement about the model, not about this transcript
 # (ENG-1095/ENG-1155). See `_verifier_latched`.
 _VERIFIER_TOKEN_BUDGETS = (2048, 4096)
+
+# Verdict-call failures that are transient by nature rather than statements about
+# the model's capability. The latch exists for a model that *cannot* produce a
+# verdict (kimi-K3 rejecting the forced `tool_choice` with a 400, ENG-1095); a
+# dropped connection or a read timeout says nothing about that, and counting it
+# would let two blips switch verification off for a whole re-probe window
+# (review: pnewsam on #299).
+#
+# `TransientProviderError` is ENG-673's typed mapping and is listed first for
+# intent, though `OSError` already covers it (it subclasses `ConnectionError`).
+# `OSError` also covers `asyncio.TimeoutError`, which is `TimeoutError` — an
+# `OSError` subclass — on 3.11+. `httpx.TransportError` covers the transport the
+# OpenAI SDK uses, which is reachable when an error escapes the SDK's own
+# wrapping. Anything outside this set still latches, bounded by the re-probe.
+_TRANSIENT_VERDICT_ERRORS: tuple[type[BaseException], ...] = (
+    TransientProviderError,
+    OSError,
+    httpx.TransportError,
+)
 
 # Turns a latched session skips before spending one verdict call to see whether
 # the cause has gone away (user switched model, gateway fix shipped). Without a
@@ -1562,8 +1582,8 @@ class ChatSession:
             log.warning(
                 "%s diagnosis returned no content — nothing appended; history "
                 "keeps its existing tail (the pre-verification reply on the "
-                "verifier paths, the final tool-call message on the "
-                "max-tool-rounds path)",
+                "verifier paths, the injected SYSTEM pause message on the "
+                "max-tool-rounds path, since the cap block appends that last)",
                 label,
             )
 
@@ -3044,15 +3064,16 @@ class ChatSession:
                     )
                     if not retrying:
                         break
-                except TransientProviderError as exc:
+                except _TRANSIENT_VERDICT_ERRORS as exc:
                     # Explicitly NOT a latch candidate. The latch is for a model
                     # that cannot produce a verdict at all (kimi-K3 rejecting the
                     # forced tool_choice with a 400, ENG-1095) — a statement about
-                    # capability. A typed transient error is the opposite claim:
-                    # retryable, and already retried upstream (ENG-673). Counting
-                    # it would let two provider blips disable verification for the
-                    # rest of the session. The turn still gets its honest
-                    # diagnosis (ENG-1079); it just doesn't latch.
+                    # capability. A transient failure is the opposite claim:
+                    # retryable, and typed ones are already retried upstream
+                    # (ENG-673). Counting them would let two provider blips
+                    # disable verification for a whole re-probe window. The turn
+                    # still gets its honest diagnosis (ENG-1079); it just doesn't
+                    # latch. See `_TRANSIENT_VERDICT_ERRORS` for what qualifies.
                     verdict_failure = "transient"
                     _verifier_log.info(
                         "completion-verifier verdict=TRANSIENT budget=%d error=%s",
@@ -3100,17 +3121,23 @@ class ChatSession:
                         # buys nothing once the cause is known to recur. One
                         # diagnosis per session (ENG-1155).
                         #
-                        # "Hard" = neither truncation nor a typed
-                        # TransientProviderError (see that except-clause above —
-                        # typed transients never latch). Untyped provider blips
-                        # (a generic ConnectionError, say) still count: two of
-                        # those in a row do latch, bounded by the re-probe
-                        # un-latching within _VERIFIER_LATCH_REPROBE_TURNS once
-                        # the provider recovers.
+                        # "Hard" = neither truncation nor anything in
+                        # `_TRANSIENT_VERDICT_ERRORS` (see that except-clause
+                        # above). Connection drops and timeouts are transient and
+                        # never latch; what remains is the shape ENG-1095 has —
+                        # a provider that rejects the call itself.
+                        #
+                        # Not strictly "consecutive" either: the counter is reset
+                        # only by a *successful* verdict, so hard → truncated →
+                        # hard still latches on the second hard failure. That is
+                        # deliberate — only a real verdict proves the model can
+                        # produce one, and a truncation in between is no evidence
+                        # that it can (review: pnewsam on #299).
                         self._verifier_latched = True
                         _verifier_log.info(
-                            "completion-verifier latched after %d consecutive hard "
-                            "failures — skipping further verification this session",
+                            "completion-verifier latched after %d hard failures with "
+                            "no successful verdict between them — skipping further "
+                            "verification this session",
                             self._verifier_hard_failures,
                         )
                         break
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1560,8 +1560,10 @@ class ChatSession:
             self._append_history({"role": "assistant", "content": text})
         else:
             log.warning(
-                "%s diagnosis returned no content — history falls back to the "
-                "pre-verification reply",
+                "%s diagnosis returned no content — nothing appended; history "
+                "keeps its existing tail (the pre-verification reply on the "
+                "verifier paths, the final tool-call message on the "
+                "max-tool-rounds path)",
                 label,
             )
 
@@ -2946,10 +2948,19 @@ class ChatSession:
             # user may switch off the broken model mid-session, or the gateway
             # fix may land. Without this, "reset on a successful verdict" is
             # unreachable — a latched session skips every verdict call, so there
-            # is never another success to reset on. A failed re-probe stays
-            # latched and does NOT re-diagnose (the >=2 branch below breaks
+            # is never another success to reset on. A hard-failing re-probe stays
+            # latched and does NOT re-diagnose (the latched branch below breaks
             # before the diagnosis), so the cost is one verdict call per
             # _VERIFIER_LATCH_REPROBE_TURNS turns.
+            # A re-probe that TRUNCATES (or hits a typed TransientProviderError)
+            # instead is intended to fall through to the honest diagnosis below,
+            # and it leaves the latch set: neither counts toward or against the
+            # latch (truncation is a tail sample, see _VERIFIER_TOKEN_BUDGETS;
+            # typed transients never latch by definition), so only a successful
+            # verdict clears it. A latched session re-probing into a
+            # persistently-verbose model therefore diagnoses once per re-probe
+            # cycle rather than never — matching "every truncated turn keeps its
+            # honest diagnosis".
             if self._verifier_latched:
                 self._verifier_latch_skips += 1
                 if self._verifier_latch_skips < _VERIFIER_LATCH_REPROBE_TURNS:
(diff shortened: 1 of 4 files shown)


alecantu7:
## Adversarial self-review

Attacked my own diff for paths the new tests don't cover. **Two real defects, both introduced by this PR**, both now fixed in `3b7e2a2` with a regression test verified to fail without its fix. Plus three findings I decided not to act on, with reasoning.

---

### 🔴 1. `_reply_persisted` was turn-scoped — dropped the answer on a text-only continuation

**The worst kind of bug for this PR to have: the same defect it exists to fix, re-introduced somewhere else.**

The flag is supposed to mean *"the reply currently in `llm_response` is already in history."* I set it once per turn. Failure path:

1. Verifier returns `INCOMPLETE` → continuation fires.
2. The continuation finishes the answer **with no tool calls** (common — the point of a continuation is often just "finish writing").
3. Next loop iteration: `tool_round` stays `0`, so the loop breaks at the `verify_min_tool_rounds` gate **before** the reply is appended.
4. `_reply_persisted` is still `True` from iteration 1 → post-loop fallback skips → **the answer the user just read is not in history at all.**

Reproduced directly — history ended on the SYSTEM continuation injection with no assistant turn:

```
[3] assistant FIRST_DRAFT
[4] user      SYSTEM: Task verification determined this task is not yet complete (attempt 1/…
              ← FINAL_ANSWER_USER_READ missing entirely
```

**The old role-based guard handled this case correctly** — I verified by reverting just that line and re-running the probe (`True` before, `False` after). So this was a clean regression, not a pre-existing hole.

Not academic: ENG-1134 documents `INCOMPLETE` over-firing, so this path carries real traffic.

**Fix:** reset the flag at the top of each verification-loop iteration — per-iteration, matching what the flag actually claims. Test: `test_text_only_continuation_answer_is_not_dropped`.

### 🔴 2. The latch's reset was unreachable — it outlived its cause

ENG-1155's "Done when" says *reset on a successful verdict*. My implementation made that **impossible**: a latched session skips every verdict call, so `verdict is not None` never happens again and `self._verifier_latched = False` was dead code.

Consequence: a user who hits two hard verifier failures and then **switches off the broken model** gets no verification for the rest of that session. Sessions are long-lived, so that's not a corner.

**Fix:** re-probe with one verdict call every `_VERIFIER_LATCH_REPROBE_TURNS` (10) skipped verifications. Success clears the latch; failure stays latched and does **not** re-diagnose (the `>=2` branch breaks before the diagnosis), so the cost is one call per 10 turns and the "one diagnosis per session" property holds. Test: `test_latched_verifier_reprobes_and_can_recover`.

---

### Considered and deliberately not changed

**A truncation between two hard failures doesn't reset the counter.** The ticket says "twice *consecutively*", so strictly a truncation in between should break the streak. Left as-is: resetting on truncation means a model alternating truncation/hard-failure never latches, which is the case the latch is for. Two hard failures in one session is a pattern regardless of what happened between them. Flagging in case you read "consecutive" more strictly than I did.

**`_stream_handback_diagnosis` doesn't persist anything if `plan_stream_with_recovery` raises.** Unchanged from #276's behaviour — the exception propagates and the turn fails visibly, which is right. Persisting a partial diagnosis would put text in history the user may never have seen.

**The empty-diagnosis case still lets the fallback append the reply.** Intentional and pre-existing (`test_empty_diagnosis_is_logged_not_circular` pins it): stale-but-real beats a blank assistant turn. With finding 1 fixed there's no double-append, because the fallback now runs only when the current reply genuinely isn't in history.

---

**Process note on finding 1:** it existed because every test I wrote drove the verifier to a *terminal* verdict. Nothing exercised `INCOMPLETE` → continuation → text-only finish. The e2e `test_continuation_limit_respected` queues a tool call every round, so it never hits it either. That gap is now closed, and it's a reminder that the continuation path is the least-tested branch of the four.

alecantu7:
@pnewsam — review please. This is the follow-up you scoped out of #276 on 2026-07-29 ("predates this PR, follow-up feedback rather than a merge blocker"), so both items here are the ones you deliberately deferred.

**Three commits, read in order** — the first is the failing baseline, so the diff shows the defect caught and then closed:

1. `fce54ea` — tests only, 3 deliberately failing. Establishes the bug.
2. `ae23b31` — the fix for both ENG-1155 items.
3. `3b7e2a2` — two defects I found reviewing my own commit 2, and their fixes.

**Where I'd most value your eyes, in priority order:**

1. **`_reply_persisted` semantics** (`_stream_and_handle_tools`). This is the riskiest thing in the PR — I got it wrong once already. It replaced the old `history[-1]["role"] != "assistant"` guard, which was too loose (double-appended on hand-backs) but happened to be correct on a path my first attempt broke: `INCOMPLETE` → continuation finishes with no tool calls → early break before the append. It's now per-iteration. **If there's a fifth exit from that loop I haven't accounted for, this is where it bites**, and the symptom is silent — a lost assistant turn, not an error.

2. **Whether the latch should exist at all in this shape.** It's a mitigation for ENG-1095 (kimi-K3 rejecting forced `tool_choice`), not a fix. Once the gateway lands the capability fix, this becomes dead weight that suppresses verification after any two hard failures. Options if you'd rather not carry it: gate it behind a setting, or drop it and let ENG-1095 fix it properly. I implemented it because the ticket asked for it, but I don't feel strongly.

3. **The re-probe interval (10 turns).** Arbitrary. It exists because without it "reset on a successful verdict" is unreachable. If you have a view on the right number, or think a model-change trigger is better than a turn count, say so.

**Two things you'll want to know that aren't in the ticket:**

- **The bug had a fourth site**, not the three the ticket predicts: the 25-round tool-cap circuit breaker. It's the highest-traffic of the four — the cap fires far more often than `STUCK` (~1.9% of verdicts). Fixed here rather than leaving the shared helper covering 3 of 4.
- **PR #293 — a Dependabot postcss bump in `/docs`** — was squash-merged from a stale base and reverted an ENG-1079 line on `staging` (the `status, reason = "COMPLETE", "verifier unavailable"` fail-safe, put back directly above the comment saying why it must not exist). **No prod impact** — it was dead code and never reached `main`. Removed here since it's in the block I rewrote. The separate question worth someone's decision: a `/docs` dependency bump was able to modify `anton/core/session.py`.

**State:** 1419 unit + 39 e2e green. The 2 `test_chat_scratchpad` failures are pre-existing on clean `staging` (scratchpad subprocess under sandboxed pytest) — verified by stashing and re-running, not assumed. Every new test was checked to fail without its fix, including reverting individual hunks.

Not merging this myself — it's yours when you're happy.

alecantu7:
## Adjacency note — a queued change lands in the same block, and it isn't in this PR

New findings on ENG-836 today overlap this PR's territory without belonging in it. Recording so review isn't done twice and so ENG-836 doesn't collide by surprise.

### The overlap

This PR modifies the **max-tool-rounds hand-back block** (the 25-round cap: append reply → inject a SYSTEM pause → stream a diagnosis), changing it to capture that diagnosis into history via the shared helper.

ENG-836 now proposes changing the **content of that same message**: today it says *"I've paused — shall I continue?"*, which is the right question for a merely-long task and the wrong invitation for an environmentally impossible one, because it asks the user to authorise more spend on something that cannot succeed. Making it feasibility-aware touches the same lines this PR rewrites.

No conflict — different concerns, and this PR lands first. Flagging the sequencing only.

### What is deliberately NOT in this PR

Three things surfaced today that a reviewer might reasonably expect to see here:

**1. Tail-biased tool-result truncation.** `_render_tool_result_content` truncates with `content[:cap]` at `tool_cap = 400` — *head* truncation, verified live on both `main` and `staging`. A Python traceback's information is at the end, so the verifier is shown the shape of a failure and denied its cause. This is the suspected reason `STUCK` fires only ~1-in-5 on the Kiranam blocker (ENG-836 recorded 4 COMPLETE / 4 INCOMPLETE / 1 STUCK on one wall).

Out of scope here on purpose: it changes *which verdict comes back*, and its acceptance test is a before/after measurement (1-in-5 → N-of-N) that this PR can't demonstrate. This PR is the failure-path layer — what happens when a verdict hands the task back — not verdict quality. Confirmed by grep that nothing here touches `_render_verify_transcript`, `_render_tool_result_content` or `tool_cap`.

**2. Making truncation direction testable.** Neither existing test can observe it: `test_includes_truncated_tool_result_evidence` has a ~25-character fixture against a 400-character cap, so nothing truncates despite the name; `test_truncates_large_tool_output` truncates `"x" * 5000`, which makes head and tail indistinguishable. **Flipping `content[:cap]` to `content[-cap:]` would break no test.**

Also deliberately not fixed here. The only assertion that is currently truthful is "head truncation happens", which pins the bug — so the fixture belongs in ENG-836's PR, asserting the *corrected* direction. Doing it here means it gets rewritten immediately, and in the meantime the suite documents a bug as intended behaviour.

**3. ENG-1211's STUCK fixture rule and N-of-N requirement.** Both now in that ticket's description. They govern an eval that doesn't exist yet; nothing to add to code.

### Why this is worth a comment rather than commits

Same reasoning that kept the `session.py` scope tight in the first place: this PR already carries four defects found by review (three mine, one from a second pass), all concentrated in the new latch state and the reply flag. Adding a verdict-quality change on top would mix a measurable behaviour change into a PR whose risk profile is already the interesting part. Better to ship this, measure that separately.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed `session.py`, `structured.py`, and the two test files. The history-desync fix, the four-path unification behind `_stream_handback_diagnosis()`, and the verifier latch/re-probe all hold up against the tests, and the two self-review defects are genuinely fixed. No blocking or major issues.

### Validation

- Traced all five terminal paths; `_reply_persisted`'s per-iteration reset prevents the double-append and preserves the text-only continuation answer.
- Latch state machine matches its tests: latch on the 2nd hard failure (no 2nd diagnosis), clear on success, re-probe on the 10th skip, failed/truncating re-probe stays latched.
- New `TransientProviderError` / `_safe_error_detail` / `StructuredOutputError` symbols are imported; `structured.py` drops a genuine duplicate import.
- CI green: `run-tests`, `CodeQL`, `Analyze`, Snyk.

### Findings

Three line-specific findings — 1 Minor (untyped provider errors can latch verification off) and 2 Nits (comment/docstring accuracy) — are posted as **inline comments**. None blocking.

**Nit (no diff line):** the PR body's "Three commits" table is stale — the branch now has 5 commits (`2044d8b`, `d4b7785` added later). Worth refreshing so the narrative matches the diff.

anton/core/session.py
```diff
@@ -2915,24 +3044,76 @@ async def _stream_and_handle_tools(
...
+                        "completion-verifier verdict=TRANSIENT budget=%d error=%s",
+                        budget, _safe_error_detail(exc),
+                    )
+                    break
                 except Exception as exc:
                     # Enough to tell the failure modes apart — four used to
                     # collapse into one "verifier unavailable" line — but never
                     # the exception message, which can carry conversation
                     # content (ENG-1081).
+                    verdict_failure = "hard"
```
pnewsam, line 3089: **Minor:** Untyped provider exceptions classify as `"hard"` and can latch verification off. The typed `TransientProviderError` is carved out just above, but a bare `asyncio.TimeoutError` / `httpx.ConnectError` reaching this generic handler counts toward the latch — two on consecutive multi-step turns with no successful verdict between them disable verification for up to `_VERIFIER_LATCH_REPROBE_TURNS` (10) turns. Bounded by the re-probe and reset on any success, so exposure is narrow and self-healing; flagging only as residual risk if other transient failures reach this path untyped. Not blocking.

anton/core/session.py
```diff
@@ -2915,24 +3044,76 @@ async def _stream_and_handle_tools(
...
+                        # A failed re-probe: the cause is still there. Stay
+                        # latched with its own log line — re-announcing "latched
+                        # after N failures" with an ever-growing N would read as
+                        # a new event each cycle. No re-diagnosis (one per
+                        # session, ENG-1155).
+                        _verifier_log.info(
+                            "completion-verifier re-probe failed — staying latched"
+                        )
+                        break
+                    if self._verifier_hard_failures >= 2:
```
pnewsam, line 3117: **Nit:** "Consecutive hard failures" is slightly imprecise — `_verifier_hard_failures` is reset only by a successful verdict, not by an intervening truncation or typed-transient turn, so `hard → truncated → hard` still latches on the second hard failure. The behavior is defensible (only a real verdict proves the model *can* produce one); the comment just reads as if the two hards must be strictly adjacent.

anton/core/session.py
```diff
@@ -1512,6 +1531,42 @@ def factory_validated():
...
+            if diagnosis_response is not None
+            else ""
+        )
+        if text:
+            self._append_history({"role": "assistant", "content": text})
+        else:
+            log.warning(
+                "%s diagnosis returned no content — nothing appended; history "
+                "keeps its existing tail (the pre-verification reply on the "
+                "verifier paths, the final tool-call message on the "
```
pnewsam: **Nit:** On the max-tool-rounds path the injected `SYSTEM` user message is appended last (the cap block above), so an empty diagnosis leaves *that user message* as the tail — not "the final tool-call message" as this line states. Cosmetic.

pnewsam (commented):
Moving the three line-specific findings from my approval summary to inline comments (they anchor to changed lines). None blocking — approval stands.

anton/core/session.py
alecantu7: Fixed in `1b19a1a` rather than accepted as residual risk — you were right that a timeout says nothing about whether the model can produce a verdict, which is the only thing the latch is meant to infer.

New `_TRANSIENT_VERDICT_ERRORS = (TransientProviderError, OSError, httpx.TransportError)`:

- `OSError` covers builtin `ConnectionError` and, on 3.11+, `asyncio.TimeoutError` — which *is* `TimeoutError`, an `OSError` subclass (checked, not assumed).
- `httpx.TransportError` covers your `ConnectError` case. `httpx` is already a dependency via the OpenAI SDK, and `openai.py` imports it directly, so this needed no new dep.
- The typed entry is kept first for intent even though `OSError` subsumes it.

Anything outside the set still latches, bounded by the re-probe — so the narrow-and-self-healing property you noted is preserved for genuinely unknown failures.

`test_transient_provider_errors_never_latch` is now parametrized over four modes (typed, asyncio timeout, connection reset, httpx connect). Narrowing the tuple back to typed-only fails exactly 3 of the 4, so the widening is pinned.

anton/core/session.py
alecantu7: Fixed in `1b19a1a`. Your reading of the behaviour is right and it stays — only a real verdict proves the model can produce one, so a truncation in between is no evidence and should not reset the counter. It was only the comment that overclaimed.

Comment and log line both corrected: the log now reads *"latched after N hard failures with no successful verdict between them"* instead of "N consecutive hard failures".

That block also asserted *"untyped provider blips (a generic `ConnectionError`, say) still count"* — which the Minor fix above makes false. Reconciled in the same commit rather than left contradicting the code, since a comment that lies about behaviour is what produced two of the other findings on this PR.

anton/core/session.py
alecantu7: Fixed in `1b19a1a`. Confirmed by re-reading the cap block: it appends the assistant reply, then the injected SYSTEM pause message, then streams the diagnosis — so on an empty diagnosis the tail is that SYSTEM *user* message, exactly as you said, not a tool-call message.

Now reads: *"the injected SYSTEM pause message on the max-tool-rounds path, since the cap block appends that last"*.

Cosmetic in effect, but worth the fix — this log line exists to explain what history holds when the diagnosis is empty, so a wrong description defeats its only purpose.

alecantu7:
Thanks @pnewsam — all three findings addressed in `1b19a1a`, replies on each thread. The PR body's commit table is refreshed too (it said three; it's six).

**The Minor I fixed rather than accepted.** You framed the untyped-transient latch as narrow and self-healing, which it was — but the underlying point stands on its own: the latch exists to infer *"this model cannot produce a verdict"*, and a socket timeout carries no information about that. So it shouldn't be counted regardless of how bounded the exposure is.

`_TRANSIENT_VERDICT_ERRORS = (TransientProviderError, OSError, httpx.TransportError)` covers your two examples: `OSError` catches `asyncio.TimeoutError` (verified — it *is* `TimeoutError`, an `OSError` subclass on 3.11+) and builtin `ConnectionError`; `httpx.TransportError` catches `ConnectError`, with no new dependency since `openai.py` already imports httpx. Unknown failures still latch, bounded by the re-probe, so the property you liked is intact.

**One thing your Nit surfaced that I'd have missed.** The latch comment block also claimed *"untyped provider blips still count"* — true when written, and made false by the Minor fix in the same commit. Reconciled rather than left standing. That's now the third comment-versus-code contradiction on this PR, after the `_VERIFIER_TOKEN_BUDGETS` "there is no latch" line and the max-tool-rounds tail description you caught. Worth noting as a pattern in this file: the behaviour is subtle enough that comments drift faster than the code, and they've been wrong in a way that would mislead the next reader every time.

**Tests:** `test_transient_provider_errors_never_latch` parametrized over four failure modes; narrowing the tuple back to typed-only fails exactly 3 of 4, so the widening is pinned rather than assumed. 1425 unit + 39 e2e pass, same 2 pre-existing `test_chat_scratchpad` failures.

Approval predates these three commits, so re-approve or re-review as you prefer — the Minor changed behaviour, not just comments. Not merging it myself either way.

tests/test_verifier_failsafe.py
```diff
@@ -196,6 +198,579 @@ def fake_plan_stream(**kwargs):
...
+        assert diagnoses == 1, (
+            f"expected exactly one diagnosis per session, got {diagnoses}"
+        )
+        # Turns 1 and 2 call the verifier (the 2nd establishes the pattern);
+        # turns 3 and 4 must not pay for it at all.
+        assert verdict_calls == 2, (
+            f"latched session kept calling the verifier ({verdict_calls} calls)"
+        )
+        assert session._verifier_latched is True
+        assert any("latched after 2 consecutive hard failures" in r.message
```
pnewsam: **Nit (optional):** this assertion was written to pin the latch *announcement* line, but that line's wording changed to "…hard failures with no successful verdict between them" in `1b19a1a`. The substring still matches — but now via the *skip* log line (`session.py:2988`), which still reads "…2 consecutive hard failures" and fires on turns 3–4. So the test passes without actually guarding the announcement message it was written for. Harmless; tweak only if you want the assertion to mean what its comment says (e.g. assert on the new announcement wording).

pnewsam (approved):
## Re-review — `1b19a1a`

**Verdict: APPROVE** — all three findings from my earlier review are resolved, CI green.

- Minor (untyped transients could latch): fixed via `_TRANSIENT_VERDICT_ERRORS`; the parametrized test covers typed/asyncio-timeout/connection-reset/httpx-connect. The `OSError` widening is broad but a deliberate, low-consequence tradeoff.
- Nit ("consecutive"): comment and announcement log line corrected to match the reset-on-success semantics.
- Nit (docstring tail): now correctly names the injected SYSTEM pause message on the max-tool-rounds path.

One new optional Nit inline. Nothing blocking.

tests/test_verifier_failsafe.py
alecantu7: Fixed in `4705ff8` — taking this one rather than leaving it optional, because you found a real hole and the cause was mine.

Verified your reading by rendering both format strings as they stood at `1b19a1a`:

- needle `"latched after 2 consecutive hard failures"` → **matches the skip line**
- same needle → **does not match the announcement**

So the assertion was passing entirely via the per-turn skip log on turns 3–4, and would not have failed if the announcement had been deleted outright.

**Root cause was an incomplete fix of your earlier Nit.** `1b19a1a` reworded the announcement but left the skip log at `session.py:2988` still saying "N consecutive hard failures" — two logs about the same counter, one accurate and one not. Both now read "…N hard failures with no successful verdict between them".

The assertion now anchors on `"skipping further verification this session"`, which only the announcement contains, asserts **exactly one** such record, and checks the reset-semantics wording. Confirmed it fails when the announcement log is removed.

Worth noting this is the fourth test on this PR that read as coverage and was not — after the e2e stub's `queue_verification_stuck()` that no scenario called, `test_continuation_limit_respected` asserting only that a SYSTEM prompt reached the model, and `test_truncates_large_tool_output` using `"x" * 5000` so head and tail truncation are indistinguishable (that last one is ENG-836's). Same shape every time: the assertion pins plumbing rather than the behaviour it names. Recorded on ENG-836 as review guidance — check what the fixture makes observable, not whether a test with a relevant name exists.

alecantu7:
Thanks @pnewsam — re-review Nit fixed in `4705ff8`, reply on the thread. Not leaving it as optional: the assertion was genuinely dead and the cause was an incomplete fix of your *earlier* Nit on my part.

**What happened.** `1b19a1a` reworded the latch *announcement* log but left the per-turn *skip* log still saying "N consecutive hard failures" — two logs about the same counter, one accurate and one not. Rendering both format strings as they stood at that commit confirms your reading exactly: the needle `"latched after 2 consecutive hard failures"` matches the skip line and **not** the announcement. So the test passed via the skip log on turns 3–4 and would not have failed if the announcement had been deleted outright.

Both logs now read "…N hard failures with no successful verdict between them". The assertion anchors on `"skipping further verification this session"` — unique to the announcement — asserts exactly one such record, and checks the reset-semantics wording. Verified it fails when the announcement is removed.

**On your `OSError` note** ("broad but a deliberate, low-consequence tradeoff") — agreed, and recording the reasoning so it isn't re-litigated later: the cost of a false transient is one wasted verdict call per turn on a session that should have latched, bounded by the existing re-probe. The cost of a false hard is verification silently off for ten turns. Asymmetric, so erring broad is the right side. If a genuinely permanent `OSError` shows up in practice, the fix is to narrow by errno rather than to re-widen the latch.

**One pattern worth carrying out of this PR.** This was the fourth test here that read as coverage and wasn't:

| test | why it didn't guard what it named |
|---|---|
| `queue_verification_stuck()` (e2e stub) | helper existed, no scenario ever called it |
| `test_continuation_limit_respected` | asserted a SYSTEM prompt *reached* the model, not what history holds |
| `test_deterministic_hard_failure_latches_after_one_diagnosis` | matched the skip log, not the announcement (this fix) |
| `test_truncates_large_tool_output` | `"x" * 5000` makes head and tail truncation indistinguishable (ENG-836's, not fixed here) |

Same shape each time: the assertion pins plumbing rather than the behaviour it names, and the name makes it look done. Recorded on ENG-836 as review guidance, and it's the argument for ENG-1211 — behavioural verifier coverage that asserts verdicts rather than prompts.

Approval stands from your side; `4705ff8` is one log string and one assertion, no behaviour change. Merge is still yours whenever you're happy.

pnewsam (approved):
## Re-review — `4705ff8`

**Verdict: APPROVE** — the one open Nit is resolved correctly. CI green (`run-tests` incl.).

`4705ff8` rewords the per-turn skip log to match the announcement ("N hard failures with no successful verdict between them") and re-anchors the test assertion on `"skipping further verification this session"` — a substring unique to the announcement line (verified: absent from the skip and re-probe-failed logs). It now asserts exactly one announcement plus the reset-semantics wording, so it genuinely guards what it was written for. No blocking or major issues; nothing further from me.

PR #300, opened 2026-08-03
ENG-972 Add stable, human-readable labels to saved connections

StpMax:
Sibling PRs:
https://github.com/mindsdb/cowork/pull/552
https://github.com/mindsdb/cowork-server/pull/264

Connections saved in the local vault get a persistent, unique `user_label` (e.g. "prod-db") instead of only the auto-generated slug (e.g. `postgres-7e8971c3`).

- `/connect` prompts for a Label on new connections; `/edit` prompts for one after a successful test
- Save/update/reconnect messages now show the label alongside the slug
- `/list` gains Label and Identity columns, renames Name to Slug
- `/remove` picker shows the label
- System prompt's connected-datasource section is restructured per-connection, showing Label, Engine, Host/Database/Account, and credential env vars
- `connect_new_datasource` tool accepts and reports `user_label`
- Fixed an env-var leak where `_`-prefixed bookkeeping fields (including labels) could be injected as `DS_*` env vars and get scrubbed as false-positive secrets

No breaking changes — connections without a label still work, falling back to the slug.

image

### DIFF
--- a/anton/tools.py
+++ b/anton/tools.py
@@ -190,6 +190,9 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
                     conn_name = uuid.uuid4().hex[:8]
                     while vault.load(engine_def.engine, conn_name) is not None:
                         conn_name = uuid.uuid4().hex[:8]
+                    from anton.utils.datasources import default_user_label, ensure_unique_user_label
+                    requested_label = tc_input.get("user_label") or default_user_label(vault, engine_def.engine)
+                    test_credentials["_user_label"] = ensure_unique_user_label(vault, requested_label)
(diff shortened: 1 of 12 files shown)


anton/tools.py
```diff
@@ -262,12 +267,16 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
 
     if new_connections:
         slug = next(iter(new_connections))
+        engine_saved, name_saved = slug.split("-", 1)
```
ea-rus: can the engine name be with hyphen ? 
for example: optima-api-a12345

StpMax: fixed

ea-rus (approved):

PR #301, opened 2026-08-04
Check uv.lock matches pyproject.toml in CI

hamishfagg:
Snyk reads `pyproject.toml` and `uv.lock` directly, so no exported requirements file is needed here. What matters is that the lockfile Snyk scans matches the manifest we install from, which `uv lock --check` enforces in the unit test workflow.

Part of [ENG-875](https://linear.app/mindsdb/issue/ENG-875/setup-snyk-security-scanning-for-repositories).


pyproject.toml
```diff
@@ -33,13 +33,15 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+clipboard = [
+    "Pillow>=12.3.0",
+]
+
+[dependency-groups]
```
lucas-koontz, line 40: **The contributing docs install an extra that no longer exists.**

`docs/docs/developer/contributing.md:55` says `pip install -e ".[dev]"`. pip just warns "does not provide the extra 'dev'" and installs without it, so the next line in that block (`pytest tests/`) fails. That page publishes to Pages, so it's what external contributors follow.

```bash
uv sync --group dev
uv run pytest tests/
```

lucas-koontz (commented):
Small and clean at 48d0eb7 — the lock diff is exactly the extras-to-dev-dependencies move, no version changes, no new packages, no resolved-URL churn. Checked the sibling repos too: cowork-server, cowork and minds all pull `anton-agent` without extras, so dropping `dev` from the published metadata doesn't reach anyone.

One thing to fix on the way in (B1) — the contributing docs still tell people to install the `dev` extra this removes. Not blocking, but it's the public contributor path.

Unrelated but worth knowing: anton is the only one of the three repos still on `astral-sh/setup-uv@v5`; auth and inference are on `@v7`.

PR #302, opened 2026-08-04
ENG-1234: Dependabot config + scope guard so a dependency bump can't smuggle source changes

alecantu7:
## What

Adds the missing `.github/dependabot.yml` **and** a CI guard that fails any dependency PR touching anything but manifests and lockfiles. The guard is the part that prevents recurrence; the config is hygiene.

## Why

anton#293 — *"chore(deps): bump postcss from 8.5.15 to 8.5.25 in /docs"* — merged into `staging` on 2026-08-03 and also reverted merged verifier code:

- `anton/core/session.py`: re-added `status, reason = "COMPLETE", "verifier unavailable"`, the silent fail-safe **ENG-1079 deliberately deleted**, placed directly above the comment explaining why it must not exist. That line is what produced the original "the agent dies silently" reports (ENG-1081: 24.9% of verdicts, 33 users).
- `anton/core/llm/structured.py`: a duplicate `from typing import Any, NoReturn`.

**No production impact** — the line was unreachable and never reached `main`; the 2026-08-02 release shipped correct code. Removed in #299. **This PR is about the mechanism.** There are **7 open Dependabot alerts**, so more of these PRs are coming.

How it happened, verified: Dependabot security PRs are based on the **default branch** (`main`); this one was **retargeted at `staging`**. Because the ENG-1081 hotfix exists as two different commits (`c51afe8` on main, `fc832d5` on staging — same code, unrelated identities), the merge base fell back to `c8de488`, the *previous* week's release. From there both sides had independently edited the same block of `session.py`, producing a real conflict, which a human resolved by **keeping both sides**. A squash merge then hid everything under the postcss title. `git merge-tree` confirms a clean merge would have been correct.

## The `target-branch` trap

The obvious fix — `target-branch: staging` — is **wrong**, and the config says so in a comment where the next person will read it. Dependabot **security** updates only run against the default branch, and GitHub's docs state that for them *"you should not specify a `target-branch`"*. Setting it would silence the security PRs this repo depends on.

So security PRs stay based on `main`. The rule that follows: **merge them into `main` and let the weekly sync carry them — do not retarget them at `staging`.**

## What's here

| file | purpose |
|---|---|
| `.github/scripts/check_dependency_pr_scope.py` | the check. In a script, not inline YAML, so it is testable against a real diff |
| `.github/workflows/dependency-scope.yml` | runs it for `dependabot`-authored or `dependencies`-labelled PRs |
| `tests/test_dependency_pr_scope.py` | 16 cases; the first is #293's actual file list |
| `.github/dependabot.yml` | npm `/docs` + `github-actions`, grouped, labelled, no `target-branch` |

Guard details worth reviewing:

- Keys off the **API file list** — the diff against the merge base, which is exactly what exposed `session.py` in #293.
- Listens to `labeled`/`unlabeled`, because the `dependencies` label is often added after opening.
- Matches manifests on **basename**, so a future workspace needs no config change. Negative cases cover the near-misses: `anton/package_json_helper.py` must not pass on a substring, and `.github/CODEOWNERS` must not pass just for being under `.github/`.

**Verified live against both real PRs:** #293 → exit 1 naming both source files; #294 (the clean `brace-expansion` security PR) → exit 0.

## Deliberately out of scope

- **Python ecosystem.** `uv` is not a valid `package-ecosystem` value, and `pip` against a `uv.lock` project risks PRs that edit `pyproject.toml` without regenerating the lock. Left for a deliberate decision in ENG-1234 rather than guessed at.
- **The root enabler.** A hotfix squashed into `main` and re-squashed into `staging` guarantees this conflict class for any branch crossing between them. Fixing that means changing the release convention — a bigger conversation, noted on the ticket.

## Timing

`dependabot.yml` is read from the **default branch**, so it takes effect after the next weekly release, not on merge. The workflow guard is active as soon as this lands, since workflows run from the PR's own branch.

## Security

Reviewed per convention. The workflow requests only `contents: read` and `pull-requests: read`, uses the default `GITHUB_TOKEN`, and runs on `pull_request` (not `pull_request_target`), so a fork PR cannot obtain write access or secrets. No PR-controlled string is interpolated into shell — author and labels arrive via `env`, and the file list is written to a file and piped to the script rather than expanded inline. The script only reads path strings; it never executes or opens the files it is judging.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,68 @@
+# Dependabot configuration (ENG-1234).
+#
+# NOTE ON `target-branch` — deliberately absent.
+#
+# PRs in this repo normally branch from `staging` and target `staging`, so
+# pointing Dependabot at `staging` looks like the obvious thing to do. It is not:
+# Dependabot *security* updates only run against the default branch, and GitHub's
+# docs state that for security updates "you should not specify a target-branch".
+# Setting it would silence the security PRs this repo actually relies on (7 open
+# alerts when this file was written).
+#
+# The consequence is that security PRs arrive based on `main`. Merge them into
+# `main` and let the weekly sync carry them into `staging` — do NOT retarget them
+# at `staging`. Retargeting is what caused anton#293, where a postcss bump in
+# /docs reverted merged verifier code in `anton/core/session.py`: the branch came
+# from `main`, `staging` had moved on, and the resulting conflict was resolved by
+# keeping both sides.
+#
+# This file does not prevent that on its own — the CI scope guard in ENG-1234
+# does. What it buys is explicit ecosystems, grouping, PR limits and labels
+# instead of defaults.
+#
+# Python is intentionally not listed. `uv` is not a valid package-ecosystem
+# value, and running `pip` against a uv.lock project risks PRs that edit
+# pyproject.toml without regenerating the lock. Tracked in ENG-1234.
+version: 2
+updates:
+  # Docs site (Docusaurus). Dev-only — these packages are not shipped in the
+  # anton wheel, which is why the blast radius of anton#293 was zero.
+  - package-ecosystem: npm
+    directory: /docs
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    labels:
+      - dependencies
+      - javascript
+    commit-message:
+      prefix: "chore(deps)"
+      prefix-development: "chore(deps-dev)"
+    # One PR per batch of routine bumps rather than one per package. Security
+    # updates are unaffected by grouping and still arrive individually.
+    groups:
+      docs-minor-and-patch:
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
+
+  # Workflow actions. Pinned action versions rot quietly and are a supply-chain
+  # surface, so keep them current.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    labels:
+      - dependencies
+      - github-actions
+    commit-message:
+      prefix: "chore(ci)"
+    groups:
+      actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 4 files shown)


hamishfagg (approved):

alecantu7 (commented):
Second review pass (post-approval, pre-merge) — **no code changes needed**; everything material verified against reality rather than the PR text. Recording the results plus the one operational gap that decides whether this actually prevents recurrence.

**Verified:**
- #293's file list via the API is exactly the test's `PR_293_FILES` — the regression case is the real incident, not a reconstruction. 16/16 tests pass on the PR head; both YAML files parse.
- The `target-branch` reasoning is correct per GitHub's docs (security updates run only against the default branch), and `main` is confirmed the default branch, so the timing note is right too.
- The workflow is injection-safe (PR-controlled strings arrive via `env`, never interpolated), read-only (`contents: read` + `pull-requests: read`, plain `pull_request`), and uses `--paginate` so a large file list can't truncate into a false pass.
- The always-run-then-early-exit design means the job reports a status on every PR to main/staging — i.e. it is *safe to mark as a required check* without blocking normal PRs. That matters below.
- `dependencies` and `javascript` labels exist; `github-actions` didn't (Dependabot skips missing labels with only a log warning) — created it just now, so the config behaves as written.

**The gap that matters — enforcement.** Neither `staging` nor `main` has any required status checks (classic protection 404s, rulesets carry none). So this guard is advisory: the same human motion that merged #293 — merge without reading — can merge past a red X. The check was designed to be requirable; the follow-through is an admin adding `Dependency PR scope / scope` to required checks on both branches. Until that happens, this PR reduces the *odds* of recurrence, not the *possibility*.

**Known holes, fine to accept but worth having on the record:**
- `.github/workflows/` is allowlisted (the actions ecosystem legitimately edits workflow files), so a conflict-smuggled *workflow* revert would pass the guard. Tightening would require content-level inspection (only `uses:` lines changed) — not worth it now, but it's the known blind spot.
- A PR that edits the guard script itself is judged by its own edited copy. The accidental variant fails closed (script deleted → `python3` errors → red); only a deliberate no-op edit passes, which is outside this threat model (a malicious human just wouldn't add the label).
- Removing the `dependencies` label un-triggers the guard for human PRs — symmetric with opting in, and Dependabot-authored PRs are still caught by author, so this is consistent rather than a leak.

PR #304, opened 2026-08-04
feat(ENG-763): stream ToolProgress markers for generic tools, with a correlated + error-safe tool_done

StpMax:
## Summary

Adds a streaming progress protocol for generic (non-scratchpad) tools, and makes the closing `tool_done` marker reliable enough for a consumer to actually build a UI on top of it.

image

with error:
image

**Producer side (`ToolDef.handler` can now be an async generator):**
- `anton/core/tools/progress.py` — new `ToolProgress(text: str)` marker dataclass.
- `ToolRegistry.dispatch_tool_stream()` / `dispatch_tool()` (`anton/core/tools/registry.py`) — a tool handler can now be a plain coroutine (as before) *or* an async generator that yields `ToolProgress` markers and returns its final result as the last item. Both dispatch through the same code path.
- `ChatSession.turn_stream()` (`anton/core/session.py`) — the generic (non-scratchpad, non-interactive) tool-dispatch branch now forwards each `ToolProgress` as `StreamTaskProgress(phase="tool_progress", message=..., id=tc.id)`.
- `anton/chat_ui.py` — renders `tool_progress` markers as permanent console lines in the CLI (previously they'd only ever update a transient spinner footer and vanish).

**Reliability fix on `tool_done` (added after manual verification surfaced two real gaps):**
- `tool_done` now carries `id=tc.id`, matching the existing `scratchpad_done` precedent — without it, nothing downstream can correlate a `tool_done` event back to the specific tool call it closes.
- `tool_done` is now guaranteed to fire even when the tool handler raises mid-stream (previously the exception jumped straight past the `tool_done` yield to the outer handler, so a failed streaming tool never signalled completion at all). This uses a local `try/except Exception` + re-raise, deliberately **not** `try/finally` with a `yield` inside it — `yield`ing while an async generator is unwinding a `GeneratorExit` raises `RuntimeError: async generator ignored GeneratorExit`, which is exactly the path Esc-cancellation takes in the CLI. Verified both failure modes directly (a broken `try/finally` version does crash on cancellation; this version doesn't, and still emits `tool_done` on a handler exception).
- `StreamTaskProgress`'s docstring updated to document `tool_done` alongside the other markers that carry a correlation `id`.

## Why

This is the producer half of a feature to show a tool's live progress in downstream UIs (cowork chat). Nothing here changes existing tool behavior.

## Testing

- `tests/test_tool_progress.py`, `tests/test_tool_registry_streaming.py`, `tests/test_turn_stream_tool_progress.py` — new/extended. Covers: progress events arrive in order with the correct tool-call id; progress markers never leak into tool-result history; a handler exception mid-stream still produces a clean `"...failed: ..."` result (not a crash); `tool_done` carries the tool call's id in both the success and error paths; closing the generator mid-progress (simulating Esc-cancellation) does not raise.
- Full suite green: `uv run --extra dev python -m pytest tests/` → 1470 passed, 17 skipped.

## Notes for reviewers

- Companion PRs: `cowork-server` (maps these events to a renderable role) and `cowork` (renders the live step). Both degrade gracefully if this hasn't shipped yet — they just stay inert.
https://github.com/mindsdb/cowork-server/pull/274
https://github.com/mindsdb/cowork/pull/565


ea-rus:
**Blocker: a failed generic tool renders as a success (green ✔) in the CLI**

The generic-tool branch in `session.py` catches the handler error into
`_tool_error`, yields `StreamTaskProgress(phase="tool_done", ...)`, then
re-raises. That re-raise is caught by the outer `except` right below
(`result_text = f"Tool '{tc.name}' failed: {exc}"; tool_ok = False`), so the
turn completes normally — nothing propagates out. `tool_ok=False` correctly
feeds error-tracking and the LLM's tool_result, but the `tool_done` event
carries no failure signal, and `chat_ui` treats *any* `tool_done` as completion.

Traced repro (generic handler raises):
1. `on_tool_use_end` prints the activity line and sets `act.printed = True`.
2. Branch yields `tool_done(message=tc.name)`; the consumer processes it before
   the re-raise unwinds → `tool_done` handler sets `act.done = True`.
3. `raise _tool_error` → caught by the outer `except` → `result_text = "Tool 'x'
   failed"`, turn finishes normally.
4. `finish()` flushes every `act.done and not done_line_printed` →
   `_print_done_line` prints `  ✔ (Worked: …)` — there is no failure variant.

Result: the user sees a green ✔ for a tool that actually failed. `cowork-server`'s
`anton_harness` stream_formatter has the same gap — it emits a `tool_done` for
the failed step with no error indication, so the web UI marks it complete too.

Fix — give `tool_done` a verdict and branch the render on it:
- add an `ok`/`failed` flag to `StreamTaskProgress` for `tool_done` (set it from
  `tool_ok`/the `_tool_error` path) and branch `_print_done_line` (`✔` vs `✗`); or
- emit a distinct `tool_failed` phase from the failure path and render it as a
  failure line.

Both consumers (CLI `_print_done_line` and the server stream formatter) need to
distinguish done-ok from done-failed.

StpMax:
@ea-rus fixed

ea-rus (approved):

PR #305, opened 2026-08-04
ENG-836: verifier detects environment walls — tail-biased truncation + STUCK definition

### LINKED ISSUE #2: anton

alecantu7:
## What

Two measured changes so the completion verifier recognizes an unrecoverable environment wall (missing OS driver, no root) and stops instead of force-continuing:

1. **Tail-biased tool-result truncation.** `_render_tool_result_content` clipped oversize tool results with `content[:cap]` at `tool_cap = 400` — head truncation. A traceback's decisive final line (`libodbc.so.2: cannot open shared object file`, `Permission denied`) is exactly what got discarded: the verifier saw the shape of a failure and was denied its cause. Now: keep a small head (`cap // 3` — what ran, the `[error]` marker), elide the middle with a marker, spend the rest on the tail.
2. **STUCK definition names environment walls** (ticket fix #2). Appended to the `_VerifierVerdict` STUCK bullet, shipped verbatim as A/B-tested: *"This includes environment walls: an OS-level package, driver, or system library the task requires but that cannot be installed in this environment (no root/sudo, package manager blocked). Repeated failed workarounds for the same underlying blocker mean STUCK, not INCOMPLETE — even if the assistant says it will try another approach."*

## Why — measured, not assumed

[ENG-836](https://linear.app/mindsdb/issue/ENG-836): a Pro user's environmentally-blocked task burned ~5M tokens on INCOMPLETE-driven continuations. Live A/B against the prod gateway on `haiku` (the tier-default coding model), production verify call byte-for-byte, Kiranam-shaped fixture, n=12/arm — full method + verdict transcripts on the ticket (2026-08-04 comment):

| Truncation | STUCK schema | STUCK verdicts |
|---|---|---|
| head-clip (prod today) | original | **0/12** |
| tail-biased (commit 1) | original | 1/12 — verifier cites "no sudo/apt access" and still says INCOMPLETE |
| tail-biased | extended (commit 3) | **12/12** |
| head-clip | extended | 12/12 (this fixture's assistant narrates the blocker; see below) |
| controls (extended schema) | | unfinished-unblocked stays INCOMPLETE 12/12; errored-but-recovered stays COMPLETE 12/12 (ENG-1134-compatible) |

The schema wording is the binding lever for the verdict; the truncation fix is the evidence guarantee for failures the assistant does NOT narrate or actively misreads (the real session's 404-page-parsed-as-valid-archive class) and is what puts the precise cause into the `reason` string feeding the user-facing STUCK diagnosis.

## Tests

- Truncation **direction** is now pinned: the old `"x" * 5000` fixture couldn't distinguish head- from tail-clipping (reverting the fix broke nothing). Re-fixtured with `HEAD_MARKER`/`CAUSE_MARKER` + elision assert, plus `test_traceback_cause_survives_truncation` (>400-char pyodbc-shaped traceback, cause at the end).
- Schema wording pinned in a new `tests/test_verifier_verdict_schema.py` — labeled honestly as wording pins (behavioural verifier coverage is ENG-1211's scope), including a guard that ENG-1134's recovered-error rule survived the edit. New file deliberately, to avoid churn against ENG-1155's #299 which extends `test_verifier_failsafe.py` heavily.
- Verifier/session suites: 164 passed. Full suite: 1462 passed; the 2 `test_chat_scratchpad.py` failures pre-exist on clean `origin/staging` (and don't fail in CI).

## Interaction with ENG-1155 (#299, merging)

No textual conflict (#299 doesn't touch the verdict schema or the renderer) and no semantic overlap: 1155's latch handles the verifier *call itself* dying (hard provider failures), this PR changes what a *working* verifier concludes. Complementary in the right direction — #299 also fixed the STUCK hand-back path this PR makes fire more often.

## Notes

- Clip output can exceed `cap` by the ~30-char elision marker — bounded, negligible against the post-ENG-1081 budgets.
- Security pass: string clipping + prompt wording in an internal verifier call — no new input surface, no secrets in logs or output.

## Still open on the ticket

Feasibility-aware cap-exhaustion message (re-scope item 3), repeated-run regression eval in cowork_evals, and the deferred original fixes #3/#4 — though the 12/12 result suggests re-measuring after this lands before building them.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -195,7 +195,19 @@ class _VerifierVerdict(BaseModel):
             "including when it implied success but the data its answer actually "
             "depends on errored or came back empty and was never recovered.\n"
             "- STUCK: a hard blocker prevents completion (missing credentials, an "
-            "unavailable service, or a permission the assistant does not have)."
+            "unavailable service, or a permission the assistant does not have). "
+            # Environment walls must be named explicitly or the verifier files
+            # them under INCOMPLETE and force-continues into the wall: measured
+            # 0-1/12 STUCK without the sentences below vs 12/12 with them, on
+            # the same blocked-task transcript, with unblocked-unfinished and
+            # errored-but-recovered controls unmoved (ENG-836 live A/B,
+            # 2026-08-04).
+            "This includes environment walls: an OS-level package, driver, or "
+            "system library the task requires but that cannot be installed in "
+            "this environment (no root/sudo, package manager blocked). Repeated "
+            "failed workarounds for the same underlying blocker mean STUCK, not "
+            "INCOMPLETE — even if the assistant says it will try another "
+            "approach."
         )
     )
     reason: str = Field(description="One brief sentence explaining the verdict.")
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -363,7 +363,15 @@ def _clip_keep_cause(text: str, cap: int) -> str:
     head = cap // 3
     tail = cap - head
     elided = len(text) - head - tail
-    return f"{text[:head]}\n[... {elided} chars elided ...]\n{text[-tail:]}"
+    marker = f"\n[... {elided} chars elided ...]\n"
+    # Near-threshold inputs: when the marker costs at least what it removes,
+    # clipping would EXPAND the text (a 401-char input at cap=400 came back
+    # ~428 chars with a "[... 1 chars elided ...]" in the middle). Pass those
+    # through whole — same worst-case output bound, and the invariant becomes
+    # "clipping never returns more characters than it was given" (#305 review).
+    if len(text) <= cap + len(marker):
+        return text
+    return f"{text[:head]}{marker}{text[-tail:]}"
 
 
 def _render_tool_result_content(content, cap: int) -> str:
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -335,15 +335,36 @@ _SOLVABILITY_CLAUSE = (
 )
 
 
+def _clip_keep_cause(text: str, cap: int) -> str:
+    """Clip ``text`` to ~``cap`` chars keeping both ends, biased to the tail.
+
+    A failing tool result names its cause at the END — a traceback's final
+    line is the one that says ``libodbc.so.2: cannot open shared object
+    file`` — while the head carries what ran and the ``[error]`` marker. A
+    plain ``text[:cap]`` therefore showed the verifier the shape of a failure
+    but discarded its cause, which is how an unrecoverable environment wall
+    kept getting judged INCOMPLETE instead of STUCK (ENG-836). Keep a small
+    head, elide the middle, and spend most of the budget on the tail.
+    """
+    if len(text) <= cap:
+        return text
+    head = cap // 3
+    tail = cap - head
+    elided = len(text) - head - tail
+    return f"{text[:head]}\n[... {elided} chars elided ...]\n{text[-tail:]}"
+
+
 def _render_tool_result_content(content, cap: int) -> str:
     """Render a tool_result's content as bounded plain text.
 
     Never serializes raw payloads: a multimodal result (e.g. read_image) can
     carry megabytes of base64, so we keep only text blocks and mark images with
     a placeholder rather than ``json.dumps``-ing the whole thing (ENG-716).
+    Oversize content is clipped tail-biased so a failure's cause survives
+    (ENG-836).
     """
     if isinstance(content, str):
-        return content[:cap] or "(empty result)"
+        return _clip_keep_cause(content, cap) or "(empty result)"
     if isinstance(content, list):
         parts: list[str] = []
         for block in content:
@@ -353,8 +374,9 @@ def _render_tool_result_content(content, cap: int) -> str:
                 parts.append((block.get("text") or "").strip())
             elif block.get("type") in ("image", "image_url"):
                 parts.append("[image]")
-        return (" ".join(p for p in parts if p)[:cap]) or "[non-text result]"
-    return str(content)[:cap]
+        joined = " ".join(p for p in parts if p)
+        return _clip_keep_cause(joined, cap) or "[non-text result]"
+    return _clip_keep_cause(str(content), cap)
 
 
 def _render_verify_transcript(
--- a/tests/test_verifier_verdict_schema.py
+++ b/tests/test_verifier_verdict_schema.py
@@ -1,0 +1,45 @@
+"""Pins the _VerifierVerdict STUCK definition's environment-wall coverage (ENG-836).
+
+Honest scope note: these are wording pins, not behavioural coverage — they
+exist so a later edit can't silently drop the sentences whose effect was
+measured live (0-1/12 STUCK without them vs 12/12 with them on a blocked-task
+transcript, controls unmoved; see the 2026-08-04 A/B on ENG-836). Behavioural
+verifier coverage is ENG-1211's scope.
+"""
+
+from __future__ import annotations
+
+from anton.core.session import _VerifierVerdict
+
+
+def _status_description() -> str:
+    return _VerifierVerdict.model_fields["status"].description
+
+
+def test_stuck_definition_names_environment_walls():
+    desc = _status_description()
+    # The uninstallable-OS-dependency case (no root, package manager blocked)
+    # must be named INSIDE the STUCK bullet — the ENG-836 incident class.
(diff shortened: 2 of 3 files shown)


alecantu7 (commented):
Self-review (adversarial pass) findings:

1. **Misleading test name left in place** — `test_includes_truncated_tool_result_evidence` truncates nothing (25-char fixture vs a 400-char cap); it actually pins the ENG-716 property that tool-result *content* reaches the verifier at all. The ticket's 2026-08-04 comment explicitly calls this out as one of three looks-like-coverage-but-isn't instances in this subsystem, and this PR is the natural place to stop perpetuating it. **Fixing in a follow-up commit**: rename to `test_includes_tool_result_evidence_verbatim_when_short` with a comment stating what it does and doesn't cover.

2. **`_clip_keep_cause` returns up to `cap + ~30` chars** (the elision marker is not charged to the budget). Considered and accepted: charging the marker to the budget complicates the arithmetic for a bounded ~30-char overshoot against a 2048/4096-token verifier budget (post-ENG-1081). Docstring says "~cap" deliberately; the re-fixtured test still bounds payload chars at 400.

3. **Tail-bias applies to successful results too**, not just failures — a long query result now shows its head and tail rather than just its head. Accepted: the head third preserves "what ran", and for judging completion the end state of an output is at least as informative as its beginning.

alecantu7:
**Live A/B result (2026-08-04, prod gateway, `haiku`, n=12/arm)** — full table and method on [ENG-836](https://linear.app/mindsdb/issue/ENG-836): head-clip (prod today) = 0/12 STUCK; this PR's tail-biased clip alone = 1/12 STUCK — the verifier now *cites* the causes ("libodbc/unixodbc with no sudo/apt access") but the original STUCK schema definition doesn't cover environment walls, so the verdict stays INCOMPLETE. With an extended STUCK definition on top: 12/12 STUCK, with clean over-correction controls (unblocked-unfinished stays INCOMPLETE 12/12, errored-but-recovered stays COMPLETE 12/12).

So: this PR is the evidence-fidelity half (and the guarantee for failures the assistant doesn't narrate or misreads — the 404-parsed-as-archive class), not the verdict fix by itself. The schema-wording change (ticket fix #2) is the binding lever and is measured, but its second sentence overlaps the ENG-1155 latch design — landing it is a scope call tracked on the ticket, deliberately not silently folded in here.

alecantu7 (commented):
## Adversarial review, round 2 (full PR, post-ENG-1155 merge)

**Finding 1 — CONFIRMED and fixed: the branch base predated #299's merge, so every prior test run and CI pass validated this PR *without* ENG-1155's code present.** GitHub reported MERGEABLE/CLEAN (no textual conflicts), but "merges cleanly" is not "works combined" — this PR makes STUCK verdicts far more frequent, and STUCK now routes through #299's rewritten hand-back path (`_stream_handback_diagnosis`, the `_reply_persisted` guard). Fixed: rebased onto post-1155 `origin/staging`, full suite run on the combination — **1483 passed, 0 failed**, including #299's extended `test_verifier_failsafe.py` and the e2e loop-safety scenarios. Force-with-lease pushed (`ad12678`).

**Finding 2 — PROBED and cleared: transient-blocker over-trigger.** The A/B's controls covered "unfinished with no errors" (E) and "one error, recovered" (F), but not the middle case this wording most plausibly regresses: *repeated* failures that would succeed on retry. If "repeated failed workarounds for the same underlying blocker mean STUCK" caught rate limits, we'd hand back tasks that plain backoff would finish. Measured live (same harness, Shopify-429 fixture with `Retry-After` headers, two consecutive rate-limit errors, n=12/schema): extended schema **12/12 INCOMPLETE** (continues, correct); original schema 11/12 INCOMPLETE + 1 WAITING. No over-trigger; results in `eng836_ab_results_G1_G2.json` on the ticket's evidence trail.

**Finding 3 — cleared: no duplicated definitions or external consumers.** Grepped anton, cowork-server, cowork, cowork_evals for the STUCK definition text and verdict-string consumers: single definition site (`_VerifierVerdict`), nothing else parses or mirrors it.

**Accepted with rationale (carried from round 1):** clip output can exceed `cap` by the ~30-char elision marker (bounded, docstring says "~cap"); tail-bias also applies to successful long results (deliberate — head third keeps "what ran"); the schema tests are wording pins, not behavioural coverage (labeled as such in-file; behavioural verifier coverage is ENG-1211's scope).

Post-rebase CI is running on `ad12678`; previous green runs were pre-rebase and shouldn't be cited as evidence for the combination.

alecantu7:
**Pendulum probe H (addendum to review round 2):** the grayest over-rotation case — repeated failures of the agent's *own fixable code bug* (Kiranam's pymssql sub-bug taken alone: same blocker twice, no environment wall, fully solvable by editing code). If "repeated failed workarounds for the same underlying blocker mean STUCK" over-rotated anywhere, it's here. Measured n=12/schema on the prod gateway: extended schema **12/12 INCOMPLETE** (keeps going and fixes its own bug — correct); original schema 9/12 INCOMPLETE + 3/12 WAITING (the WAITING drift is arguably a small pre-existing misread; the extension doesn't inherit it).

Cumulative over-rotation evidence: 48/48 correct across the four non-wall fixtures (unfinished-unblocked, errored-but-recovered, repeated transient 429s, repeated self-inflicted bug), 12/12 STUCK only on the genuine wall.

anton/core/session.py
```diff
@@ -335,15 +347,36 @@ def _safe_error_detail(exc: BaseException) -> str:
...
+    but discarded its cause, which is how an unrecoverable environment wall
+    kept getting judged INCOMPLETE instead of STUCK (ENG-836). Keep a small
+    head, elide the middle, and spend most of the budget on the tail.
+    """
+    if len(text) <= cap:
+        return text
+    head = cap // 3
+    tail = cap - head
+    elided = len(text) - head - tail
+    return f"{text[:head]}\n[... {elided} chars elided ...]\n{text[-tail:]}"
```
pnewsam: **Nit:** for text only slightly over `cap`, the clipped output is *longer* than the input (e.g. a 401-char input at cap=400 yields ~133+267+marker ≈ 428 chars, plus a grammatically-off `[... 1 chars elided ...]`). Bounded and harmless for the traceback case this targets, but a `cap + margin` short-circuit would avoid expanding near-threshold results. Not blocking.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (3 files, +125/-8): the `_VerifierVerdict` STUCK wording extension, the new `_clip_keep_cause` tail-biased truncation helper and its wiring into `_render_tool_result_content`, and the test changes. The truncation logic is correct and the guard prevents head/tail overlap; the schema change is a prompt edit backed by a documented n=12/arm A/B with over-correction controls, and the tests genuinely pin the new behavior. CI is green (run-tests, CodeQL, Snyk). No actionable defects.

### Validation

- `_clip_keep_cause`: guard `len(text) <= cap` guarantees `head+tail = cap < len(text)`, so `text[:head]`/`text[-tail:]` never overlap and `elided > 0`. cap=400 -> head=133, tail=267 (tail-biased as intended).
- Only caller is `_render_verify_transcript`, which budgets tool events separately (`max_tool=12`), so the ~30-char marker overrun per result is bounded.
- `test_recovered_error_rule_is_intact` asserts against strings present verbatim in base session.py ("RECOVERED", "do NOT mark a turn incomplete..."); STUCK-slice tests match the appended wording.
- `test_truncates_large_tool_output_keeping_both_ends`: x-count = 375 <= 400 and both end markers survive.
- CI: run-tests pass (3m24s), CodeQL/Snyk pass.

### Findings

**Nit (cross-cutting):** The tail-bias now elides the *middle* of every large tool result, including successful ones whose salient data may sit mid-payload (previously head-truncation kept the first `cap` chars). The A/B controls (errored-but-recovered stays COMPLETE 12/12, unfinished stays INCOMPLETE 12/12) argue this doesn't regress the verdict, so flagging only as a documented behavioral tradeoff, not a defect.

One inline nit below. Nothing blocks merge.

anton/core/session.py
alecantu7: Fixed in b2f8f7d — pass-through when the marker would cost at least what it removes, so the invariant is now "clipping never returns more characters than it was given" (and the `1 chars elided` grammar case can no longer occur). Pinned by `test_near_threshold_output_passes_through_whole` (405-char input at cap=400 comes back verbatim). No effect on the A/B evidence or ENG-1211's eval fixtures — all of those are far past the threshold. Full suite green (1445 passed).

PR #307, opened 2026-08-04
ENG-1211: behavioural eval for the completion verifier's verdict quality

alecantu7:
# ENG-1211: behavioural eval for the completion verifier's verdict quality

The verifier's judgment rubric changed four times in a week (ENG-716, ENG-1081, ENG-1134, ENG-1155) guarded only by prompt-string assertions — tests that prove the wording reaches the model, never the judgment the model then makes. This PR adds the missing layer: real verdict calls over committed transcript fixtures, asserting the returned `status`.

[Linear ENG-1211](https://linear.app/mindsdb/issue/ENG-1211)

## What's in here

1. **`_build_verify_request()` extraction** (`anton/core/session.py`) — the verifier's prompt construction moves verbatim from the verify loop to a module-level helper. No behaviour change (full unit suite green). The eval calls this helper + `generate_object_code` with the real `_VerifierVerdict` schema and the shipped `_VERIFIER_TOKEN_BUDGETS`, so it exercises the exact production prompt by construction — a rubric edit can't drift away from what the eval measures.

2. **`tests/test_verifier_verdict_live.py`** — five incident-derived transcript fixtures × {`haiku`, `mindshub_air`} + the ENG-1081 narrating-budget guard (11 tests):

   | case | expected | source |
   |---|---|---|
   | tool errored early, recovered another way | COMPLETE | ENG-1134 |
   | implied success, data never arrived | INCOMPLETE | ENG-1134's safeguard |
   | genuine clarifying question | WAITING | ENG-716 |
   | environment wall, repeated failed workarounds | STUCK | ENG-836 |
   | stopped partway through multi-part request | INCOMPLETE | baseline |
   | narrating model reaches a verdict at shipped budgets | any verdict | ENG-1081 |

   Auto-skips without `MINDSHUB_API_KEY` (repo `.env` or env), so the default CI run is untouched. Non-determinism policy is N-of-N: 3-of-3 per case, 6-of-6 for STUCK (its pre-fix base rate was ~1-in-5, so single-run luck is a real failure mode).

   The STUCK fixture is an 845-char pyodbc traceback with the decisive `libodbc.so.2` line **last** — asserted at import time to fall entirely outside a 400-char head clip, so the case can't silently go green against the head-truncation bug ENG-836 fixed (see the 2026-08-04 comment on ENG-1211).

3. **`.github/workflows/verifier-eval.yml`** — runs the eval on PRs touching `anton/core/session.py` or the eval itself. Needs the `MINDSHUB_API_KEY` repo secret (doesn't exist yet — needs an admin); until it's added, pytest skips everything and the job stays green.

## Live measurements (prod gateway, 2026-08-04)

**Merged-world state (this branch + #305): all 11 tests pass** — 4m16s wall, including STUCK 6/6 on both models.

**The eval bites (both Done-when spot-checks):**

- **ENG-836 before/after** — `environment_wall`, 6 runs per cell:

  | rubric state | haiku | mindshub_air |
  |---|---|---|
  | pre-#305 (head clip + old STUCK wording) | 3/6 STUCK | 0/6 STUCK (all INCOMPLETE) |
  | post-#305 | **6/6** | **6/6** |

  The haiku row matches ENG-836's measured coin-flip base rate (Kiranam: 1-in-5), and the mindshub_air row extends #305's haiku-only 12/12 A/B to the narrating-model population — the wording flips both.

- **ENG-1134 revert check**: with `_VERIFIER_JUDGMENT_RUBRIC` reverted to the pre-#280 wording, `recovered_tool_error` on haiku **fails** (INCOMPLETE verdicts) — ENG-1134's own first success criterion, which shipped with no test, now has one that demonstrably regresses on revert.

- **The fixture-strength trap is real, demonstrated live**: with a first-draft 436-char traceback (cause only 36 chars past the clip, so `libodbc.so.2` partially survived head-truncation), haiku passed 6/6 STUCK *on the pre-fix rubric* — green against the live ENG-836 bug. Lengthening the traceback to 845 chars (cause fully outside the head window, asserted at import time) surfaced the real 3/6. Exactly the failure mode ENG-1211's 2026-08-04 comment predicted a tidy fixture would cause.

## Merge order

**After #305** (approved). The STUCK case asserts the A/B-validated wording #305 ships; on staging without #305, `environment_wall×mindshub_air` fails (that's the "before" measurement above, not a flake). No file conflicts — #305's session.py hunks don't overlap the extraction.

## Security pass

Test-only + a pure refactor. The eval reads `MINDSHUB_API_KEY` from env/`.env` and never logs it; fixtures are synthetic transcripts with no real credentials/PII; the workflow grants `contents: read` only and exposes the secret solely to the pytest step. No new endpoints, no input parsing.

### DIFF
--- a/tests/test_verifier_verdict_live.py
+++ b/tests/test_verifier_verdict_live.py
@@ -461,6 +461,15 @@ def _runs_for(case: Case) -> int:
 # ---------------------------------------------------------------------------
 
 
+# If this reds on a PR whose diff can't plausibly change a verdict: re-run the
+# failing case against origin/staging with the same key. Still failing -> this
+# is live gateway/model drift, i.e. the PRODUCTION verifier now judges this
+# fixture differently — a real regression that deserves its own ticket, not an
+# eval flake to be quieted (the per-run `status: reason` strings in the failure
+# message say what the model now thinks). Passing on staging -> suspect the PR
+# after all, or a transient (#307 review: the accepted lever for transients is
+# a one-retry wrapper on non-StructuredOutputError failures in `_verdict`,
+# never looser assertions).
 @pytest.mark.parametrize("model", _MODELS)
 @pytest.mark.parametrize("case", _CASES, ids=lambda c: c.name)
 async def test_verdict(model: str, case: Case):
--- a/tests/test_verifier_verdict_live.py
+++ b/tests/test_verifier_verdict_live.py
@@ -33,7 +33,10 @@ one blocker), so a single run passes ~20% of the time by luck. The STUCK case
 therefore also doubles as ENG-836's before/after measurement.
 
 Cost note: a full run is ~50 verdict calls over ~2-4k-token transcripts on
-cheap aliases.
+cheap aliases, ~4 minutes wall. Because the key is also picked up from the
+repo-root ``.env``, a plain ``pytest tests/`` on a machine with a keyed
+``.env`` runs this live — same behaviour as ``test_web_tools_live.py``.
+Deselect with ``-k "not verdict_live"`` or unset the key to skip.
 """
 
 from __future__ import annotations
@@ -168,7 +171,6 @@ class Case:
     history: list[dict]
     expected: str
     source: str
-    runs: int = 0  # 0 → module default (_RUNS, or _STUCK_RUNS for STUCK)
 
 
 # --- 1. Tool errored early, model recovered another way → COMPLETE ----------
@@ -414,7 +416,6 @@ _STUCK = Case(
     ],
     expected="STUCK",
     source="ENG-836 (Kiranam environment wall, base rate 1-in-5)",
-    runs=0,  # resolved to _STUCK_RUNS below
 )
 
 # --- 5. Stopped partway through a multi-part request → INCOMPLETE -----------
@@ -453,8 +454,6 @@ _CASES = [_RECOVERED, _IMPLIED_SUCCESS, _WAITING, _STUCK, _PARTIAL]
 
 
 def _runs_for(case: Case) -> int:
-    if case.runs:
-        return case.runs
     return _STUCK_RUNS if case.expected == "STUCK" else _RUNS
 
 
--- a/tests/test_verifier_verdict_live.py
+++ b/tests/test_verifier_verdict_live.py
@@ -108,7 +108,7 @@ def _client(model: str) -> LLMClient:
     return LLMClient.from_settings(settings)
 
 
-async def _verdict(llm: LLMClient, case: Case) -> str:
+async def _verdict(llm: LLMClient, case: Case) -> _VerifierVerdict:
     """One verdict call, replicating session.py's budget-escalation loop:
     first budget, retry once on truncation with the bigger one (ENG-1081).
     Anything else propagates — a hard failure here is a test failure, which is
@@ -117,10 +117,9 @@ async def _verdict(llm: LLMClient, case: Case) -> str:
     system, messages = _build_verify_request(case.history, case.user_message)
     for attempt, budget in enumerate(_VERIFIER_TOKEN_BUDGETS):
         try:
-            verdict = await llm.generate_object_code(
+            return await llm.generate_object_code(
                 _VerifierVerdict, system=system, messages=messages, max_tokens=budget
             )
-            return verdict.status
         except StructuredOutputError as exc:
             if exc.truncated and attempt + 1 < len(_VERIFIER_TOKEN_BUDGETS):
                 continue
@@ -128,7 +127,7 @@ async def _verdict(llm: LLMClient, case: Case) -> str:
     raise AssertionError("unreachable: budget loop exhausted without raising")
 
 
-async def _verdicts(llm: LLMClient, case: Case, n: int) -> list[str]:
+async def _verdicts(llm: LLMClient, case: Case, n: int) -> list[_VerifierVerdict]:
     # Sequential, not gathered: the gateway rate-limits per key (ENG-878
     # counts full context per request), and a burst of parallel verdict calls
     # 429ing would fail the eval for reasons that say nothing about the rubric.
@@ -467,10 +466,15 @@ def _runs_for(case: Case) -> int:
 async def test_verdict(model: str, case: Case):
     llm = _client(model)
     n = _runs_for(case)
-    statuses = await _verdicts(llm, case, n)
+    verdicts = await _verdicts(llm, case, n)
+    statuses = [v.status for v in verdicts]
+    # The verifier's own `reason` strings are the diagnostic for a red run —
+    # "why did the rubric flip" — so surface them in the failure message
+    # instead of just the statuses.
+    detail = "; ".join(f"{v.status}: {v.reason}" for v in verdicts)
     assert statuses == [case.expected] * n, (
-        f"{case.name} on {model}: expected {case.expected} x{n}, got {statuses} "
-        f"(source: {case.source})"
+        f"{case.name} on {model}: expected {case.expected} x{n} "
+        f"(source: {case.source}), got [{detail}]"
     )
 
 
@@ -481,7 +485,12 @@ async def test_narrating_model_reaches_a_verdict_at_shipped_budgets():
     which the fail-safe upstream turned into silent task death. Verdict
     *quality* on this alias is covered by the matrix above; this asserts only
     that the call survives the model's narration.
+
+    Behaviourally this overlaps the matrix (a truncation escape on the
+    recovered fixture would fail `test_verdict` too) — kept as one cheap,
+    named call so the ENG-1081 regression class stays traceable in the test
+    report even if the matrix's cases or models are later reshaped.
     """
     llm = _client(_NARRATING_MODEL)
-    status = await _verdict(llm, _RECOVERED)
-    assert status in ("COMPLETE", "WAITING", "INCOMPLETE", "STUCK")
+    verdict = await _verdict(llm, _RECOVERED)
+    assert verdict.status in ("COMPLETE", "WAITING", "INCOMPLETE", "STUCK")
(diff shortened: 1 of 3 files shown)


alecantu7 (commented):
Self-review (adversarial pass) — findings before requesting human review:

1. **`Case.runs` is dead generality.** No fixture sets it non-zero; `_runs_for` already special-cases STUCK. Speculative knobs on a test dataclass are noise — removing. *(fixed in follow-up commit)*

2. **Local-dev foot-gun.** `pytest tests/` on a machine whose repo-root `.env` carries `MINDSHUB_API_KEY` silently fires ~50 live prod-gateway calls (~4 min). `test_web_tools_live.py` set this precedent and the module docstring carries a cost note, but the .env-pickup consequence deserves an explicit sentence. *(fixed in follow-up commit)*

3. **Transient gateway blips can red an unrelated session.py PR.** A 5xx mid-eval fails the case even though it says nothing about the rubric. Accepted for now: the provider layer already retries typed transients (ENG-673), `_verdicts` runs sequentially to stay under the per-key rate limit, and the truncation retry mirrors production. If flakes show up in practice, the fix is a one-retry wrapper on non-StructuredOutput errors in `_verdict` — not looser assertions.

4. **`test_narrating_model_reaches_a_verdict_at_shipped_budgets`'s membership assertion is nearly tautological** (pydantic's Literal already guarantees it). Its real assertion is that no `StructuredOutputError` escapes the shipped budget loop; the docstring says so explicitly. Leaving as-is.

5. **Merge-order reminder:** the STUCK case asserts #305's wording — measured 3/6 (haiku) and 0/6 (mindshub_air) without it. Do not merge before #305; the workflow can't go red on that today only because the `MINDSHUB_API_KEY` secret doesn't exist yet.

alecantu7 (commented):
Adversarial review, round 2 — attacking the final diff rather than re-reading my earlier findings. Five findings, three fixed in the follow-up commit:

**1. (fixed) The path filter guarded the rubric, not the verdict call path.** `anton/core/llm/structured.py` owns truncation classification (`looks_truncated`, `raise_unusable_tool_call`) and `client.py` owns the forced-`tool_choice` call — a regression in either re-creates the ENG-1081 silent-death class *without touching session.py*, and the eval would never have fired. Both added to the workflow paths. Consciously excluded: `openai.py` — it changes for many non-verifier reasons and the wire format is exercised transitively on every eval run anyway.

**2. (fixed) No concurrency control.** Once the secret exists, every force-push to a session.py PR would stack a full live run (~50 prod calls each). Added a per-ref concurrency group with cancel-in-progress.

**3. (fixed) Red runs discarded the best diagnostic.** The assertion showed `got ['INCOMPLETE', ...]` but threw away the verifier's own `reason` strings — the thing you'd actually read to understand why a rubric edit flipped a verdict. Failure messages now carry `status: reason` per run. Validated live (1 case).

**4. (recommendation, needs an admin anyway) Secret hygiene.** The workflow uses `pull_request`, not `pull_request_target`, so fork PRs cannot read the key — but any same-repo branch can. When adding `MINDSHUB_API_KEY`, use a dedicated low-tier key for CI, not a prod admin key. Blast radius if leaked: someone else's LLM calls on our gateway (ENG-1189 family), so key choice matters.

**5. (accepted, documented) `test_narrating_model_reaches_a_verdict_at_shipped_budgets` is behaviourally redundant** — the matrix already runs that fixture on the narrating alias 3×, and a truncation escape fails there too. Kept deliberately: one cheap named call that keeps the ENG-1081 regression class visible in the test report even if the matrix is later reshaped. Docstring now says so.

Also verified while attacking: `user_message` at the production call site is always a str (`user_msg_str`), so the extracted helper's `str | None` contract is honest; fixture `source` labels (ticket IDs) never reach the model — only `history` and `user_message` flow through `_build_verify_request`; and module-level asserts run even when the key-skip is active, so the fixture-geometry guard can't be silently bypassed.

alecantu7:
#305 merged to staging (`701a91f`) — the merge-order dependency in the PR body is now satisfied. Since `pull_request` CI runs against the merge ref (staging + this branch), the verdict-eval job here now exercises the merged rubric; once the `MINDSHUB_API_KEY` secret exists, the STUCK case asserts the shipped wording directly. This PR is ready for human review.

alecantu7:
@lucas-koontz — heads-up, this PR needs one admin action from you after (or before) it merges: **a new Actions repository secret on `mindsdb/anton` named `MINDSHUB_API_KEY`**, containing a MindsHub gateway API key (an `mdb_` key for `api.mindshub.ai`).

**Why:** this PR adds `.github/workflows/verifier-eval.yml`, a CI job that makes **real LLM calls** — it fires the completion verifier at the prod MindsHub gateway over committed transcript fixtures and asserts the verdict that comes back (behavioural coverage for the verifier rubric, which changed 4× in a week guarded only by prompt-string tests — see ENG-1211). Real calls need a real key, hence the secret.

**When it runs:** only on PRs that touch the verifier's code path — `anton/core/session.py`, `anton/core/llm/structured.py`, `anton/core/llm/client.py`, or the eval file itself. Not on every PR, not on pushes.

**Cost per run:** ~50 verdict calls over ~2–4k-token transcripts on the `haiku` and `mindshub_air` aliases, ~4 minutes wall clock — pennies. A concurrency group cancels superseded runs on force-push, so rapid iteration doesn't stack charges.

**Behaviour without the secret (current state):** pytest detects the missing key and skips every case; the job reports green. So nothing is broken today — but the gate is effectively **unarmed**: a rubric edit that regresses a verdict would sail through looking covered. Adding the key is what makes the eval actually bite on PRs.

**Key choice recommendations:**
- Use a **dedicated key created for CI**, not a personal key and not anything admin/prod-privileged. If it ever leaks, the blast radius should be "someone can make LLM calls billed to that key" and the remediation should be "revoke that one key" — no shared credentials, easy rotation.
- If our gateway supports low-tier/limited keys, that tier is plenty — the eval only ever calls cheap aliases with small budgets.
- Exposure surface is deliberately narrow: the workflow runs on `pull_request` (NOT `pull_request_target`), so **fork PRs never receive the secret**; it's passed only to the single pytest step, the workflow has `contents: read` permissions only, and the test code never logs the key.

Happy to walk through the workflow file if useful — it's 45 lines. Once the secret exists, the next PR touching `session.py` will run the eval live automatically (both #305's STUCK wording and the ENG-1134 recovered-error rule are among the guarded behaviours).

tests/test_verifier_verdict_live.py
```diff
@@ -0,0 +1,496 @@
...
+async def test_verdict(model: str, case: Case):
+    llm = _client(model)
+    n = _runs_for(case)
+    verdicts = await _verdicts(llm, case, n)
+    statuses = [v.status for v in verdicts]
+    # The verifier's own `reason` strings are the diagnostic for a red run —
+    # "why did the rubric flip" — so surface them in the failure message
+    # instead of just the statuses.
+    detail = "; ".join(f"{v.status}: {v.reason}" for v in verdicts)
+    assert statuses == [case.expected] * n, (
```
pnewsam, line 484: **Minor (non-blocking, and you already flagged it):** once the `MINDSHUB_API_KEY` secret exists, this exact-match `statuses == [expected]*n` fires on every PR touching `session.py`/`structured.py`/`client.py`. N-of-N repetition guards against single-run luck but not against gateway-side model drift — a future model update on the alias could turn STUCK 6/6 red on an otherwise-unrelated PR. Your accepted fix (a one-retry wrapper on non-`StructuredOutputError` errors in `_verdict`) is the right lever if flakes show up; flagging here as the line to watch after enablement.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed all three files against `session.py`, `client.py`, `provider.py`, and `settings.py` at head `a0cd228`. The verdict holds because the source change is a byte-identical extraction (verified against the removed inline block) and the rest is a live-only test that auto-skips in CI — no production behaviour changes. No blocking findings.

### Validation

- Diffed the extracted `_build_verify_request()` against the deleted inline block: identical except one reworded comment. Prompt output unchanged.
- Confirmed every imported production symbol, signature, and `AntonSettings` field the test uses exists at head SHA (`StructuredOutputError.truncated/output_tokens/stop_reason`, `generate_object_code` and `from_settings` signatures, `minds-cloud` normalisation).
- Confirmed `asyncio_mode = "auto"` → the bare `async def` tests actually run (not vacuous), and the fixture-geometry `assert` is module-level so it runs even under the key-skip.
- CI: `verdict-eval` is green via the **skip path** (no `MINDSHUB_API_KEY` secret), not a real eval run; unit `run-tests` passed.

### Non-blocking notes (all already disclosed in your two self-reviews — reinforcing, not new)

**Nit — enablement:** the eval delivers zero protection until an admin adds the `MINDSHUB_API_KEY` secret; today's green check is a skip. Worth tracking as the key follow-up so the PR doesn't read as "verifier now guarded" before it actually is.

**Minor — merge order:** the STUCK case asserts #305's rubric wording (measured 3/6 haiku, 0/6 mindshub_air without it). Merging before #305 is only safe because the absent secret keeps the job skipped; once the secret exists, order matters. Documented in the PR body.

The line-anchored flakiness note is inline on `test_verdict`. Nothing here blocks merge.

tests/test_verifier_verdict_live.py
alecantu7: Agreed on the mechanism, with one reframe I've now encoded next to the assertion (docs-only commit): **drift-red is signal, not flake**. If a gateway model update turns the STUCK case red, the *production* verifier just started judging environment walls differently — that's a real regression we'd otherwise discover through the next Kiranam, and the eval firing on an unrelated PR is the alarm working, just with awkward timing. The comment above `test_verdict` now gives whoever hits it the 2-minute triage: re-run the failing case against `origin/staging` with the same key — still failing → live drift, file a ticket against the verifier (the `status: reason` strings in the failure message say what the model now thinks), don't quiet the eval; passing → suspect the PR or a transient, and the transient lever stays the one-retry wrapper, never looser assertions.

PR #308, opened 2026-08-04
ENG-1276: explicit tool-failure outcomes drive the error streak, not text search

alecantu7:
# ENG-1276: explicit tool-failure outcomes drive the error streak, not text search

anton decided whether a tool result was a failure by searching its text for five phrases (`[error]`, `Task failed:`, `failed`, `timed out`, `Rejected:`). That drives the per-tool failure streak → resilience nudge (at 2) → circuit breaker (at 5), and it misclassified in both directions. The false-negative direction is the mechanism behind the ENG-836 runaway: genuine failures whose text lacked the phrases **reset** the streak, so ~50 minutes of driver ping-pong with interleaved false "successes" never counted to five. The error taxonomy was being carried in prose — ENG-350's rejection message literally had a comment saying the word "failed" was there so the matcher would count it.

[Linear ENG-1276](https://linear.app/mindsdb/issue/ENG-1276)

## What changed

1. **`ToolOutcome` envelope** (`registry.py`): `dispatch_tool` now always returns `ToolOutcome(content, ok, reason)`. Handlers returning plain `str | list` wrap as `ok=None` (unmigrated); handlers can declare their verdict by returning an outcome. `reason` is a short machine-comparable cause — unused by the streak, but it is the input the ENG-1286 root-cause thrash breaker will key on.
2. **`_apply_error_tracking(..., ok=None)`** (`session.py`): an explicit verdict decides outright; `ok=None` falls back to the legacy substring match **and logs when the fallback classifies an error**, so remaining unmigrated handlers are discoverable rather than assumed. Both turn paths (non-streaming + streaming) thread the outcome through; the session's own `except` branches are definitive `ok=False`; the streaming ACC `tool_result` emit prefers the verdict over its old text heuristic.
3. **Scratchpad migrated end-to-end** (the ENG-836 tool):
   - exec: `ok` comes from the runtime's own cell verdict (`cell.error`), `reason` is the traceback's **last line** (the cause). stderr-only warnings are not failures; stdout containing "failed" is not a failure.
   - ENG-350's empty-code rejection: counted by `ok=False`, the wording no longer matters (comment updated; the "must contain *failed*" workaround is dead).
   - single-scratchpad challenge: explicitly `ok=True` — guidance, not a failure (previously depended on its wording happening to avoid the markers).
   - view/dump: `ok=True` — rendering a notebook whose *old* cells contain `[error]` is a success; under substring matching a successful view of a failed cell counted as a fresh tool failure.

## Done-when → test map (`tests/test_tool_outcome_tracking.py`, 15 tests)

| Done-when | test |
|---|---|
| classification from the tool's own outcome | `TestExplicitOutcomeClassification` + registry envelope tests |
| success containing "failed" doesn't increment | `test_success_containing_failed_does_not_increment` |
| genuine failure lacking all five markers increments | `test_failure_without_any_marker_increments`, `test_five_explicit_failures_trip_the_breaker` |
| ENG-350 message counted without the word "failed" + still trips breaker | `test_empty_code_rejection_is_explicit_failure`, `test_empty_code_rejection_trips_breaker_without_wording` |
| fallback discoverability | `test_fallback_classification_is_logged` |

Also: `test_interleaved_false_success_no_longer_resets` pins the exact ENG-836 shape.

## Compatibility

- Unmigrated handlers (web tools, datasource, skills, …) keep byte-identical behavior via the `ok=None` fallback; the log line makes them findable for organic migration.
- Existing tests that call `handle_scratchpad`/`dispatch_tool` directly unwrap via a small shim; assertions unchanged.
- Full suite: 1458 passed locally (the 2 `test_chat_scratchpad` failures are a pre-existing local-sandbox quirk in this checkout — they fail identically on clean `origin/staging` and pass in CI).

## Security pass

No new inputs parsed, no secrets touched. `ToolOutcome.reason` carries at most the last line of a traceback (≤160 chars) and stays in-process — it is not logged and not sent to the model beyond what the result text already contained. The new fallback log line contains only the tool name, never result content.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2141,6 +2141,12 @@ class ChatSession:
                         else b
                         for b in result
                     ]
+                    # NOTE for whoever migrates the first multimodal handler to declare
+                    # ok=False: the streak climbs here, but the resilience-nudge and
+                    # circuit-breaker TEXT are only appended by _apply_error_tracking
+                    # on the string path — this branch skips it, so the model is never
+                    # told to stop retrying. When that handler exists, append the nudge
+                    # as an extra {"type": "text"} block here (#308 review).
                     if outcome.ok is False:
                         error_streak[tc.name] = error_streak.get(tc.name, 0) + 1
                     else:
@@ -2930,6 +2936,12 @@ class ChatSession:
                             else b
                             for b in result_text
                         ]
+                        # NOTE for whoever migrates the first multimodal handler to declare
+                        # ok=False: the streak climbs here, but the resilience-nudge and
+                        # circuit-breaker TEXT are only appended by _apply_error_tracking
+                        # on the string path — this branch skips it, so the model is never
+                        # told to stop retrying. When that handler exists, append the nudge
+                        # as an extra {"type": "text"} block here (#308 review).
                         if tool_ok is False:
                             error_streak[tc.name] = error_streak.get(tc.name, 0) + 1
                         else:
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2131,18 +2131,21 @@ class ChatSession:
                 if isinstance(result, list):
                     # Multimodal tool result — scrub credentials from text
                     # blocks; image-block payloads are raw bytes and have
-                    # nothing to scrub. A list result signals success, so
-                    # mirror the success branch of `_apply_error_tracking`
-                    # and reset the streak instead of running the full
-                    # string-only nudge logic.
+                    # nothing to scrub. A list result signals success unless
+                    # the handler explicitly said otherwise (ENG-1276), so
+                    # mirror the matching branch of `_apply_error_tracking`
+                    # instead of running the full string-only nudge logic.
                     content: "str | list[dict]" = [
                         {**b, "text": scrub_credentials(b.get("text", ""))}
                         if b.get("type") == "text"
                         else b
                         for b in result
                     ]
-                    error_streak[tc.name] = 0
-                    resilience_nudged.discard(tc.name)
+                    if outcome.ok is False:
+                        error_streak[tc.name] = error_streak.get(tc.name, 0) + 1
+                    else:
+                        error_streak[tc.name] = 0
+                        resilience_nudged.discard(tc.name)
                 else:
                     result = scrub_credentials(result)
                     result = self._apply_error_tracking(
@@ -2917,18 +2920,21 @@ class ChatSession:
                     if isinstance(result_text, list):
                         # Multimodal tool result — scrub credentials from text
                         # blocks (image payloads carry no secrets). A list
-                        # result signals success, so mirror the success
-                        # branch of `_apply_error_tracking` and reset the
-                        # streak instead of running the full string-only
-                        # nudge logic.
+                        # result signals success unless the handler explicitly
+                        # said otherwise (ENG-1276), so mirror the matching
+                        # branch of `_apply_error_tracking` instead of running
+                        # the full string-only nudge logic.
                         scrubbed_blocks = [
                             {**b, "text": scrub_credentials(b.get("text", ""))}
                             if b.get("type") == "text"
                             else b
                             for b in result_text
                         ]
-                        error_streak[tc.name] = 0
-                        resilience_nudged.discard(tc.name)
+                        if tool_ok is False:
+                            error_streak[tc.name] = error_streak.get(tc.name, 0) + 1
+                        else:
+                            error_streak[tc.name] = 0
+                            resilience_nudged.discard(tc.name)
                         if self._episodic is not None:
                             self._episodic.log_turn(
                                 self._turn_count + 1,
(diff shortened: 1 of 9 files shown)


alecantu7 (commented):
Self-review (adversarial pass) — findings and dispositions:

1. **CONFIRMED and fixed: the multimodal branch swallowed explicit failure verdicts.** Both turn paths hard-coded a list result as success (streak reset + ACC `success: True`), so a future multimodal handler returning `ToolOutcome(content=[...], ok=False)` would have been silently misread — the exact class of lie this PR exists to remove. Fixed in 29cb89c: the reset only happens when the outcome isn't an explicit failure, and the streaming ACC emit now reports the verdict.

2. **Cross-repo consumer sweep: clean.** Per the workspace-wide-grep rule, checked cowork-server, cowork, and cowork_evals (fresh `origin/staging`, not the possibly-stale local clones) for `dispatch_tool` / `handle_scratchpad` / `prepare_scratchpad_exec` / `_apply_error_tracking` consumers. Only hit: a cowork-server test *docstring* mentioning the names. Hosts drive anton through ChatSession stream events; the envelope change is invisible to them.

3. **Accepted: reset/remove/install/unknown-action scratchpad branches stay unmigrated** (plain str → `ok=None` fallback). `No scratchpad named 'x'` and `Unknown scratchpad action` are genuine failures the fallback still misses (no marker phrase → resets the streak) — the pre-existing false-negative, and note the fallback does NOT log these (it only logs when it classifies an error, and these misread as success). Left deliberately: each unmigrated site is one obvious `ToolOutcome` wrap away, and the fallback log line makes the *error-shaped* ones discoverable. Migrating every branch here would have doubled the diff for paths with no incident history.

4. **Accepted: `install_packages` still substring-classified — but now in exactly one place.** The handler's check of the runtime's message ("Install failed" / "timed out") predates ToolOutcome; the verdict it derives is explicit from that point outward. Restructuring the runtime's return shape is out of scope.

5. **Noted for ENG-1286: `reason` granularity.** Exception path uses the exception type name; scratchpad uses the traceback's last line (≤160 chars). Whether that's the right root-cause key granularity (too fine: the pyodbc→apt→pymssql wall looks like 3+ causes; too coarse: distinct bugs collide) is precisely ENG-1286's design problem — deliberately not pre-decided here.

Verified while attacking: the challenge outcome (`ok=True`) matches its previous de-facto behavior (its wording avoided every marker, so it always reset the streak — now that's structural, not lexical); `test_second_distinct_name_is_challenged`'s wording assertion still holds; the dump re-prefix path in streaming carries the outcome's `ok=True` correctly; no circular imports (registry has no runtime deps).

anton/core/session.py
```diff
@@ -2863,23 +2913,28 @@ async def _stream_and_handle_tools(
...
                         scrubbed_blocks = [
                             {**b, "text": scrub_credentials(b.get("text", ""))}
                             if b.get("type") == "text"
                             else b
                             for b in result_text
                         ]
-                        error_streak[tc.name] = 0
-                        resilience_nudged.discard(tc.name)
+                        if tool_ok is False:
+                            error_streak[tc.name] = error_streak.get(tc.name, 0) + 1
```
pnewsam, line 2946: **Minor (forward-looking):** a multimodal (`list`) result with `ok is False` increments the streak here, but the resilience nudge and circuit-breaker text are only produced by `_apply_error_tracking` on string results — which this branch skips via `continue`. So a migrated multimodal handler failing `_max_consecutive_errors` times would climb the counter without the model ever being told to stop retrying. No current handler returns a multimodal `ok=False`, so this is latent, and nudge text can't be appended to image blocks anyway. Worth a comment here so whoever migrates the first multimodal handler knows the breaker message won't reach the model on this path. Same shape in the non-streaming `turn` path.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff at head `29cb89c`: the `ToolOutcome` envelope, both the non-streaming (`turn`) and streaming (`_stream_and_handle_tools`) verdict-threading paths, the scratchpad migration, and the 15 new tests. The design is sound — the handler's own verdict replaces text matching, unmigrated handlers keep byte-identical behavior via the `ok=None` fallback, and every `dispatch_tool` caller (session + four test files) is updated for the envelope return. No merge-blocking issues; one minor forward-looking gap inline.

### Validation

- Confirmed `dispatch_tool` has no callers outside `session.py` and the migrated tests — the return-type change (`str|list` → `ToolOutcome`) is fully absorbed.
- Verified `Cell.error` is a field distinct from `stderr`, so `tool_ok = not (cell.error or "").strip()` correctly treats stderr-only warnings as success.
- Traced the streaming multimodal branch — it ends in `continue`, so the string-only nudge logic below is correctly skipped.
- ACC severity change is a no-op for unmigrated handlers (`ok=None` → `tool_ok is not False` → success, severity 1); only `ok=False` changes it.
- CI: all checks pass, including `run-tests`, CodeQL, and Snyk.

anton/core/session.py
alecantu7: Good catch — annotated at both sites (streaming + non-streaming) in the follow-up commit, including the concrete fix for whoever migrates the first multimodal handler: the nudge/breaker text *can* reach the model on this path as an extra `{"type": "text"}` block appended to the content list, it just isn't wired because no handler exercises it and the messages are tuned for string results. Deliberately a comment rather than the wiring itself — behavior added to a latent path with no handler to test against is exactly the speculative-generality shape this PR is trying to avoid.

PR #309, opened 2026-08-05
ENG-1288: per-turn token cost counter — log line + turn_completed analytics event

alecantu7:
# ENG-1288: per-turn token cost counter — anton finally knows what a turn spent

Nobody knows what a turn costs: the Kiranam session (~3.72M tokens, ~50 min) was discovered weeks later via manual Langfuse archaeology. This PR gives the runtime its own books — the primitive under the ENG-1286 spend ceiling, cost-per-task analytics, and runaway alerting.

[Linear ENG-1288](https://linear.app/mindsdb/issue/ENG-1288) — the full measurement contract (turn = measured unit; round = read-point + shape stats; conversation = derived; task v1 ≔ conversation; two-ledger doctrine) lives on the ticket.

## What's in here

1. **`Usage` gains cache components, normalized across providers.** `cache_read_tokens` / `cache_creation_tokens` + a `context_tokens` property (cache tokens ARE context — a warm-cache 190k call is not a 2k call). Anthropic passes its native fields through (its `input_tokens` already excludes cache activity); OpenAI-shaped usage **subtracts `cached_tokens` out of `prompt_tokens`** so `input` means fresh tokens on every provider (`_split_cached_input`, junk-tolerant for gateway quirks/test doubles). `context_pressure` keeps its exact old value on the OpenAI path (fresh + cached) — no compaction behavior change.

2. **`LLMClient.usage_listener`** — one narrow waist. `plan` / `plan_stream` (planning), `code` + `_generate_object_with` (coding — the completion verifier's verdict calls, **counted before the no-tool-call raise**: a failed structured call still costs money, ENG-1081's shape), `summarize` / `gate` (router). A listener that raises is swallowed with a warning — accounting must never kill the turn it's counting.

3. **`TurnCost`** (`anton/core/turn_cost.py`) — four raw components + `llm_calls` + shape stats (`rounds`, `peak_context_tokens`, `continuations`) + `ended_by` + duration. `total_tokens` is ENG-1286's mid-turn read surface.

   **Per-role attribution** (added 2026-08-06 on request): `<role>_model` / `<role>_tokens` / `<role>_calls` for the closed role set `planning` / `coding` / `router`. Answers two questions the blended total can't:
   - *"which model was the user on"* — `planning_model` reports the model that **actually ran** the conversational loop, falling back to the configured name only if the role never ran.
   - *correct dollar math* — a turn routinely mixes an expensive planning model with the cheap coding model the verifier runs on, so one token count can't be priced at any single rate; dollars need `(model, tokens)` pairs. Router cost is recorded here for the first time (it appeared in no sink before).

   Role slices are asserted to reconcile with the turn total; two models in one role join with `|` so ambiguous per-role pricing is visible rather than one model being silently dropped. **Limit:** this is the alias anton *requested* — the gateway can resolve or fail over server-side and doesn't report that back (`LLMResponse` carries no model field), so Langfuse stays truth for what served a call.

4. **Session wiring** — books open at turn start (both `turn_stream` and CLI `turn()`), terminal sites mark `ended_by` (`round_cap`, `handback_stuck`, `handback_budget`, `handback_verifier_failure`), and `_emit_turn_cost` closes the books in the turn's `finally`: in-flight exception → `error`, cancel/GeneratorExit → `cancelled`, explicit marks survive clean exits, else `completed`. Emission disarms the listener so post-turn background work (cerebellum flush, identity extraction) can't leak into the next turn's books.

5. **Two reporting sinks** — a structured `turn_cost` log line, and a `turn_completed` analytics event through the existing `send_event` channel (same settings-resolution pattern as the `ds_connect_*` events; `analytics_enabled` opt-out and CI-drop apply). Event carries **numbers, names, and IDs only** — components, shape, `ended_by`, models, provider, harness, `anton_version`, and `conversation_id` + `turn_index` **join** keys that address the turn's Langfuse trace. (A join key, *not* a unique event key: `_turn_count` advances only after a turn completes, so an abandoned turn and its retry legitimately share an index — and the Langfuse trace context derives its `turn_id` from the same expression, so both sides agree it's two attempts at one conversational turn.)

6. **Test-suite guard**: `conftest.py` sets `ANTON_ANALYTICS_ENABLED=false` suite-wide — full-turn stub tests now reach the emission path, and no test run may ever fire a real collector event.

## Deviations from the ticket (will comment there)

- **The analytics channel is anton's anonymous collector (aid-keyed), not an identified PostHog client** — anton has no PostHog surface and its analytics module is anonymous by deliberate design. User-level joins go event → `conversation_id`/`turn_index` → Langfuse trace → user identity. The ticket's "Keycloak sub on the event" line was wrong at this layer.
- ~~**`cache_creation_tokens` = 0 on OpenAI-compatible surfaces** (incl. the gateway) — the wire format exposes only the read side.~~ **Retracted 2026-08-05 — this was factually wrong.** `mindshub_inference@origin/main` publishes `prompt_tokens_details.{cached_tokens, cache_write_tokens}` with `prompt_tokens = fresh + reads + writes`; the split now reads both buckets (review finding 3, fixed in `c0d5a30`). Third-party endpoints that omit the write field report 0 for it.
- **`origin` (interactive/channel/scheduled) not emitted** — not derivable inside anton today; `harness` is. Fast follow if hosts start passing it.

## Tests

14 new in `test_turn_cost.py`: accumulator math incl. cache-blind-peak guard; listener wiring per role; failed-verifier-call counting; listener-exception isolation; `ended_by` resolution per terminal (error / cancelled / explicit marks / completed); books close + double-emit no-op; bare-`CoreSettings` fallback; fresh-books-per-turn on a long-lived session (distinct instances, non-cumulative totals). Full suite: 1472 passed (+e2e 39) — the 2 `test_chat_scratchpad` failures are the known pre-existing local-sandbox quirk, failing identically on clean staging.

## Security pass

No content ever leaves the process: the event is integers, model/provider names, harness, version, and opaque IDs; the log line is the same. Analytics respects the existing opt-out + CI drop; emission is fire-and-forget and exception-isolated both directions (listener → call, emission → turn). No new inputs parsed, no secrets touched.

### DIFF
--- a/anton/core/llm/client.py
+++ b/anton/core/llm/client.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import AsyncIterator
 from typing import TYPE_CHECKING
 
-from .provider import LLMProvider, LLMResponse, StreamEvent
+from .provider import LLMProvider, LLMResponse, StreamComplete, StreamEvent
 
 if TYPE_CHECKING:
     from anton.config.settings import AntonSettings
@@ -56,6 +57,26 @@ class LLMClient:
         self._router_provider = router_provider or coding_provider
         self._router_model = router_model or coding_model
         self._max_tokens = max_tokens
+        # ENG-1288: optional per-call usage observer. Every LLM call this
+        # client makes — plan/plan_stream (planning), code + structured
+        # coding calls like the completion verifier (coding), summarize/gate
+        # (router) — reports (role, model, usage) here. The session installs
+        # its per-turn cost accumulator; None means nobody is counting.
+        # Accounting must never break a call: notification is wrapped and
+        # swallowed (see _notify_usage).
+        self.usage_listener = None  # Callable[[str, str, Usage], None] | None
+
+    def _notify_usage(self, role: str, model: str, usage) -> None:
+        listener = self.usage_listener
+        if listener is None:
+            return
+        try:
+            listener(role, model, usage)
+        except Exception:
+            # A broken accumulator must never kill the turn it's counting.
+            logging.getLogger(__name__).warning(
+                "usage_listener raised — turn cost undercounted", exc_info=True
+            )
 
     async def plan(
         self,
@@ -66,7 +87,7 @@ class LLMClient:
         max_tokens: int | None = None,
         native_web_tools: set[str] | None = None,
     ) -> LLMResponse:
-        return await self._planning_provider.complete(
+        response = await self._planning_provider.complete(
             model=self._planning_model,
             system=system,
             messages=messages,
@@ -74,6 +95,8 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
             native_web_tools=native_web_tools,
         )
+        self._notify_usage("planning", self._planning_model, response.usage)
+        return response
 
     async def plan_stream(
         self,
@@ -92,6 +115,10 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
             native_web_tools=native_web_tools,
         ):
+            if isinstance(event, StreamComplete):
+                self._notify_usage(
+                    "planning", self._planning_model, event.response.usage
+                )
             yield event
 
     @property
@@ -109,6 +136,11 @@ class LLMClient:
         trusted at the cap (ENG-1082).
         """
         return self._max_tokens
+
+    @property
+    def planning_model(self) -> str:
+        """The model name used for planning / the user-facing turn loop."""
+        return self._planning_model
 
     @property
     def coding_provider(self) -> LLMProvider:
@@ -139,7 +171,7 @@ class LLMClient:
         max_tokens: int | None = None,
         native_web_tools: set[str] | None = None,
     ) -> LLMResponse:
-        return await self._coding_provider.complete(
+        response = await self._coding_provider.complete(
             model=self._coding_model,
             system=system,
             messages=messages,
@@ -147,6 +179,8 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
             native_web_tools=native_web_tools,
         )
+        self._notify_usage("coding", self._coding_model, response.usage)
+        return response
 
     async def summarize(
         self,
@@ -161,12 +195,14 @@ class LLMClient:
         (the router_* kwargs default to the coding role in __init__), so
         this is behavior-preserving unless a distinct model is selected.
         """
-        return await self._router_provider.complete(
+        response = await self._router_provider.complete(
             model=self._router_model,
             system=system,
             messages=messages,
             max_tokens=max_tokens or self._max_tokens,
         )
+        self._notify_usage("router", self._router_model, response.usage)
+        return response
 
     async def gate(
         self,
@@ -182,7 +218,7 @@ class LLMClient:
         No ``native_web_tools``: the thalamus must never do work itself,
         only answer from context or delegate.
         """
-        return await self._router_provider.complete(
+        response = await self._router_provider.complete(
             model=self._router_model,
             system=system,
             messages=messages,
@@ -190,6 +226,8 @@ class LLMClient:
             tool_choice=tool_choice,
             max_tokens=max_tokens or self._max_tokens,
         )
+        self._notify_usage("router", self._router_model, response.usage)
+        return response
 
     async def _generate_object_with(
         self,
@@ -226,6 +264,15 @@ class LLMClient:
             tools=[tool],
             tool_choice={"type": "tool", "name": tool["name"]},
             max_tokens=budget,
+        )
+        # Count BEFORE the no-tool-call raise below: a structured call that
+        # failed (and its bigger-budget retry) still spent real tokens
+        # (ENG-1288). Role is coding for both generate_object variants today;
+        # derive it from the provider/model actually used.
+        self._notify_usage(
+            "planning" if model == self._planning_model else "coding",
+            model,
+            response.usage,
         )
 
         if not response.tool_calls:
--- a/anton/core/llm/client.py
+++ b/anton/core/llm/client.py
@@ -66,8 +66,18 @@ class LLMClient:
         # swallowed (see _notify_usage).
         self.usage_listener = None  # Callable[[str, str, Usage], None] | None
 
-    def _notify_usage(self, role: str, model: str, usage) -> None:
-        listener = self.usage_listener
+    def _notify_usage(self, role: str, model: str, usage, listener=None) -> None:
+        """Report one call's usage to the turn-cost accumulator.
+
+        ``listener`` must be the reference captured when the call was ISSUED,
+        not read here at completion time. End-of-turn fire-and-forget work
+        (cerebellum flush, identity update, scratchpad consolidation) shares
+        this client, so on a long-lived-session host a fast follow-up message
+        arms turn N+1's listener while turn N's flush is still in flight —
+        resolving late booked that usage into the wrong turn (#309 review).
+        Captured-at-issue means background calls started while disarmed book
+        nowhere, which is what ``turn_cost``'s docstring already promises.
+        """
         if listener is None:
             return
         try:
@@ -87,6 +97,7 @@ class LLMClient:
         max_tokens: int | None = None,
         native_web_tools: set[str] | None = None,
     ) -> LLMResponse:
+        listener = self.usage_listener
         response = await self._planning_provider.complete(
             model=self._planning_model,
             system=system,
@@ -95,7 +106,7 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
             native_web_tools=native_web_tools,
         )
-        self._notify_usage("planning", self._planning_model, response.usage)
+        self._notify_usage("planning", self._planning_model, response.usage, listener)
         return response
 
     async def plan_stream(
@@ -107,6 +118,7 @@ class LLMClient:
         max_tokens: int | None = None,
         native_web_tools: set[str] | None = None,
     ) -> AsyncIterator[StreamEvent]:
+        listener = self.usage_listener
         async for event in self._planning_provider.stream(
             model=self._planning_model,
             system=system,
@@ -117,7 +129,7 @@ class LLMClient:
         ):
             if isinstance(event, StreamComplete):
                 self._notify_usage(
-                    "planning", self._planning_model, event.response.usage
+                    "planning", self._planning_model, event.response.usage, listener
                 )
             yield event
 
@@ -171,6 +183,7 @@ class LLMClient:
         max_tokens: int | None = None,
         native_web_tools: set[str] | None = None,
     ) -> LLMResponse:
+        listener = self.usage_listener
         response = await self._coding_provider.complete(
             model=self._coding_model,
             system=system,
@@ -179,7 +192,7 @@ class LLMClient:
             max_tokens=max_tokens or self._max_tokens,
             native_web_tools=native_web_tools,
         )
-        self._notify_usage("coding", self._coding_model, response.usage)
+        self._notify_usage("coding", self._coding_model, response.usage, listener)
         return response
 
     async def summarize(
@@ -195,13 +208,14 @@ class LLMClient:
         (the router_* kwargs default to the coding role in __init__), so
         this is behavior-preserving unless a distinct model is selected.
         """
+        listener = self.usage_listener
         response = await self._router_provider.complete(
             model=self._router_model,
             system=system,
             messages=messages,
             max_tokens=max_tokens or self._max_tokens,
         )
-        self._notify_usage("router", self._router_model, response.usage)
+        self._notify_usage("router", self._router_model, response.usage, listener)
         return response
 
     async def gate(
@@ -218,6 +232,7 @@ class LLMClient:
         No ``native_web_tools``: the thalamus must never do work itself,
         only answer from context or delegate.
         """
+        listener = self.usage_listener
         response = await self._router_provider.complete(
             model=self._router_model,
             system=system,
@@ -226,7 +241,7 @@ class LLMClient:
             tool_choice=tool_choice,
             max_tokens=max_tokens or self._max_tokens,
         )
-        self._notify_usage("router", self._router_model, response.usage)
+        self._notify_usage("router", self._router_model, response.usage, listener)
         return response
 
     async def _generate_object_with(
@@ -257,6 +272,7 @@ class LLMClient:
 
         budget = max_tokens or self._max_tokens
 
+        listener = self.usage_listener
         response = await provider.complete(
             model=model,
             system=system,
@@ -273,6 +289,7 @@ class LLMClient:
             "planning" if model == self._planning_model else "coding",
             model,
             response.usage,
+            listener,
         )
 
         if not response.tool_calls:
--- a/anton/core/llm/client.py
+++ b/anton/core/llm/client.py
@@ -250,6 +250,7 @@ class LLMClient:
         *,
         provider: LLMProvider,
         model: str,
+        role: str,
         system: str,
         messages: list[dict],
         max_tokens: int | None,
@@ -283,14 +284,13 @@ class LLMClient:
         )
         # Count BEFORE the no-tool-call raise below: a structured call that
         # failed (and its bigger-budget retry) still spent real tokens
-        # (ENG-1288). Role is coding for both generate_object variants today;
-        # derive it from the provider/model actually used.
-        self._notify_usage(
-            "planning" if model == self._planning_model else "coding",
-            model,
-            response.usage,
-            listener,
-        )
+        # (ENG-1288). The role is PASSED, not inferred from model equality:
+        # deployments exist where planning and coding resolve to the same
+        # model (cowork-server's Gemini defaults are identical across all
+        # three roles), and inference collapsed every generate_object_code
+        # call — verifier verdicts, compaction, identity extraction — into
+        # `planning`, defeating the split this exists to provide (#309 review).
+        self._notify_usage(role, model, response.usage, listener)
 
         if not response.tool_calls:
             # Shared with the scratchpad's sync twin so both paths classify the
@@ -369,6 +369,7 @@ class LLMClient:
             schema_class,
             provider=self._planning_provider,
             model=self._planning_model,
+            role="planning",
             system=system,
             messages=messages,
             max_tokens=max_tokens,
@@ -395,6 +396,7 @@ class LLMClient:
             schema_class,
             provider=self._coding_provider,
             model=self._coding_model,
+            role="coding",
             system=system,
             messages=messages,
             max_tokens=max_tokens,
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -718,6 +718,38 @@ def build_chat_completion_kwargs(
     return kwargs
 
 
+def _split_cached_input(usage) -> tuple[int, int]:
+    """Return (fresh_input_tokens, cache_read_tokens) from an OpenAI-shaped usage.
+
+    OpenAI's prompt/input token count INCLUDES cached tokens, unlike
+    Anthropic's which excludes them — subtract the cached share out so
+    ``Usage`` components mean the same thing on every provider (ENG-1288).
+    Handles both naming schemes: chat.completions (``prompt_tokens`` +
+    ``prompt_tokens_details.cached_tokens``) and the Responses API
+    (``input_tokens`` + ``input_tokens_details.cached_tokens``). Cache
+    WRITES are not exposed on OpenAI-compatible surfaces at all — creation
+    stays 0 there, a stated ENG-1288 gap.
+    """
+    if usage is None:
+        return 0, 0
+
+    def _as_int(value) -> int:
+        # Gateways and test doubles put non-numeric junk in usage fields;
+        # anything that isn't a real number counts as 0, never a crash.
+        return value if isinstance(value, int) else 0
+
+    total = _as_int(getattr(usage, "prompt_tokens", None)) or _as_int(
+        getattr(usage, "input_tokens", None)
+    )
+    details = (
+        getattr(usage, "prompt_tokens_details", None)
+        or getattr(usage, "input_tokens_details", None)
+    )
+    cached = _as_int(getattr(details, "cached_tokens", None)) if details is not None else 0
+    cached = min(cached, total)  # defensive: never negative fresh input
+    return total - cached, cached
+
+
 class OpenAIProvider(LLMProvider):
     name: str = "openai"
 
@@ -974,14 +1006,17 @@ class OpenAIProvider(LLMProvider):
         )
 
         usage_obj = response.usage
-        input_tokens = usage_obj.prompt_tokens if usage_obj else 0
+        input_tokens, cache_read_tokens = _split_cached_input(usage_obj)
         return LLMResponse(
             content=content_text,
             tool_calls=tool_calls,
             usage=Usage(
                 input_tokens=input_tokens,
                 output_tokens=usage_obj.completion_tokens if usage_obj else 0,
-                context_pressure=compute_context_pressure(model, input_tokens),
+                context_pressure=compute_context_pressure(
+                    model, input_tokens + cache_read_tokens
+                ),
+                cache_read_tokens=cache_read_tokens,
             ),
             stop_reason=choice.finish_reason,
         )
@@ -1032,6 +1067,7 @@ class OpenAIProvider(LLMProvider):
         tool_calls: list[ToolCall] = []
         input_tokens = 0
         output_tokens = 0
+        cache_read_tokens = 0
         stop_reason: str | None = None
 
         # Track tool call deltas by index
@@ -1048,7 +1084,7 @@ class OpenAIProvider(LLMProvider):
             stream_started = True
             async for chunk in stream:
                 if chunk.usage:
-                    input_tokens = chunk.usage.prompt_tokens
+                    input_tokens, cache_read_tokens = _split_cached_input(chunk.usage)
                     output_tokens = chunk.usage.completion_tokens
 
                 if not chunk.choices:
@@ -1220,7 +1256,10 @@ class OpenAIProvider(LLMProvider):
                 usage=Usage(
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
-                    context_pressure=compute_context_pressure(model, input_tokens),
+                    context_pressure=compute_context_pressure(
+                        model, input_tokens + cache_read_tokens
+                    ),
+                    cache_read_tokens=cache_read_tokens,
                 ),
                 stop_reason=stop_reason,
             )
@@ -1339,6 +1378,7 @@ class OpenAIProvider(LLMProvider):
         tool_calls: list[ToolCall] = []
         input_tokens = 0
         output_tokens = 0
+        cache_read_tokens = 0
         stop_reason: str | None = None
 
         # Map output_index → in-flight function-call state. Responses API uses
@@ -1424,7 +1464,7 @@ class OpenAIProvider(LLMProvider):
                     if final_response is not None:
                         usage = getattr(final_response, "usage", None)
                         if usage is not None:
-                            input_tokens = getattr(usage, "input_tokens", 0) or 0
+                            input_tokens, cache_read_tokens = _split_cached_input(usage)
                             output_tokens = getattr(usage, "output_tokens", 0) or 0
                         stop_reason = getattr(final_response, "status", None)
         except openai.BadRequestError as exc:
@@ -1490,7 +1530,10 @@ class OpenAIProvider(LLMProvider):
                 usage=Usage(
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
-                    context_pressure=compute_context_pressure(model, input_tokens),
+                    context_pressure=compute_context_pressure(
+                        model, input_tokens + cache_read_tokens
+                    ),
+                    cache_read_tokens=cache_read_tokens,
                 ),
                 stop_reason=stop_reason,
             )
@@ -1531,7 +1574,7 @@ def _parse_response_object(response, model: str) -> LLMResponse:
     # `or 0` guards an explicit None (the attr is present but null) — the
     # Responses API returns usage.input_tokens=None on web-search responses,
     # which a bare getattr default does NOT catch. Mirrors the streaming path.
-    input_tokens = (getattr(usage, "input_tokens", 0) or 0) if usage else 0
+    input_tokens, cache_read_tokens = _split_cached_input(usage)
     output_tokens = (getattr(usage, "output_tokens", 0) or 0) if usage else 0
 
     status = getattr(response, "status", None)
@@ -1545,7 +1588,10 @@ def _parse_response_object(response, model: str) -> LLMResponse:
         usage=Usage(
             input_tokens=input_tokens,
             output_tokens=output_tokens,
-            context_pressure=compute_context_pressure(model, input_tokens),
+            context_pressure=compute_context_pressure(
+                model, input_tokens + cache_read_tokens
+            ),
+            cache_read_tokens=cache_read_tokens,
         ),
         stop_reason=status,
     )
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -718,20 +718,29 @@ def build_chat_completion_kwargs(
     return kwargs
 
 
-def _split_cached_input(usage) -> tuple[int, int]:
-    """Return (fresh_input_tokens, cache_read_tokens) from an OpenAI-shaped usage.
+def _split_cached_input(usage) -> tuple[int, int, int]:
+    """Return (fresh_input, cache_read, cache_write) from an OpenAI-shaped usage.
 
-    OpenAI's prompt/input token count INCLUDES cached tokens, unlike
-    Anthropic's which excludes them — subtract the cached share out so
+    OpenAI-dialect prompt counts are INCLUSIVE of cache activity, unlike
+    Anthropic's which excludes it — subtract both cached buckets out so
     ``Usage`` components mean the same thing on every provider (ENG-1288).
+
+    Both cache buckets are real on our own gateway: mindshub_inference
+    publishes ``prompt_tokens_details.{cached_tokens, cache_write_tokens}``
+    and composes ``prompt_tokens = fresh + reads + writes``
+    (``minds/schemas/chat.py``, ``minds/inference/types.py``). Reading only
+    ``cached_tokens`` left the write share misfiled as fresh input and
+    reported ``cache_creation_tokens`` as a constant 0 on the majority
+    production surface — which breaks per-component dollar weighting (writes
+    bill at a premium) and any "is caching working" query (#309 review).
+
     Handles both naming schemes: chat.completions (``prompt_tokens`` +
-    ``prompt_tokens_details.cached_tokens``) and the Responses API
-    (``input_tokens`` + ``input_tokens_details.cached_tokens``). Cache
-    WRITES are not exposed on OpenAI-compatible surfaces at all — creation
-    stays 0 there, a stated ENG-1288 gap.
+    ``prompt_tokens_details``) and the Responses API (``input_tokens`` +
+    ``input_tokens_details``). A third-party endpoint that omits the write
+    field simply reports 0 for it.
     """
     if usage is None:
-        return 0, 0
+        return 0, 0, 0
 
     def _as_int(value) -> int:
         # Gateways and test doubles put non-numeric junk in usage fields;
@@ -745,9 +754,15 @@ def _split_cached_input(usage) -> tuple[int, int]:
         getattr(usage, "prompt_tokens_details", None)
         or getattr(usage, "input_tokens_details", None)
     )
-    cached = _as_int(getattr(details, "cached_tokens", None)) if details is not None else 0
-    cached = min(cached, total)  # defensive: never negative fresh input
-    return total - cached, cached
+    read = _as_int(getattr(details, "cached_tokens", None)) if details is not None else 0
+    write = (
+        _as_int(getattr(details, "cache_write_tokens", None)) if details is not None else 0
+    )
+    # Clamp jointly: a malformed payload must never yield negative fresh input.
+    if read + write > total:
+        read = min(read, total)
+        write = max(0, total - read)
+    return total - read - write, read, write
 
 
 class OpenAIProvider(LLMProvider):
@@ -1006,17 +1021,22 @@ class OpenAIProvider(LLMProvider):
         )
 
         usage_obj = response.usage
-        input_tokens, cache_read_tokens = _split_cached_input(usage_obj)
+        input_tokens, cache_read_tokens, cache_creation_tokens = _split_cached_input(
+            usage_obj
+        )
         return LLMResponse(
             content=content_text,
             tool_calls=tool_calls,
             usage=Usage(
                 input_tokens=input_tokens,
                 output_tokens=usage_obj.completion_tokens if usage_obj else 0,
+                # All prompt-side tokens, exactly as before this split existed
+                # (the OpenAI dialect's prompt_tokens is cache-inclusive).
                 context_pressure=compute_context_pressure(
-                    model, input_tokens + cache_read_tokens
+                    model, input_tokens + cache_read_tokens + cache_creation_tokens
                 ),
                 cache_read_tokens=cache_read_tokens,
+                cache_creation_tokens=cache_creation_tokens,
             ),
             stop_reason=choice.finish_reason,
         )
@@ -1068,6 +1088,7 @@ class OpenAIProvider(LLMProvider):
         input_tokens = 0
         output_tokens = 0
         cache_read_tokens = 0
+        cache_creation_tokens = 0
         stop_reason: str | None = None
 
         # Track tool call deltas by index
@@ -1084,7 +1105,11 @@ class OpenAIProvider(LLMProvider):
             stream_started = True
             async for chunk in stream:
                 if chunk.usage:
-                    input_tokens, cache_read_tokens = _split_cached_input(chunk.usage)
+                    (
+                        input_tokens,
+                        cache_read_tokens,
+                        cache_creation_tokens,
+                    ) = _split_cached_input(chunk.usage)
                     output_tokens = chunk.usage.completion_tokens
 
                 if not chunk.choices:
@@ -1257,9 +1282,10 @@ class OpenAIProvider(LLMProvider):
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
                     context_pressure=compute_context_pressure(
-                        model, input_tokens + cache_read_tokens
+                        model, input_tokens + cache_read_tokens + cache_creation_tokens
                     ),
                     cache_read_tokens=cache_read_tokens,
+                    cache_creation_tokens=cache_creation_tokens,
                 ),
                 stop_reason=stop_reason,
             )
@@ -1379,6 +1405,7 @@ class OpenAIProvider(LLMProvider):
         input_tokens = 0
         output_tokens = 0
         cache_read_tokens = 0
+        cache_creation_tokens = 0
         stop_reason: str | None = None
 
         # Map output_index → in-flight function-call state. Responses API uses
@@ -1464,7 +1491,11 @@ class OpenAIProvider(LLMProvider):
                     if final_response is not None:
                         usage = getattr(final_response, "usage", None)
                         if usage is not None:
-                            input_tokens, cache_read_tokens = _split_cached_input(usage)
+                            (
+                                input_tokens,
+                                cache_read_tokens,
+                                cache_creation_tokens,
+                            ) = _split_cached_input(usage)
                             output_tokens = getattr(usage, "output_tokens", 0) or 0
                         stop_reason = getattr(final_response, "status", None)
         except openai.BadRequestError as exc:
@@ -1531,9 +1562,10 @@ class OpenAIProvider(LLMProvider):
                     input_tokens=input_tokens,
                     output_tokens=output_tokens,
                     context_pressure=compute_context_pressure(
-                        model, input_tokens + cache_read_tokens
+                        model, input_tokens + cache_read_tokens + cache_creation_tokens
                     ),
                     cache_read_tokens=cache_read_tokens,
+                    cache_creation_tokens=cache_creation_tokens,
                 ),
                 stop_reason=stop_reason,
             )
@@ -1574,7 +1606,7 @@ def _parse_response_object(response, model: str) -> LLMResponse:
     # `or 0` guards an explicit None (the attr is present but null) — the
     # Responses API returns usage.input_tokens=None on web-search responses,
     # which a bare getattr default does NOT catch. Mirrors the streaming path.
-    input_tokens, cache_read_tokens = _split_cached_input(usage)
+    input_tokens, cache_read_tokens, cache_creation_tokens = _split_cached_input(usage)
     output_tokens = (getattr(usage, "output_tokens", 0) or 0) if usage else 0
 
     status = getattr(response, "status", None)
@@ -1589,9 +1621,10 @@ def _parse_response_object(response, model: str) -> LLMResponse:
             input_tokens=input_tokens,
             output_tokens=output_tokens,
             context_pressure=compute_context_pressure(
-                model, input_tokens + cache_read_tokens
+                model, input_tokens + cache_read_tokens + cache_creation_tokens
             ),
             cache_read_tokens=cache_read_tokens,
+            cache_creation_tokens=cache_creation_tokens,
         ),
         stop_reason=status,
     )
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1778,7 +1778,19 @@ class ChatSession:
         if tc is None or tc.emitted:
             return
         tc.emitted = True  # the double-emit guard, now on the books themselves
-        if self._turn_cost is tc:
+        is_owner = self._turn_cost is tc
+        if tc.ended_monotonic is None:
+            # The owning turn is ending right now; a late finalizer is not — so
+            # it reports up to the last LLM call rather than up to whenever
+            # asyncio ran it (#309 review follow-up).
+            import time as _t
+
+            tc.ended_monotonic = (
+                _t.monotonic()
+                if is_owner or tc.last_activity_monotonic is None
+                else tc.last_activity_monotonic
+            )
+        if is_owner:
             self._turn_cost = None
             try:
                 self._llm.usage_listener = None
@@ -1809,7 +1821,9 @@ class ChatSession:
         elif exc is not None:
             tc.ended_by = "error"
 
-        turn_index = self._turn_count + 1
+        # Stamped at books-open; the live lookup remains only for bare-session
+        # tests and the legacy non-streaming path.
+        turn_index = tc.turn_index or (self._turn_count + 1)
         logger.info(
             "turn_cost session=%s turn=%d ended_by=%s tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
@@ -2256,7 +2270,7 @@ class ChatSession:
         # emission happens at both returns; an exception propagating out of
         # this method skips emission (stated gap: the CLI path's error exits
         # go unreported rather than restructuring the method around it).
-        self._turn_cost = TurnCost()
+        self._turn_cost = TurnCost(turn_index=self._turn_count + 1)
         self._llm.usage_listener = self._turn_cost.add
 
         user_msg_str = (
@@ -2517,7 +2531,7 @@ class ChatSession:
         # Open the turn's cost books and listen at the LLM-client narrow
         # waist — planning, coding (incl. verifier verdicts), and router
         # calls all report here (ENG-1288).
-        self._turn_cost = TurnCost()
+        self._turn_cost = TurnCost(turn_index=self._turn_count + 1)
         _turn_cost_books = self._turn_cost
         self._llm.usage_listener = self._turn_cost.add
 
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -467,6 +467,22 @@ def _render_verify_transcript(
     return "\n".join(line for _, line in kept) or "(no conversation)"
 
 
+def _role_model(tc: TurnCost, role: str) -> str:
+    """Model that ran ``role`` this turn ("" if the role never ran)."""
+    slice_ = tc.by_role.get(role)
+    return slice_.model if slice_ else ""
+
+
+def _role_tokens(tc: TurnCost, role: str) -> str:
+    slice_ = tc.by_role.get(role)
+    return str(slice_.tokens if slice_ else 0)
+
+
+def _role_calls(tc: TurnCost, role: str) -> str:
+    slice_ = tc.by_role.get(role)
+    return str(slice_.calls if slice_ else 0)
+
+
 def _build_verify_request(
     history: list[dict], user_message: str | None
 ) -> tuple[str, list[dict]]:
@@ -1777,11 +1793,16 @@ class ChatSession:
         logger.info(
             "turn_cost session=%s turn=%d ended_by=%s tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
-            "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d",
+            "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d "
+            "by_role=%s",
             self._session_id, turn_index, tc.ended_by, tc.total_tokens,
             tc.input_tokens, tc.output_tokens, tc.cache_read_tokens,
             tc.cache_creation_tokens, tc.llm_calls, tc.rounds,
             tc.continuations, tc.peak_context_tokens, tc.duration_ms,
+            ",".join(
+                f"{role}={s.model}:{s.tokens}/{s.calls}"
+                for role, s in sorted(tc.by_role.items())
+            ) or "-",
         )
 
         # Analytics sink — same settings-resolution pattern as the
@@ -1813,8 +1834,25 @@ class ChatSession:
                 continuations=str(tc.continuations),
                 peak_context_tokens=str(tc.peak_context_tokens),
                 duration_ms=str(tc.duration_ms),
-                planning_model=str(self._llm.planning_model or ""),
-                coding_model=str(self._llm.coding_model or ""),
+                # Per-role attribution: which model actually ran each role and
+                # what it spent. `<role>_model` is the model the USER was on
+                # for the conversational loop (planning) — the configured names
+                # can drift from what ran, and dollars are only computable from
+                # (model, tokens) pairs since a turn mixes rates. Roles are a
+                # closed set, so these stay flat and queryable.
+                planning_model=_role_model(tc, "planning") or str(
+                    self._llm.planning_model or ""
+                ),
+                planning_tokens=_role_tokens(tc, "planning"),
+                planning_calls=_role_calls(tc, "planning"),
+                coding_model=_role_model(tc, "coding") or str(
+                    self._llm.coding_model or ""
+                ),
+                coding_tokens=_role_tokens(tc, "coding"),
+                coding_calls=_role_calls(tc, "coding"),
+                router_model=_role_model(tc, "router"),
+                router_tokens=_role_tokens(tc, "router"),
+                router_calls=_role_calls(tc, "router"),
                 # Configured provider name (anthropic / openai /
                 # openai-compatible): separates gateway traffic from BYOK in
                 # queries. The finer per-model provider split is a follow-up.
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1726,7 +1726,7 @@ class ChatSession:
         self.hard_truncate_history()
         return await self._llm.plan(messages=factory_validated(), **kwargs)
 
-    def _emit_turn_cost(self) -> None:
+    def _emit_turn_cost(self, expected: TurnCost | None = None) -> None:
         """Close the turn's cost books: disarm the listener, resolve the
         terminal path, and emit the two reporting sinks (ENG-1288).
 
@@ -1745,6 +1745,13 @@ class ChatSession:
         tc = self._turn_cost
         if tc is None:
             return
+        if expected is not None and tc is not expected:
+            # A late async-generator finalizer for an ABANDONED turn: asyncio
+            # runs it in a fresh task long after the fact, by which point a
+            # newer turn may own the books. Closing those would emit the next
+            # turn's partial totals mid-turn and leave it uncounted. Only the
+            # turn that opened these books may close them (#309 review).
+            return
         self._turn_cost = None
         try:
             self._llm.usage_listener = None
@@ -1755,7 +1762,13 @@ class ChatSession:
         cancelled = bool(
             getattr(self, "_cancel_event", None) and self._cancel_event.is_set()
         )
-        if isinstance(exc, GeneratorExit) or cancelled:
+        # asyncio.CancelledError is the SHAPE USER STOP TAKES on the primary
+        # host: cowork-server's RunHandle.cancel() calls task.cancel(), which
+        # raises it inside the suspended generator frame — and nothing there
+        # sets `_cancel_event` (that's anton's CLI only). Without it every
+        # Stop filed as "error" (#309 review). GeneratorExit covers the
+        # narrower case where the cancel lands exactly at a yield boundary.
+        if isinstance(exc, (GeneratorExit, asyncio.CancelledError)) or cancelled:
             tc.ended_by = "cancelled"
         elif exc is not None:
             tc.ended_by = "error"
@@ -2232,9 +2245,14 @@ class ChatSession:
 
         while response.tool_calls:
             tool_round += 1
-            if self._turn_cost is not None:
-                self._turn_cost.rounds = tool_round
+            if self._turn_cost is not None and tool_round <= self._max_tool_rounds:
+                self._turn_cost.rounds += 1
             if tool_round > self._max_tool_rounds:
+                # Mirror turn_stream's cap mark (#309 review) — this path is
+                # public API even without in-repo callers, and the books are
+                # wired here too.
+                if self._turn_cost is not None:
+                    self._turn_cost.ended_by = "round_cap"
                 self._append_history(
                     {"role": "assistant", "content": response.content or ""}
                 )
@@ -2437,6 +2455,7 @@ class ChatSession:
         # waist — planning, coding (incl. verifier verdicts), and router
         # calls all report here (ENG-1288).
         self._turn_cost = TurnCost()
+        _turn_cost_books = self._turn_cost
         self._llm.usage_listener = self._turn_cost.add
 
         try:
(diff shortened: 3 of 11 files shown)


alecantu7 (commented):
Self-review (adversarial pass):

1. **Verified empirically, not assumed: `sys.exc_info()` inside a generator's `finally`.** The whole `ended_by` resolution for `cancelled`/`error` hinges on the finally seeing the in-flight exception during propagation and on `GeneratorExit` being visible during `.close()`. Probed both on this interpreter: GeneratorExit and propagating exceptions are both visible. The unit tests cover the except-context path; the close path is covered by the probe and the isinstance branch.

2. **Pre-existing quirk inherited, not fixed: CLI `turn()` never increments `_turn_count`** (the only increment site is turn_stream's tail), so `turn_index` on the non-streaming path reports 1 forever. Every production surface (CLI included) drives `turn_stream`, where indexing is correct — noted rather than fixed because touching turn-count semantics on the legacy path is out of scope for a counter PR.

3. **Emission ordering vs post-turn work is deliberate**: `_emit_turn_cost` runs in the finally, BEFORE `_schedule_cerebellum_flush`/identity extraction fire — and it disarms the listener, so background LLM usage lands in nobody's books instead of the next turn's. That's the stated-gap trade: slightly undercounted totals, never mis-attributed ones.

4. **Counting failed verifier calls happens above the raise** in `_generate_object_with` — a truncated verdict + its 4096 retry counts as two calls, which is what the gateway bills. Covered by `test_failed_structured_call_still_counts`.

5. **Accepted gaps, stated in module docstring + ticket**: `cache_creation_tokens`=0 on OpenAI-compatible surfaces (wire format exposes no write side); `origin` (interactive/channel/scheduled) not derivable in anton today — `harness` ships instead; per-role token split deferred (the listener contract carries role/model, so adding it later is additive); CLI `turn()` error exits skip emission (no wrapping finally on that path — restructuring a 250-line legacy method for a reporting sink failed the cost/benefit test).

6. **The analytics event rides the anonymous channel by design** — reviewed the privacy posture explicitly: aid fingerprint (existing), opaque conversation/turn IDs, token integers, model/provider/harness/version names. No prose, no paths, no user identity. Identity joins go through the Langfuse trace, which already holds them under its own access model.

7. **Test-suite guard double-checked**: `ANTON_ANALYTICS_ENABLED=false` in conftest uses `setdefault`, so a developer deliberately testing analytics can override from their shell; send_event's CI-drop covers workflows independently.

anton/core/session.py
```diff
@@ -1720,6 +1726,99 @@ def factory_validated():
...
+        try:
+            self._llm.usage_listener = None
+        except Exception:
+            pass
+
+        exc = sys.exc_info()[1]
+        cancelled = bool(
+            getattr(self, "_cancel_event", None) and self._cancel_event.is_set()
+        )
+        if isinstance(exc, GeneratorExit) or cancelled:
```
alecantu7: **Self-review · HIGH — user Stop on cowork-server lands in the `error` branch, never `cancelled`** (confirmed by repro)

cowork-server's Stop path is `RunHandle.cancel()` → `self.task.cancel()` (`cowork/streaming/registry.py:63`), which raises `asyncio.CancelledError` inside the suspended generator frame. This classifier recognizes only `GeneratorExit` or `_cancel_event` — and nothing in cowork-server ever sets `_cancel_event` (grep of origin/staging: zero hits; only anton's CLI `chat.py:1865` / `goal.py:116` set it). So every user Stop on the primary production host emits `ended_by="error"`.

Repro against the real `turn_stream` (task.cancel() mid-LLM-await): `ended_by = error`, `cancel_event was set? False`. The `GeneratorExit` path only fires when the cancel lands exactly at a yield boundary — not the common case.

**Fix:** `if isinstance(exc, (GeneratorExit, asyncio.CancelledError)) or cancelled:` — safe, since `asyncio.wait_for` converts its internal cancels to `TimeoutError` before they'd propagate here.
**Test:** cancel a task consuming `turn_stream`; assert `send_event` kwargs have `ended_by == "cancelled"`. Fails on HEAD.

anton/core/session.py
```diff
@@ -2499,6 +2616,7 @@ async def turn_stream(
                     "".join(assistant_text_parts)[:2000]
                 )
             reset_trace_context(_trace_token)
+            self._emit_turn_cost()
```
alecantu7: **Self-review · MEDIUM — CLI ESC-cancel emits nothing: deferred finalizer dies before `_emit_turn_cost`** (confirmed by repro)

ESC in the CLI (`chat.py:1864-1866`) abandons the generator without `aclose()`. asyncio's async-gen finalizer later runs this `finally` in a *new task with a copied context*, and the pre-existing `reset_trace_context(_trace_token)` on the line above raises `ValueError: Token ... created in a different Context` — aborting the `finally` before this emit line runs. Result per ESC: no `turn_cost` log, no event, plus a "Task exception was never retrieved" in the log. Repro: abandoned mid-iteration → 0 events + the ValueError; control with same-task `aclose()` works fine.

Second hazard: `_emit_turn_cost` reads the session-shared `self._turn_cost`, so a late finalizer that *did* survive would close the **next** turn's books mid-turn.

**Fix:** (1) make `self._emit_turn_cost()` the first statement of the `finally` (and/or wrap `reset_trace_context` in try/except ValueError); (2) capture `tc = self._turn_cost` at books-open and pass `_emit_turn_cost(expected=tc)`, no-op when a newer turn owns the books.
**Test:** break out of the async-for without aclose, drive the loop briefly, assert exactly one `send_event` with `ended_by == "cancelled"`. Fails on HEAD (never called).

anton/core/llm/openai.py
```diff
@@ -718,6 +718,38 @@ def build_chat_completion_kwargs(
...
+def _split_cached_input(usage) -> tuple[int, int]:
+    """Return (fresh_input_tokens, cache_read_tokens) from an OpenAI-shaped usage.
+
+    OpenAI's prompt/input token count INCLUDES cached tokens, unlike
+    Anthropic's which excludes them — subtract the cached share out so
+    ``Usage`` components mean the same thing on every provider (ENG-1288).
+    Handles both naming schemes: chat.completions (``prompt_tokens`` +
+    ``prompt_tokens_details.cached_tokens``) and the Responses API
+    (``input_tokens`` + ``input_tokens_details.cached_tokens``). Cache
+    WRITES are not exposed on OpenAI-compatible surfaces at all — creation
```
alecantu7: **Self-review · MEDIUM — gateway cache-*write* tokens are misfiled as fresh input; the PR's "wire format doesn't expose it" deviation note is factually wrong** (confirmed against gateway source)

mindshub_inference origin/main *does* expose writes on the OpenAI-compatible surface: `PromptTokensDetails{cached_tokens, cache_write_tokens}` (`minds/schemas/chat.py`), with `prompt_tokens = fresh + reads + writes` (`minds/inference/types.py:262-264`; `providers/anthropic.py` populates `cache_write_tokens` from `cache_creation_*`). `_split_cached_input` subtracts only `cached_tokens`, so the write share stays inside `input` and `cache_creation_tokens` reads a constant 0 on the majority production surface.

Executed: `{prompt_tokens:100, prompt_tokens_details:{cached_tokens:30, cache_write_tokens:10}}` parsed through the real openai SDK (extra field survives `model_validate`) → anton reports input=70, cache_read=30, creation=0; truth is 60/30/10. Consequence: per-component dollar weighting (writes bill 1.25×) and any "is caching working" query over the analytics are silently wrong, and PR-body deviation #2 + the ticket comment need correcting.

**Fix:** also read `getattr(details, 'cache_write_tokens', None)`, subtract both (clamped), pass `cache_creation_tokens` through at the four `Usage` construction sites (1013/1256/1530/1588); fix the docstrings here and in `provider.py`. `context_pressure` stays exact since it sums input+read+creation.
**Test:** usage double with the payload above → assert `(60, 30, 10)` split and unchanged `context_pressure`.

anton/core/session.py
```diff
@@ -2499,6 +2616,7 @@ async def turn_stream(
                     "".join(assistant_text_parts)[:2000]
```
alecantu7, line 2737: **Self-review · MEDIUM — failed turns report `ended_by="completed"`** (confirmed by repro; real site is the exhausted-retries branch, ~2575-2609)

When every LLM call raises a generic (non-transient) error, the retry loop exhausts `_max_auto_retries`, yields the "An unexpected error occurred…" apology, and `break`s. No exception is in flight when this `finally` runs, so `sys.exc_info()` is None and the `TurnCost` default `completed` holds. Repro with all calls raising: event said `ended_by: completed, rounds: 0, llm_calls: 0` — an error apology as the turn's entire output, filed as a clean completion. Error-rate monitoring on this event systematically undercounts the most common failure mode (absorbed mid-turn provider errors).

**Fix:** in the exhausted-retries branch, `if self._turn_cost is not None: self._turn_cost.ended_by = "retry_exhausted"` (mirroring the round_cap/handback_* marks). The finally's error/cancelled overrides still win when an exception genuinely propagates.
**Test:** patch `_stream_and_handle_tools` and `plan_stream` to raise, run `turn_stream` to exhaustion, assert `ended_by != "completed"`. Fails on HEAD.

anton/core/session.py
```diff
@@ -2711,8 +2829,12 @@ async def _stream_and_handle_tools(
 
             while llm_response.tool_calls:
                 tool_round += 1
+                if self._turn_cost is not None:
+                    self._turn_cost.rounds = tool_round
```
alecantu7: **Self-review · MEDIUM — `rounds` is overwritten per verifier continuation; reports only the last iteration** (confirmed by trace; both writers enumerated)

`self._turn_cost.rounds = tool_round` assigns the loop-local counter, which resets to 0 on every verifier-forced continuation (`continuation += 1` → back to the top of the verification loop → `tool_round = 0`). A 20-round turn with a 3-round continuation emits `rounds=3`. Also, the cap fires at `tool_round == max+1` *after* this assignment, so capped turns report `cap+1` (26 for a 25-cap). `llm_calls` in the same books accumulates correctly across continuations, confirming the per-turn intent — `rounds` is the odd one out, and it breaks the metric for exactly the expensive-continuation shape it exists to classify (the ENG-578 thrash split quoted in `turn_cost.py`'s own comment).

**Fix:** increment (`self._turn_cost.rounds += 1`) instead of assigning, here and at the `turn()` writer (~2236).
**Test:** 3 tool rounds → INCOMPLETE verdict → 2 rounds → COMPLETE; assert emitted `rounds == 5` (currently 2).

anton/core/session.py
```diff
@@ -1720,6 +1726,99 @@ def factory_validated():
...
+                # openai-compatible): separates gateway traffic from BYOK in
+                # queries. The finer per-model provider split is a follow-up.
+                llm_provider=str(getattr(settings, "planning_provider", "") or ""),
+                harness=str(self._harness or ""),
+                anton_version=_anton_version,
+                # Join keys: the same session/turn identity the MindsHub
+                # trace headers carry, so an analytics row links back to
+                # its Langfuse trace (and through it, to the user) for
+                # forensics.
+                conversation_id=str(self._session_id or ""),
```
alecantu7, line 1907: **Self-review · MEDIUM — `turn_completed` ships a user-joinable `conversation_id` through the collector whose contract says "anonymous — no PII"** (confirmed wire-format probe; decision needed before merge)

This is the first anton analytics event carrying a per-user-joinable identifier — every pre-PR `send_event` caller (checked at base c77dd88) sends only action + hashed-MAC `aid` + engine/provider names, and `anton/analytics.py:4-6` promises exactly that. On cowork-server this value is `str(conversation.id)` — the same identity emitted as `Langfuse-Session-Id`, and Langfuse traces are tagged with user emails; the code comment on this line says so itself. It rides a GET query string (probe: `...?action=turn_completed&aid=...&conversation_id=...`) into an API-Gateway lambda, i.e. access logs.

It's the join key the design wants, so this may well be the right call — but it changes the collector's anonymity posture and must be decided visibly, not shipped as a side effect.

**Fix (pick one):** drop it (aid + turn_index + timestamp still correlate loosely); send a salted hash so joins require the salt; or keep it and update the `analytics.py` docstring + PR body + get explicit privacy sign-off.
**Test (if dropped/hashed):** assert the emitted kwargs carry no raw session id.

anton/core/llm/openai.py
```diff
@@ -718,6 +718,38 @@ def build_chat_completion_kwargs(
...
+    total = _as_int(getattr(usage, "prompt_tokens", None)) or _as_int(
+        getattr(usage, "input_tokens", None)
+    )
+    details = (
+        getattr(usage, "prompt_tokens_details", None)
+        or getattr(usage, "input_tokens_details", None)
+    )
+    cached = _as_int(getattr(details, "cached_tokens", None)) if details is not None else 0
+    cached = min(cached, total)  # defensive: never negative fresh input
+    return total - cached, cached
```
alecantu7: **Self-review · MEDIUM — the cache-subtraction invariant has zero test coverage: the no-subtract mutation survives all 1474 tests** (confirmed by mutation run)

Reverting `return total - cached, cached` → `return total, cached` passes the full suite (`1474 passed, 28 skipped`). No test in the repo sets `cached_tokens` / `prompt_tokens_details` (grep: nothing). Control: gutting the helper to `return 0, 0` fails 4 tests — so the untested part is specifically the cached-share semantics, i.e. both headline invariants of this PR ("input = fresh on all providers", "context_pressure keeps its exact old value") are unpinned. A merge-conflict resolution here would silently double-count cache reads into `TurnCost.total_tokens` (the ENG-1286 ceiling read) and inflate `context_pressure` → premature compaction on gateway traffic.

**Fix:** add the missing tests — chat shape and Responses-API shape doubles with `prompt_tokens=1000, cached_tokens=800` asserting `(200, 800)`, plus one provider-level test asserting `Usage.input_tokens==200`, `cache_read_tokens==800`, and `context_pressure == compute_context_pressure(model, 1000)` (the exact-old-value claim).

anton/core/llm/client.py
```diff
@@ -56,6 +57,26 @@ def __init__(
...
+        # client makes — plan/plan_stream (planning), code + structured
+        # coding calls like the completion verifier (coding), summarize/gate
+        # (router) — reports (role, model, usage) here. The session installs
+        # its per-turn cost accumulator; None means nobody is counting.
+        # Accounting must never break a call: notification is wrapped and
+        # swallowed (see _notify_usage).
+        self.usage_listener = None  # Callable[[str, str, Usage], None] | None
+
+    def _notify_usage(self, role: str, model: str, usage) -> None:
+        listener = self.usage_listener
```
alecantu7: **Self-review · MEDIUM — listener resolved at call-*completion* time: end-of-turn background LLM work books into the NEXT turn** (confirmed by repro, twice independently)

`_notify_usage` reads `self.usage_listener` when the provider call *returns*, not when it was issued. The end-of-turn fire-and-forget tasks — cerebellum flush (`session.py:2070`), identity update (2644), scratchpad consolidation (3592) — all route through this same shared client (`Cerebellum(cortex=..., llm=self._llm)`, `session.py:686-689`). On a long-lived-session host (CLI chat loop, dispatch local_runtime's cached sessions), a fast follow-up message arms turn N+1's listener while turn N's flush is still in flight → its usage lands in turn N+1's books. Repro: turn 1 emitted `llm_calls=0/tokens=0`, turn 2 emitted turn 1's flush (`llm_calls=1`). This contradicts `turn_cost.py`'s docstring ("post-turn background work … is not attributed to any turn") — the disarm-in-finally only protects the idle gap, not overlap. cowork-server escapes only because it builds a fresh session per stream.

**Fix:** snapshot the listener at call-*issue* time in each client method (`listener = self.usage_listener` before the `await`, notify the captured ref). Background calls issued while disarmed then book nowhere — which is what the docstring already promises.
**Test:** schedule `client.code()` with a provider that sleeps past the turn boundary, arm a fresh `TurnCost`, await the task, assert the new books show `llm_calls == 0` (currently 1).

anton/core/session.py
```diff
@@ -2124,6 +2232,8 @@ async def turn(self, user_input: str | list[dict]) -> str:
 
         while response.tool_calls:
             tool_round += 1
+            if self._turn_cost is not None:
+                self._turn_cost.rounds = tool_round
```
alecantu7: **Self-review · LOW — non-streaming `turn()` round-cap path never marks `ended_by="round_cap"`** (confirmed by trace + repro)

Only `turn_stream` marks the cap (2836-2837); this branch appends the pause message and breaks, so a capped non-streaming turn emits `ended_by="completed"`. Low because `turn()` has no in-repo callers — but it's public API, this PR deliberately wired books into it, and the PR body's "terminal sites mark ended_by (round_cap, …)" claim doesn't hold for this path.

**Fix:** mirror `turn_stream`: `if self._turn_cost is not None: self._turn_cost.ended_by = "round_cap"` inside this cap branch.
**Test:** drive `turn()` with a plan that always returns tool_calls and `_max_tool_rounds=1`; assert `ended_by == "round_cap"` (currently "completed").

alecantu7 (commented):
## Deep adversarial self-review (pre-human-review)

Ran the full adversarial pass: 5 independent finders (cancellation/partial-failure, cross-repo consumer sweep, accounting math, mutation-testing the new tests, data exposure) + one adversarial verifier per finding, everything execution-confirmed against the real `ChatSession`/`LLMClient` in a scratch worktree. 15 raw findings → 10 survived verification, 0 refuted. The 9 line-anchored comments below are the substantive ones; fixes + regression tests to follow as commits on this branch.

**One finding that can't be line-anchored** (file untouched by this PR):
- **LOW — CLI token toolbar silently understates warm-cache input ~10× for OpenAI BYOK** (`anton/chat.py:1887`): the toolbar still sums `.input_tokens`, whose meaning this PR changed from whole-prompt to fresh-only. For a warm-cache 100k call the displayed "in" drops 100000 → 10000 — exactly the understatement `provider.py:38-40`'s own docstring warns about. Fix: display `usage.context_tokens` (this PR's new property) instead, preserving the old meaning on OpenAI and improving it on Anthropic.

**What came back clean (executed, for the human reviewer's benefit):** `context_pressure` exact-old-value proven at all four construction sites old-vs-new; verifier-verdict usage confirmed counted before the no-tool-call raise; double-emit no-op; nothing in the emit `finally` can mask the turn's real exception; `send_event` fire-and-forget (3s-timeout daemon thread, never blocks the turn); `Usage` serialization/kwargs compat across repos; no user text on either sink; opt-out + CI-drop honored; `_split_cached_input` fuzzed with 14 junk shapes without crash or negative fresh-count; mutation runs confirmed the disarm, listener-exception-swallow, and cancelled-branch tests genuinely detect their mutations; cross-repo sweep (cowork-server, renderer, hermes, cowork_evals — all against origin, not local clones) found no consumer broken by the `input_tokens` semantics change besides the CLI toolbar above.

Severity distribution: 1 high, 7 medium, 2 low — all introduced by this PR (none pre-existing), all with a repro or mutation run attached in the comment.

alecantu7:
All nine review findings addressed in `c0d5a30`. **Eight fixed with tests; one needs your decision (the privacy posture, #6).**

Every fix was mutation-verified — reverting it fails its test:

| # | finding | fix | test |
|---|---|---|---|
| 1 (HIGH) | Stop → `error` | classifier accepts `asyncio.CancelledError` | `test_task_cancel_reports_cancelled_not_error` (also asserts the fix does NOT rely on `_cancel_event`, which the hosted path never sets) |
| 2 | ESC-cancel emitted nothing | emit moved FIRST in the `finally`; `reset_trace_context` now ValueError-tolerant; `expected=` books-identity guard | `test_abandoned_generator_still_emits_exactly_once`, `test_late_finalizer_cannot_close_a_newer_turns_books` |
| 3 | cache-WRITE misfiled as fresh input | `_split_cached_input` → 3 components, threaded to all 4 `Usage` sites | `test_gateway_shape_splits_reads_and_writes` (your exact 100/30/10 payload) |
| 4 | failed turns → `completed` | exhausted-retries marks `retry_exhausted` | `test_exhausted_retries_is_not_reported_as_completed` |
| 5 | `rounds` overwritten per continuation | accumulate, counted before the cap check | `test_rounds_accumulate_across_verifier_continuations` (2+1=3), `test_round_cap_reports_the_cap_not_cap_plus_one` |
| 7 | cache split untested | 6 tests incl. provider-level `context_pressure` exactness | your `return total, cached` mutation now **fails 7 tests** (was: passed all 1474) |
| 8 | listener resolved at completion time | captured at call-**issue** time in all 6 client paths | `test_background_call_started_before_disarm_books_nowhere` |
| 9 | `turn()` cap unmarked | mirrors turn_stream | `test_non_streaming_turn_marks_round_cap` |

**Confirmed your finding 3 against the source** rather than taking it on trust: `mindshub_inference@origin/main` publishes `PromptTokensDetails{cached_tokens, cache_write_tokens}` and `total_prompt_tokens = input + cached_input + cache_write`. **The PR body's deviation #2 was factually wrong** — corrected there, in the `Usage` docstring, and on the ticket. `context_pressure` still sums all prompt tokens, so its value is byte-identical to pre-PR (asserted).

Test hygiene: the terminal-path tests stub `dispatch_tool` rather than running the real scratchpad handler — no venv/subprocess per test, no event-loop teardown noise.

**Finding 6 — needs an explicit call before merge.** You're right that it changes the collector's posture and shouldn't ship as a side effect. What I did *not* do is quietly pick an option; I updated `analytics.py`'s module contract to state the truth (these events are now **pseudonymous, not anonymous**: an opaque join key that resolves to a user only for someone holding both collector data and Langfuse access; content still never sent). My recommendation is to **keep it** — the join is the entire point of the field, dropping it makes cost-per-conversation impossible, and hashing it breaks the Langfuse join it exists for. But it's a privacy decision, so: confirm, or say drop/hash and it's a one-line change plus a test.

Full suite: 1486 passed, e2e 39 passed (the 2 `test_chat_scratchpad` failures are the known pre-existing local-sandbox quirk; they fail identically on clean staging and pass in CI).

alecantu7:
**Finding 6 resolved: keeping `conversation_id` and the event. Verified the destination end-to-end against PostHog project 355390 (`8fa6cf4` documents it in code).**

Alejandro pointed out these events reach PostHog, which I hadn't verified. They do, and it settles the design:

1. **The collector relays into PostHog** — anton's events are there tagged `source = mindsdb-zoominfo-lambda` (`ds_connect_success` 19,991 / `anton_query` 802 / `anton_started` 310 in the last 30d).
2. **Every `extra` kwarg becomes a queryable event property — not an allowlist.** Proven on a live event: `ds_connect_success` carries `engine = postgres`. So `turn_completed`'s full token breakdown arrives intact, and **the parameter names are now a published schema** — renaming one breaks downstream queries. Written into the module docstring so the next caller knows.
3. **One event per turn makes turn VOLUME countable**, not just per-turn cost — `count()` = turns, `sum(tokens_total)` = spend, `avg` = cost/turn, `group by ended_by` = the runaway/hand-back distribution, `group by conversation_id` = cost per task (v1). That's the whole ENG-1288 contract, live.

**Two honest limits, now documented rather than discovered later:**
- **`distinct_id` is the `aid` fingerprint → per-INSTALL, not per-user.** These do NOT join to the Keycloak `sub` the console, the desktop renderer's PostHog client, and the billing mirror key on. Per-user/per-plan cost needs either that identified client or the `conversation_id` → Langfuse → user hop. (My earlier PR-body deviation #1 claimed no PostHog surface existed at all — wrong; corrected.)
- **PostHog type-infers property values** (`has_mdb_key` "1"/"0" arrives as `Nullable(Bool)`), so the smoke query should cast explicitly rather than trust inference on `tokens_total`.

On the privacy question specifically — Alejandro's argument, which I agree with: **Langfuse already holds the full conversation content for the same session id**, so shipping an opaque id on this channel adds no disclosure that path doesn't already have. The framing in `analytics.py` is updated accordingly (from "is this anonymous" to "here's where it lands and what the schema contract is").

**Net: no code change to the event. #309 stands as-is, all nine review findings closed.** Ready to merge from my side.

anton/core/session.py
alecantu7: Fixed in `c0d5a30`: the classifier now accepts `asyncio.CancelledError` alongside `GeneratorExit`. Test `test_task_cancel_reports_cancelled_not_error` drives the real `turn_stream`, cancels the consuming task mid-LLM-await, and additionally asserts `_cancel_event` is NOT set — so the test would fail if someone later 'simplified' the fix back to depending on it. Mutation-verified: reverting the isinstance tuple fails that test.

anton/core/session.py
alecantu7: Fixed in `c0d5a30`, both halves. (1) `self._emit_turn_cost(...)` is now the FIRST statement of the `finally`, and `reset_trace_context` is wrapped in `except ValueError` (the cross-context token is unrestorable and the ContextVar copy dies with the task anyway). (2) Books identity guard: `_emit_turn_cost(expected=...)` receives the `TurnCost` captured at books-open and no-ops when a newer turn owns `self._turn_cost`. Tests: `test_abandoned_generator_still_emits_exactly_once` (asserts exactly one event, `cancelled`) and `test_late_finalizer_cannot_close_a_newer_turns_books`.

anton/core/llm/openai.py
alecantu7: Confirmed against the source before fixing — you're right and my PR body was wrong. `mindshub_inference@origin/main`: `PromptTokensDetails{cached_tokens, cache_write_tokens}` (`minds/schemas/chat.py`) and `total_prompt_tokens = input + cached_input + cache_write` (`minds/inference/types.py`), populated from `cache_creation_input_tokens` in `providers/anthropic.py`. Fixed in `c0d5a30`: `_split_cached_input` returns three components, joint-clamped so a malformed payload can't yield negative fresh input, threaded to all four `Usage` sites. `context_pressure` now sums all three (= its exact pre-split value). Test uses your payload verbatim: 100/30/10 → `(60, 30, 10)`. **PR body deviation #2 retracted** there, in the `Usage` docstring, and on ENG-1288.

anton/core/session.py
alecantu7: Fixed in `c0d5a30`: the exhausted-retries branch marks `ended_by = "retry_exhausted"` before yielding the apology, mirroring the `round_cap`/`handback_*` marks. The finally's `error`/`cancelled` overrides still win when an exception genuinely propagates. Test `test_exhausted_retries_is_not_reported_as_completed` drives `turn_stream` with `plan_stream` always raising; mutation-verified (removing the mark fails it).

anton/core/session.py
alecantu7: Fixed in `c0d5a30`: `rounds += 1` instead of assigning the loop-local counter, at both writers (`turn_stream` and `turn`). Also moved the increment *before* the cap check and gated it on `tool_round <= _max_tool_rounds`, so a capped turn reports the cap rather than cap+1 — your second observation. Tests: `test_rounds_accumulate_across_verifier_continuations` (3 rounds + INCOMPLETE verdict + 2 rounds → asserts 3, was 2) and `test_round_cap_reports_the_cap_not_cap_plus_one`.

anton/core/session.py
alecantu7: **Resolved as: keep the field** — but not on my own authority, and with the destination verified rather than inferred. Alejandro flagged that these events reach PostHog; I checked, and they do: project 355390, tagged `source = mindsdb-zoominfo-lambda`. Two things that came out of the check are now documented in `analytics.py` (`8fa6cf4`): every `extra` kwarg arrives as a queryable property (so **param names are a published schema** — `ds_connect_success` carries `engine=postgres` through), and `distinct_id` is the `aid` fingerprint, so these events are **per-install, not per-user** — they do NOT join to the Keycloak `sub` (my PR-body deviation #1 was wrong in the other direction: a PostHog surface does exist, one layer up, in cowork's renderer). On the privacy question itself, the deciding argument is Alejandro's: Langfuse already holds the full conversation content for the same session id, so an opaque id here adds no disclosure that path doesn't already have. The docstring framing changed from 'is this anonymous' to 'here's where it lands and what the schema contract is'.

anton/core/llm/openai.py
alecantu7: Fixed in `c0d5a30` — and I re-ran your mutation to confirm the coverage actually bites. `return total - read - write` → `return total, read, write` now **fails 7 tests** (it passed all 1474 before). Added `TestCacheSplitAcrossProviders`: chat shape, Responses-API shape, gateway read+write shape, no-details passthrough, malformed-payload clamp, plus a provider-level test asserting `input_tokens == 200`, `cache_read_tokens == 800` AND `context_pressure == compute_context_pressure(model, 1000)` — pinning the 'exact old value' claim explicitly, since that was the invariant most likely to rot silently.

anton/core/llm/client.py
alecantu7: Fixed in `c0d5a30`: the listener is captured at call-**issue** time in all six client paths (`plan`, `plan_stream`, `code`, `summarize`, `gate`, `_generate_object_with`) and passed into `_notify_usage`, which no longer reads `self.usage_listener`. Background calls issued while disarmed now book nowhere — which is what `turn_cost.py`'s docstring already promised. Test `test_background_call_started_before_disarm_books_nowhere` reproduces your overlap exactly (a provider that blocks on an Event until after the turn boundary) and asserts the new books show `llm_calls == 0`.

anton/core/session.py
alecantu7: Fixed in `c0d5a30`: `turn()`'s cap branch now marks `ended_by = "round_cap"`, mirroring `turn_stream`. Agreed on the severity reasoning — no in-repo callers — but the PR body claimed all terminal sites mark, and this PR is what wired books into that path, so leaving it would have made the claim false. Test `test_non_streaming_turn_marks_round_cap`.

tests/test_turn_cost_terminals.py
```diff
@@ -0,0 +1,259 @@
...
+        return _gen()
+
+    mock_llm.plan_stream = _stream
+    session = ChatSession(ChatSessionConfig(llm_client=mock_llm, workspace=workspace))
+
+    with patch("anton.analytics.send_event") as send:
+        gen = session.turn_stream("start then abandon")
+        async for _ in gen:
+            break  # abandon mid-iteration, no aclose()
+        await gen.aclose()  # what asyncio's finalizer eventually does
```
alecantu7: **Fix-verification · MEDIUM — this test does not detect the bug it was written for; it passes on the unfixed source** (confirmed by execution both directions)

Reverting only `anton/` to `ecd5c55` and re-running the new tests fails 13 of 14 — **this is the one that still passes.** So the F2a fix (emit-before-`reset_trace_context` + `try/except ValueError`) is currently unpinned: a future refactor that moves the emit back below the reset, or drops the `except`, reintroduces "an ESC-cancelled turn is never counted" with the suite green.

The cause is this line, and specifically its comment:

```python
await gen.aclose()  # what asyncio's finalizer eventually does
```

It isn't. Same-task `aclose()` runs the `finally` in the **original** context, where `reset_trace_context(_trace_token)` succeeds — so the `ValueError` never fires and there is nothing for the fix to survive. The docstring above describes the real mechanism correctly ("runs in a copied context where `reset_trace_context` raises ValueError"), but the body never creates that condition. The original review's own repro had already identified same-task `aclose()` as the **control that works fine** — so the control got encoded as the regression test.

Worth naming the pattern, because it's the same shape as that review's refuted finding #3: a mechanism verified in isolation, without reproducing the condition that breaks it.

**Replacement — executed, fails on `ecd5c55`, passes on `8fa6cf4`:**

```python
async def test_emit_survives_reset_trace_context_valueerror(workspace):
    """Reproduces the real cross-context condition: reset_trace_context raises
    (as it does in a deferred finalizer's copied context) and the turn's books
    must still be emitted.
    """
    mock_llm = make_mock_llm()
    session = ChatSession(ChatSessionConfig(llm_client=mock_llm, workspace=workspace))

    def _boom(_tok):
        raise ValueError("Token was created in a different Context")

    with patch("anton.core.session.reset_trace_context", side_effect=_boom), \
         patch("anton.analytics.send_event") as send:
        async for _ in session.turn_stream("hello"):
            pass
        assert send.call_count == 1, "books must emit even when reset raises"
```

Patching the raise is the right level here: the ValueError *is* the mechanism, and forcing a genuine deferred finalizer (drop the ref, let asyncio schedule it in a fresh task) is timing-dependent and flaky by comparison. Keep the existing test too if you like — it covers the same-task abandon path and `ended_by == "cancelled"` — just don't count it as coverage of the cross-context fix.

alecantu7 (commented):
## Fix-verification pass on the review findings (executed, not read)

Re-verified all 10 findings from the 03:54 review against the three follow-up commits (`c0d5a30`, `31bf348`, `8fa6cf4`) in an isolated worktree. **Nine are genuinely closed; one regression test is hollow (detail inline); F6 is signed off below.**

Method, since "the tests pass" is not evidence on its own: full suite on head = **1527 passed, 28 skipped** (up from the 1474 baseline). Then reverted **only** `anton/` to the pre-fix commit `ecd5c55` while keeping the new tests → **13 of the 14 new tests fail**. That before/after is what makes them regression tests rather than decoration.

| # | Finding | Sev | Fix | Test pins it? |
|---|---|---|---|---|
| 1 | User Stop lands in `error` | HIGH | ✅ `(GeneratorExit, asyncio.CancelledError)` | ✅ |
| 2a | Finalizer `ValueError` aborts emit | MED | ✅ emit-before-reset + `try/except` | ❌ **see inline** |
| 2b | Late finalizer closes newer books | MED | ✅ `expected=` guard | ✅ |
| 3 | Cache-*write* misfiled as fresh | MED | ✅ all 4 `Usage` sites + both docstrings | ✅ |
| 4 | Failed turns report `completed` | MED | ✅ `retry_exhausted` | ✅ |
| 5 | `rounds` overwritten (+ cap+1) | MED | ✅ both writers, both halves | ✅ |
| 6 | `conversation_id` privacy posture | MED | ✅ documented + **signed off** | n/a |
| 7 | Cache split untested | MED | ✅ 6 new tests | ✅ |
| 8 | Listener resolved at completion | MED | ✅ all 6 call sites | ✅ |
| 9 | `turn()` cap unmarked | LOW | ✅ | ✅ |
| 10 | CLI toolbar understates | LOW | ✅ `context_tokens` | ➖ none (display-only, fine) |

Three fixes went past what was asked, worth noting for the human reviewer: **F2** implemented *both* offered remedies rather than either; **F5** also closed the cap+1 sub-issue via `tool_round <= self._max_tool_rounds` before the increment; **F3** retracted the PR body's deviation #2 with a visible strikethrough instead of quietly editing it — the wrong claim stays legible next to its correction, which is the right call.

Completeness checks that came back clean: all six `_notify_usage` call sites now pass a listener captured before the `await` (F8 — a partial fix here would have been invisible); `Usage.context_tokens` exists and sums all three components (F10's fix depends on it); `_split_cached_input`'s joint clamp cannot produce negative fresh input for any read/write/total combination.

### F6 — signed off

The finding asked for the collector's anonymity-posture change to be **decided visibly rather than shipped as a side effect**. Alejandro reviewed it and approved keeping `conversation_id` (2026-08-06). Recording that here so the decision is on the record where the code is, which is the whole ask. The supporting argument in `anton/analytics.py` stands: the id is opaque, resolving it to a person requires Langfuse, and Langfuse already holds the full conversation content for that same session — so it adds no disclosure that path doesn't already have. The docstring no longer claims "anonymous / no PII", and it now documents that every `extra` kwarg becomes a queryable PostHog property (i.e. the kwarg names are a published schema) and that `distinct_id` is per-install `aid`, not per-user.

One claim in that docstring I did **not** independently verify: "verified 2026-08-06 against project 355390". Project 355390 is indeed Anton, so it's consistent — just flagging that the destination assertion is untested here rather than confirmed.

<sub>Verified in a scratch worktree on a separate ref so the in-flight `review/pr-309` worktree was left untouched. Full suite 1527/28 on head; 13/14 new tests confirmed failing on `ecd5c55` with tests held constant.</sub>

tests/test_turn_cost_terminals.py
alecantu7: Correct, and confirmed independently before fixing — thanks, this is the most useful kind of finding: a green suite that isn't. Fixed in `8d7d97d`.

**Reproduced your result with a targeted mutation** (restore the `ecd5c55` ordering — emit after reset, no `except` — while leaving everything else intact): **only the new test fails; the abandon test keeps passing.** So it never covered this mechanism, exactly as you said.

```
FAILED test_emit_survives_reset_trace_context_valueerror
1 failed, 7 passed
```

What landed:
- **`test_emit_survives_reset_trace_context_valueerror`** — your version, verbatim in substance. Agreed on patching the raise rather than forcing a real deferred finalizer: the ValueError *is* the mechanism, and a genuine finalizer is timing-dependent.
- **Kept the abandon test**, with a scope note stating what it does NOT cover and a pointer to the new one — and **corrected the comment that caused the gap**: `# what asyncio's finalizer eventually does` was the false claim that made same-task `aclose()` look like the real condition. A wrong comment beside a passing test is how this stays invisible, so it had to go, not just be supplemented.

**Applied the same skepticism to the sibling fix in that finding.** A full revert to `ecd5c55` fails `test_late_finalizer_cannot_close_a_newer_turns_books` for a *signature* reason (no `expected=` param yet), which would prove nothing about the guard. So I mutated the guard **body** with the signature intact — it fails only its own test. That one bites for the right reason.

Point taken on the pattern, too: same shape as the refuted finding #3 in the earlier round — mechanism verified in isolation, condition never reproduced. Worth carrying into how I write the next batch of these.

Full suite: 1487 passed (the 2 `test_chat_scratchpad` failures are the known pre-existing local-sandbox quirk).

anton/core/llm/client.py
```diff
@@ -227,6 +281,16 @@ async def _generate_object_with(
             tool_choice={"type": "tool", "name": tool["name"]},
             max_tokens=budget,
         )
+        # Count BEFORE the no-tool-call raise below: a structured call that
+        # failed (and its bigger-budget retry) still spent real tokens
+        # (ENG-1288). Role is coding for both generate_object variants today;
+        # derive it from the provider/model actually used.
+        self._notify_usage(
+            "planning" if model == self._planning_model else "coding",
```
ea-rus: **MEDIUM · correctness — Role is inferred by comparing model names, so it collapses when a deployment runs one model for both roles.**

`_generate_object_with` doesn't receive a role — it derives one from `model == self._planning_model`. When planning and coding resolve to the same model, every `generate_object_code` call (verifier verdicts, memory compaction, identity extraction, scratchpad post-mortem) books to **planning**.

That's not hypothetical: cowork-server's Gemini defaults are identical across all three roles (`cowork/common/settings/app_settings.py:63-73` — `PLANNING_MODEL_DEFAULTS["gemini"] == CODING_MODEL_DEFAULTS["gemini"] == "gemini-3.6-flash"`).

Executed against the real client with `planning_model == coding_model == "gemini-3.6-flash"`, calling `generate_object_code`:

```
by_role: {'planning': RoleCost(model='gemini-3.6-flash', tokens=1100, calls=1)}
```

`coding` never appears. Dollar math survives (one model, one rate), but `coding_tokens` under-reports and `planning_tokens` over-reports — which defeats the split this commit exists to provide: separating the expensive planning loop from the cheap verifier.

**Fix:** both callers already know the role, so pass it instead of guessing:

```python
async def _generate_object_with(self, ..., role: str, ...):
    ...
    self._notify_usage(role, model, response.usage, listener)
# generate_object      -> role="planning"
# generate_object_code -> role="coding"
```

**Test:** construct an `LLMClient` with planning == coding, call `generate_object_code`, assert `"coding" in tc.by_role`. Fails today.

anton/core/session.py
```diff
@@ -1720,6 +1742,134 @@ def factory_validated():
...
+
+        Post-turn background work (cerebellum flush, identity extraction)
+        runs after this and is deliberately NOT attributed to any turn —
+        the listener is disarmed here precisely so late usage can't leak
+        into the next turn's books (stated ENG-1288 gap).
+        """
+        tc = self._turn_cost
+        if tc is None:
+            return
+        if expected is not None and tc is not expected:
```
ea-rus: **MEDIUM · correctness — The books-identity guard is right to refuse the newer turn's books, but it also discards the abandoned turn's own — so that turn is never counted at all.**

F2b's fix no-ops when `self._turn_cost` has moved on. The stale `TurnCost` passed in as `expected` holds a complete, valid set of totals for a real turn — and it is dropped on the floor. Repro: consume one event, `break`, start a second `turn_stream` on the same session, then close the first generator → **1 `turn_completed` event for 2 turns**.

`test_late_finalizer_cannot_close_a_newer_turns_books:200` currently codifies the drop (`session._emit_turn_cost(expected=stale); assert not send.called`). The invariant it should encode is narrower: *a late finalizer must not close a newer turn's books* — not *a late finalizer must emit nothing*.

Narrow window in practice (cowork-server builds a fresh `ChatSession` per turn; the CLI usually finalizes before the next turn opens), but the failure mode is the runaway turn the user just cancelled — exactly the one ENG-1288 exists to catch.

**Fix:** emit whichever books were handed in; only clear the shared slot when it matches.

```python
@dataclass
class TurnCost:
    emitted: bool = False

def _emit_turn_cost(self, expected: TurnCost | None = None) -> None:
    tc = expected if expected is not None else self._turn_cost
    if tc is None or tc.emitted:
        return
    tc.emitted = True
    if self._turn_cost is tc:          # only the owning turn clears the slot
        self._turn_cost = None
        self._llm.usage_listener = None
    ...
```

The `emitted` flag preserves the existing double-emit no-op that `self._turn_cost = None` used to provide.

**Test:** flip the existing test to `assert send.called` with the stale books' totals, and add `assert session._turn_cost is current`.

anton/core/session.py
```diff
@@ -1720,6 +1742,134 @@ def factory_validated():
...
+            # turn's partial totals mid-turn and leave it uncounted. Only the
+            # turn that opened these books may close them (#309 review).
+            return
+        self._turn_cost = None
+        try:
+            self._llm.usage_listener = None
+        except Exception:
+            pass
+
+        exc = sys.exc_info()[1]
```
ea-rus: **LOW · latent — `sys.exc_info()` returns whatever the thread is currently handling, including the caller's own exception — so a clean turn can report `ended_by="error"`.**

The 02:43 self-review probe established the true-positive direction (propagating exceptions and `GeneratorExit` are visible in the `finally`). The false-positive direction isn't covered. Executed:

```python
try:
    raise ValueError("caller's own, already-handled error")
except ValueError:
    async for _ in session.turn_stream("hello"):
        pass
# ended_by == 'error'   (expected 'completed')
```

Not reachable from today's callers — `chat.py`, `anton_harness/harness.py` and `connectors/probe.py` all iterate inside `try`, not inside `except`. Filing it as latent rather than live, but `ended_by` is the field every error-rate query keys on, and the trap costs one line to remove.

**Fix:** capture the exception explicitly instead of asking the interpreter what it happens to be handling.

```python
_turn_exc: BaseException | None = None
try:
    ...
except BaseException as e:
    _turn_exc = e
    raise
finally:
    self._emit_turn_cost(expected=_turn_cost_books, exc=_turn_exc)
```

`_emit_turn_cost` then takes `exc` as a parameter (defaulting to `sys.exc_info()[1]` keeps the `turn()` path unchanged).

anton/core/session.py
```diff
@@ -3165,6 +3368,8 @@ async def _stream_and_handle_tools(
                 yield StreamTaskProgress(
                     phase="analyzing", message="Diagnosing incomplete task..."
                 )
+                if self._turn_cost is not None:
+                    self._turn_cost.ended_by = "handback_budget"
```
ea-rus, line 3429: **MEDIUM · test coverage — None of the three `handback_*` marks is pinned by a test — all three can be deleted with the suite green.**

Mutation run on head, one at a time, each replaced with `pass`:

```
SURVIVED  handback_budget           (session.py:3372)  35 passed
SURVIVED  handback_stuck            (session.py:3609)  35 passed
SURVIVED  handback_verifier_failure (session.py:3569)  35 passed
```

`test_explicit_terminal_marks_survive_clean_exit` looks like coverage but isn't — it sets `s._turn_cost.ended_by = "handback_stuck"` **on the object** and asserts `_emit_turn_cost` preserves it. That pins the preservation logic, not the three sites that are supposed to write the mark.

That's 3 of the 7 `ended_by` values, and specifically the "expensive turn ended badly" ones — the `group by ended_by` distribution named as the ENG-1288 contract in the 04:32 comment. For contrast, every other terminal (`round_cap`, `retry_exhausted`, `cancelled`, `error`, `completed`) is mutation-caught.

**Fix:** three tests on the pattern of `test_task_cancel_reports_cancelled_not_error` — drive `turn_stream` into each hand-back and assert `_ended_by(send)`.

anton/core/turn_cost.py
```diff
@@ -0,0 +1,136 @@
...
+        can answer "which model was this user actually on" and "what did each
+        model cost" rather than only a blended sum.
+        """
+        self.llm_calls += 1
+        self.input_tokens += usage.input_tokens
+        self.output_tokens += usage.output_tokens
+        self.cache_read_tokens += usage.cache_read_tokens
+        self.cache_creation_tokens += usage.cache_creation_tokens
+        self.peak_context_tokens = max(self.peak_context_tokens, usage.context_tokens)
+        slice_ = self.by_role.setdefault(role or "unknown", RoleCost())
```
ea-rus: **LOW · robustness — The `unknown` role bucket is written but never read, so its tokens vanish from the per-role breakdown while still counting toward `tokens_total`.**

`role or "unknown"` creates a fourth bucket, but the event only reads `planning` / `coding` / `router`. If it ever fills, `planning_tokens + coding_tokens + router_tokens != tokens_total` — silently breaking the reconciliation `test_role_slices_track_model_tokens_and_calls` asserts. Unreachable today (every `_notify_usage` call site passes a literal role), and the mutation `role or "unknown"` → `role or "planning"` survives the suite.

**Fix:** pick one — drop the `or "unknown"` so an empty role raises a `KeyError` at the point of the bug, or emit `unknown_tokens`/`unknown_calls` alongside the other three so the sum always reconciles. Second option is safer if a future caller ever passes a computed role.

anton/core/llm/client.py
alecantu7: **Confirmed and fixed in `024980c`.** The Gemini evidence is what makes this a real bug rather than a theoretical one — I inferred the role because `_generate_object_with` is shared, and never checked whether the two models can coincide. They do, by default, on a whole provider.

Both callers pass `role=` now (`generate_object` → planning, `generate_object_code` → coding), so nothing is derived from model identity. Test `test_structured_coding_call_books_to_coding_when_models_match` builds a client with `planning_model == coding_model == "gemini-3.6-flash"` and asserts `"coding" in tc.by_role` and `"planning" not in`; mutation-verified — restoring the inference fails only that test.

Worth noting the consequence you named, since it's the reason this matters beyond tidiness: with the roles collapsed, "what is the verifier costing us vs the user-facing loop" is unanswerable on exactly the deployments where cost questions get asked.

anton/core/session.py
alecantu7: **Right on both counts — fixed in `024980c`, and the test flipped.** You've correctly separated the two invariants I'd conflated: *don't close a newer turn's books* (what I meant) vs *emit nothing* (what I implemented, and then codified in the test).

Took your shape: `tc = expected if expected is not None else self._turn_cost`, an `emitted` flag on `TurnCost` as the double-emit guard, and the shared slot cleared only when `self._turn_cost is tc`. The test now asserts the abandoned books ARE reported (`tokens_total == "780"` from the stale set), that re-emitting the same books is still a no-op, and that the current turn keeps its slot until its owner closes it.

Your last line is the part I'd underweighted: the narrow window is real, but the turn it drops is disproportionately the cancelled runaway — the single case ENG-1288 exists to make visible. Mutation-verified.

anton/core/session.py
alecantu7: **Fixed in `024980c` — and your suggested signature needed one adjustment worth flagging, because I implemented it literally first and it didn't work.**

`exc` defaulting to `None` conflates two different things: *"the caller captured no exception"* (a clean turn — authoritative) and *"the caller didn't tell us"* (fall back to `sys.exc_info()`). With a plain `None` default, a clean `turn_stream` passes `exc=None`, the fallback fires anyway, and the leak survives untouched — my new test failed on my own fix and caught it. So the parameter takes a module-level `_EXC_UNSET` sentinel; only the legacy non-streaming `turn()`, which has no wrapping handler to capture from, reaches the lookup.

`turn_stream` now does `except BaseException as _e: _turn_exc = _e; raise` — `BaseException` deliberately, so `GeneratorExit` and `CancelledError` (both real turn endings, both used by the cancel classifier) are captured too. Test reproduces your snippet exactly: raise, catch, then iterate the turn inside the `except`, assert `completed`. Mutation-verified: restoring `if exc is None` fails 4 tests.

anton/core/session.py
alecantu7: **This one lands hardest — thank you.** Three of seven `ended_by` values unpinned, and the mutation table is unarguable. Fixed in `024980c`.

You also correctly diagnosed *why* it looked covered: `test_explicit_terminal_marks_survive_clean_exit` sets `ended_by` **on the object** and asserts `_emit_turn_cost` preserves it — that pins the preservation logic, not the three sites that write the mark. A test that exercises the neighbouring working path and passes for the wrong reason.

Three tests now drive `turn_stream` into each hand-back for real: a STUCK verdict, INCOMPLETE with `_max_continuations = 0`, and a verdict call that raises (ENG-1079's fail-safe). All three mutation-verified individually — deleting any one mark now fails exactly its own test.

The uncomfortable part: I wrote a memory note yesterday that a test isn't done until I've watched it fail on the unfixed code, after being caught by this same class twice. I mutation-verified `round_cap`, `retry_exhausted`, `cancelled` and `error` — and then didn't extend the same check to the three marks I'd added earlier in the PR. Your per-site table is the discipline actually applied.

anton/core/turn_cost.py
alecantu7: **Fixed in `024980c`, taking your second option.** Emitting `unknown_tokens`/`unknown_calls` rather than raising, for the reason you gave: a future caller passing a computed role should degrade to a visible non-zero bucket, not a `KeyError` in the middle of a turn's accounting — reporting must never be able to break the turn it measures.

Two tests: an empty role lands in `unknown` and is emitted, and reconciliation holds with all four buckets populated (`planning + coding + router + unknown == tokens_total`). The second is the one that matters — it makes the sum an asserted property rather than an incidental one, so a fifth bucket added later can't quietly break it. Mutation-verified: dropping the two emitted properties fails both.

alecantu7:
@ea-rus — **all five fixed in `024980c`, ready for another pass.** Inline replies are on each thread (I resolved them, so they're collapsed — the detail is in there). Summary so the re-review doesn't need archaeology:

| # | finding | fix | pinned by |
|---|---|---|---|
| F1 | role inferred from `model == planning_model`, collapses on shared-model deployments | both callers pass `role=` explicitly | `test_structured_coding_call_books_to_coding_when_models_match` (planning==coding==gemini-3.6-flash) |
| F2 | guard discarded the abandoned turn's books | emit the handed-in books; only the owner clears the slot; `emitted` flag as the double-emit guard | flipped `test_late_finalizer_cannot_close_a_newer_turns_books` to assert it IS emitted |
| F3 | `sys.exc_info()` can read the caller's own handled exception | `turn_stream` captures via `except BaseException`; sentinel param | `test_callers_already_handled_exception_is_not_reported_as_error` |
| F4 | three `handback_*` marks unpinned | 3 tests driving `turn_stream` into each hand-back | one test per mark, each mutation-verified individually |
| F5 | `unknown` bucket written, never read | emitted, so per-role sum always reconciles | `TestUnknownRoleReconciles` (2 tests) |

**Your mutation table, re-run after the fixes** — each fix reverted on its own now fails only its own test:

```
F1 revert (infer role)                 -> FAILED test_structured_coding_call_books_to_coding...
F2 revert (drop abandoned books)       -> FAILED test_late_finalizer_cannot_close_a_newer_turns_books
F3 revert (exc is None fallback)       -> FAILED ...already_handled_exception... (+3 more)
F4 revert handback_budget              -> FAILED test_exhausted_continuations_report_handback_budget
F4 revert handback_stuck               -> FAILED test_stuck_verdict_reports_handback_stuck
F4 revert handback_verifier_failure    -> FAILED test_verifier_failure_reports_handback_verifier_failure
F5 revert (unknown not emitted)        -> FAILED TestUnknownRoleReconciles (both)
```

**Two things worth your eyes specifically, because they diverge from what you proposed:**

1. **F3 needed a sentinel, not `exc=None`.** I implemented your signature literally first and my own new test failed on it: a clean `turn_stream` passes `exc=None`, which a `None` default reads as "not told" → falls back to `sys.exc_info()` → leak intact. So `_emit_turn_cost(exc=_EXC_UNSET)`, where an explicit `None` means "clean turn, and I'm authoritative". Only the legacy non-streaming `turn()` reaches the lookup.
2. **F5 took your "second option"** (emit `unknown_*`) over raising — reporting must never be able to break the turn it measures, so a computed role should degrade to a visible non-zero bucket rather than a `KeyError` mid-accounting.

Suite: 1501 passed + 39 e2e (the 2 `test_chat_scratchpad` failures are pre-existing on clean staging — they pass in a fresh worktree, which I verified separately).

**Heads-up on CI:** the red checks on this PR are **GitHub's Actions outage**, not code. All are `cancelled` with zero failing steps, killed at exactly 15m03s; GitHub's status page has an active critical incident (Actions `major_outage`, opened 15:22 UTC). They'll need one re-run once it clears.

anton/core/session.py
```diff
@@ -1720,6 +1748,153 @@ def factory_validated():
...
+        # raises it inside the suspended generator frame — and nothing there
+        # sets `_cancel_event` (that's anton's CLI only). Without it every
+        # Stop filed as "error" (#309 review). GeneratorExit covers the
+        # narrower case where the cancel lands exactly at a yield boundary.
+        if isinstance(exc, (GeneratorExit, asyncio.CancelledError)) or cancelled:
+            tc.ended_by = "cancelled"
+        elif exc is not None:
+            tc.ended_by = "error"
+
+        turn_index = self._turn_count + 1
```
ea-rus: **LOW · correctness — Follow-up on F2: the late emit reads live session state, so the abandoned turn is stamped with a later turn's index and an inflated duration. Not a merge blocker.**

F2 is right and verified — the abandoned turn's books now reach the sink. But two event fields are still resolved at *emit* time from state that has moved on: `turn_index` reads `self._turn_count` here, and `duration_ms` measures `time.monotonic()` against `started_monotonic`. For books a late finalizer emits, both belong to a turn that ended long ago.

Abandon turn 1 -> complete turn 2 -> fire turn 1's finalizer -> complete turn 3:

```
(ended_by, turn_index):  [('completed','1'), ('cancelled','2'), ('completed','2')]
```

The abandoned turn takes index 2 and turn 3 takes it again — `(conversation_id, turn_index)`, the Langfuse hop, is no longer unique. Same shape for duration: a turn abandoned ~10 ms in, finalized 400 ms later, reports `duration_ms=426` — the field a runaway query sorts on. (That half is partly pre-existing; F2 turned it from "never reported" into "always reported".)

**Low, honestly:** cowork-server can't reach it (fresh `ChatSession` per turn → the stale session's `_turn_count` is frozen at the right value), and on the CLI the finalizer normally fires while the prompt awaits input — I had to order `break` / full turn / `aclose()` by hand to reproduce. Token counts are unaffected; they live on the `TurnCost`. Follow-up is fine.

**Fix — stamp per-turn facts at books-open, not at read:**

```python
@dataclass
class TurnCost:
    turn_index: int = 0
    ended_monotonic: float | None = None

self._turn_cost = TurnCost(turn_index=self._turn_count + 1)   # turn_stream / turn
```

Set `ended_monotonic` where `ended_by` resolves; `duration_ms` prefers it. `_emit_turn_cost` uses `tc.turn_index`, keeping the current lookup as fallback for bare-session tests and legacy `turn()`. **Test:** the sequence above, asserting three distinct indices — fails today.

ea-rus (approved):

anton/core/turn_cost.py
alecantu7: Fixed in `0caaf82` — thanks for the approve, and this was worth doing now rather than as a follow-up, because it's the cost→Langfuse hop pointing at the wrong turn.

Took your shape: `TurnCost` gains `turn_index` stamped at construction by both books-open sites (live lookup kept as fallback for bare-session tests and legacy `turn()`), plus `ended_monotonic` + `last_activity_monotonic`. The owning turn is ending now so it uses now; a late finalizer can't know the real end, so it uses the last LLM call — understating, but bounded and truthful rather than measuring to whenever asyncio scheduled it.

**One correction to the framing, and it changed the test.** Your suggested assertion — three *distinct* indices — isn't achievable, and I don't think it should be. `_turn_count` advances only **after** a turn completes, so an abandoned turn and its successor legitimately share an index. Running your sequence I got `['1','1','2']`: the collision is turn 1 vs turn **2**, not turn 3.

And that collision is *correct*: the Langfuse trace context derives its `turn_id` from the **same** `self._turn_count + 1` expression, so both sides independently agree this is "two attempts at one conversational turn." Forcing three distinct indices would desynchronise the cost event from the trace it's supposed to join to.

So what was actually broken — and is now fixed and pinned — is narrower than uniqueness: **the abandoned turn drifting onto an unrelated later turn's index.** `test_abandoned_turn_keeps_its_own_index_not_a_later_turns` asserts the cancelled event still says `1` (reverting the stamp makes it say `2`), and `test_late_emit_duration_is_bounded_by_last_activity` covers the duration half. Both mutation-verified.

Consequence I've written into the test docstring rather than leaving implicit: **`(conversation_id, turn_index)` is a join key, not a unique event key** — my earlier PR-body and ticket wording overclaimed that, and I've corrected it. Two events sharing it is meaningful information (a retried turn), not corruption.

Suite: 1503 + 39 e2e.

alecantu7:
@ea-rus — **two commits landed after your approval, and I'd like you to re-confirm rather than have me ride an approval that predates them.** Both are on your finding's surface, but one contradicts your suggested test and the other you haven't seen at all.

### `0caaf82` — your F2 follow-up (turn_index / duration stamped at books-open)

Implemented as you proposed: `TurnCost` gains `turn_index` stamped at construction, plus `ended_monotonic` + `last_activity_monotonic` (the owning turn uses now; a late finalizer uses the last LLM call — understating, but bounded rather than measuring to whenever asyncio ran it).

**Where I diverged, and why — this is the part worth your disagreement.** Your suggested test asserted **three distinct turn indices**. That isn't achievable, and I don't think it should be. Running your sequence I got `['1','1','2']` — the collision is turn 1 vs turn **2**, not turn 3, because `_turn_count` advances only *after* a turn completes, so an abandoned turn and its successor legitimately share an index.

And that shared index is *correct*: the trace context derives `turn_id` from the **same** `self._turn_count + 1` expression, so both sides independently agree it's "two attempts at one conversational turn." Forcing distinct indices would desynchronise the cost event from the trace it exists to join to. So I narrowed the assertion to the defect that is real — **the abandoned turn drifting onto an unrelated later turn's index** — and documented that `(conversation_id, turn_index)` is a **join** key, not a unique event key. My earlier PR body and ticket comment overclaimed uniqueness; both corrected.

If you think uniqueness is worth having, that's a different (bigger) change — moving the `_turn_count` increment, which would ripple into episodic memory, explainability and trace ids. I'd rather you weigh in than have me pick.

### `d98f75d` — a defect **in my own fix**, found by reviewing it before asking you

`turn_stream` accepts an explicit `turn_id` ("so downstream telemetry can correlate"), and the trace context prefers it: `turn_id if turn_id is not None else self._turn_count + 1`. My books derived their index from `_turn_count` alone — so the cost event and the trace used **different expressions**, reintroducing exactly the defect one layer down.

Not live: verified against `origin/staging` that cowork-server's `anton_harness` builds `turn_kwargs` with only `trace_tags`/`trace_metadata`, so no host passes `turn_id` today and the two happen to agree. But it was correct *by accident*; now the books stamp the identical expression and it's correct *by construction*. Pinned by `test_host_supplied_turn_id_is_what_the_event_reports` (`turn_id=4242` must surface as `turn_index`), mutation-verified.

Also documented one residual I probed rather than fixed: a turn abandoned *before* its first LLM call has no activity watermark, so its duration still runs to finalizer time. It carries `llm_calls == 0` / `tokens_total == 0`, so it's excluded from the cost and runaway queries that are duration's only consumer.

### Why I'm asking rather than assuming

Your approval technically still stands, and both commits are test-covered and mutation-verified. But you've reviewed this file twice today and caught real defects both times, I found a third by re-reading my own follow-up, and I've reinterpreted the conclusion of a finding you raised. That's three signals that one pass over this area isn't enough. A glance at two commits is cheap; a wrong join key isn't.

### CI caveat, so you don't chase it

The PR currently reads `mergeState=CLEAN` with **only 3 check-runs on head** — CodeQL only. `run-tests`, `build`, `scope` and `verdict-eval` were never triggered: GitHub's Actions incident still has webhook triggers throttled ("many push and pull request events are not yet triggering new workflow runs"). So CLEAN here means *no checks ran*, not *checks passed* — worth knowing before anyone reads it as green. Locally on `d98f75d`: **1504 passed + 39 e2e** (the 2 `test_chat_scratchpad` failures are pre-existing on clean staging).

anton/core/turn_cost.py
```diff
@@ -0,0 +1,172 @@
...
+        self.input_tokens += usage.input_tokens
+        self.output_tokens += usage.output_tokens
+        self.cache_read_tokens += usage.cache_read_tokens
+        self.cache_creation_tokens += usage.cache_creation_tokens
+        self.peak_context_tokens = max(self.peak_context_tokens, usage.context_tokens)
+        self.last_activity_monotonic = time.monotonic()
+        # An empty role lands in `unknown`, which the event emits alongside the
+        # three real roles — so the per-role sum always reconciles with
+        # `total_tokens` instead of silently losing a bucket (#309 review).
+        slice_ = self.by_role.setdefault(role or UNKNOWN_ROLE, RoleCost())
```
pnewsam: **Nit:** the `unknown` bucket only catches a *falsy* role (`role or UNKNOWN_ROLE`), not an unexpected *non-empty* one. If a future caller ever passes a novel role string (e.g. `"verifier"`), its tokens land in `by_role["verifier"]`, which the event never reads — so they silently drop from the per-role breakdown and quietly break the `planning + coding + router + unknown == tokens_total` reconciliation *without* tripping the non-zero `unknown` alarm. The comment at `session.py` ("reconciles… even if a future caller passes an unexpected role") slightly overstates the guarantee: it holds for empty roles, not arbitrary ones.

Unreachable today — every `_notify_usage` site passes a literal `"planning"`/`"coding"`/`"router"` — so this is a Nit, not an ask. If you want the invariant to actually hold, fold any role not in the closed `EVENT_ROLES` set into `unknown` here.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Independent pass over the full diff (11 files, +1766/-20): the `Usage` cache split, `_split_cached_input` clamp math, the `LLMClient` listener waist, `TurnCost` accumulation, and all session wiring (`turn`, `turn_stream`, `_emit_turn_cost`, terminal marks). The PR already carries ea-rus's approval at `024980c`; I re-derived the two post-approval commits (`0caaf82`, `d98f75d`) from scratch rather than trusting them. No merge-blocking defect. One Nit inline.

### Validation

- Traced the non-streaming `turn()` cap branch: it `break`s to the shared tail and reaches `_emit_turn_cost()` — no missed emission or listener leak on that path.
- Verified `_split_cached_input`'s joint clamp: `fresh + read + write == total` for all combinations, so `context_pressure` (recomputed on the summed value) is byte-identical to pre-split on every OpenAI site.
- Confirmed the books `turn_index` and the trace context `turn_id` use the identical `turn_id if turn_id is not None else self._turn_count + 1` expression — the cost→Langfuse join can't drift even when a host supplies `turn_id`.
- Confirmed the listener is captured at call-issue in all six client paths and disarmed in `_emit_turn_cost`, so post-turn background work books nowhere.
- CI on head `d98f75d`: `Analyze (python)`, `Analyze (actions)`, CodeQL, and all three Snyk checks pass.

The documented gaps (`turn()` error-exit skips emission; abandoned-before-first-call duration) are each explicitly accepted with sound rationale, and keeping the opaque `conversation_id` join key is the right call given Langfuse already holds the conversation under the same session id.

anton/core/turn_cost.py
alecantu7: Fixed rather than left as a nit — the code half is small, but **the comment half is the real defect** and I'd rather not ship it.

`TurnCost.add` now folds anything outside the closed `EVENT_ROLES` set into `unknown`, so the reconciliation holds for **any** role string, not just `""`. And you're right that the `session.py` comment overstated it — it claimed the sum reconciles "even if a future caller passes an unexpected role", which was true only for falsy ones. Corrected, because a comment asserting a guarantee the code doesn't provide is exactly the shape that has already bitten this PR twice: the false `aclose()` comment earlier, and the "keep them aligned" comments now filed as [ENG-1322](https://linear.app/mindsdb/issue/ENG-1322).

Trade-off, stated in the code: the folded role's **name** is lost, only its tokens survive. Deliberate — tokens reconciling is the invariant that matters, and a non-zero `unknown` is the alarm that someone invented a role. Preserving names would be a bigger change than this guarantee is worth.

`test_novel_role_folds_into_unknown_rather_than_vanishing` asserts a `"verifier"` role lands in `unknown` (`unknown_tokens == "320"`), never keys its own bucket, and that all four buckets still sum to `tokens_total`. Mutation-verified: restoring `role or UNKNOWN_ROLE` fails only that test. Fixed in `77e57e3`.

Suite: 1505 + 39 e2e. Thanks for both passes on this one — and apologies for resolving the thread a moment before this reply landed; my API call for it 404'd on a bad id.

anton/core/session.py
alecantu7: Fixed in `0caaf82`, with a follow-up in `d98f75d` — and apologies, my original reply to this landed on the wrong thread (it went to the earlier `unknown`-bucket one; I selected the comment id by "last comment by author" instead of by thread). Reposting it here where it belongs.

Took your shape: `TurnCost` gains `turn_index` stamped at construction, plus `ended_monotonic` + `last_activity_monotonic` — the owning turn uses now; a late finalizer uses the last LLM call, understating but bounded rather than measuring to whenever asyncio ran it.

**Where I diverged, and it's worth your disagreement.** Your suggested test asserted **three distinct turn indices**. That isn't achievable, and I don't think it should be. Running your sequence I got `['1','1','2']` — the collision is turn 1 vs turn **2**, not turn 3, because `_turn_count` advances only *after* a turn completes, so an abandoned turn and its successor legitimately share an index.

And that shared index is *correct*: the trace context derives `turn_id` from the **same** `self._turn_count + 1` expression, so both sides independently agree it's "two attempts at one conversational turn." Forcing distinct indices would desynchronise the cost event from the trace it exists to join to. So I narrowed the assertion to the defect that is real — **the abandoned turn drifting onto an unrelated later turn's index** — and documented that `(conversation_id, turn_index)` is a **join** key, not a unique event key. My earlier PR body and ticket comment overclaimed uniqueness; both corrected.

If you think uniqueness is worth having, that's a bigger change — moving the `_turn_count` increment, which ripples into episodic memory, explainability and trace ids. Your call rather than mine.

**`d98f75d` — a defect in my own fix, found reviewing it before asking you to re-confirm.** `turn_stream` accepts an explicit `turn_id` and the trace prefers it; my books derived the index from `_turn_count` alone, so the two used *different* expressions — the same defect one layer down. Not live (verified against `origin/staging`: cowork-server's `anton_harness` passes only `trace_tags`/`trace_metadata`, so no host sends `turn_id` today), but correct by accident. Both now use the identical expression. Pinned by `test_host_supplied_turn_id_is_what_the_event_reports`.

Tests: `test_abandoned_turn_keeps_its_own_index_not_a_later_turns` (the cancelled event must still say `1`; reverting the stamp makes it say `2`) and `test_late_emit_duration_is_bounded_by_last_activity`. Both mutation-verified.

PR #310, opened 2026-08-05
ENG-1274: Fix: compaction fail can clear history for anton

ea-rus:
The problem:
In rare case, if there is a network error and compaction fail: a history was replaced by "(Earlier conversation with N turns — summarization failed)"
- for anton history is lost. 
- for cowork server history is restored on next turn

Changes:
- don't replace history with '... summarization failed' content
- mark `_compacted_this_turn`  if compaction was success

Fixes: https://linear.app/mindsdb/issue/ENG-1274/a-failed-history-summarisation-silently-replaces-the-conversation-with


alecantu7 (commented):
Reviewed the diff and dug into your three points from chat, because two of them change how I'd weigh this and one I'd push back on. Short version: **the fix is right, your read of the blast radius is better than mine was, and the caller-side change introduces a new failure mode worth closing before this lands.**

## The fix does what it says, plus one thing you didn't claim

Same probe on both branches — 6 tool rounds, summarizer raising persistently:

| | `origin/staging` | this PR |
|---|---|---|
| history preserved | **no** — placeholder written | **yes** |
| `StreamContextCompacted` emitted | **1** | 0 |
| summarize LLM calls in the turn | 1 | **7** |

Mutation-tested rather than eyeballed: putting the placeholder line back fails `test_failed_summarize_reports_no_compaction`. That half is properly pinned.

The middle row is a bonus the PR body doesn't mention and is worth keeping in the description: on staging a *failed* compaction still emitted "Context was getting long — older history has been summarized." So the user was told their history was summarized at the exact moment it was deleted. That's gone now.

## 1. A failed compaction is now re-attempted every tool round

This is the one I'd fix before merge.

`_compacted_this_turn` was doing two jobs. The obvious one is "record that we compacted". The incidental one is "stop this from being attempted again this turn". Gating it on success fixes the first and silently removes the second — so when compaction fails, the gate stays open and the pressure check re-fires on the next round.

Measured with the probe above: **7 summarize calls in a 6-round turn** (one pre-loop, one per round) against 1 on staging. `max_tool_rounds` defaults to **25**, so worst case is ~26 real LLM round-trips per turn, with no backoff. The comment above all six call sites still reads *"gated to at most once per turn"*, which is no longer true on the failure path.

The case that makes it more than waste is a **429**: compaction fails *because* we're being rate-limited, and this makes us hit the limiter ~25× harder for the rest of the turn. The failure is correlated with its own trigger — high context is what throttles us and what triggers compaction — so this isn't a hypothetical pairing.

Smallest fix that keeps the good behaviour (a genuine transient blip still gets a second chance) while bounding the bad one:

```python
# in __init__ / turn reset, alongside _compacted_this_turn
self._compaction_failed_this_turn = False
```

```python
if (
    not self._compacted_this_turn
    and not self._compaction_failed_this_turn
    and llm_response.usage.context_pressure > self._context_pressure_threshold
):
    compacted = await self._summarize_history()
    self._compact_scratchpads()
    if compacted:
        self._compacted_this_turn = True
        yield StreamContextCompacted(...)
    else:
        self._compaction_failed_this_turn = True
```

Whatever shape you prefer, the six "at most once per turn" comments should end up describing what the code does.

## 2. The caller half of the diff has no test

Replacing all six `if compacted:` with `if True:` — i.e. reverting the entire caller-side change — leaves **143 tests passing**.

Worth knowing *why*, because it looks like it's covered: `TestContextCompaction` in `test_chat.py` stubs `session._summarize_history = AsyncMock()`, and an `AsyncMock`'s default return value is a truthy `MagicMock`. So those tests behave identically whether the caller honours the return value or ignores it entirely. They read like coverage of exactly this behaviour and are structurally unable to test it.

What's missing is one caller-level test: summarizer raises → `_compacted_this_turn` stays `False`, **no** `StreamContextCompacted` is emitted, and (with the fix above) compaction is attempted at most twice in a multi-round turn.

## 3. Done-when item 3 is still open

The ticket's third bullet — route the summariser through `generate_with_truncation_retry` (`anton/core/llm/structured.py`, `DEFAULT_STRUCTURED_BUDGETS = (2048, 4096)`) — isn't in this PR. Fine to defer, but let's say so on the ticket rather than close it implicitly. It also partly subsumes #1: if the summariser retried internally with a bigger budget, the outer per-round retry would matter less.

## On your three points from chat

**"In cowork server the history is restored on next turn" — correct, and I verified it.** `_persist_history_compaction` returns early when `session.last_compaction is None`, and `_last_compacted_count` is only assigned inside the `try`, so a failed compaction persists nothing and `_seed_history` rebuilds from the DB rows next turn. Two caveats: it bounds *persistence*, not *damage* — the rest of that turn (up to 25 rounds) runs against the placeholder, and whatever the agent does or says in those rounds **is** persisted as real rows; and it doesn't cover the anton CLI/library path, which is what your PR title calls out.

**"The Langfuse search returned nothing" — fair, and I gave it more weight than I would have.** I didn't search for the failure string first; I searched for its sibling. Success and failure write into the same field of the same payloads, so I checked whether the successful marker shows up at all: in 120 generations since Aug 3, **51 carry `[COMPACTED CONTEXT — REFERENCE ONLY]` and 0 carry `summarization failed`**. That makes your null a real signal rather than a possibly-blind instrument — the search that finds 51 successes would find failures.

One limit: that sample contained exactly **1** actual summarize call. The 51 markers are replays of one summary, not 51 attempts. So it says "not observed", which it does honestly, but it can't bound a per-call failure rate. If we want that number, the measurement is failures per summarize call over a longer window.

**"It should fail only on transient network error, a rare case" — this is the one I'd push back on.** `except Exception` catches more than that, and the ones that matter are the ones the code deliberately **never retries**:

- **429 rate-limited** → `TransientProviderError(session_backoff=False)` (`provider.py:516-523`), explicitly kept out of the retry loop so a billing 429 can't loop. Raises straight into this `except`.
- **5xx** → also `session_backoff=False`, fail fast.
- **402 / `insufficient_quota`** → the mapper returns `None`, so the raw SDK error propagates.
- **empty 200** → `raise_on_empty_response` **now raises** where it previously handed back empty content. That's your own #297, merged this week — it made this path strictly more reachable, which is a nice illustration that "rare" isn't stable across releases.

And the detail that changes the risk profile most: `summarize()` runs on the **router role**, not the model the user picked. `ROUTER_MODEL_DEFAULTS["minds_cloud"] = "kimi"` (cowork-server `common/settings/app_settings.py:79`), and the one summarize call I found in Langfuse ran on **`kimi-k3` with `max_tokens=2048`**, asked for a seven-section structured record. Kimi is in the family we already have tickets about for deterministic 400s and for narrating past tight budgets (ENG-1081/1082). A deterministic model-side failure isn't rare — it's *every* compaction for that user, which is exactly the case where #1 turns into ~26 calls a turn.

So I'd land this (the discard is unrecoverable and silent; carrying history you can't compact is recoverable and the overflow ladder already exists for it), and I'd treat "not observed in traffic" as a reason to lower the priority rather than to leave the discard in.

Happy to approve once #1 and #2 are in — both are small. #3 can be a follow-up if you'd rather.

ea-rus:
Addressed the review:

**#1 — failed compaction no longer re-attempted every round.** Added `_compaction_failed_this_turn` (reset per turn alongside `_compacted_this_turn`). All four proactive gates now check it and set it in the `else` branch, so a failed proactive compaction is attempted at most once per turn instead of once per tool round. Worst case drops from ~26 summarize calls/turn to ≤2 (one proactive + at most one reactive on a real overflow). The misleading "gated to at most once per turn" comments now read "…and not re-attempted after a failure"; the full rationale (429 correlated with its own trigger, `max_tool_rounds` with no backoff) lives at the flag definition.

**#2 — caller-side change now has teeth.** Added two caller-level tests in `TestContextCompaction` that drive `_summarize_history` to return `False` (not the truthy `AsyncMock` default that made the old tests unable to tell honouring the return from ignoring it):
- `test_failed_compaction_emits_no_event_and_closes_gate` — False → no `StreamContextCompacted`, `_compacted_this_turn` stays False, `_compaction_failed_this_turn` set.
- `test_failed_compaction_not_retried_every_round` — overflow + high-pressure retry → 2 summarize calls, no event.

Mutation-checked: reverting one `if compacted:` gate to unconditional fails the first test (the old truthy-mock tests stayed green).

**#3 — deferred to ENG-1291, not dropped.** Routing the summariser through `generate_with_truncation_retry` is a distinct change (the summariser uses plain-text `complete()`, the ladder wraps structured generation) and only covers the truncation failure mode — 429/5xx/quota are already bounded by #1. Noted on ENG-1274 and tracked in ENG-1291 with the related input-side `old_text[:8000]` redesign.

Also folding the "failed compaction still emitted 'older history has been summarized'" fix into the PR description, as suggested.

alecantu7 (commented):
All three items addressed, and I verified each rather than reading the summary. One item left, then I'll approve.

## Verified good

**#1 works, and better than your estimate.** Re-ran the same probe from my review — 6 tool rounds, persistently failing summarizer:

```
before: 7 summarize calls    after: 1
```

You said ≤2; it's 1 here because the reactive path never fired without a real overflow. Which is the right design and I checked it specifically: the guard gates only the four *proactive* pressure checks and leaves the reactive `ContextOverflowError` recovery ungated, so a genuine overflow can still attempt compaction after a proactive failure.

I also went looking for the classic bug shape with a new flag — reset in fewer places than the flag it shadows, so one failure disables compaction for the rest of the session. It isn't there: `_compaction_failed_this_turn = False` sits immediately after `_compacted_this_turn = False` at all three sites (init, `turn`, `turn_stream`), no site resets one without the other.

**#2 is properly closed.** Mutation: all six `if compacted:` → unconditional now fails **2 tests**. The same mutation left 143 green before. And you named the actual cause — the truthy `AsyncMock` default — rather than just adding tests around it.

**#3 — ENG-1291 is a better ticket than the item I raised.** Folding in that `old_text[:8000]` keeps the *head* and drops the recent turns, and generalising `[:N]`-on-tail-meaning across ENG-836 and ENG-1116, is the more useful framing. Sequenced behind this one correctly.

## The one item: the new guard is load-bearing but nothing pins it

Same defect class you just fixed in #2, one item later.

Mutation — remove `and not self._compaction_failed_this_turn` (restoring per-round retry), change nothing else:

```
my probe (6 tool rounds):   1 call  →  7 calls     ← the guard is doing real work
test_failed_compaction_not_retried_every_round:     passes either way
```

Why it can't fail: the test drives overflow → one high-pressure retry, so the turn has exactly **two** compaction opportunities, one reactive and one proactive. The guard only gates proactive checks, and there's a single proactive check in that scenario — so `call_count == 2` holds with or without it. The `_compaction_failed_this_turn is True` assertion survives too, because it observes the *write* (`= True`), not the suppression (`and not ...`).

So the test names the behaviour in a scenario that structurally cannot exhibit it. That's the same shape as the truthy-mock problem: the name is right, the assertion can't go red for that reason.

What would catch it is a **tool loop**, which is where the 26-calls-per-turn case actually lived. Roughly the probe I used:

```python
async def test_failed_compaction_not_retried_across_tool_rounds(self):
    """N tool rounds above the pressure threshold must still yield at most
    one proactive attempt — the guard, not the once-per-turn flag, is what
    stops it (ENG-1274 #1)."""
    rounds, calls = 6, 0
    async def _plan_stream(**kwargs):
        nonlocal rounds
        rounds -= 1
        if rounds > 0:
            yield StreamComplete(response=LLMResponse(
                content="", tool_calls=[ToolCall(id=f"t{rounds}", name="no_such_tool", input={})],
                usage=Usage(context_pressure=0.9), stop_reason="tool_use"))
        else:
            yield StreamComplete(response=LLMResponse(
                content="done", usage=Usage(context_pressure=0.9), stop_reason="end_turn"))
    ...
    assert session._summarize_history.call_count <= 2
```

The load-bearing part is that the tool loop runs several rounds each reporting pressure above the threshold — that's what turns into one call per round without the guard. `<=` rather than `==` so it doesn't become a tripwire on unrelated recovery-path changes.

Also 4 commits behind `origin/staging`.

Nothing else from me — ping when that's in and I'll approve.

alecantu7 (approved):
Pinned. Approving.

Verified by mutation rather than by reading it — removing `and not self._compaction_failed_this_turn` (restoring per-round retry, changing nothing else) now fails `test_failed_compaction_not_retried_across_tool_rounds`. The same mutation left the whole suite green before this commit. The caller-gating mutation still trips two tests, so #2 stays closed.

Two things you did beyond the ask, both worth noting:

**You renamed the old test instead of leaving it.** `test_failed_compaction_overflow_plus_proactive_bounded` now says what it actually checks, and its docstring points at the across-rounds test as the one that pins the guard. That matters more than the new test does: the old name claimed a property it couldn't verify, which is how a reviewer skims past a gap. Fixing the name closes the same failure mode the truthy-`AsyncMock` problem did.

**The `filterwarnings` carries its reason.** Marking the never-awaited coroutine as a test-double artifact — the real provider awaits it — rather than silencing it bare is the difference between a suppression someone can evaluate later and one they have to re-derive. I hit the same warning building the probe.

`<=` over `==` is the right call for exactly the reason in your docstring.

Full picture on this PR, for the record: history is preserved on a failed summarisation (the data-loss fix), the false "older history has been summarized" event no longer fires on failure, a persistent failure costs at most two summarize calls per turn instead of one per tool round, and both halves are now pinned by tests that fail when the behaviour is removed. Done-when #3 is deferred to ENG-1291 with a written rationale on the ticket, which is the right shape — it's a design change, not a line.

2 commits behind `origin/staging`; worth a refresh before merge so what lands is what was reviewed. Merge is Konstantin's call.

PR #311, opened 2026-08-05
Fix duplicated wording in Anthropic/OpenAI 404 error messages (ENG-1139)

StpMax:
## Summary

- 404 responses from the raw Anthropic provider previously had no dedicated classification and fell through to a generic `ConnectionError("Server returned 404 — the LLM endpoint may be temporarily unavailable. Try again in a moment.")`. Because this exception is not one of the "curated" types the session loop fails fast on, it exhausted the auto-retry budget and then got wrapped a second time by the turn loop's generic fallback (`"An unexpected error occurred: {e}. Please try again or rephrase your request."`), producing a doubled, confusing message with a stray double period and two redundant "try again" instructions.
- Extracted the existing OpenAI-compatible 404 classification logic (model-not-found vs. endpoint-misconfiguration) out of `openai.py` into a shared `classify_404()` helper in `provider.py`, extended it to recognize Anthropic's own `error.type == "not_found_error"` signal, and wired it into `anthropic.py`'s status-error mapper.
- A 404 is now consistently raised as `ModelUnavailableError` (permanent, "switch models" remedy) or `EndpointConfigurationError` (permanent, "check endpoint config" remedy) across both providers, instead of being misclassified as a transient, retryable failure.
- Added `EndpointConfigurationError` to the session turn loop's fail-fast exception tuple, alongside `ModelUnavailableError`: without this, an endpoint-misconfiguration 404 (the sibling outcome of `classify_404()`, not model-specific) would still get auto-retried against the same doomed endpoint and double-wrapped by the generic fallback prose — reproducing the exact bug this PR fixes, just for that branch (caught in review).
- Added regression tests pinning the new Anthropic 404 behavior and asserting the model name/reason isn't duplicated in the final message.

## Why

ENG-1139 reported that a 404 from the LLM endpoint surfaced as a confusing, duplicated error message to the user. Root cause: the Anthropic provider had no 404-specific handling, so the error was treated as transient/retryable and got double-wrapped by the turn loop's catch-all error text before reaching the user, instead of being classified as the permanent, actionable error it actually is. The same double-wrap risk existed for the endpoint-misconfiguration outcome of the shared classifier, since the session loop's fail-fast list didn't know about `EndpointConfigurationError` either.

## Changes

- `anton/core/llm/provider.py`: add `classify_404()` — shared model-vs-endpoint 404 classifier.
- `anton/core/llm/openai.py`: replace inline 404 classification with a call to `classify_404()` (behavior-preserving).
- `anton/core/llm/anthropic.py`: add a 404 branch (previously missing) that calls `classify_404()` before falling through to the generic transient/unavailable mapping.
- `anton/core/session.py`: add `EndpointConfigurationError` to the turn loop's fail-fast tuple (alongside `TokenLimitExceeded`, `ModelUnavailableError`) so both `classify_404()` outcomes skip auto-retry and the generic error-wrap, consistent with how `chat.py` already treats both types as "permanent, use `setup`" rather than "retry".
- `tests/test_status_error_mapper.py`: new test covering the Anthropic 404 → `ModelUnavailableError` mapping.

## Test plan

- [x] `pytest tests/test_status_error_mapper.py -q` — 38 passed
- [x] `pytest tests/ -q` — 1500 passed, 28 skipped (full suite)
- [x] Manually reproduced via `anton chat` with `ANTON_PLANNING_PROVIDER=anthropic` + a nonexistent model — confirmed the old doubled/misleading message is gone and the new message is a single, permanent, actionable `ModelUnavailableError`

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -25,6 +25,7 @@ from .provider import (
     ToolCall,
     TransientProviderError,
     Usage,
+    classify_404,
     classify_transient,
     compute_context_pressure,
     wallet_denial_code,
@@ -49,6 +50,9 @@ def _raise_for_status_error(
     - 402/429 with an M3 gate wallet code (``wallet_empty`` /
       ``included_allowance_exhausted``, body or X-MindsHub-Reason header)
       → TokenLimitExceeded (ENG-1169). Code-exact: BYOK 402s stay generic.
+    - 404 → ModelUnavailableError or EndpointConfigurationError via the
+      shared ``classify_404`` (ENG-1139) — permanent, non-duplicated copy
+      instead of the generic "temporarily unavailable, try again".
     - overloaded/api_error (incl. the mid-stream HTTP-200 case), 5xx, plain 429
       → TransientProviderError (retryable — see ENG-673).
     - anything else → the generic "temporarily unavailable" ConnectionError.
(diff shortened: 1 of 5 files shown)


ea-rus:
Cross-consumer sweep: `session.py` has two sites catching curated failures; this PR updates only one.

- `session.py:2643` (fail-fast) — updated ✅
- `session.py:2759` (post-retry-exhaustion summary `plan_stream`) — still only `(TokenLimitExceeded, ModelUnavailableError)`; anything else → generic `"An unexpected error occurred... try again"` wrap.

`ModelUnavailableError` is there because the summary call can re-fail on a gate — same applies to `EndpointConfigurationError`: a 404 during summary → swallowed → double "try again" prose (the bug this PR fixes) + the server's endpoint-config card never fires.

Fix: add `EndpointConfigurationError` to the 2759 tuple too, symmetric with 2643. Low severity (narrow path), but one line and same bug class — worth doing in this PR.

ea-rus (approved):
Added not-blocking comment

PR #312, opened 2026-08-05
[ENG-578] Watchdog and liveness heartbeat

tino097:
https://linear.app/mindsdb/issue/ENG-578/anton-inactivity-watchdog-kills-throttled-batch-loops-forces-per-item
 
 - Added heartbeat and the scratchpad signals "alive" every 10s while a cell runs
- Watchdog improvement so no more killed cells if the code waits, sleeps, or blocks (throttled email loops now run in one cell)
- Slow package auto-installs no longer killed mid-install
- Killed/crashed cells keep the output they printed before dying
- Kill messages now say the real cause: worker dead vs out of time
- Postkill advice and saved lessons match the cause no more "make it smaller" for a cell that was just waiting
- Updated scratchpad tool description to the new rules
- Dead/frozen workers still killed after 30s total time limit unchanged


alecantu7 (commented):
Reviewed against the ticket's fix list and the T2 comment. Leading with the credit because it's earned: **all three of the false statements ENG-578 demanded be corrected are corrected** — the kill message, the `compute_timeouts` comment (now explicitly says cell prints go to a per-cell `StringIO` and never reach the timer), and the tool description's "hard timeout of 120 seconds / break the work into smaller cells". The heartbeat is a real OS thread writing to `_real_stdout`, which is exactly the right way around the rebinding the probe exposed. And T2's model-free test pair exists precisely as specified (`test_liveness_kill_does_not_claim_too_heavy` / `test_total_budget_kill_keeps_too_heavy`). 120 tests pass. The `shipped`-before-`buf` ordering in the salvage swap, with its "worst case must be a duplicate chunk, never a skipped one" note, is the kind of care this area needed.

The partial-stdout salvage isn't in the ticket at all and is a genuine addition — keeping the *tail* rather than the head after a kill is the right call, since "sent 47/50" is the part that tells you which side effects happened.

## 1. The heartbeat also disables hang detection — and that isn't stated anywhere

Because the beat is unconditional and independent of the main thread, the inactivity timer no longer distinguishes *waiting* from *wedged*. Probed against the real runtime at a 1s silence window and an 8s total budget:

```
deliberate sleep(4)  [the fix]        4.0s  err=(none)                 ← works
busy infinite loop   [the hang]       8.0s  Cell timed out after 8s total
double lock acquire  [deadlock]       8.0s  Cell timed out after 8s total
```

Before this PR both of the bottom two died at the silence window. Three consequences, in order of how much they'd cost:

**(a) The kill message asserts a cause again.** It now reads *"the process looks dead or wedged"*. A wedged process no longer reaches that timer — only a dead/EOF'd one, or one whose threads are all blocked below Python (a C extension holding the GIL). That message was rewritten specifically to stop naming a cause it couldn't know, so it's worth not reintroducing one. "…without a liveness signal — the worker process appears to have died" would be true of what actually gets here.

**(b) The inversion T2 removes has moved rather than gone.** A deadlocked or spinning cell is killed on the total budget → `"Cell timed out after Ns total"` → `_select_resilience_nudge` → `SCRATCHPAD_TIMEOUT_NUDGE` → *"the work is too heavy… make the next cell smaller."* That's the same wrong lesson for the same wrong reason, now arriving via the other timer. The wedge case is the one where "too heavy" is most misleading, because shrinking the batch does nothing to a deadlock.

Worth considering a third branch, or at least a signal to split on: a cell that produced steady output and then hit its budget genuinely was too heavy; one that produced nothing at all and hit its budget was probably wedged. The salvage buffer already tells you which — `self._salvage` being empty at the total-timeout is a decent wedge signature, and it's free.

**(c) `cell_total_max` is now the only hang backstop, and it defaults to off.** Latency to detect a wedge goes from ≤60s to the total budget: 120s with no estimate, `estimate × 2` with one. So `estimated_execution_time_seconds=600` buys a deadlocked cell 20 minutes, and the value comes from the agent — the same actor whose mistakes the bound exists to contain. The `utils.py` comment describes `cell_total_max` as a backstop "for a runaway that keeps producing output forever (which the inactivity cap can't catch)"; after this PR the inactivity cap can't catch a silent hang either, so that sentence understates its new job. I'd give it a non-zero default in the same PR that takes its safety net away.

None of this is an argument against the change — deliberate waits had to become survivable, and this is the right mechanism. It's an argument for saying out loud what it trades.

## 2. `detect_kill_loop` defaults an unreasoned kill to the harmful lesson

`is_liveness` reads `detail["reason"]`, and anything missing or empty is falsy → the event lands in `heavy`. `crosses_threshold(heavy)` is then evaluated first and returns immediately. So:

- a kill event whose `reason` never got populated → "too heavy"
- a turn of 2 budget kills and 20 liveness kills → "too heavy"

Given the whole point is that this rule is the expensive one to write into durable memory, I'd rather it defaulted the other way, or compared group sizes instead of short-circuiting on `heavy`. The docstring's "a mixed turn resolves to it as the safer rule to write down" is the call I'd question — ENG-578 measured ~$30 a session from that rule specifically.

## 3. A third coupling the lockstep comment doesn't name

`local.py` warns that the wording is load-bearing for the nudge and the ACC detector. There's a third link: `observe_scratchpad_cell` stores `reason=err[:120]`, so ACC routing depends on `"liveness"` surviving that truncation. It does — index 32 of 225, verified — but a future prefix on the message ("Error: ", a wrapper) could push it past 120 and silently return every kill to the too-heavy lesson with all tests green. Worth naming the slice in that comment.

## 4. T2 item 4 is still open

The ticket's fourth T2 item was deciding whether to clean up rules already written into users' durable memory. Real users have "two kills in a turn mean the approach is too heavy" stored and re-read on every later scratchpad call. This PR fixes the writer, not the written. Fine to defer, but it should be an explicit decision on the ticket rather than fall off with the PR.

Also 2 commits behind `origin/staging`.

Nothing above is a defect in what's here — (1) is a tradeoff that needs stating and a default changing, (2) is a tiebreak I'd flip, (3) and (4) are follow-through. Happy to approve once (1c) has a default and (a) is reworded.

tino097:
Review round addressed in 89baaa2c (branch also rebased onto origin/staging @ 560d29e5).

**1(a) — reworded.** The silence-kill message now names only what actually reaches that timer: the worker process died, or a native call is stuck holding it below Python. The same overstated claim in `SCRATCHPAD_STUCK_NUDGE` is corrected too.

**1(b) — implemented, with one adjustment to the suggested signal.** Empty-salvage-at-budget ≈ wedge has a false positive: a legitimately heavy *silent* cell (a multi-minute `df.merge` prints nothing) is indistinguishable from a deadlock by that signature. So the third branch says the ambiguity out loud instead of guessing: a budget kill with zero salvaged output reads *"…without producing any output — either a call is stuck or the work is heavier than estimated"* and routes to a new `SCRATCHPAD_SILENT_TIMEOUT_NUDGE` that prescribes instrumentation (realistic estimate, prints, progress(), per-call timeouts) and only on a silent repeat says treat it as stuck. A budget kill that *was* producing output keeps the plain message and the too-heavy nudge — the one case where that diagnosis is genuinely right.

**1(c) — `cell_total_max` now defaults to 3600.** Reasoning for the number: it bounds a beating-but-stuck cell (deadlock/spin — which now heartbeats like a working cell) at 1h, while leaving room for the throttled-campaign shape this PR exists to enable (50 sends × 30s ≈ 25min). Estimates under ~30min are unaffected by the cap. If you want it tighter, it's one line + one test.

**2 — majority-wins.** `detect_kill_loop` compares group sizes; ties go to liveness (a wrong reset-and-retry costs one retry, a wrong "smaller batches" costs the measured per-item pattern). Unreasoned kills and silent-budget kills count toward **neither** class — for a rule written into durable memory, no rule beats a wrong one. Your 2-budget+20-liveness example now writes the liveness lesson; 20-budget+2-liveness still writes too-heavy.

**3 — the `err[:120]` slice is now load-bearing by test, not just comment.** The lockstep comment names all three consumers, and tests assert the routing keywords ("liveness", "without producing any output") sit inside the first 120 chars of their kill messages — a future prefix breaks a test instead of silently reverting every kill to the too-heavy lesson.

**4 — deferred to the ticket as an explicit decision.** Cleaning up already-written rules touches user-owned memory files; that call is Konstantin's to make on ENG-578 rather than something this PR should do silently.

Suite: full run green except the pre-existing `test_namespace_survives_restart` failure, which fails identically on untouched staging (ENG-1124 function-persistence — separate issue, will flag on the ticket).

alecantu7 (approved):
All four addressed. Approving — and one of them you were right to reject.

**(b) — your objection to my signal is correct and the result is better than what I asked for.** I proposed empty-salvage-at-budget as a wedge signature. You're right that a legitimately heavy *silent* cell — a multi-minute `df.merge` that prints nothing — is indistinguishable from a deadlock under it, so the heuristic would confidently mislabel exactly the case it was meant to catch. Saying the ambiguity out loud instead ("either a call is stuck or the work is heavier than estimated; the runtime cannot tell which"), prescribing instrumentation, and only escalating to "treat it as stuck" on a silent *repeat* is the honest version. Keeping the too-heavy nudge for a budget kill that *was* producing output is the right split — that's the one case where the diagnosis is genuinely earned.

**(a) — the message is now precisely true.** "or wedged" gone, replaced by "the worker process died, or a native call is stuck holding it below Python." That matches what my probe actually showed reaches that timer: a Python-level deadlock heartbeats like a working cell and never gets there, so the old wording was claiming a cause it couldn't observe. This one can.

**(c) — checked the number against the use case this PR exists for**, not just that a default appeared:

```
estimate=1500s (50 sends × 30s)  -> total 3000s   uncapped ✔
estimate=1800s                   -> total 3600s   at the cap
estimate=2400s                   -> total 4800s   clipped to 3600
```

So the throttled-campaign shape fits comfortably and clipping only starts above a 30-minute estimate. The bound is real and the thing it exists to permit still runs. Fine at 3600 from my side.

**(2) — my actual complaint is closed, not just the example.** The fix I cared about is that a missing `reason` now classifies as `"ambiguous"` instead of falling into `heavy` — the old code defaulted an unclassifiable kill to the expensive lesson. Ambiguous counting toward neither class is the right call for something written into durable memory. Verified the arithmetic: 2 heavy + 20 liveness now writes liveness (it wrote too-heavy before), 20 heavy + 2 liveness still writes too-heavy, ties go to liveness.

**(3) — now enforced rather than described.** `assert "liveness" in cell.error[:120].lower()` on the silence path, the same for `"without producing any output"` on the silent-budget path, plus the negative assertion that the budget message carries no "liveness" so the two can't cross-contaminate. I prefixed the kill message to confirm it bites, and it fails a test. That's the difference between a comment asking future readers to be careful and a tripwire.

**(4)** Right call to leave with Konstantin on the ticket — it touches users' own memory files, and doing it silently inside a code PR is how you'd get it wrong.

35 tests green.

## One thing to re-check, not blocking

The "pre-existing `test_namespace_survives_restart` failure" doesn't reproduce here. It passes on this branch **and** on untouched `origin/staging` — in isolation and across the full `tests/test_scratchpad.py` (91 passed each way). So there's nothing attributable to this PR either direction; it looks environmental, or stale from before the rebase onto `560d29e5`. Worth confirming before it gets carried around as a known baseline failure, since a real one is worth its own ticket and a phantom one costs the next person an afternoon.

alecantu7 (commented):
## Review — validated against `89baaa2`

Reviewed specifically for the two things ENG-578 flags as part of the fix: the `print()`-vs-`progress()` contract, and the cause-aware kill lesson. **Both are correct. No merge blocker.** Two notes below, one recommendation and one test gap, neither requiring code changes to land.

Executed: `test_scratchpad_watchdog_contracts.py` + `test_scratchpad.py` + `test_resilience_nudge.py` → **126 passed** in an isolated worktree. Base verified fresh (`origin/staging` = `560d29e`), branch 0 behind.

### The two contract fixes are right

**The `compute_timeouts` comment no longer asserts the false version.** It used to claim *"stdout/progress() reset this window"*; the 2026-07-28 probe disproved that (`print()` killed at 30.0s, `progress()` survived 45.3s, daemon-thread `print()` heartbeat also killed — `sys.stdout` is rebound to a per-cell `StringIO` before `exec`). The new comment states the real mechanism, and the fix matches it: `_heartbeat_loop` writes to `_real_stdout` under `_wire_lock`, which is exactly why the old daemon-thread workaround died and this one doesn't. `tool_defs.py` and `prompts.py` are corrected too, with `test_description_does_not_claim_progress_is_survival_critical` / `test_system_prompt_does_not_mandate_splitting_cells` pinning them.

**`detect_kill_loop` being cause-aware is the right call, and three details stand out.** Ambiguous kills writing *no* rule — *"no rule beats a wrong one"* — is correct for something that persists in durable memory and gets re-read on later calls. Ties resolving toward liveness with the cost argued rather than asserted (one wasted retry vs. the ENG-578 per-item pattern at ~$30/session) is the right asymmetry. And lockstepping the phrases across the `local.py` kill messages, the nudge routing, and the lesson classifier addresses the obvious failure mode, which is those three drifting apart.

---

### Note 1 — the heartbeat changes which timer catches a hang (recommendation, not a defect)

There are two timers answering different questions: **inactivity** asks *"is this cell producing anything?"*, **total** asks *"is this longer than the work should take?"* The heartbeat thread is unconditional — a plain `while not stop.wait(10)` loop with no knowledge of whether the cell is progressing — so inactivity now effectively means *"is this process alive?"* And since the beat interval (10s) is below the silence window (30–60s), the beat always wins: inactivity can no longer fire for a live process.

So hangs fall through to the total timer, which was sized for legitimate runtime rather than hang tolerance. The settings comment already says this (*"a userland deadlock or infinite loop beats like a working cell, and this ceiling is the only thing that ends it"*) — this note is only about the consequence for the user.

**Calibrating it, since `3600` is a ceiling and not a duration** — `total = min(max(est*2, est+30), 3600)`:

| declared estimate | wedge dies after |
| -- | -- |
| omitted | 2 min |
| 30s | 1 min |
| 120s | 4 min |
| 300s | 10 min |
| 1500s | 50 min |
| ≥1800s | 60 min |

For ordinary cells that's ~60s → 60–120s, i.e. negligible. The exposure is largest exactly on long-batch cells — the motivating workload (50 throttled sends × 30s ≈ 25 min → est ≈ 1500) sits at ~50 min. Mitigating factors: no LLM calls happen while a cell runs, so this costs wall-clock rather than tokens, and repeats are bounded by ENG-1276's breaker at five kills.

**The harm is that it's silent, not that it's long — so I'd suggest surfacing the heartbeat rather than lowering the ceiling.** See the inline comment on `local.py:887`: a liveness signal arrives every 10s proving the worker is alive, and it's discarded. Lowering `cell_total_max` would re-break the long batch this PR exists to enable.

### Note 2 — the new stop path has no behavioural test

`cell_total_max` went from `0` (off) to `3600`, which makes it **load-bearing for the first time** — it is now the only thing that ends a hang. Its coverage is `assert CoreSettings().cell_total_max == 3600` plus a clamp test: value assertions, not behaviour. The closest behavioural test, `test_beatless_silent_cell_still_killed`, disables the heartbeat and checks the *old* inactivity path still works. Nothing exercises **heartbeat on + cell wedged + total budget ends it**.

That path's downstream behaviour is what this PR spent its care on: a total-budget kill classifies as `heavy` and teaches "make the next cell smaller", which for a genuine wedge is also wrong advice — a smaller batch doesn't fix a deadlock — and it lands in durable memory. Given the deliberate choice to have ambiguous kills write no rule, this gap sits right next to reasoning already done here.

Cheap to close: set `cell_total_max` low (~2s), run `while True: pass`, then assert the cell is killed, assert the kill message, and assert which lesson `detect_kill_loop` produces. The lesson assertion is the valuable one — it pins the classification on the path that now catches every hang.

### One caveat worth a line in the PR body

The heartbeat needs the GIL. A non-releasing C call or a pathological `re.match` blocks the thread, so a *genuinely active* cell can still take a **liveness** kill — and the new lesson then says retry the SAME cell unchanged, which reproduces it exactly. Bounded by the breaker at five kills, and `test_blocking_call_survives` covers the GIL-*releasing* case explicitly, so this is a caveat rather than a defect.

### Merge-order note (#315)

[#315](https://github.com/mindsdb/anton/pull/315) is the ENG-578 sibling and is **not** stacked — both branch from `560d29e`, per convention 3, and both are approved. They share exactly one file, `anton/core/session.py`, in different regions (this PR in nudge routing, #315 at the system-prompt build ~L1128). Merge order is semantically free; **whichever lands second wants a mergeability re-check on `session.py`.** Separately: these two cover ticket fixes 1, 3, 5 and 7 — fix 2 (gateway prompt caching + cache telemetry) and fix 6 (ENG-626 tool-description bloat) remain open, so merging both isn't ticket closure.

<sub>Reviewed on a session-unique ref and cleaned up after; the in-flight `wt312b` worktree and `pr-312b` branch were left untouched.</sub>

anton/core/backends/local.py
```diff
@@ -816,24 +850,61 @@ async def _read_result(
...
-                    "stdout": "",
+                    "stdout": "".join(self._salvage),
                     "stderr": "",
                     "error": "Process exited unexpectedly.",
                 }
                 return
 
             line = raw.decode("utf-8", errors="replace").rstrip("\r\n")
 
+            if line.startswith(HEARTBEAT_MARKER):
```
alecantu7, line 887: **Suggestion (Note 1's actual fix): the liveness signal exists here and is thrown away.**

```python
if line.startswith(HEARTBEAT_MARKER):
    # Liveness only: arrival already re-armed the readline timer.
    continue
```

Every 10 seconds this proves the worker is alive, and it's dropped. Combined with the unconditional heartbeat, that means a wedged long-estimate cell can run for tens of minutes (≈50 min at est≈1500s) with **no signal to the user at all** — where before this PR it failed in ~60s.

The duration itself is defensible: it scales with the agent's declared estimate and the ceiling exists to bound it. What isn't defensible is the silence, and the information needed to end it is already at this line.

Surfacing something like *"still running (12m)"* on a heartbeat tick would turn tens of minutes of silence into visible progress the user can act on — and the manual escape hatch now works properly, since anton#309 fixed user Stop landing in `error` instead of `cancelled` (`asyncio.CancelledError` is the shape `RunHandle.cancel()` produces). So a visible long-running cell becomes a user decision instead of a hang.

This is a suggestion rather than a finding — it's additive, out of scope if you'd rather keep this PR to the watchdog contract, and worth its own ticket if so. Flagging it here because this is the one line where the fix belongs, and `continue` makes it easy to miss that the signal was ever available.

PR #315, opened 2026-08-05
[ENG-578] Scratchpad discovery on startup

tino097:
[ENG-578:](https://linear.app/mindsdb/issue/ENG-578/anton-inactivity-watchdog-kills-throttled-batch-loops-forces-per-item) scratchpad discovery at turn start |

- Added a small "Workspace state" block the agent sees at the start of each turn
- It lists the scratchpads that already exist for this conversation (with "active" or snapshot age) and the files in the project root
- Fixes the blind cold start: the agent stops inventing new scratchpad names and rebuilding data it already has (measured ~$30 of wasted rebuilding in one session; 12+ pad names including typos)
- Only agent-created scratchpads are shown — system pads never appear
- Small by design: max 10 pads and 30 files, hidden files excluded, a few hundred tokens
- Placed after the prompt's cached prefix, so it never breaks prompt caching (pinned by a test)
- If anything fails while building the block, it's simply skipped — a turn can never break because of it

### DIFF
--- a/anton/core/backends/manager.py
+++ b/anton/core/backends/manager.py
@@ -59,6 +59,32 @@ class ScratchpadManager:
         """Read-only view of the active scratchpad runtimes."""
         return self._pads
 
+    @property
+    def workspace_path(self) -> Path | None:
+        """Project root this manager serves — discovery needs it for listings."""
+        return self._workspace_path
+
+    def pad_snapshot_mtime(self, name: str) -> float | None:
+        """Epoch mtime of `name`'s ENG-1124 snapshot, or None when unknowable.
+
+        None covers every non-answer (unscoped session, no snapshot dir,
+        never persisted): discovery renders those pads without an age rather
+        than failing the block.
+        """
+        if not self._session_id or not name:
+            return None
+        try:
+            from anton.core.backends.local import default_venvs_base, snapshot_file
+
+            path = snapshot_file(
+                default_venvs_base(self._workspace_path), self._session_id, name
+            )
+            return path.stat().st_mtime if path is not None else None
+        except OSError:
+            return None
+        except Exception:
+            return None
+
     def _agent_pads_file(self):
         """Where this conversation's agent-chosen pad names are recorded, or None.
 
--- a/anton/core/llm/prompt_builder.py
+++ b/anton/core/llm/prompt_builder.py
@@ -138,6 +138,7 @@ class ChatSystemPromptBuilder:
         self_awareness_context: str = "",
         datasource_context: str = "",
         skill_store: "SkillStore | None" = None,
+        workspace_context: str = "",
     ) -> str:
         visualizations_section = self._build_visualizations_section(
             proactive_dashboards=proactive_dashboards,
@@ -199,6 +200,10 @@ class ChatSystemPromptBuilder:
         )
         if memory_context:
             prompt += memory_context
+        # Workspace discovery (ENG-578) is volatile too — pads and files
+        # change between turns, so it must never sit in the cached prefix.
+        if workspace_context:
+            prompt += workspace_context
 
         return prompt
 
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -74,6 +74,7 @@ from anton.core.tools.tool_defs import (
 )
 from anton.core.interaction.selection import SelectionElicitor
 from anton.core.utils.scratchpad import (
+    build_workspace_discovery_context,
     prepare_scratchpad_exec,
     format_cell_result,
     observe_scratchpad_cell,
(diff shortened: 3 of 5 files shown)


alecantu7 (commented):
Reviewed against fix #3 in ENG-578 and the two comments on it. This is clean — the thing I went looking for is the thing it gets right.

## The part that matters

Fix #3's trap was recorded on the ticket: the `confirm_new_scratchpad` guard *existed* and fired **0 times across 676 cells**, because its `seen` set lived on the `ChatSession` and cowork-server rebuilds that per user message — so it was always empty exactly when it was needed. A discovery block built the same way would have been dead on arrival in the hosted and desktop products while passing every test locally.

This reads `manager.agent_pads()`, which is disk-backed (`_agent_pads.json` under the conversation's snapshot dir), so it survives the per-turn rebuild. And it's `agent_pads()` rather than `self._pads` or the snapshot listing, so the artifact backend's slug pad can't leak in and can't be presented to the agent as something it created. Both of those are the correct choices for non-obvious reasons.

Two other things verified rather than assumed:

- **Cache placement is genuinely pinned.** I moved the block into the cached prefix and `test_workspace_context_lands_after_datetime_marker` fails. (My first attempt moved it ahead of `memory_context` and it passed — correctly, since both already sit in the volatile tail. Not a gap.) Given ENG-1122 measured ~99% of long-session context arriving as cache reads, a turn-start block in the wrong place would have been an expensive mistake, and it's the kind that shows up as a bill rather than a test failure.
- **Failure is partial, not total.** Two independent `try/except` blocks mean a broken pad listing still yields the root listing and vice versa, and a total failure yields `""` rather than a raised turn. Matches the body's claim.

11 tests pass.

## Two minor notes, neither blocking

**1. The root listing materialises the whole directory before slicing.** `sorted(p.name … for p in root.iterdir())` builds every entry, then takes 30. Fine for a normal project root; a workspace with a large root (an unpacked dataset, a vendored `node_modules` at top level) pays a full scan on every turn for 30 names. Not worth restructuring unless it's easy — flagging because it's per-turn work on a path that can't fail loudly.

**2. Dotfiles are excluded, so `.env` isn't listed.** The example block in the ticket included it. I think excluding is the better call — a context block that gets shipped to the model is not where filenames hinting at secret locations belong — so this is a deliberate improvement on my own example rather than a miss. Worth a word in the code comment so nobody "fixes" it back later.

One thing the block doesn't carry that the ticket's example did: a cell count per pad (`catanah (37 cells, last used 2m ago)`). `(active)` / snapshot age covers the decision the agent is actually making — reuse this or not — so I don't think the count earns its tokens. Mentioning only so the difference is on the record as intentional.

Approvable from my side once rebased (2 behind `origin/staging`). Not approving here only because the ordering across this and #312 is a human call.

tino097:
Both notes addressed in the latest commit; branch rebased onto origin/staging @ 560d29e5.

**1 — listing cost:** switched to `os.scandir`, which gets dir-ness from the dirent instead of a stat per entry — that was the real per-entry cost, and it runs every turn. The full iteration itself stays: a sorted, deterministic first-30 can't be produced without seeing all names, and capping the scan instead would make the listing order depend on filesystem enumeration order.

**2 — dotfile exclusion:** now stated in a comment at the exclusion site — this block ships to the model, and filenames hinting at secret locations don't belong in LLM-visible context even names-only, superseding the ticket example's `.env`. The cell-count omission from the ticket's example stays intentional per your note.

Suite: 11/11 discovery tests; scratchpad suite green except the pre-existing `test_namespace_survives_restart` baseline failure (fails on untouched staging).

alecantu7 (approved):
Both notes addressed. Approving.

**Listing cost — `os.scandir` is the right fix, and your pushback on the rest is correct.** Dir-ness from the dirent instead of a stat per entry is where the real per-entry cost was, and this runs every turn. And you're right that the full iteration is inherent rather than incidental: a deterministic sorted first-30 can't be produced without seeing all the names, and capping the scan would make which files the agent sees depend on filesystem enumeration order — which is worse than the cost it saves, because it would vary between machines for no visible reason. Good call to keep it and say why.

**Dotfile exclusion — documented at the site, which is what I wanted.** The reason is the load-bearing part: this block ships to the model, so filenames pointing at secret locations don't belong in it even names-only. That's a better rule than the `.env` in my own ticket example, and having it written where the `startswith(".")` lives is what stops someone "fixing" it back later. Cell counts staying out is right too — `(active)` / snapshot age covers the decision the agent is actually making.

11/11 discovery tests green, 0 behind `origin/staging`.

Re-stating the thing I checked on the first pass, since it's the property that makes this PR work at all and it's invisible in the diff: `agent_pads()` is disk-backed under the conversation's snapshot dir, so it survives cowork-server rebuilding the `ChatSession` per user message. A version of this built on session state would have passed every test locally and shown an empty block in the hosted and desktop products — which is exactly how the `confirm_new_scratchpad` guard managed to fire 0 times across 676 cells. Worth keeping that in mind if anyone ever refactors where the pad list comes from.

Same note as on #312, non-blocking: the "pre-existing `test_namespace_survives_restart` failure" doesn't reproduce here — passes on this branch and on untouched `origin/staging`, in isolation and across the full file.

PR #317, opened 2026-08-06
ENG-1140: resolve MindsHub setup models from the live catalogue, surface probe errors

alecantu7:
Fixes [ENG-1140](https://linear.app/mindsdb/issue/ENG-1140/antons-cloud-setup-hardcodes-a-non-existent-model-alias-fails-with-a).

## Problem

Anton's Minds setup flow hardcoded the legacy mdb.ai smart-router aliases `_reason_`/`_code_`, which don't exist in the MindsHub catalogue. Setup therefore **always failed with a valid key**, and `test_llm()` collapsed the server's precise `404 model_not_found` into a bare `False`, rendered as *"Could not connect. Check your API key and URL."* — pointing the user at the one thing that was fine.

## What changed

**Durable fix, not just the string** (per the ticket):

- **`minds_client.resolve_minds_models()`** picks the (planning, coding) pair from the live `/v1/models` catalogue, preferring the tier defaults `sonnet`/`haiku`, with `opus`/`fable`/first-listed as fallbacks. When the listing route isn't deployed it falls back to the defaults instead of blocking — same caution cowork-server's `validate_minds` documents (listing routes 404 on some hosts even for valid keys).
- **`test_llm()`** now probes with the model setup is about to configure (so a passing probe validates the actual config) and returns a structured `LLMTestResult` carrying the provider's own error message — the local instance of the ENG-1283 pattern. The setup flow prints that message (`Could not connect: model_not_found: ...`) instead of the generic advice; the generic line remains only when no detail is available.
- **Provider menu now matches the Electron onboarding**: `MindsHub` + `Bring your own key`. The `Minds-Enterprise-Cloud` / `Minds-Enterprise-Server` split is gone (the product no longer exists); mdb.ai links/copy → MindsHub (`console.mindshub.ai` for upgrade messages). Applied in first-run onboarding, `/setup`, and the `/minds` connect flow in `chat.py` (the third hardcoded-alias write site the ticket didn't mention).
- The router role (`router_model`, history summarization) intentionally stays unset — it falls back to the coding model by design, which now resolves to the cheap tier default.

## Verification

- Live against prod `api.mindshub.ai`: catalogue resolves `('sonnet', 'haiku')`; probe with `haiku` passes; probing the old alias reports `model_not_found: The model '_code_' does not exist or you do not have access to it.` — the exact message that was being masked.
- New regression tests: error-message propagation, 429 → rate-limited, unparseable body fallback, catalogue resolution (defaults present / missing / route 404), never-returns-dead-aliases, and an end-to-end `_setup_minds` test asserting the resolved pair is persisted instead of `_reason_`/`_code_`.
- Full suite: 1521 passed; the 2 `test_chat_scratchpad` failures are pre-existing on `origin/staging` (verified by stashing).

## Security pass

Reviewed the changed surface: the propagated error text comes from the provider's response body and is printed only to the local user's own console (no keys echoed, nothing logged); API-key handling paths are unchanged; no new endpoints or deserialization.

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -274,10 +274,8 @@ async def _handle_connect(
 
     # --- Test if the Minds server also supports LLM endpoints ---
     # (silenced: was printing "Testing LLM endpoints..." and "not available" messages)
-    planning_model, coding_model = resolve_minds_models(
-        minds_url, api_key, verify=ssl_verify
-    )
-    llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=coding_model)
+    models = resolve_minds_models(minds_url, api_key, verify=ssl_verify)
+    llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=models.probe)
 
     if not llm_result.ok and llm_result.error:
         console.print(
@@ -291,15 +289,15 @@ async def _handle_connect(
         )
         settings.planning_provider = "openai-compatible"
         settings.coding_provider = "openai-compatible"
-        settings.planning_model = planning_model
-        settings.coding_model = coding_model
+        settings.planning_model = models.planning
+        settings.coding_model = models.coding
         # openai_api_key and openai_base_url are derived at runtime from
         # minds_api_key and minds_url via model_post_init — no need to persist them.
         settings.model_post_init(None)
         global_ws.set_secret("ANTON_PLANNING_PROVIDER", "openai-compatible")
         global_ws.set_secret("ANTON_CODING_PROVIDER", "openai-compatible")
-        global_ws.set_secret("ANTON_PLANNING_MODEL", planning_model)
-        global_ws.set_secret("ANTON_CODING_MODEL", coding_model)
+        global_ws.set_secret("ANTON_PLANNING_MODEL", models.planning)
+        global_ws.set_secret("ANTON_CODING_MODEL", models.coding)
     else:
         # Check if Anthropic key is already configured
         has_anthropic = settings.anthropic_api_key or os.environ.get(
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -95,6 +95,7 @@ from prompt_toolkit import PromptSession
 from prompt_toolkit.formatted_text import HTML
 from prompt_toolkit.shortcuts import CompleteStyle
 from prompt_toolkit.styles import Style as PTStyle
+from rich.markup import escape
 from rich.prompt import Prompt
 from anton.memory.manage import MemoryManage, MEMORY_COMMANDS
 from anton.commands.goal import parse_goal_args, run_goal_loop
@@ -277,13 +278,24 @@ async def _handle_connect(
     models = resolve_minds_models(minds_url, api_key, verify=ssl_verify)
     llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=models.probe)
 
-    if not llm_result.ok and llm_result.error:
+    # A 429 proves the endpoint is THERE — it answered, it just throttled this
+    # probe. Treating it as unavailable dropped a throttled-but-valid MindsHub
+    # endpoint and fell through to the BYOK prompt (#317 review); the previous
+    # truthy "rate_limited" string had adopted it.
+    llm_available = llm_result.ok or llm_result.rate_limited
+
+    if not llm_available and llm_result.error:
         console.print(
-            f"[anton.muted]Minds LLM endpoints unavailable ({llm_result.error}) — "
-            "keeping the current LLM provider.[/]"
+            f"[anton.muted]Minds LLM endpoints unavailable "
+            f"({escape(llm_result.error)}) — keeping the current LLM provider.[/]"
+        )
+    elif llm_result.rate_limited:
+        console.print(
+            "[anton.muted]Minds LLM endpoints are rate-limited right now — "
+            "using them anyway; limits clear on their own.[/]"
         )
 
-    if llm_result.ok:
+    if llm_available:
         console.print(
             "[anton.success]LLM endpoints available — using Minds server as LLM provider.[/]"
         )
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -80,6 +80,7 @@ from anton.minds_client import (
     describe_minds_connection_error,
     list_minds,
     list_datasources,
+    resolve_minds_models,
     test_llm,
 )
 from anton.core.datasources.data_vault import LocalDataVault
@@ -273,23 +274,26 @@ async def _handle_connect(
 
     # --- Test if the Minds server also supports LLM endpoints ---
     # (silenced: was printing "Testing LLM endpoints..." and "not available" messages)
-    llm_ok = test_llm(minds_url, api_key, verify=ssl_verify)
+    planning_model, coding_model = resolve_minds_models(
+        minds_url, api_key, verify=ssl_verify
+    )
+    llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=coding_model)
 
-    if llm_ok:
+    if llm_result.ok:
         console.print(
             "[anton.success]LLM endpoints available — using Minds server as LLM provider.[/]"
         )
         settings.planning_provider = "openai-compatible"
         settings.coding_provider = "openai-compatible"
-        settings.planning_model = "_reason_"
-        settings.coding_model = "_code_"
+        settings.planning_model = planning_model
+        settings.coding_model = coding_model
         # openai_api_key and openai_base_url are derived at runtime from
         # minds_api_key and minds_url via model_post_init — no need to persist them.
         settings.model_post_init(None)
         global_ws.set_secret("ANTON_PLANNING_PROVIDER", "openai-compatible")
         global_ws.set_secret("ANTON_CODING_PROVIDER", "openai-compatible")
-        global_ws.set_secret("ANTON_PLANNING_MODEL", "_reason_")
-        global_ws.set_secret("ANTON_CODING_MODEL", "_code_")
+        global_ws.set_secret("ANTON_PLANNING_MODEL", planning_model)
+        global_ws.set_secret("ANTON_CODING_MODEL", coding_model)
     else:
         # Check if Anthropic key is already configured
         has_anthropic = settings.anthropic_api_key or os.environ.get(
@@ -378,7 +382,7 @@ async def _handle_remote(
         console.print("  [anton.muted]To use remote scratchpad you need a free Minds account.[/]")
         console.print()
         has_key = await prompt_or_cancel(
-            "  Do you have an mdb.ai API key?",
+            "  Do you have a MindsHub API key?",
             choices=["y", "n"],
             choices_display="y/n",
             default="y",
@@ -441,7 +445,7 @@ async def _handle_publish(
         console.print("  [anton.muted]To publish dashboards you need a free Minds account.[/]")
         console.print()
         has_key = await prompt_or_cancel(
-            "  Do you have an mdb.ai API key?",
+            "  Do you have a MindsHub API key?",
             choices=["y", "n"],
             choices_display="y/n",
             default="y",
@@ -1917,7 +1921,7 @@ async def _chat_loop(
                     console.print(
                         f"[anton.warning]Approaching token limit: {last_token_status.used:,} / "
                         f"{last_token_status.limit:,} tokens used ({pct}%). "
-                        "Visit mdb.ai to upgrade your plan or top up your tokens.[/]"
+                        "Visit console.mindshub.ai to upgrade your plan or top up your tokens.[/]"
                     )
                     console.print()
                 if _query_count == 1:
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -80,7 +80,7 @@ from anton.minds_client import (
     describe_minds_connection_error,
     list_minds,
     list_datasources,
-    resolve_minds_models,
+    resolve_and_probe,
     test_llm,
 )
 from anton.core.datasources.data_vault import LocalDataVault
@@ -275,8 +275,7 @@ async def _handle_connect(
 
     # --- Test if the Minds server also supports LLM endpoints ---
     # (silenced: was printing "Testing LLM endpoints..." and "not available" messages)
-    models = resolve_minds_models(minds_url, api_key, verify=ssl_verify)
-    llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=models.probe)
+    models, llm_result = resolve_and_probe(minds_url, api_key, verify=ssl_verify)
 
     # A 429 proves the endpoint is THERE — it answered, it just throttled this
     # probe. Treating it as unavailable dropped a throttled-but-valid MindsHub
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -279,6 +279,12 @@ async def _handle_connect(
     )
     llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=coding_model)
 
+    if not llm_result.ok and llm_result.error:
+        console.print(
+            f"[anton.muted]Minds LLM endpoints unavailable ({llm_result.error}) — "
+            "keeping the current LLM provider.[/]"
+        )
+
     if llm_result.ok:
         console.print(
             "[anton.success]LLM endpoints available — using Minds server as LLM provider.[/]"
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -36,7 +36,7 @@ from anton.commands.datasource import (
     handle_list_data_sources,
     handle_test_datasource
 )
-from anton.minds_client import resolve_minds_models, test_llm
+from anton.minds_client import minds_v1_base, resolve_minds_models, test_llm
 
 
 def _build_scratchpad_manager(
@@ -735,16 +735,13 @@ def _setup_minds(settings, ws) -> None:
     llm_ok = False
     rate_limited = False
     error_detail: str | None = None
-    planning_model = coding_model = None
 
     with Live(Spinner("dots", text="  Connecting...", style="anton.cyan"), console=console, transient=True):
         # Resolve the models from the live catalogue first, then probe with
         # the model we're about to configure — a passing probe validates the
         # actual configuration, not a hardcoded alias (ENG-1140).
-        planning_model, coding_model = resolve_minds_models(
-            minds_url, api_key, verify=True
-        )
-        result = test_llm(minds_url, api_key, verify=True, model=coding_model)
+        models = resolve_minds_models(minds_url, api_key, verify=True)
+        result = test_llm(minds_url, api_key, verify=True, model=models.probe)
         if result.rate_limited:
             rate_limited = True
             error_detail = result.error
@@ -756,11 +753,9 @@ def _setup_minds(settings, ws) -> None:
             error_detail = result.error
         else:
             error_detail = result.error
-            planning_model, coding_model = resolve_minds_models(
-                minds_url, api_key, verify=False
-            )
+            models = resolve_minds_models(minds_url, api_key, verify=False)
             result_no_ssl = test_llm(
-                minds_url, api_key, verify=False, model=coding_model
+                minds_url, api_key, verify=False, model=models.probe
             )
             if result_no_ssl.rate_limited:
                 rate_limited = True
@@ -785,16 +780,19 @@ def _setup_minds(settings, ws) -> None:
         console.print("  [anton.success]Connected[/]")
         settings.planning_provider = "openai-compatible"
         settings.coding_provider = "openai-compatible"
-        settings.planning_model = planning_model
-        settings.coding_model = coding_model
+        settings.planning_model = models.planning
+        settings.coding_model = models.coding
         settings.minds_ssl_verify = ssl_verify
-        derived_base_url = f"{minds_url}/v1"
+        # Same host rule as the probe (test_llm) — the persisted base must be
+        # the URL the probe validated, or "Connected" lies about the config
+        # that gets saved (e.g. a /v1-suffixed ANTON_MINDS_URL would double).
+        derived_base_url = minds_v1_base(minds_url)
         settings.openai_api_key = api_key
         settings.openai_base_url = derived_base_url
         ws.set_secret("ANTON_PLANNING_PROVIDER", "openai-compatible")
         ws.set_secret("ANTON_CODING_PROVIDER", "openai-compatible")
-        ws.set_secret("ANTON_PLANNING_MODEL", planning_model)
-        ws.set_secret("ANTON_CODING_MODEL", coding_model)
+        ws.set_secret("ANTON_PLANNING_MODEL", models.planning)
+        ws.set_secret("ANTON_CODING_MODEL", models.coding)
         ws.set_secret("ANTON_MINDS_SSL_VERIFY", "true" if ssl_verify else "false")
         ws.set_secret("ANTON_OPENAI_API_KEY", api_key)
         ws.set_secret("ANTON_OPENAI_BASE_URL", derived_base_url)
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -13,6 +13,7 @@ from pathlib import Path
 import typer
 from rich.console import Console
 from rich.live import Live
+from rich.markup import escape
 from rich.panel import Panel
 from rich.prompt import Confirm
 from rich.spinner import Spinner
@@ -800,8 +801,13 @@ def _setup_minds(settings, ws) -> None:
         # A 429 here is usually the TPM/RPM limiter, not an exhausted
         # allowance (wallet-empty is a 402) — so lead with the provider's own
         # message and keep the top-up advice secondary.
+        # escape(): the provider's text is interpolated into Rich markup, and
+        # a message containing [...] that Rich reads as a bad style tag raises
+        # MarkupError — crashing setup instead of reporting the error it was
+        # added to surface (#317 review).
         console.print(
-            f"  [anton.error]{error_detail or 'Rate limit or token limit exceeded.'}[/]"
+            f"  [anton.error]"
+            f"{escape(error_detail) if error_detail else 'Rate limit or token limit exceeded.'}[/]"
         )
         console.print(
             "  [anton.muted]Rate limits clear on their own — try again in a moment. "
@@ -810,7 +816,9 @@ def _setup_minds(settings, ws) -> None:
         raise _SetupRetry()
     else:
         if error_detail:
-            console.print(f"  [anton.error]Could not connect: {error_detail}[/]")
+            console.print(
+                f"  [anton.error]Could not connect: {escape(error_detail)}[/]"
+            )
         else:
             console.print(
                 "  [anton.error]Could not connect. Check your API key and URL.[/]"
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -36,7 +36,7 @@ from anton.commands.datasource import (
     handle_list_data_sources,
     handle_test_datasource
 )
-from anton.minds_client import test_llm
+from anton.minds_client import resolve_minds_models, test_llm
 
 
 def _build_scratchpad_manager(
@@ -443,9 +443,9 @@ def _onboard(settings) -> None:
     _INTRO_LINES = [
         "Hi Boss! I'm Anton, your AI coworker.",
         "",
-        "For the best experience, I recommend Minds-Enterprise-Cloud as your LLM Provider:",
+        "For the best experience, I recommend MindsHub as your LLM Provider:",
         "",
-        "  \u2713 Smart model routing",
+        "  \u2713 Latest frontier models",
         "  \u2713 Faster responses",
         "  \u2713 Cost optimized",
         "  \u2713 Secure data connectors",
@@ -555,21 +555,12 @@ async def _animate_onboard(
     console.print()
     console.print(f"[anton.glow] {'━' * 40}[/]")
     console.print()
-    console.print(
-        "  [bold]1[/]  [link=https://mdb.ai][anton.cyan]Minds-Enterprise-Cloud[/][/link] [anton.success](recommended)[/]"
-    )
-    console.print(
-        "  [bold]2[/]  [anton.cyan]Minds-Enterprise-Server[/] [anton.muted]self-hosted[/]"
-    )
-    console.print(
-        "  [bold]3[/]  [anton.cyan]Bring your own key[/] [anton.muted]Anthropic / OpenAI / Gemini[/]"
-    )
-    console.print()
+    _print_provider_choices()
 
     while True:
         choice = await prompt_or_cancel(
             "(anton) Choose LLM Provider",
-            choices=["1", "2", "3"],
+            choices=["1", "2"],
             default="1",
             allow_cancel=False,
         )
@@ -578,22 +569,11 @@ async def _animate_onboard(
             if choice == "1":
                 _setup_minds(settings, ws)
             elif choice == "2":
-                _setup_minds(settings, ws, default_url=None)
-            elif choice == "3":
                 _setup_other_provider(settings, ws)
             break  # success
         except _SetupRetry:
             console.print()
-            console.print(
-                "  [bold]1[/]  [link=https://mdb.ai][anton.cyan]Minds-Enterprise-Cloud[/][/link] [anton.success](recommended)[/]"
-            )
-            console.print(
-                "  [bold]2[/]  [anton.cyan]Minds-Enterprise-Server[/] [anton.muted]self-hosted[/]"
-            )
-            console.print(
-                "  [bold]3[/]  [anton.cyan]Bring your own key[/] [anton.muted]Anthropic / OpenAI / Gemini[/]"
-            )
-            console.print()
+            _print_provider_choices()
             continue
 
     # Reload env vars so the scratchpad subprocess inherits them
@@ -607,9 +587,10 @@ async def _animate_onboard(
     model_label = settings.planning_model
     if provider_label == "openai-compatible":
         base = settings.openai_base_url or ""
-        if settings.minds_url and "mdb.ai" in settings.minds_url:
-            provider_label = "Minds-Enterprise-Cloud"
-            model_label = "smart_router"
+        if settings.minds_url and (
+            "mindshub.ai" in settings.minds_url or "mdb.ai" in settings.minds_url
+        ):
+            provider_label = "MindsHub"
         elif url_hostname(base) == "generativelanguage.googleapis.com":
             provider_label = "Google Gemini"
         elif base:
@@ -695,8 +676,19 @@ def _setup_prompt(
     return result
 
 
+def _print_provider_choices() -> None:
+    """Print the LLM-provider menu (shared by first-run onboarding and retry)."""
+    console.print(
+        "  [bold]1[/]  [link=https://mindshub.ai][anton.cyan]MindsHub[/][/link] [anton.success](recommended)[/]"
+    )
+    console.print(
+        "  [bold]2[/]  [anton.cyan]Bring your own key[/] [anton.muted]Anthropic / OpenAI / Gemini[/]"
+    )
+    console.print()
+
+
 def _setup_minds(settings, ws, *, default_url: str | None = "https://api.mindshub.ai") -> None:
-    """Set up Minds as the LLM provider (cloud or enterprise)."""
+    """Set up MindsHub as the LLM provider."""
     console.print()
 
     is_cloud = default_url and default_url.endswith(("mindshub.ai", "mdb.ai"))
@@ -717,7 +709,7 @@ def _setup_minds(settings, ws, *, default_url: str | None = "https://api.mindshu
         )
         console.print()
         has_key = Confirm.ask(
-            "  Do you have an mdb.ai API key?",
+            "  Do you have a MindsHub API key?",
             default=True,
             console=console,
         )
@@ -745,20 +737,36 @@ def _setup_minds(settings, ws, *, default_url: str | None = "https://api.mindshu
     ssl_verify = True
     llm_ok = False
     rate_limited = False
+    error_detail: str | None = None
+    planning_model = coding_model = None
 
     with Live(Spinner("dots", text="  Connecting...", style="anton.cyan"), console=console, transient=True):
-        result = test_llm(minds_url, api_key, verify=True)
-        if result == "rate_limited":
+        # Resolve the models from the live catalogue first, then probe with
+        # the model we're about to configure — a passing probe validates the
+        # actual configuration, not a hardcoded alias (ENG-1140).
+        planning_model, coding_model = resolve_minds_models(
+            minds_url, api_key, verify=True
+        )
+        result = test_llm(minds_url, api_key, verify=True, model=coding_model)
+        if result.rate_limited:
             rate_limited = True
-        elif not result:
-            result_no_ssl = test_llm(minds_url, api_key, verify=False)
-            if result_no_ssl == "rate_limited":
+        elif result.ok:
+            llm_ok = True
+        else:
+            error_detail = result.error
+            planning_model, coding_model = resolve_minds_models(
+                minds_url, api_key, verify=False
+            )
+            result_no_ssl = test_llm(
+                minds_url, api_key, verify=False, model=coding_model
+            )
+            if result_no_ssl.rate_limited:
                 rate_limited = True
-            elif result_no_ssl:
+            elif result_no_ssl.ok:
                 ssl_verify = False
                 llm_ok = True
-        else:
-            llm_ok = True
+            else:
+                error_detail = result_no_ssl.error or error_detail
 
     if llm_ok and not ssl_verify:
         console.print("  [anton.warning]SSL certificate verification failed.[/]")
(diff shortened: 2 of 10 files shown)


alecantu7 (commented):
Adversarial self-review findings:

1. **No-SSL retry fired on HTTP-level failures** (found, fixed in 482904a). The original flow retried with `verify=False` on *any* first-probe failure. A 404/401 response proves TLS worked, so that retry could only repeat the same error while doubling the probe and catalogue requests. `LLMTestResult` now carries `http_status` and the retry only runs on transport-level failures. Regression test added.

2. **`resolve_minds_models` falls back to `ids[0]` when sonnet/opus/fable are all absent** (considered, kept). The prod catalogue currently lists `muse-spark` first, so this could theoretically pick an odd planning model — but only if all three preferred ids vanish, and a served-but-odd model still beats a hardcoded id that 404s (the exact failure this PR removes). Falling back to the bare defaults instead would recreate the ticket's bug shape on hosts whose catalogue genuinely lacks them.

3. **Known limitation — existing installs are not migrated.** Users who ran the old setup have `ANTON_PLANNING_MODEL=_reason_` / `ANTON_CODING_MODEL=_code_` persisted in `~/.anton/.env` and keep failing until they re-run `/setup` (which now works and heals the config). An auto-heal on boot (detect the dead aliases, re-resolve) would be a separate, deliberate change — flagging for a follow-up decision rather than smuggling it in here.

alecantu7 (commented):
Second adversarial review pass. The core move is right — resolve from the live catalogue instead of hardcoding, and propagate the provider's message. Two findings block, though: the replacement is grounded in the catalogue's *ids* but discards the two fields that say whether a model is actually usable, and the probe itself 400s on the models a restricted key can reach. Net effect is that setup still deterministically fails for free-tier / wallet-empty users — the same user-visible outcome the ticket opened on, with a better message.

Base `staging`, 0 behind, CI green, 39/39 tests pass locally.

---

## 🔴 1. Probing with `haiku` reintroduces ENG-576 for free-tier / empty-wallet users

`resolve_minds_models` prefers `sonnet`/`haiku`, and `test_llm` probes with `coding_model` — so the probe sends **`haiku`**, a paid model. A prior incident already removed exactly that from the probe. From `cowork-server/tests/test_provider_probe_model.py:1`:

> "MindsHub gates paid models per plan tier: a free-tier key gets a 403 for haiku/sonnet/etc. The Settings health probe (`ping_provider`) and onboarding validation (`validate_minds`) used to POST `CODING_MODEL_DEFAULTS["minds_cloud"]` = "haiku" (paid) → free-tier accounts saw "MindsHub failed its last test" / "Invalid API key" even though chat worked on mindshub_air. Both must now probe `mindshub_air` (the free baseline, present in every tier)."

cowork-server settled on `MINDS_PROBE_MODEL = "mindshub_air"` (`cowork/services/providers.py:28`) for this reason. This PR moves the CLI back toward the pattern that caused ENG-576 — and dead-ends harder now, since the new `http_status is not None` branch (correctly) suppresses the no-SSL retry.

**The catalogue already answers this and the code discards the answer.** `list_models()` keeps only `id`. The live prod payload carries two more fields per entry:

```json
{"id": "sonnet",      "label": "Claude Sonnet 5",           "enabled": true, "embedding": false, ...}
{"id": "embed-small", "label": "Text Embedding 3 (small)",  "enabled": true, "embedding": true,  ...}
```

`enabled` is not decoration. `auth/entitlements/serializers/entitlements_me.py:45`:

> "`enabled` is the access decision auth owns: the wallet can pay, or the model is a free-bucket model with allowance left. Clients render a disabled model as a locked upsell … **they must not recompute access themselves.**"

`mindshub_inference`'s `/v1/models` is a straight relay of that (`minds/api/v1/endpoints/models.py:7` — "auth computes each model's `enabled` flag (wallet- and allowance-aware)"), and `auth/configs/model_policy.json` sets `free_bucket: true` only for `mindshub_air`. So a free-tier or wallet-empty key sees `sonnet`/`haiku` with `enabled: false` and `mindshub_air` with `enabled: true` — and this code picks `haiku` regardless, ignoring the one model it could actually have used.

`embedding` matters for the `ids[0]` fallback too. The self-review considered `ids[0]` and kept it — "a served-but-odd model still beats a hardcoded id that 404s" — which is fair, but `embed-small` is in the prod list, so "odd" can mean an embeddings model configured as both planning and coding. Filtering keeps that argument intact and makes the fallback safe:

```python
for entry in entries or []:
    if not isinstance(entry, dict) or not entry.get("id"):
        continue
    if entry.get("embedding") or entry.get("enabled") is False:
        continue
    ids.append(str(entry["id"]))
```

Then add `mindshub_air` to the `pick` preference chain as the last resort before `ids[0]`, so a restricted key lands on the model it is entitled to.

## 🔴 2. `max_tokens=1` returns HTTP 400 on `mindshub_air` and the `gpt-*` family

Measured live against prod `api.mindshub.ai`:

| model | max=1 | max=16 |
|---|---|---|
| `haiku` | 200 | 200 |
| `sonnet` / `gemini-flash` | 200 | 200 |
| **`mindshub_air`** | **400** | 200 |
| **`gpt-nano`** | **400** | 200 |
| **`embed-small`** | **400** | — |

```
HTTP 400 {"error":{"message":"Error code: 400 - {'error': {'message': \"Invalid 'max_output_tokens':
integer below minimum value. Expected a value >= 16, but got 1 instead.\", ...
'code': 'integer_below_min_value'}}"}}
```

So the probe cannot succeed on the free-tier model — fixing #1 alone still dead-ends setup, now with `Could not connect: integer_below_min_value: Invalid 'max_output_tokens'…`. Honest message, still a false negative on a perfectly good key. `validate_minds` already uses `max_tokens: 20` (`cowork/services/providers.py:471`), above the floor. Match it.

This is the ticket's own thesis turned on the fix: `max_tokens=1` is another constant that was correct when written and silently wrong once the catalogue grew OpenAI-backed aliases.

---

## 🟡 3. No remaining path to a non-prod or self-hosted MindsHub URL

Removing Minds-Enterprise-Server was the agreed scope addition and the Electron-parity argument holds. But it was the only writer of a custom `minds_url`: `_setup_minds` is now called exclusively with the `https://api.mindshub.ai` default, so the `is_cloud` false branch — the `Server URL` prompt at `anton/cli.py:697-701` — is unreachable dead code, and no live caller passes anything but the default for `default_url`.

Worth a deliberate decision rather than a silent loss:

- **This ticket's own repro can no longer be executed.** Steps 3–4 are "choose option 2, enter `https://api.staging.mindshub.ai`". QA has no UI path to staging; only `ANTON_MINDS_URL`, which nothing in the flow mentions.
- BYOK → "OpenAI-compatible (custom endpoint)" is not a substitute: it never sets `minds_url` / `minds_api_key`, so `/minds`, the remote scratchpad and `/publish` stay unconfigured.
- `/minds` connect (`anton/chat.py:150`) reads `settings.minds_url` and never prompts for it, so it can't fill the gap either.

Cheapest fix that keeps the menu clean: prompt for the URL only when an env override or the entered key indicates non-prod, or keep an unlisted "advanced" entry. At minimum, delete the dead branch and document `ANTON_MINDS_URL`.

## 🟡 4. `test_never_returns_dead_smart_router_aliases` asserts a guarantee the code doesn't have

The docstring says the legacy aliases must never be picked "**whatever the server says**", but the fixture only serves `["sonnet", "haiku"]` — the assertion cannot fail regardless of the implementation. And the promised property is false:

```
['_reason_', '_code_']                     -> ('_reason_', '_reason_')
['text-embedding-3-small', 'mindshub_air'] -> ('text-embedding-3-small', 'text-embedding-3-small')
```

Either serve `["_reason_", "_code_"]` and add an explicit denylist, or rewrite the docstring to the property the code actually has. As written it reads as coverage for this ticket's central regression and provides none.

## 🟡 5. The 429 branch still hardcodes the misleading advice this ticket is about

`test_llm` returns `error=None` for 429 and `cli.py:806` unconditionally prints `Token limit exceeded. Visit https://console.mindshub.ai to upgrade or to top up your tokens.` But a 429 on MindsHub is usually the TPM/RPM limiter, not an exhausted allowance — wallet-empty is a **402** (`code: wallet_empty`). So a rate-limited user is told to buy tokens they already have. The body is right there: run 429 through `_http_error_detail` too, and either append the provider's message or key the copy on the error code.

---

## Smaller things

- **Docs still teach the removed thing.** `docs/docs/start/pick-a-provider.md:17,27` documents "Minds-Enterprise-Cloud (mdb.ai)" with "`_reason_` / `_code_` smart routing"; `docs/docs/connect/web-search.md:15` and `web-fetch.md:16` list the removed provider name. For a PR whose thesis is "the assumption moved and nothing checked", leaving the docs asserting the dead aliases is the same failure one layer up.
- **Reuse (convention 7).** `list_models`/`test_llm` hardcode `/v1`, while the `/v1`-vs-`/api/v1` host rule already exists three times in anton (`settings.py:181`, `core/backends/local.py:545`) and once as a named helper in cowork-server (`minds_chat_base_url`). The new `list_models` inherits the assumption and hides it behind swallow-and-fallback. There is also a shell twin doing this exact job in `anton_services/snapshots/cowork/setup.sh:110` (`mh_fetch_model_catalog` / `mh_pick_model`) — worth a pointer comment so the two don't drift, especially as that one resolves `latest:`-prefixed forms and this one bare ids.
- **`chat.py` drops the error it now has.** `_handle_connect` uses `llm_result.ok` and discards `llm_result.error`, silently falling back to Anthropic. The ENG-1283 pattern got applied to one of the two callers.
- **Worst-case spinner time doubles.** The catalogue call is `timeout=30` like the probe, and the transport-failure path re-resolves before re-probing — 4 × 30s on a blackholed host vs 2 × 30s before. A shorter timeout on the catalogue (it is advisory) keeps the hang where it was.
- **`settings.py:144`** still defaults `minds_url = "https://mdb.ai"`, which `model_post_init` turns into `https://mdb.ai/api/v1` for anyone with a key but no URL. Same species of stale default; not this PR's job, but worth confirming that host is still live.
- Pre-existing: `ws.set_secret("ANTON_MINDS_API_KEY", …)` runs before the probe, so a failed setup still persists credentials.

---

**Recommendation:** changes requested on #1 and #2 — both are small (filter `list_models` on `enabled`/`embedding` + add `mindshub_air` to the preference chain; raise the probe to `max_tokens=20`), and without them setup still fails for the users most likely to hit it. #3 needs a call on whether losing the custom-URL path is acceptable. The third hardcoded write site in `chat.py`, the structured `LLMTestResult`, and the `http_status` fix in 482904a all hold up.

alecantu7 (commented):
All findings addressed in d74a984. Point by point:

**🔴 1 — probe/resolution ignores entitlements (ENG-576 shape): fixed.** `list_models()` filters on `embedding` and `enabled is False` (absent flags are kept, so older hosts that don't send them still work), denylists `_reason_`/`_code_` outright, and `mindshub_air` joins both preference chains as the last resort before `ids[0]`. A free-tier key whose catalogue disables the paid models now resolves `('mindshub_air', 'mindshub_air')` — test added for exactly that scenario. The `ids[0]` fallback survives, but it can no longer land on an embeddings model.

**🔴 2 — `max_tokens=1` 400s on the free-tier model: fixed.** Probe raised to 20, matching `validate_minds`. Verified live: `mindshub_air` probe now returns ok (was `integer_below_min_value` at 1).

**🟡 3 — custom-URL path: decided with Alejandro — dead code removed.** No self-hosted deployment exists now or short-term, and Electron never had an editable URL either (the host is baked per build via `MINDS_API_BASE`). The Server-URL prompt branch is deleted; `ANTON_MINDS_URL` is the single path to a non-default host, and `_setup_minds` now *honors* it instead of clobbering it back to prod — so the ticket's staging repro stays executable (`export ANTON_MINDS_URL=https://api.staging.mindshub.ai`, run setup). Documented in `pick-a-provider.md` and covered by a test.

**🟡 4 — vacuous denylist test: fixed.** The fixture now serves `["_reason_", "_code_"]` and the denylist makes the docstring's "whatever the server says" actually true — that catalogue falls back to the defaults.

**🟡 5 — 429 copy: fixed.** `test_llm` runs 429s through `_http_error_detail` too, and the setup flow leads with the provider's message; the top-up advice is now secondary and phrased conditionally (a 429 is usually the TPM/RPM limiter — wallet-empty is a 402).

**Smaller items:** docs updated (pick-a-provider, web-search, web-fetch); `minds_v1_base()` centralizes the /v1-vs-/api/v1 rule for the new call sites with pointers to the settings.py rule, cowork-server's `minds_chat_base_url`, and the `setup.sh` shell twin; `chat.py` connect now prints the probe error instead of silently falling back; catalogue calls get a 10s advisory timeout so the worst-case spinner hang stays where it was. Left alone as flagged: the stale `minds_url = "https://mdb.ai"` settings default and the persist-before-probe behavior (both pre-existing, both worth their own small tickets if we care).

Full suite 1,529 passed. Live against prod: usable-id filtering drops `embed-*`, pair resolves `('sonnet', 'haiku')`, both probes pass.

alecantu7 (commented):
Second review pass, at `d74a984`. All five numbered findings and the smaller items are addressed, and addressed properly — the tests that were vacuous now actually bind. Local run: 1530 passed / 2 failed, and both failures (`tests/test_chat_scratchpad.py`) reproduce identically on `origin/staging`, so they are pre-existing and unrelated.

## Resolved

| # | Fix | Verified |
|---|---|---|
| 🔴 1 | `list_models` filters on `enabled` / `embedding` and drops `LEGACY_SMART_ROUTER_ALIASES`; `MINDS_FREE_TIER_MODEL = "mindshub_air"` sits last in both `pick` chains | `test_free_tier_key_lands_on_the_free_bucket_model`, `test_embedding_models_are_never_picked` — real fixtures now, not tautologies |
| 🔴 2 | `max_tokens=20`, with the ≥16 floor and `integer_below_min_value` recorded in the docstring | matches `validate_minds`; re-probed prod — `mindshub_air` / `gpt-nano` pass at 20 |
| 🟡 3 | Dead `default_url` parameter gone; `ANTON_MINDS_URL` honored and documented (`pick-a-provider.md:39-41`, `test_setup_minds_honors_env_url_override`) | ✅ |
| 🟡 4 | Fixture is now `["_reason_", "_code_"]` and asserts the fallback pair — the property is real and enforced in `list_models` | ✅ |
| 🟡 5 | 429 carries `_http_error_detail(e)`; copy leads with the provider's own message and demotes the top-up advice | ✅ |
| — | `minds_v1_base()` extracted with pointers to the cowork-server helper and the `anton_services` shell twin; catalogue `timeout=10`; docs de-`_reason_`-ed; `chat.py` now surfaces `llm_result.error` | ✅ |

## Retracting one of my own findings

**The `mdb.ai` SPA-200 hazard I raised does not reproduce through this client.** I measured it with curl's default `Accept: */*`; `minds_request` sends `Accept: application/json`, and the host content-negotiates:

```
GET https://mdb.ai/api/v1/models
  Accept: */*               -> HTTP 200  text/html   (SPA catch-all)
  Accept: application/json  -> HTTP 404               (what anton actually gets)
  Accept: text/html         -> HTTP 200  text/html
```

So there is no false-positive "Connected" and no HTML reaching `json.loads`. Nothing to fix — my earlier measurement used the wrong instrument.

## 🔴 One new finding, introduced by the fix

**`anton/cli.py:791` still writes `derived_base_url = f"{minds_url}/v1"` — the one place that didn't get `minds_v1_base()`.** The probe and the persisted runtime base URL are now computed by two different rules, and `ANTON_MINDS_URL` — which this commit just made the documented escape hatch — feeds both. Live against prod:

```
ANTON_MINDS_URL=https://api.mindshub.ai      probe=ok  persisted=…/v1      runtime=200 OK
ANTON_MINDS_URL=https://api.mindshub.ai/v1   probe=ok  persisted=…/v1/v1   runtime=HTTP 404 Not Found
                                                       *** SETUP SAYS CONNECTED, RUNTIME BROKEN ***
```

`minds_v1_base` explicitly anticipates a `/v1`-suffixed host — there is a test for exactly that (`test_explicit_v1_suffix_kept_as_is`) — so the probe absorbs the suffix and succeeds while the persist path doubles it. Same shape for `ANTON_MINDS_URL=https://mdb.ai`: probe → `/api/v1`, persisted → `/v1`. The docs at `pick-a-provider.md:40` say "set `ANTON_MINDS_URL` … setup honors it" without constraining the form, and `/v1` is the natural thing to paste since that is the OpenAI base URL.

It also breaks this PR's own thesis that "a passing probe validates the actual configuration" — here the probe validates a *different* URL than the one that gets saved. `chat.py` is immune because `_handle_connect` persists via `settings.model_post_init(None)`, which uses the settings.py rule; `cli.py` is the lone divergent copy.

One line:

```python
derived_base_url = minds_v1_base(minds_url)
```

Worth adding a test that pins the invariant — the persisted `ANTON_OPENAI_BASE_URL` must equal the base the probe hit. Nothing currently asserts it, which is why the helper landed in two of the three call sites.

## 🟡 Residual worth one line of thought

When `/v1/models` is unreachable, `resolve_minds_models` falls back to `("sonnet", "haiku")` and the probe then sends `haiku` — so the ENG-576 shape returns for a free-tier key on any host that doesn't serve the listing route (the exact caution the docstring cites). Not prod, and it fails safe-ish, but the clean resolution is to stop conflating the two roles: **probe with `mindshub_air` when there is no catalogue to trust, configure `sonnet`/`haiku`.** That keeps "the probe validates the config" whenever the catalogue *is* available, and keeps the probe universal when it isn't.

Caveat on evidence: I could not test the free-tier path live — no free-tier key — so this one rests on the auth contract (`entitlements_me.py:45`) and the ENG-576 record, not on a measurement.

Still open from the first pass, both explicitly deferred and fine to leave: `settings.py:144` defaults `minds_url = "https://mdb.ai"` (fails loudly with a 404, no false Connected), and `core/backends/local.py:545` keeps its own copy of the `/v1`-vs-`/api/v1` rule that `minds_v1_base` now supersedes.

---

**Verdict:** ship-ready once `derived_base_url` uses `minds_v1_base`.

alecantu7 (commented):
Both items addressed in d71c4c3:

**🔴 `derived_base_url` divergence: fixed with the suggested one-liner, plus the invariant test.** The persist path now uses `minds_v1_base(minds_url)` — the same rule the probe applies — and `test_setup_minds_persists_the_base_url_the_probe_validated` pins the invariant with the exact failure input (`ANTON_MINDS_URL=…/v1`): the persisted `ANTON_OPENAI_BASE_URL` must equal the base the probe hit. Verified live: bare and `/v1`-suffixed hosts both probe ok and persist the identical validated base; no more `…/v1/v1`.

**🟡 Residual (paid probe on catalogue-less hosts): taken.** `resolve_minds_models` now returns a `MindsModels` resolution that carries its own probe model — the coding model when the catalogue vouched for the pair (the probe validates the actual config), `mindshub_air` when there was no catalogue to trust (a paid default would 403 for free-tier keys on exactly those hosts). Verified live: a catalogue-less path resolves the `sonnet`/`haiku` defaults but probes `mindshub_air`. Same caveat as yours — no free-tier key to test the 403 side live; this rests on the auth contract.

Also acknowledged as deliberately left: the `settings.py` `mdb.ai` default (fails loudly) and `core/backends/local.py:545`'s own copy of the host rule — the latter is a candidate for a tiny follow-up migration to `minds_v1_base` per convention 7, but it touches the scratchpad backend, which this PR otherwise doesn't.

Full suite: 1,530 passed (same 2 pre-existing scratchpad failures on origin/staging excluded).

alecantu7 (commented):
Third pass, at `d71c4c3`. Both findings from the second pass are fixed and I verified them live. Suite: 1531 passed / 2 failed, both pre-existing on `origin/staging` (`test_chat_scratchpad.py`). 0 behind base.

## Verified fixed

**🔴 `derived_base_url` now uses `minds_v1_base()`** — the probe and the persisted runtime base are one rule. Re-ran the exact case that broke before, live against prod:

```
ANTON_MINDS_URL=https://api.mindshub.ai      probe=ok  persisted=…/v1  runtime=200 OK  CONSISTENT
ANTON_MINDS_URL=https://api.mindshub.ai/v1   probe=ok  persisted=…/v1  runtime=200 OK  CONSISTENT   (was …/v1/v1 -> 404)
ANTON_MINDS_URL=https://api.mindshub.ai/     probe=ok  persisted=…/v1  runtime=200 OK  CONSISTENT
```

`test_setup_minds_persists_the_base_url_the_probe_validated` pins the invariant against the probe's own base rather than a literal, which is the right shape — it can't drift with the helper.

**🟡 `MindsModels` splits probe from config.** Behaviour across catalogue states:

```
prod (live)            planning=sonnet        coding=haiku         probe=haiku
free tier (paid off)   planning=mindshub_air  coding=mindshub_air  probe=mindshub_air
no catalogue (404)     planning=sonnet        coding=haiku         probe=mindshub_air   <-- probe != configured
self-hosted, no air    planning=sonnet        coding=haiku         probe=haiku
```

## 🟡 One correction to my own recommendation

I suggested the free-bucket probe fallback, and measurement says I aimed it at an empty population. Both MindsHub envs serve the listing route:

```
api.mindshub.ai          /v1/models  -> HTTP 401 unauthenticated (route exists)
api.staging.mindshub.ai  /v1/models  -> HTTP 401 unauthenticated (route exists)
```

A 401 without credentials means the route is there and auth-gated, and with a valid key prod returns all 20 entries with `enabled` / `embedding`. So on prod and staging the catalogue path is **always** taken and the entitlement filter always runs — the free-tier-cloud case that `probe=MINDS_FREE_TIER_MODEL` protects against cannot reach the no-catalogue branch on any env we operate.

The population that *does* reach it is whatever lacks the listing route: self-hosted, legacy `mdb.ai`, or a host behind a proxy that blocks it. Those are the hosts least likely to serve `mindshub_air` at all — it's a MindsHub-Cloud free-bucket alias out of auth's `model_policy.json`, not something a self-hosted deployment inherits. On such a host that serves `sonnet`/`haiku` fine, the probe now sends `mindshub_air`, 404s, and setup fails — where `d74a984` probed `haiku` and would have passed. So the change trades a cloud false negative that can't occur for a self-hosted one that can, on the branch that exists *because* of self-hosted hosts.

Secondary point, smaller: in that branch the probe validates a model the setup does not configure, which is the same "probe validated something other than what we saved" divergence the base-URL fix in this very commit set out to remove.

Two ways out, either fine:

- **Smallest, evidence-backed:** put `probe=coding` back in the no-catalogue branch and record the measurement above in the comment — the motivating scenario is unreachable on prod/staging.
- **Thorough:** probe the configured coding model first, and only on a model-not-found / entitlement failure retry with the free-bucket model — then configure whichever one passed. That removes both false negatives and keeps the invariant that what we persist is what we validated.

Not blocking: relative to `staging` (which hardcodes `_code_` and 404s everywhere) any of these is an improvement, and QA will be on staging, which has the catalogue.

## Everything else

No leftover tuple-unpacking callers of `resolve_minds_models`; `models` is unconditionally bound before use in both `cli.py` and `chat.py`; the removed `planning_model = coding_model = None` init is safe. The new tests assert `(planning, coding, probe)` triples rather than just the pair, so the split is actually covered.

Still open and fine to leave, both flagged earlier and deliberately out of scope: `settings.py:144` defaults `minds_url = "https://mdb.ai"` (fails loudly — 404, no false Connected), and `core/backends/local.py:545` keeps its own copy of the `/v1`-vs-`/api/v1` rule that `minds_v1_base` now supersedes. Worth a follow-up ticket to collapse the remaining copies onto the helper, since its docstring is now the place that says "keep them aligned".

---

**Verdict:** ship it. The no-catalogue probe choice is worth 5 minutes of thought before merge, but it doesn't gate — it can't fire on prod or staging.

anton/core/llm/client.py
```diff
@@ -12,7 +12,7 @@
 def _resolve_openai_compatible_flavor(settings: AntonSettings) -> str:
     """Distinguish mdb.ai passthrough from a generic openai-compatible endpoint.
 
-    The "Minds-Enterprise-Cloud" setup path writes ``openai_base_url =
+    The MindsHub setup path writes ``openai_base_url =
```
pnewsam: **Minor / question.** This docstring now reads "The MindsHub setup path writes `openai_base_url = f"{minds_url.rstrip('/')}/api/v1"`", but MindsHub hosts write `/v1` (via `minds_v1_base` / `model_post_init`). The detection below only matches `base == minds or base == f"{minds}/api/v1"`, so a MindsHub `.../v1` base falls through to `FLAVOR_OPENAI_COMPATIBLE_GENERIC` — the native MindsHub web-search/fetch passthrough may silently degrade to the generic fallback for the provider this PR makes canonical.

The logic looks pre-existing (setup already wrote `/v1`), but the rename here intersects it and the docs now misstate it. Please confirm passthrough is still detected for `api.mindshub.ai`, or add the `f"{minds}/v1"` case.

anton/chat.py
```diff
@@ -273,23 +274,30 @@ async def _handle_connect(
 
     # --- Test if the Minds server also supports LLM endpoints ---
     # (silenced: was printing "Testing LLM endpoints..." and "not available" messages)
-    llm_ok = test_llm(minds_url, api_key, verify=ssl_verify)
+    models = resolve_minds_models(minds_url, api_key, verify=ssl_verify)
+    llm_result = test_llm(minds_url, api_key, verify=ssl_verify, model=models.probe)
 
-    if llm_ok:
+    if not llm_result.ok and llm_result.error:
```
pnewsam: **Minor.** A rate-limited probe now sets `ok=False`, so Anton treats the Minds LLM as unavailable and falls through to the Anthropic-key prompt below. Previously `test_llm` returned the truthy `"rate_limited"` string, so a throttled-but-valid endpoint was still adopted. The muted line also says "keeping the current LLM provider" immediately before the `else` branch may prompt the user to set a *new* one.

Consider treating `llm_result.rate_limited` as "endpoint present" here, and rewording the message.

anton/cli.py
```diff
@@ -774,31 +780,44 @@ def _setup_minds(settings, ws, *, default_url: str | None = "https://api.mindshu
...
         ws.set_secret("ANTON_MINDS_SSL_VERIFY", "true" if ssl_verify else "false")
         ws.set_secret("ANTON_OPENAI_API_KEY", api_key)
         ws.set_secret("ANTON_OPENAI_BASE_URL", derived_base_url)
     elif rate_limited:
+        # A 429 here is usually the TPM/RPM limiter, not an exhausted
+        # allowance (wallet-empty is a 402) — so lead with the provider's own
+        # message and keep the top-up advice secondary.
         console.print(
-            "[anton.error]Token limit exceeded. Visit https://mdb.ai to upgrade or to top up your tokens.[/]"
+            f"  [anton.error]{error_detail or 'Rate limit or token limit exceeded.'}[/]"
```
pnewsam: **Nit.** Provider-supplied error text is interpolated directly into Rich markup here (and at `cli.py:813`, `chat.py:282`). `console` has markup enabled, so an error message containing `[…]` that Rich reads as a bad style tag raises `MarkupError` and crashes setup with a traceback — the opposite of the error-surfacing this PR adds. Low probability for OpenAI-shaped messages, but cheap to harden with `rich.markup.escape(error_detail)`.

pnewsam (commented):
## Code Review

**Verdict: COMMENT**

Reviewed the full diff plus `cli.py`, `minds_client.py`, `chat.py`, `settings.py`, and `core/llm/client.py` at head `d71c4c3`. The core fix is correct and well-tested: setup resolves real catalogue models, probes the config it is about to persist, and surfaces the provider's own error. No blocking or major defects. Three smaller items are inline; one is worth confirming before merge because this PR makes MindsHub the sole Minds path.

### Validation

- Traced both write paths (`_setup_minds`, `_handle_connect`) end to end, including the SSL-fallback and retry branches.
- Confirmed `minds_v1_base` matches `AntonSettings.model_post_init`'s `/v1`-vs-`/api/v1` rule, so `_handle_connect`'s reliance on `model_post_init` stays consistent with the probe.
- Checked every `test_llm` reference: the only production callers (`chat.py`, `cli.py`, `minds_client.py`) are all updated; the remaining hits are pytest method names, so the `bool → LLMTestResult` change breaks nothing untouched.
- Reviewed the new tests. I did not run the suite myself — the "1521 passed" claim is unverified here.

pnewsam (approved):

anton/core/llm/client.py
alecantu7: **Confirmed broken, not just misdocumented — good catch.** Fixed in `12d3537`.

Checked it empirically rather than by reading, and the whole chain holds:

```
minds_url=https://api.mindshub.ai      -> base=https://api.mindshub.ai/v1   generic     <-- what setup writes
minds_url=https://api.mindshub.ai/v1   -> base=https://api.mindshub.ai/v1   PASSTHROUGH
minds_url=https://mdb.ai               -> base=https://mdb.ai/api/v1        PASSTHROUGH
```

Setup writes the bare host (`cli.py:700`), `model_post_init` derives `/v1`, and the resolver matched neither — so **every MindsHub install got `OPENAI_COMPATIBLE_GENERIC`**, whose `native_web_tools()` returns `set()`. The session then routes `web_search` to a handler ToolDef that needs an Exa/Brave key MindsHub users don't have: web search just stops working, silently, no error anywhere. Pre-existing, but this PR makes MindsHub the canonical provider, so it's now 100% of that path.

**Verified the passthrough is actually correct before enabling it** — I didn't want to trade a silent degradation for a hard failure. anton's passthrough sends `{"type": "web_search"}` / `{"type": "fetch"}` (`openai.py:690-697`), and `mindshub_inference@origin/main`'s `GenericToolType` is exactly `WEB_SEARCH = "web_search"` / `FETCH = "fetch"` (`minds/inference/types.py:399`). Same contract, so the gateway understands it; per-alias `WebSearchMode` then decides native-vs-external-vs-drop server-side, which is its call to make.

Fix: `base in (minds, f"{minds}/v1", f"{minds}/api/v1")`, docstring corrected (it described the `/api/v1` derivation as if it were the only one). `TestMindsPassthroughFlavorDetection` covers all three shapes, keeps a third-party endpoint on generic (guard against over-matching), and asserts the property that actually matters — `native_web_tools() == {web_search, web_fetch}`. Mutation-verified: dropping the `/v1` case fails `test_mindshub_host_is_passthrough`.

anton/chat.py
alecantu7: Fixed in `12d3537` — both halves. `llm_available = llm_result.ok or llm_result.rate_limited`: a 429 proves the endpoint answered and throttled *this probe*, so it's adopted with the catalogue-resolved models, and the throttled case gets its own message ("rate-limited right now — using them anyway; limits clear on their own") instead of the unavailable line. You're right that the old truthy `"rate_limited"` string had adopted it, so `LLMTestResult` had quietly regressed this.

Also took the wording note: the "keeping the current LLM provider" line no longer prints immediately before offering to set a new one — it's now gated on genuinely-unavailable.

Test `test_connect_adopts_a_rate_limited_minds_endpoint` drives the real `_handle_connect` with a 429 `LLMTestResult` and asserts the providers/models are persisted; mutation-verified (reverting to `ok`-only fails it).

anton/cli.py
alecantu7: Fixed in `12d3537` — `escape()` at all three sites (`cli.py` ×2, `chat.py`).

Worth recording what I found while testing it, because it changes how likely this is: **Rich tolerates an unmatched OPENING tag and only raises on a CLOSING one.** Verified directly —

```
ok      | no access to [sonnet-4-6] on this plan
ok      | unexpected [ in body
RAISES  | unknown route [/v1/chat/completions]   -> closing tag has nothing to close
RAISES  | validation error at [/model]           -> doesn't match any open tag
```

So the realistic trigger is a **route or JSON-pointer echoed in a gateway error** — quite plausible from a validation path — not the bracketed-model-name shape I'd have guessed. My first two fixtures used the tolerated shape and **passed against the unfixed code**; I caught it only by mutating the fix. Same trap as anton#309's abandon test, second time in two days, so it's a habit I need rather than a one-off. Replaced with the two shapes above; mutation-verified that removing `escape()` now fails both.

anton/minds_client.py
```diff
@@ -185,24 +218,162 @@ def list_datasources(
...
+    """
+    try:
+        ids = list_models(base_url, api_key, verify=verify)
+    except Exception:
+        ids = []
+    if not ids:
+        return MindsModels(
+            planning=MINDS_DEFAULT_PLANNING_MODEL,
+            coding=MINDS_DEFAULT_CODING_MODEL,
+            probe=MINDS_FREE_TIER_MODEL,
```
pnewsam: **Minor.** This no-catalogue path breaks the PR's own "a passing probe validates the actual config" invariant. It returns `planning=sonnet, coding=haiku` but `probe=mindshub_air`. A free-tier key then passes the `mindshub_air` probe, setup prints "Connected," and every real request afterward uses `sonnet`/`haiku` and 403s — the exact success-masking class ENG-1140 set out to remove.

On prod this is reachable when the catalogue call hits its (best-effort, 10s) timeout for a free-tier key mid-onboarding, not only on hosts that lack the listing route.

The docstring shows this is a deliberate trade-off — probing a paid default would wrongly *block* free-tier keys — so I'm not blocking. But consider persisting the probed model (`mindshub_air`) when it's the free-bucket fallback rather than the unvalidated paid pair, or printing a softer "connected, but model access wasn't confirmed" note on this branch specifically.

pnewsam (commented):
## Re-review

Re-reviewed at `12d3537`, focused on whether the three findings from the prior pass landed and whether the final commit introduced anything new. **All three are correctly fixed**, verified against the real code, not just the diff:

- **`/v1` passthrough detection** — fixed; `TestMindsPassthroughFlavorDetection` pins all three derivable base URLs, the generic-endpoint guard, and the downstream `native_web_tools()` property.
- **Dropped throttled endpoint** — fixed via `llm_available = ok or rate_limited` with its own message; covered by `test_connect_adopts_a_rate_limited_minds_endpoint`.
- **`MarkupError` crash** — `escape()` at all three sites, and the markup-safety tests use *closing*-tag fixtures (the shape that actually raises), avoiding the false-pass trap.

Also confirmed out-of-diff: `minds_request` accepts the `timeout` kwarg `list_models` now passes, `describe_minds_connection_error` returns the 2-tuple `test_llm` unpacks, and `settings.model_post_init` derives the base URL with the same host rule as `minds_v1_base` (so chat.py's runtime-derived config matches the URL the probe validated), and `minds_v1_base` is idempotent on a `/v1` suffix.

One non-blocking observation inline. Nothing here blocks merge.

---

**Nit (no action needed):** `minds_v1_base()` and the inline block in `settings.py::model_post_init` are twins — the code comment already flags "keep them aligned," and a settings→minds_client import cycle may be why they weren't unified.

anton/minds_client.py
alecantu7: **You're right, and it's worth more than a softer message — fixed in `26c7d40`.** This was the PR's own invariant broken on its own fallback branch, and your reachability point (best-effort catalogue timeout, not just hosts missing the listing route) is what makes it worth doing properly.

**I took a third option rather than either of yours, because both trade one bad outcome for another:**
- *Persist the probed free model* → a **paid** user whose catalogue call times out gets silently configured with `mindshub_air`. Working, but quietly downgraded to the cheapest model, and far harder to notice than a 403. That's a quieter bug than the one we're fixing.
- *Softer "access wasn't confirmed" note* → the broken config still gets written; we'd just be warning the user we might have broken their install.

So `resolve_and_probe()` **probes the model it would configure and escalates once** to the free bucket on a model-access denial, returning the pair that was actually validated:

| case | probes | persisted |
|---|---|---|
| paid key, no catalogue | `haiku` ✓ | sonnet/haiku — validated |
| free key, no catalogue | `haiku` 403 → `mindshub_air` ✓ | mindshub_air — validated, and not blocked (ENG-576) |
| bad key / throttled / transport | one probe | reports itself, no escalation |

That's the catalogue path's semantics (configure what the key can actually use) reconstructed with probes where no listing exists — so the invariant now holds on **every** branch, at the cost of one extra probe only in the free-tier-plus-no-catalogue case.

Deliberate details: only **403/404** escalate (a 401 is a bad key, a 429 is throttling, a transport failure says nothing about entitlement — escalating any of those would burn a call and could misreport the cause); when both probes are denied it reports the **original** denial, since that names the model the user asked to be set up with; and a catalogue-derived pair carries `free_fallback=None`, so it can never escalate — `/v1/models` is already access-aware.

Mutation-verified both directions: reverting to the free-model probe fails 8 tests, and escalating on *any* failure fails exactly the three guard tests. One structural side effect worth flagging since it touched your review's seams: both call sites now use the single helper, which collapsed two monkeypatch points into one across 8 test sites — mechanical, and the tests are tighter for it. Full suite 1509 + e2e 39.

alecantu7:
@pnewsam — your re-review's finding is fixed in `26c7d40`, and it deserves another pass because **I didn't take either of the two options you suggested** and the change is structural rather than local.

**Your finding:** the no-catalogue branch returned `planning=sonnet, coding=haiku` but `probe=mindshub_air`, so a free-tier key passed the probe, setup said "Connected," and every real request afterward 403'd — this PR's own invariant broken on its own fallback path.

**Why I rejected both suggested options:** *persist the probed free model* silently downgrades a **paid** user to the cheapest model whenever the catalogue call times out — a quieter bug than the one being fixed; *print a softer note* still writes the broken config and just warns about it.

**What landed instead — `resolve_and_probe()`:** probes the model it would actually configure and **escalates once** to the free bucket on a model-access denial, returning the pair that was validated.

| case | probes | persisted |
|---|---|---|
| paid key, no catalogue | `haiku` ✓ | sonnet/haiku — validated |
| free key, no catalogue | `haiku` 403 → `mindshub_air` ✓ | mindshub_air — validated, not blocked (ENG-576) |
| bad key / throttled / transport | one probe | reports itself, no escalation |

That reconstructs the catalogue path's semantics (configure what the key can actually use) with probes where no listing exists, so the invariant holds on every branch. Cost: one extra probe only in the free-tier-plus-no-catalogue case.

**The parts I'd most like you to check:**
- **Escalation trigger is 403/404 only.** 401 is a bad key, 429 is throttling, `http_status is None` is transport — none is an entitlement signal, and escalating any of them would burn a call and could misreport the cause. Tests assert non-escalation for all three.
- **Both probes denied → reports the ORIGINAL denial**, since it names the model the user asked to be set up with.
- **A catalogue-derived pair carries `free_fallback=None`** and can never escalate — `/v1/models` is already access-aware.
- **Structural churn you should know about:** both call sites now use the one helper, which collapsed two monkeypatch seams (`resolve_minds_models` + `test_llm`) into one across **8 test sites**. Mechanical, and the tests are tighter for it, but it's the kind of diff you'd rather be told about than discover.

Mutation-verified both directions: reverting to the free-model probe fails 8 tests; escalating on *any* failure fails exactly the three guard tests. Suite 1509 + e2e 39.

**CI caveat:** the three red checks here are GitHub's active Actions outage — all `cancelled`, zero failing steps, killed at exactly 15m03s. `run-tests`, `Analyze (python)` and `build` passed on that same commit. They need one UI re-run after the incident clears (the API refuses retries on cancelled runs).

pnewsam (approved):
## Re-review at `26c7d40`

**Verdict: APPROVE**

Re-reviewed the one commit added since my last pass (`12d3537`), which resolves the sole inline finding I left: the no-catalogue branch persisted `sonnet`/`haiku` while probing `mindshub_air`, so a free-tier key could pass the probe and then 403 on every real request — the exact "Connected"-masking-a-broken-install class ENG-1140 exists to remove. The fix is correct and the rest of the PR still holds. No blocking or major defects.

### Validation

- **The finding is fixed properly.** `resolve_and_probe` now probes the model it would *configure* (`haiku`) and escalates once to the free bucket only on a `403`/`404`, returning whichever pair actually validated — so "a passing probe validates the persisted config" now holds on every branch, including no-catalogue.
- **Escalation guards are right.** `401` (bad key), `429` (throttle), and transport failures (`http_status is None`) do *not* escalate — each surfaces as itself rather than being retried as a model problem. A catalogue-derived pair carries `free_fallback=None` and never escalates, correct since `/v1/models` is already access-aware.
- **Both callers migrated; no stragglers.** Confirmed against base `staging` that `chat.py` and `cli.py` were the only `test_llm` truthiness callers — both now go through `resolve_and_probe`, and neither unpacks the old contract.
- **Tests bind the behavior.** `TestResolveAndProbeEscalation` covers all eight branches (paid-passes, denied→escalate, 404→escalate, 401/429/transport no-escalate, both-denied reports the original denial, catalogue-never-escalates), with `seen == [...]` assertions pinning the probe count.
- **CI:** `run-tests` passes at this head — resolving the "suite unverified here" caveat from my prior pass; `build`, `Analyze (python)`, and the Snyk checks pass too. The `scope` / `verdict-eval` / `CLAssistant` / `Analyze (actions)` failures are administrative/infra checks, not code — worth clearing before merge but not review-blocking.

No actionable findings. The only item still open is the one both sides already deferred: `minds_v1_base()` and the inline block in `settings.py::model_post_init` remain twins (a settings→minds_client import cycle is the likely reason they aren't unified), flagged in-code with a "keep them aligned" comment.

alecantu7:
@pnewsam thanks for the approve — and I've filed your deferred `minds_v1_base` twins item as [ENG-1322](https://linear.app/mindsdb/issue/ENG-1322) rather than leaving it as a comment, because checking it turned up more than twins.

**It's four implementations across three repos plus a shell script**, not two:

| # | location | form |
|---|---|---|
| 1 | `anton/minds_client.py::minds_v1_base()` | function |
| 2 | `anton/config/settings.py::model_post_init` | inline block (ENG-436) |
| 3 | `cowork-server/cowork/services/providers.py::minds_chat_base_url()` | same logic, different name |
| 4 | `anton_services/snapshots/cowork/setup.sh` | shell — the *inverse* derivation (`${LLM_API_BASE_URL%/v1*}`) |

The three Python ones are byte-equivalent in behaviour today, and each carries a comment telling the reader to keep it aligned with the others.

**Which is the argument for doing it: that comment is the mitigation that already failed.** ENG-1140's bug was exactly a *consumer* knowing one derivation and not the other — the flavor resolver matched the `mdb.ai` shape but not the `/v1` shape copy #2 actually produces, so every MindsHub install silently lost native web search. The hazard was documented and the documentation didn't help.

Ticket notes the constraint you inferred (the `settings` → `minds_client` cycle is why #1 and #2 were never unified, so the single home has to be dependency-free), that #4 can't import Python and therefore needs an *agreement test* rather than pretend consolidation, and that `TestMindsPassthroughFlavorDetection` from this PR should extend to cover the consolidated helper's full output set — so a new host shape can never again outrun its consumers. Also flagged that none of the four has an explicit case for `api.<env>.mindshub.ai`.

Not blocking this PR; #317 is done pending the CI re-run once GitHub's Actions incident clears.

PR #318, opened 2026-08-06
ENG-764: defer tool schemas behind recalled skills (unlock_skill)

ea-rus:
What

Generic mechanism to keep rarely-needed tool schemas out of the LLM request until they're actually needed. A tool tagged with unlock_skill="<label>" is not sent up front; it's registered only when the model recalls that skill. Host-driven — anton hardcodes nothing about which tools defer.

This is the reusable engine; the first consumer is cowork-server's data-source connection tools (see cowork-server#283).

How

- ToolDef.unlock_skill (tool_defs.py) — optional label marking a tool as deferred.
- _build_tools (session.py) — tools with unlock_skill go into a per-session _deferred_bundles map keyed by label instead of registering eagerly; everything else registers as before.
- _register_tool_bundle(label) — registers a label's deferred tools on demand; sticky for the session, deduped.
- Recall wiring — both handle_recall_skill (model self-recall) and _inject_recalled_skills (thalamus preload) unlock the bundle for the resolved skill label.
- Same-turn availability — tools is rebuilt inside both tool-call loops after dispatch, so a tool unlocked by a recall_skill this round is visible to the follow-up request in the same turn.
- Resume — _replay_tool_bundles_from_history re-unlocks bundles from prior recall_skill calls in the loaded history (survives server restart); runs after eager tools so tool order stays cache-stable.
- Host skill roots — SkillStore(extra_roots=…) + AntonSettings.skills_extra_roots: hosts can ship read-only skills alongside their own tools. Precedence: user → host → anton built-ins.

Impact

- No behavior change unless a tool sets unlock_skill (default None).
- Deferring a tool also defers its prompt text (the system-prompt section is built from registered tools only) — extra savings, no dangling references.
- Sticky registration → at most one prompt-cache invalidation per session (on unlock); stable tail ordering avoids repeat busts.

Files

tool_defs.py, session.py, recall_skill.py, skills.py, config/settings.py (+115 lines).

Fixes: https://linear.app/mindsdb/issue/ENG-764/token-usage-by-tools


StpMax (changes requested):
# Review: mindsdb/anton#318 — ENG-764: defer tool schemas behind recalled skills (unlock_skill)

## Summary
Solid, well-scoped mechanism: `ToolDef.unlock_skill` + a per-session `_deferred_bundles` map, unlocked on `recall_skill` (live call or thalamus preload), with same-turn visibility and allowlist coexistence. The design reasoning in the PR body is sound and matches the code. Test coverage for the "happy path" scenarios (allowlist coexistence, replay, same-turn visibility) is good. One correctness gap in the resume/replay path, plus a small piece of dead code.

## Findings

### 1. `_replay_tool_bundles_from_history` can fail to re-unlock a bundle that *was* unlocked live (medium confidence)
`anton/core/session.py:1357-1375`

The live unlock paths (`handle_recall_skill` in `recall_skill.py:162-166`, `_inject_recalled_skills` in `session.py:2431`) always call `_register_tool_bundle(skill.label)` — the **resolved, canonical** label returned by `SkillStore.load()`/`closest_match()`. This means a model call like `recall_skill(label="connect_datasource")` (underscore) or a typo that gets fuzzy-matched to `connect-datasource` still unlocks the right bundle, because `store.load()` normalizes/resolves before the label reaches `_register_tool_bundle`.

`_replay_tool_bundles_from_history`, however, reads the **raw, as-typed** `label` straight out of the historical `tool_use` block's `input` (`label = (block.get("input") or {}).get("label")`, only `.strip()`-ed) and calls `_register_tool_bundle` with that raw string — it never goes through `store.load()`/`closest_match()`. `_register_tool_bundle` does an exact dict lookup (`self._deferred_bundles.get(label, [])`), so if the historical raw label differs at all from the tool's declared `unlock_skill` string (underscore vs hyphen, case, or anything that needed fuzzy-matching live), the bundle silently fails to re-unlock.

Concrete failure scenario: user says "connect my postgres db" → model calls `recall_skill(label="connect_datasource")` → live call resolves to `connect-datasource` (skills use hyphens per `skills.py`'s own docstring, but tool/function names elsewhere use underscores, so this is a plausible model output) → bundle unlocks, trio of tools becomes available for the rest of the session. Server restarts (the exact scenario this method's docstring says it's for: *"survive a session rebuild (e.g. server restart)"*). On rebuild, `_replay_tool_bundles_from_history` sees `input.label == "connect_datasource"`, looks up `_deferred_bundles.get("connect_datasource")` → miss (key is `"connect-datasource"`) → tools stay hidden even though the user's session history clearly shows they were already unlocked.

Suggest resolving the historical label through the skill store the same way the live path does (`store.load(label)` → use `skill.label`) before calling `_register_tool_bundle`, instead of using the raw stored string directly.

The existing test (`test_bundle_replays_from_prior_recall_in_history`) only exercises the case where the historical label already matches the bundle key exactly, so it doesn't catch this.

### 2. Dead code: `_builtin_dirs()` (low, nit)
`anton/core/memory/skills.py:438`

`_iter_skill_dirs()` was changed from calling `self._builtin_dirs()` to `self._readonly_dirs()`, but `_builtin_dirs()` itself was left in place and is no longer called anywhere in the codebase (verified with a repo-wide grep, including tests). Worth deleting rather than leaving an orphaned method.

### 3. `Skill.provenance` type comment not updated (cosmetic)
`anton/core/memory/skills.py:105`

The field comment still lists `"manual" | "consolidator" | "builtin"`, but host roots now legitimately produce `provenance == "host"` (as declared in cowork-server's new `SKILL.md`). Not used in any conditional logic today (checked both repos), so purely a documentation nit.

## Not a bug, but worth double-checking
- `_build_tools()` rebuilds are now called twice in some turns (once at turn start, once after tool dispatch) plus once in `_build_system_prompt`. Each rebuild after the first only touches allowlist bookkeeping and `dict.dump()`, so this looks cheap — just flagging that it's on a hot path in case tool_registry ever grows a more expensive `dump()`.
- The "sticky, no rollback" model (a tool once unlocked never re-locks under the allowlist) is well matched by `test_deferred_tool_left_out_of_allowlist_stays_filtered`, no issue found there.

ea-rus:
Fixed:
 1. (replay label resolution, medium): _replay_tool_bundles_from_history now resolves the raw historical label through the skill store (store.load(raw) → skill.label) before calling _register_tool_bundle, exactly as the live recall paths do, with a fallback to the raw label so a missing store never regresses an already-matching key. This fixes the reported case where a live recall_skill(label="connect_datasource") (underscore) unlocked the connect-datasource bundle but, after a session rebuild, replay missed it on the exact-match dict lookup and left the tools hidden. Added a regression test test_replay_resolves_non_canonical_label_via_store (history carries the underscore label, a real hyphen skill on disk) — it's red without the fix. Also updated the now-inaccurate comment in the existing test_bundle_replays_from_prior_recall_in_history to describe the store-resolve-then-fallback behavior.

2. (dead _builtin_dirs, nit): Deleted — _iter_skill_dirs uses _readonly_dirs now and nothing else references it (grep-confirmed across both repos, including tests).

3. (provenance comment, cosmetic): Updated to "manual" | "consolidator" | "builtin" | "host".

StpMax (approved):

PR #320, opened 2026-08-07
Feature/cloud memory

ZoranPandovski:
Each cloud turn  request carries a memory block, the pod stages it as files and reads from there, and anything the agent memorizes is reported back for cowork-server to save. The pod never stores memory itself.

Two supporting changes in core. A `background_memory` flag (default on) so cloud can skip the automatic end-of-turn passes — they'd spend LLM calls on writes that vanish when the pod exits. And Hippocampus now strips newlines and `-->` from stored entry text, because slot files are one entry per line and a newline could
turn a lesson into a binding rule.

Desktop is unchanged, with tests covering it. Full suite matches the baseline. Fixes [ENG-1342](https://linear.app/mindsdb/issue/ENG-1342/enable-memory-in-cloud)

### DIFF
--- a/anton/cloud_turn/__main__.py
+++ b/anton/cloud_turn/__main__.py
@@ -23,7 +23,7 @@ import os
 import sys
 
 from anton.cloud_turn.contract import TurnRequestV1
-from anton.cloud_turn.session import build_cloud_chat_session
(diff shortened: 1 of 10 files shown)


tino097 (approved):
LGTM

PR #326, opened 2026-08-10
Feature/cloud skills

ZoranPandovski:
The pod now receives the org's skills in the turn request and makes them available to the agent, matching how per-turn memory already works. [Fixes ENG-1379](https://linear.app/mindsdb/issue/ENG-1379/skills-as-payload-to-anton-turn)


tino097 (approved):
LGTM

PR #328, opened 2026-08-10
ENG-1334: stop the verifier eval reporting green without running

alecantu7:
## What

`verdict-eval` passes while executing nothing. `pytestmark = skipif(not _KEY)` means a missing `MINDSHUB_API_KEY` skips all 11 cases and the job goes green in ~15 seconds — against ~4 minutes for a real run.

This makes a missing key **fail** on first-party PRs, while keeping the skip for developers and forks.

## Why now

Measured 2026-08-10 — eight consecutive `verdict-eval` runs, all green, all 13–24 seconds, every one the skip path:

| time (UTC) | duration | branch |
| -- | -- | -- |
| 16:03 | 15s | tino097/eng-1310-… |
| 13:58 | 16s | feat/ENG-969 |
| 12:14 | 14s | feat/ENG-969 |
| 12:05 | 13s | feat/eng-764-db-tools |
| 12:00 | 24s | feat/ENG-969 |
| 11:53 | 21s | feat/eng-764-db-tools |
| 11:46 | 19s | feat/eng-1092-move-current-date |
| 11:25 | 15s | feat/eng-1092-move-current-date |

Eight PRs merged past a check asserting nothing about a rubric that changed **five times in nine days** (ENG-716, ENG-1081, ENG-1134, ENG-1155, ENG-836).

## The actual bug: two callers, opposite needs

The old design conflated them:

- A developer running `pytest tests/` **should skip** — no key, not testing the verifier, failing their whole run would be rude.
- CI **must fail** — a gate that goes green when it cannot run is worse than no gate, because the board looks guarded.

So the skip stays the default and CI opts out explicitly via `VERIFIER_EVAL_REQUIRE_LIVE=1`, which turns a missing key into a collection error carrying an explanation.

## Four cases, all verified locally

| case | result | verified |
| -- | -- | -- |
| developer, no key | 11 skipped, **exit 0** | ✅ unchanged |
| first-party PR, no key | collection error, **exit 2** | ✅ |
| fork PR | skips + step-summary note | ✅ executed, all four IS_PR/IS_FORK combos |
| key present | runs | ✅ 11/11 in 78s on CI, and twice locally against prod |

Forks are the one place skipping is correct: `pull_request` withholds secrets from them by design, so a contributor cannot fix this and must not be failed for it. anton is public with **116 forks** (though **0 fork PRs in the last 60**), so the branch is reachable. The step summary says plainly that *"the verifier rubric is unguarded on this PR"* rather than letting a green tick imply otherwise.

## Second guard: assert cases actually ran

The key check can't see other zero-execution paths — a stray `-k`, a fixture skipping at setup, some future condition. So the job also asserts against the junit report that `tests - skipped > 0`.

Validated against three reports:

| report | result |
| -- | -- |
| real all-skipped run (11/11 skipped) | **caught**, exit 1 |
| real run (11 executed) | allowed, exit 0 |
| empty collection (0 collected) | **caught**, exit 1 |

Attribute names were confirmed against pytest's actual output, not assumed.

## The two fixture bugs are now fixed here too

An earlier revision of this description said the red check was expected and the fixtures were a separate PR. **Both are fixed in this PR now**, because a check that is randomly red ~1 run in 5 carries no more information than one that is falsely green — which is this PR's whole thesis.

**`implied_success_data_never_arrived` over-specified its outcome.** ENG-1134's requirement is "never accept a hallucinated success as done". INCOMPLETE (keep working) and STUCK (stop and explain) both satisfy it — neither ships the invented `$1.24M`. Pinning INCOMPLETE alone flaked ~1 in 3 on a STUCK whose reasoning was correct. Now accepts either; COMPLETE and WAITING still must never appear, and a test asserts the set cannot grow to include them.

**`recovered_tool_error` asked an ambiguous question.** *"…compared to Tesla before Tesla's IPO"* leaves "before Tesla's IPO" free to bind one figure or both, and the answer reads it as one (`$9.8B to date` vs `Tesla (pre-IPO): $783M`). Both readings are defensible, so the verifier split ~1 in 6. The case exists to prove a *recovered* tool error does not force a continuation; the ambiguity smuggled in a second, unintended assertion about answer-responsiveness, and that is what flaked. Rephrased as two explicit asks — **the answer text is untouched.**

The transcript repeats the question as its first user turn, and only changing one copy would have left the verifier reading the old phrasing. Both changed, and a test pins them identical.

**Reliability:** 3 consecutive live runs green (73s / 72s / 81s), against 1-in-5 failing before. **Three runs is a small sample and does not prove the flake is gone.** Watch the gate over the next several PRs before making the check required.

## Deliberately not in this change

Adding `pyproject.toml` or this workflow to the `paths:` triggers. Both would catch invocation drift like the `--extra dev` → `--group dev` break earlier today (PEP 735 migration on staging), but both widen the red-check surface before the key exists. Worth a follow-up once it does.

## Security

No new credential, no new network call, no secret in output. The failure message names the env var but never its value. The fork branch **reduces** exposure by making it explicit that secrets are absent there.

Refs ENG-1334.

### DIFF
--- a/.github/workflows/verifier-eval.yml
+++ b/.github/workflows/verifier-eval.yml
@@ -50,6 +50,10 @@ concurrency:
 jobs:
   verdict-eval:
     runs-on: ubuntu-latest
+    # A real run is ~90s. The retry budget bounds throttle sleeping to ~5min,
+    # so anything past this is hung rather than slow — and the GitHub default
+    # of 360 would hold a runner for six hours to find that out.
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/verifier-eval.yml
+++ b/.github/workflows/verifier-eval.yml
@@ -64,10 +64,18 @@ jobs:
       - name: Decide whether live mode is required
         id: mode
         env:
+          # Fail CLOSED. `IS_FORK` is derived from a pull_request payload, so on
+          # any other event it compares an empty string and comes out "true" —
+          # which would set require_live=0 and silently restore the exact
+          # green-while-skipping bug this workflow exists to prevent. Only a
+          # positively-identified fork PR is allowed to skip; anything else
+          # (a workflow_dispatch added later, a schedule, an unknown event)
+          # requires live mode and fails loudly if it cannot run.
+          IS_PR: ${{ github.event_name == 'pull_request' }}
           IS_FORK: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
         run: |
           set -euo pipefail
-          if [ "$IS_FORK" = "true" ]; then
+          if [ "$IS_PR" = "true" ] && [ "$IS_FORK" = "true" ]; then
             echo "require_live=0" >> "$GITHUB_OUTPUT"
             {
               echo "### Verifier eval skipped — fork PR"
@@ -136,7 +144,11 @@ jobs:
           if starved and len(starved) == len(skips):
               detail = starved[0][:300]
               print(f"::warning::Verifier eval could NOT run — the CI key hit 402/429. {detail}")
-              with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as fh:
+              # .get, not [], so a missing GITHUB_STEP_SUMMARY cannot KeyError
+              # inside the branch that handles "we could not run" — that would
+              # turn an honest warning into a red for a third unrelated reason.
+              summary = os.environ.get("GITHUB_STEP_SUMMARY")
+              with open(summary or os.devnull, "a") as fh:
                   fh.write(
                       "### ⚠️ Verifier eval did not run — CI key out of money\n\n"
                       "Every case skipped on a gateway 402/429, so **the verifier "
--- a/.github/workflows/verifier-eval.yml
+++ b/.github/workflows/verifier-eval.yml
@@ -101,18 +101,57 @@ jobs:
         if: steps.mode.outputs.require_live == '1'
         run: |
           python3 - <<'PY'
-          import sys, xml.etree.ElementTree as ET
+          import os, sys, xml.etree.ElementTree as ET
+
+          # Three outcomes, not two. The point of this gate is that its result
+          # carries information, so "could not run" must not look like "passed"
+          # and must not look like "the rubric regressed" either.
+          #
+          #   executed > 0                  -> green. It ran.
+          #   0 executed, all out-of-money  -> green + loud warning. The CI key
+          #                                    hit 402/429. Nobody's PR caused
+          #                                    it and nobody can fix it from a
+          #                                    PR, so it must not block — but it
+          #                                    must not read as a pass either.
+          #   0 executed, any other reason  -> red. Misconfiguration.
+          MARKER = "GATEWAY_UNAVAILABLE"  # set in tests/test_verifier_verdict_live.py
+
           root = ET.parse("eval-junit.xml").getroot()
-          total = skipped = 0
-          for suite in root.iter("testsuite"):
-              total += int(suite.get("tests", 0))
-              skipped += int(suite.get("skipped", 0))
-          executed = total - skipped
-          print(f"collected={total} skipped={skipped} executed={executed}")
-          if executed == 0:
-              print(
-                  "::error::verdict-eval passed without executing a single case. "
-                  "A green gate that ran nothing is worse than no gate. See ENG-1334."
-              )
-              sys.exit(1)
+          total = executed = 0
+          skips: list[str] = []
+          for case in root.iter("testcase"):
+              total += 1
+              skip = case.find("skipped")
+              if skip is None:
+                  executed += 1
+              else:
+                  skips.append(skip.get("message") or "")
+
+          starved = [m for m in skips if MARKER in m]
+          print(f"collected={total} executed={executed} skipped={len(skips)} out_of_money={len(starved)}")
+
+          if executed:
+              sys.exit(0)
+
+          if starved and len(starved) == len(skips):
+              detail = starved[0][:300]
+              print(f"::warning::Verifier eval could NOT run — the CI key hit 402/429. {detail}")
+              with open(os.environ["GITHUB_STEP_SUMMARY"], "a") as fh:
+                  fh.write(
+                      "### ⚠️ Verifier eval did not run — CI key out of money\n\n"
+                      "Every case skipped on a gateway 402/429, so **the verifier "
+                      "rubric is unguarded on this PR.** This is not a failure of "
+                      "your change and not a rubric regression.\n\n"
+                      f"First reason reported:\n\n```\n{detail}\n```\n\n"
+                      "Top up or raise the limit on the key behind "
+                      "`MINDSHUB_API_KEY`, then re-run this job. Tracking: ENG-1334.\n"
+                  )
+              sys.exit(0)
+
+          print(
+              "::error::verdict-eval passed without executing a single case, and not "
+              "because the key ran out of money. A green gate that ran nothing is "
+              "worse than no gate. See ENG-1334."
+          )
+          sys.exit(1)
           PY
--- a/.github/workflows/verifier-eval.yml
+++ b/.github/workflows/verifier-eval.yml
@@ -154,20 +154,25 @@ jobs:
               summary = os.environ.get("GITHUB_STEP_SUMMARY")
               with open(summary or os.devnull, "a") as fh:
                   fh.write(
-                      "### ⚠️ Verifier eval did not run — CI key out of money\n\n"
-                      "Every case skipped on a gateway 402/429, so **the verifier "
+                      "### ⚠️ Verifier eval did not run — the CI key hit 402/429\n\n"
+                      "Every case skipped on a gateway denial, so **the verifier "
                       "rubric is unguarded on this PR.** This is not a failure of "
                       "your change and not a rubric regression.\n\n"
                       f"First reason reported:\n\n```\n{detail}\n```\n\n"
-                      "Top up or raise the limit on the key behind "
-                      "`MINDSHUB_API_KEY`, then re-run this job. Tracking: ENG-1334.\n"
+                      "Two different causes land here, so read the reason above "
+                      "before acting: an empty wallet or spent allowance "
+                      "(`wallet_empty` / `included_allowance_exhausted`) needs "
+                      "credit, while a velocity throttle (`rate_limited`) that "
+                      "outlived the retry budget needs a higher rate limit or a "
+                      "quieter moment — not money. Then re-run this job. "
+                      "Tracking: ENG-1334.\n"
                   )
               sys.exit(0)
 
           print(
               "::error::verdict-eval passed without executing a single case, and not "
-              "because the key ran out of money. A green gate that ran nothing is "
-              "worse than no gate. See ENG-1334."
+              "because of a gateway denial. A green gate that ran nothing is worse "
+              "than no gate. See ENG-1334."
           )
           sys.exit(1)
           PY
--- a/.github/workflows/verifier-eval.yml
+++ b/.github/workflows/verifier-eval.yml
@@ -6,9 +6,25 @@ name: Verifier verdict eval
 # the eval itself) — a rubric edit that regresses a case should be caught by
 # the PR, not by a customer.
 #
-# Requires the MINDSHUB_API_KEY repo secret. When it is absent (forks, or
-# before an admin adds it), pytest skips every case and the job stays green —
-# the gate is the key's presence, not workflow logic.
+# Requires the MINDSHUB_API_KEY repo secret.
+#
+# That used to mean "no key -> every case skips -> job green", which made the
+# check worthless: measured 2026-08-10, eight consecutive green runs at 13-24s
+# each, all of them the skip path, against a real run of ~4 minutes. Eight PRs
+# merged past a gate that asserted nothing (ENG-1334).
+#
+# Now the job distinguishes the two reasons a key can be missing:
+#
+#   First-party PR, no key  -> HARD FAIL. Somebody has to add the secret; until
+#                              then this gate is not guarding the rubric and
+#                              should not claim to be.
+#   Fork PR                 -> skip, and say so in the summary. `pull_request`
+#                              deliberately withholds secrets from forks, so a
+#                              contributor cannot fix this and must not be
+#                              failed for it.
+#
+# The distinction is made here, not in pytest, because only the workflow knows
+# whether the run is from a fork.
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


alecantu7 (commented):
## Self-review (adversarial) — head `4e45275`, base `origin/staging` `5f3f2f7`

Validated rather than read: extracted the guard script verbatim and ran it against synthetic junit reports, exercised all three pytest paths locally, and traced the 429 mapping through the provider the eval actually uses.

**The mechanism is correct.** All five guard branches behave exactly as documented, and all three local paths do what the description's table claims:

| path | result |
|---|---|
| developer, no key | **11 skipped, exit 0** ✅ |
| `REQUIRE_LIVE=1`, no key | **collection error with the intended message, exit 2** ✅ |
| `REQUIRE_LIVE=0`, no key (fork) | **11 skipped, exit 0** ✅ |

(Confirmed there is no `.env` in the worktree, so the no-key results are real rather than an accidental skip.)

Five findings. Findings 2 and 3 share one fix. Finding 1 is the urgent one and it is about this description, not the code.

---

### 1. `MINDSHUB_API_KEY` exists now — this description's headline warning is inverted, and the red is real

**high · confirmed from CI logs**

The description says:

> **⚠️ This PR's own verdict-eval will be RED** … First-party PR, no secret → collection error. **That is the change working**

That is not what happened:

| run | sha | actual |
|---|---|---|
| 19:24 | `e9a3d02` | `VERIFIER_EVAL_REQUIRE_LIVE: 1`, **11 collected, 11 PASSED in 78s** |
| 19:39 | `4e45275` | **1 failed, 10 passed in 71s** |

No `RuntimeError` and no collection error, so `_KEY` was truthy. **The secret is in the repo** — it landed somewhere between the 16:03 measurement in the table above and 19:24. Neither this description nor ENG-1334 reflects that step 2 is done.

The current red is a genuine eval failure:

```
FAILED test_verdict[implied_success_data_never_arrived-haiku]
  expected INCOMPLETE x3 (source: ENG-1134 (hallucinated-success safeguard))
  got [STUCK: …; INCOMPLETE: …; INCOMPLETE: …]
```

**This recreates the failure mode the PR exists to eliminate, one level up.** A reviewer reads "red is expected, that's the change working," ticks past it, and merges over a rubric signal — the eval carrying no information because the *prose* said to ignore it rather than because the *job* skipped. The flipping case is the ENG-1134 hallucinated-success safeguard.

Two consequences: the whole **"⚠️ Needs a decision before merge"** section — permanently-red check, mark it non-required, wait for the key — is moot and should come out. And the failing case needs triage before merge.

---

### 2. The out-of-money escape hatch produces a green gate that executed zero cases, reachable by a transient

**medium-high · confirmed — executed + full code trace**

The last commit added a third outcome the description never mentions. Guard extracted verbatim from the workflow and run against synthetic reports:

| junit fixture | exit | |
|---|---|---|
| 1 executed, 1 starved | 0 | ✅ intended |
| **all skipped, all `GATEWAY_UNAVAILABLE`** | **0** | **green, zero executed** |
| all skipped, no marker | 1 | ✅ |
| mixed marker / non-marker | 1 | ✅ |
| empty collection | 1 | ✅ |

The logic matches its own documentation. The problem is what can reach the starved branch. The eval goes `minds-cloud` → openai-compatible, so `anton/core/llm/openai.py:107` is its path:

```python
if exc.status_code == 429 and isinstance(detail, str) and detail:
    msg = f"Server returned 429 — {detail}"
    raise TokenLimitExceeded(msg) from exc
```

**Any 429 carrying a string `detail` becomes `TokenLimitExceeded`**, and it is checked *before* the wallet-code branch at `:143`. The same file's docstring at `:71` states *"The gateway's 429 is FastAPI-style (`{"detail": …}`)"*. So a **TPM-velocity throttle** — not a dead wallet — maps to `TokenLimitExceeded` → the new `except` → `pytest.skip(GATEWAY_UNAVAILABLE)` → green with nothing run.

`_verdicts` already knows this risk — *"the gateway rate-limits per key (ENG-878 counts full context per request)"* — which is why calls are sequential. But `_verdict` has no backoff on the new path: the first throttled call skips instantly and the next test fires straight into the same throttle window. **The all-skip cascade is the natural shape, not the unlikely one.**

This violates the ticket's own acceptance criterion: *"No configuration can produce a green `verdict-eval` that executed zero cases."*

*Refutation attempted:* checked whether anything separates quota-429 from velocity-429 before `:107` — nothing does, the wallet-code check is 36 lines later. Checked whether a partial throttle saves it — it does (`executed > 0` → honest green; mixed → red), so the bad outcome needs a sustained window. That is why this is medium-high rather than high.

**Fix — the discriminator is time, not the exception type.** The file already names the lever: *"the accepted lever for transients is a one-retry wrapper on non-`StructuredOutputError` failures in `_verdict`."* Retry once after a short sleep — a velocity 429 clears and the test runs; a dead wallet persists and the honest starved path fires. That also fixes finding 3.

---

### 3. The eval is flaky at n=2, and this PR is what makes that matter

**medium · confirmed · pre-existing, newly load-bearing**

`implied_success_data_never_arrived-haiku` **passed at 19:24 and failed at 19:39**, on a commit whose only change is `except TokenLimitExceeded` — which cannot alter a verdict. That is flake, not regression.

Those are the only two runs in the last 30 that ever executed. 11:25–17:08 is the silent skip path this PR fixes; 09:09–10:48 is the `--extra dev` PEP 735 break. **The executed-run sample is n=2 and they disagree.**

The assertion is strict across samples:

```python
_RUNS = 3 ; _STUCK_RUNS = 6
assert statuses == [case.expected] * n
```

~38 verdict calls per run, every one required to land exactly. At 1% per-sample flake the job flakes ~32% of the time; at 2%, ~53%.

*Refutation attempted:* the file anticipates this, documents the triage (*"re-run the failing case against `origin/staging` with the same key"*) and forbids the wrong fix (*"never looser assertions"*). Both right — but the remedy it names is not implemented, so the gate's reliability is currently unmeasured.

Why it matters now: this description argues a permanently-red required check is how checks get made non-required and then ignored. A **flaky** required check reaches that same end state by a different road, and this PR is what turns the flake from invisible into blocking.

**Fix:** same one-retry wrapper. Then re-run this branch several times to get n > 2 before calling the gate reliable.

---

### 4. `paths:` still excludes the workflow and `pyproject.toml`, but the stated reason has expired

**low-medium · confirmed**

The description defends the omission: *"both would widen the red-check surface before the key exists."* The key exists. The `--extra dev` → `--group dev` break the ticket documents at length is exactly what those triggers would catch, and the ticket flags that hazard as **still armed** (`main` and `staging` disagree on both halves).

**Fix:** add both here, or note in the description that the reason changed and it is being deferred anyway.

---

### 5. The marker string is a two-file contract held by comments — but it fails safe

**low · confirmed**

`_GATEWAY_UNAVAILABLE` in the test module and `MARKER` in the workflow heredoc, with "keep in sync" comments in both. If they drift, starved runs stop matching and fall to the **red** branch, which is the correct direction. Worth one line saying so — "keep in sync" reads scarier than the consequence warrants.

---

### Checked and produced nothing

- **`if: steps.mode.outputs.require_live == '1'`** — implicit `success()` means a red pytest skips the guard, matching the stated intent (*"a red pytest is already honest"*). Correct.
- **Missing or unparseable `eval-junit.xml`** → `ET.parse` raises → non-zero → red. Fail-safe.
- **Fork detection** — a null `head.repo` evaluates to "fork" → skip, the safe direction.
- **The 11th test** goes through the same `_verdict`, so it inherits the escape hatch consistently rather than erroring differently.
- **Security** — no new credential, no secret in output, the failure message names the env var and never its value; `contents: read`, `pull_request` not `pull_request_target`.

Reviewed: both diff files plus `anton/core/llm/openai.py` (429/402 branch ordering), `anton/core/llm/provider.py` (`classify_transient`), the full workflow, and the last 30 `verdict-eval` runs classified by why they ended.

alecantu7 (commented):
## Re-review — head `2ec9a51` (was `4e45275`)

Verified each fix by execution and by checking the claims against the other repo, rather than reading the commit messages.

### Verified fixed

| # | evidence |
| -- | -- |
| **1** stale description | now states the key exists, and that *"An earlier version of this description said 'red is the fail-loud mechanism demonstrating itself' — that is **false now and has been removed**"* | 
| **2** escape hatch | `_gateway_denial` keys on the gateway's real taxonomy — **verified against `mindshub_inference/minds/inference/errors.py` on `origin/main`**: `rate_limited` / `wallet_empty` / `included_allowance_exhausted` all exist, are emitted on `X-MindsHub-Reason`, and `Retry-After` rides only on the velocity one. Starved skips immediately; a throttle retries once honouring `Retry-After`, capped |
| **5** two-file marker contract | now enforced by a test that greps `MARKER` out of the workflow — and it runs in `run-tests`, so a workflow edit that breaks it fails on every PR |

Also confirmed the mechanism holds end to end: `raise ... from exc` appears 5× in `openai.py`, and nothing in `structured.py` / `client.py` re-wraps `TokenLimitExceeded`, so `__cause__` survives to where `_gateway_denial` reads it.

Full suite **1888 passed, 28 skipped** (up 48). The 11 new gate tests pass. `verdict-eval` executed for real (90s) on both new commits.

**Two things caught here that I missed in the first pass**, both real:

- **`IS_PR` fail-closed.** On any non-`pull_request` event `IS_FORK` compares an empty string and comes out `true` → `require_live=0` → the silent-skip bug restored. I read that block and did not see it.
- **`GITHUB_STEP_SUMMARY` via `.get`.** A `KeyError` inside the *we-could-not-run* branch would have turned an honest warning into a red for a third unrelated reason.

---

### One new finding

**A. The retry budget is per-call, not per-run — worst case ~37 minutes of sleeping**

**medium-low · confirmed by trace**

`retried_after_throttle` is initialised inside `_verdict` (`:234`), and `_verdicts` calls it `n` times per case. Per model: 4 cases × 3 + the STUCK case × 6 = 18; two models = 36; plus the narrating-budget test = **~37 `_verdict` calls**, each able to sleep up to `_THROTTLE_RETRY_CAP_S = 60`.

So under a *sustained* throttle the job sleeps **12 minutes** (at the 20s default) to **37 minutes** (at the cap), and then still reports green with zero cases executed. The retry genuinely fixes the common case — a short window that clears — but it converts the sustained case from a fast wrong-green into a slow one. There is no `timeout-minutes` on the job, so the default 360 applies.

**Fix:** a module-level counter bounding total retries per session — after K, stop pausing and skip. Keeps the per-case retry that does the work; bounds the pathological run.

---

### Still open from the first review

- **3 (flake).** Better quantified now. Five runs have genuinely executed: 19:24 ✅ · 19:39 ❌ (1 case) · 21:36 ✅ · 22:52 ✅ · 23:16 ✅ — **1 in 5 ≈ 20% job-level flake**. The retry addresses *throttles*, not rubric nondeterminism, so this one is untouched. The failing case passed in the four other runs, which tilts it toward flake rather than live drift.
- **4 (`paths:`).** Still excludes the workflow and `pyproject.toml`. Partly compensated now — the marker test reads the workflow and runs on every PR, so the highest-value drift is caught. The `--extra dev` → `--group dev` class still is not.

---

Net: in good shape. `mergeStateStatus: CLEAN`, all checks green. Finding A is a small bound rather than a redesign, and 3 is the one that decides whether this gate gets trusted or routed around once it is blocking.

alecantu7 (commented):
## Re-review — head `ba01dc5` (was `2ec9a51`)

**Both remaining findings closed.** I re-ran the mutation claims myself rather than taking the commit messages for them. Branch is now rebased onto current staging, so it carries #329 (ENG-1366) and #330 (ENG-1390).

### Finding A — closed

Session-wide budget (default 5) replaces the per-call one: worst case drops from ~37 minutes of sleeping to ~5 pauses. Plus `timeout-minutes: 20` on the job, which I had flagged as absent.

### Finding 3 (flake) — closed, and diagnosed rather than papered over

Two distinct fixture bugs, each with a stated cause:

1. **`_IMPLIED_SUCCESS` over-specified its outcome.** Now accepts `INCOMPLETE` or `STUCK`; `COMPLETE` / `WAITING` still forbidden.
2. **`_RECOVERED` asked an ambiguous question.** *"…compared to Tesla before Tesla's IPO"* could bind one figure or both; the answer read it one way and the verifier fairly split. Rephrased as two explicit asks, answer text untouched.

**I pushed hard on the widening**, because "accept two verdicts" is exactly what this file's own rule forbids. It survives the challenge: `_STUCK` (6 runs) and `_PARTIAL` are still single-valued, so the INCOMPLETE-vs-STUCK distinction remains pinned by dedicated cases. The widening is scoped to the one fixture where both verdicts genuinely prevent the originating incident, and a test forbids the set growing to include `COMPLETE`/`WAITING`.

### Mutations I ran independently

| mutation | caught by |
| -- | -- |
| add `COMPLETE` to the safeguard set | `test_only_the_hallucinated_success_case_accepts_two_verdicts` |
| widen a single-valued case (`WAITING`) | `test_single_valued_cases_still_demand_that_exact_verdict` (+1) |
| delete `timeout-minutes` | `test_the_workflow_bounds_the_job_so_a_hang_cannot_hold_a_runner` |
| never increment the session budget | `test_a_sustained_throttle_stops_pausing_once_the_session_budget_is_spent` |
| drift the transcript from `user_message` | `test_the_recovered_case_asks_one_unambiguous_question` |

All five caught by a **named** test. Full suite **1909 passed, 28 skipped**; 16 gate tests.

Worth recording a trap I hit: my first attempt at the transcript mutation silently failed to match and reported "16 passed" — I nearly wrote the pin up as broken. **A no-op mutation is indistinguishable from a surviving one**, so the mutation itself has to be confirmed applied before its result means anything.

### One small thing, not a finding

`_RECOVERED`'s comment still describes it as *"ENG-1134's originating incident shape (the SpaceX-vs-Tesla run)"*, but the user message is now a cleaned-up variant rather than the incident's actual text. The tool-error mechanic the case exists to guard is untouched, so this does not weaken it — one line noting the phrasing was disambiguated would stop a future reader treating it as verbatim from the incident.

And the residual the commit states itself, which is the right posture: three consecutive green runs does not prove the flake is gone. Watching it over the next several PRs before making the check required is the correct call.

Nothing left blocking from my side.

.github/workflows/verifier-eval.yml
```diff
@@ -43,7 +63,111 @@ jobs:
...
+          if starved and len(starved) == len(skips):
+              detail = starved[0][:300]
+              print(f"::warning::Verifier eval could NOT run — the CI key hit 402/429. {detail}")
+              # .get, not [], so a missing GITHUB_STEP_SUMMARY cannot KeyError
+              # inside the branch that handles "we could not run" — that would
+              # turn an honest warning into a red for a third unrelated reason.
+              summary = os.environ.get("GITHUB_STEP_SUMMARY")
+              with open(summary or os.devnull, "a") as fh:
+                  fh.write(
+                      "### ⚠️ Verifier eval did not run — CI key out of money\n\n"
```
pnewsam: **Nit (non-blocking):** a sustained velocity throttle gets reported to operators as "out of money."

When a `rate_limited` (429 velocity) denial outlives the retry budget, `_verdict` skips with the generic `_GATEWAY_UNAVAILABLE` marker (`tests/test_verifier_verdict_live.py:813`). If every case ends that way, this all-marked-skips branch fires and writes `CI key out of money` plus `Top up or raise the limit on the key` — guidance that misdirects for a throttle, which is not a wallet problem.

The behavior is the correct safe direction (green + non-blocking warning, never a false pass or false red), and the `GATEWAY_UNAVAILABLE` marker is deliberately generic — only this hardcoded human-facing text over-specifies the cause. Consider softening it (e.g. "the CI key hit 402/429 — out of money or throttled") so the honest-signal intent this PR is built on extends to the message text.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

I reviewed all four changed files: the `verdict-eval` workflow rewrite, the `_verdict` gateway-denial/retry logic and acceptable-set change in `test_verifier_verdict_live.py`, the new `test_verifier_eval_gate.py`, and the `.gitignore` line. The fail-closed mode decision, the three-outcome junit assertion, and the throttle-vs-starved classification are internally consistent, well-guarded, and backed by unit tests plus a real green CI run. No merge-blocking issues; one non-blocking cosmetic note left inline.

### Validation

- `gh pr checks 328` — all green, notably `verdict-eval` **pass 1m37s** (a real ~90s run, not the old ~15s skip path) and `run-tests` **pass 3m54s** (covers the new gate tests). Because `pull_request` uses the PR's own workflow, the new mode-decision + junit-assertion steps were exercised end-to-end here.
- Traced the mode logic across all four `IS_PR`/`IS_FORK` combinations: only a positively-identified fork PR reaches `require_live=0`; every other event falls to `require_live=1`. Fails closed as documented.
- Traced the junit assertion's three branches (`executed>0` → green; all-marked-skips → green+warning; else → red) against empty/all-skip/partial collections. A rubric regression stays red — it fails `test_verdict`, so the assertion step is skipped by the implicit `success()` and the job is red regardless.
- Traced `_verdict`'s restructured `while` loop: the ENG-1081 truncation escalation still advances `attempt`; the throttle retry holds `attempt` and is bounded per-call (`retried_after_throttle`) and per-session (`_throttle_retries_used < _THROTTLE_RETRY_BUDGET`). Terminates via return, `raise`, or `pytest.skip` — no unbounded loop.
- Confirmed the gate tests match the fixtures: `recovered_tool_error` stays `expected="COMPLETE"`, only `implied_success_data_never_arrived` widens to `("INCOMPLETE","STUCK")`, and the marker/timeout regex assertions match the shipped workflow.

The PR's stated risk — `verdict-eval` goes red on first-party PRs touching the watched files if `MINDSHUB_API_KEY` is ever absent — is the explicit, ticket-tracked design intent, and the key is clearly present today (the check ran green live).

.github/workflows/verifier-eval.yml
alecantu7: Fixed in 86cd969 — good catch, and it lands exactly where this PR is supposed to be sensitive.

The marker stays generic on purpose; it was only the prose that over-specified. The summary now names both causes and points at the reason string, which already distinguishes them:

> Two different causes land here, so read the reason above before acting: an empty wallet or spent allowance (`wallet_empty` / `included_allowance_exhausted`) needs credit, while a velocity throttle (`rate_limited`) that outlived the retry budget needs a higher rate limit or a quieter moment — not money.

Same over-specification was in the `::error::` branch — it said *"not because the key ran out of money"* where it meant *"not because of a gateway denial"*. Fixed too.

No logic change. Verified by extracting the guard from the shipped workflow and running it against a junit report carrying a `rate_limited` skip: still exits 0 with the warning, and the summary now sends the operator to the reason rather than to a credit card.

pnewsam (approved):
## Code Review (re-review — head `86cd969`)

**Verdict: APPROVE** (reaffirms my prior approval at `1b684b07`)

Re-reviewed only what changed since my last approval: one commit (`86cd969`) touching `.github/workflows/verifier-eval.yml` (+11/−6). It is a prose-only edit to two operator-facing message strings and resolves the cosmetic nit from my previous pass — the starved-run warning no longer blames "out of money" for what may be a velocity throttle. No logic, control flow, or exit code changed, so the earlier approval stands.

### Validation

- Diffed `1b684b07...86cd969` — the only hunk rewords the step-summary warning (now separates `wallet_empty` / `included_allowance_exhausted` → needs credit from `rate_limited` → needs a higher limit or a quieter moment) and the `::error::` text ("ran out of money" → "gateway denial"). Both sit inside `fh.write(...)` / `print(...)` string literals; `sys.exit(0)` and `sys.exit(1)` are untouched.
- `gh pr checks 328` — all green at `86cd969`: `verdict-eval` **pass 1m30s** (a real run, which exercises the reworded branch since `pull_request` runs the PR's own workflow), `run-tests` **pass 3m59s** (the gate tests), plus CodeQL / Snyk / build.
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

No actionable findings. The nit from my previous review is fixed.

Carryover, not a blocker (already flagged in the PR body): three consecutive green runs is a small sample. Worth watching the gate over the next several PRs before marking the check required.

PR #329, opened 2026-08-10
ENG-1366: load scratchpad snapshots that hold agent-authored modules

alecantu7:
Fixes [ENG-1366](https://linear.app/mindsdb/issue/ENG-1366/scratchpad-snapshots-fail-to-load-when-the-agent-writes-its-own-helper).

## The bug

ENG-1124 made the scratchpad namespace survive a turn boundary. The **write** side works. The **load** side discarded the entire namespace whenever it held an object whose defining module was a `.py` file the agent wrote into its own workspace.

Measured at **8% of cells** for one production user (22 of 29 failures naming the same helper, `campaign_engine`). It never surfaced as an error — the agent absorbs the loss by re-reading from disk — so the only symptom was a cache-cold reload costing ~$0.15–0.28 each time. That is why it survived the fix meant to end it.

## Root cause — two independent causes

**1. The module was unresolvable at load time.** `local.py` pins the subprocess cwd to the workspace, but Python puts the *script's* directory on `sys.path`, never the cwd. The agent adds cwd itself on turn 1 (`sys.path.insert(0, os.getcwd())`) and imports fine; that lives in memory and dies with the process. The next turn's loader runs before any cell, with a bare path — and dill stores those objects **by reference to their defining module**, so the import raises.

Broader than the ticket described: I verified all four reference shapes fail identically — the module object, an instance of an agent-defined class, the class, and a bare function. A namespace holding only `c = Campaign(...)`, never binding the module, is equally poisoned.

**2. The load was all-or-nothing.** A pickle stream is a sequential program; an exception part-way through kills everything after it. One bad reference destroyed every unrelated variable.

## The fix

**Cause 1 — `local.py` + `scratchpad_boot.py`.** `local.py` exports the resolved workspace root; the loader appends it to `sys.path` for the duration of the load.

- **Appended, never prepended.** The workspace is the agent's own directory and may hold files named after stdlib modules. I verified a workspace `types.py` shadows the stdlib when prepended and does not when appended.
- **Load-scoped** (`try/finally`), so the agent's own cell imports resolve exactly as before.
- **Set only from the explicit workspace arg, never `os.getcwd()`** — with no workspace bound, cwd is cowork-server's install directory. This matters because `ANTON_SCRATCHPAD_PERSIST_SESSION` is process-global, so the loader is reachable from contexts that never opted in (the credential probe among them); keying on the explicit arg keeps it inert there.

**Cause 2 — snapshot format.** The snapshot is now an envelope of individually-pickled values instead of one stream. A value that cannot be rebuilt is dropped and **named on the cell's `logs`** (never `error` — a snapshot problem must not feed the consecutive-error circuit breaker); everything else loads.

I tried to avoid the format change first, with an unpickler that substitutes a placeholder for any unresolvable reference. It got past all three bad references and died on the next instruction (`TypeError: Missing.__setstate__() missing 1 required positional argument`). No placeholder can satisfy every reconstruction protocol, so partial recovery genuinely requires the values to be independent on disk.

**Backwards compatible:** pre-existing snapshots still load, so a conversation in flight across the upgrade keeps its state instead of paying a cold turn.

**Simplification, not an added layer:** isolating values makes an unpicklable object ordinary rather than exceptional. This replaces the two-phase recovery that re-walked the whole namespace after a failed bulk dump. The size cap now aborts during encoding, before anything is written — strictly earlier than before.

## Why this format, measured first

The risk of per-value framing is snapshot inflation pushing a namespace past the 100 MB cap, whose failure mode is total (no snapshot at all). I measured both sides before committing:

| | Result |
|---|---|
| Snapshots that have ever hit the cap | **0** (399 turns, 12 users, 3 weeks) |
| Inflation, realistic mixed namespace | **1.03×** |
| Inflation, no sharing / slices / plain data | 1.00× |
| Inflation, worst realistic case (aliasing) | 2.00× |

So the tail this could push over does not exist. **Top-level alias de-duplication was considered and dropped**: measurement showed inflation comes from *nested* sharing, where top-level `id()` checks find nothing to dedupe — it would have added complexity for the cheapest case only.

**Known cost, stated plainly:** cross-key aliasing no longer survives a turn boundary (`df2 = df` restores as two objects), and peak memory during a dump is now ~snapshot-size higher, since values are encoded before writing. Both are bounded by the 100 MB cap and neither has cap pressure today.

## Tests

Three regression tests, **each watched failing on unfixed code** with the production traceback (`ModuleNotFoundError: No module named 'campaign_engine'` via dill's `_import_module`):

- `test_namespace_survives_an_agent_written_local_module` — the reported bug.
- `test_an_unresolvable_value_drops_only_itself` — helper deleted between turns, so no `sys.path` entry can save it; the rest of the namespace must survive and the loss must be named.
- `test_the_workspace_is_not_left_on_syspath` — restored *and* not left on the path; guards the load-scoped design against drifting process-wide.

Plus `test_a_snapshot_in_the_old_format_still_loads`, which passes on unfixed code by construction — it guards the new loader's backwards compatibility, not the bug.

`1899 passed, 28 skipped`. Two files fail identically with and without this change on clean `staging` (`test_chat_scratchpad.py`, `test_cloud_turn_process.py`) — pre-existing, verified by stashing.

## Security pass

Performed, covering:

- **Widening what the loader resolves.** Pickle load is already arbitrary code execution by design. The added path is the agent's own workspace, whose code already runs as the user, so no new execution surface. It is host-supplied (`conversation.project.path`), never model-influenced — the pad name reaches the snapshot *filename*, not this path — re-resolved with `realpath` and required to be a real directory.
- **Appending, not prepending**, so a workspace file cannot shadow the stdlib or site-packages.
- **The credential-probe path.** `ANTON_SCRATCHPAD_PERSIST_SESSION` is process-global, so the probe inherits persistence. It has no session id, so it gets no snapshot and the loader early-returns; the new env var is unused there.
- **No new secrets persisted.** The format change reframes the same values into the same per-conversation directory under the same retention.

## Delivery note

anton is vendored by cowork-server pinned to `branch = "main"`, so merging here ships nothing on its own — this needs a cowork-server bump and a desktop release to reach users.

**ENG-1366's Done-when #4** ("re-measure `aortiz`, load-failure rate drops to ~0") is currently **blocked by [ENG-1392](https://linear.app/mindsdb/issue/ENG-1392)** — since 2026-08-09, ~66% of production trace payloads are replaced by `[REDACTED:masking-error]`, and the cell `logs` carrying the failure text are exactly what is being redacted. Filed separately while measuring for this PR.

## Follow-up

ENG-1124's description states *"dill is not a second obstacle."* Its probe covered agent-defined functions declared **inside the exec namespace** (pickled by value, always fine) but never one whose home is an agent-written `.py` file. That sentence should be corrected on ENG-1124 so the next reader does not trust it.

### DIFF
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -596,6 +596,27 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
             # persist, rather than silently pretending it does.
             env.pop("ANTON_SCRATCHPAD_SESSION_PATH", None)
 
+        # The workspace root, for resolving agent-authored modules when the snapshot
+        # is loaded (ENG-1366). Python puts the *script's* directory on `sys.path`,
+        # never the cwd, so a helper the agent wrote into the workspace and imported
+        # on turn 1 (via its own `sys.path.insert`) is unimportable in the fresh
+        # process that loads the snapshot — and dill stores those objects by
+        # reference to their defining module, so the import failure discarded the
+        # whole namespace.
+        #
+        # Set ONLY from the explicit workspace arg, never from `os.getcwd()`: with no
+        # workspace bound the cwd is whatever launched the parent server (cowork's own
+        # install directory), which must never go on the scratchpad's `sys.path`.
+        # Persistence is process-global (`ANTON_SCRATCHPAD_PERSIST_SESSION`), so the
+        # loader is reachable from contexts that never opted in — the credential probe
+        # among them — and this keeps it inert for all of them.
+        if self._explicit_workspace_path is not None:
+            env["ANTON_SCRATCHPAD_WORKSPACE_PATH"] = str(
+                Path(self._explicit_workspace_path).resolve()
+            )
+        else:
+            env.pop("ANTON_SCRATCHPAD_WORKSPACE_PATH", None)
+
         _anton_root = str(Path(__file__).resolve().parent.parent.parent.parent)
         python_path = env.get("PYTHONPATH", "")
         if _anton_root not in python_path:
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -1,3 +1,4 @@
+import contextlib
 import io
 import json
 import os
@@ -26,6 +27,11 @@ PERSIST_SESSION = os.environ.get("ANTON_SCRATCHPAD_PERSIST_SESSION", "false").lo
 # aiming writes at somewhere unwritable. The caller owns the path: `local.py` composes
 # a per-pad, per-conversation one under the workspace and creates the directory.
 SESSION_PATH = os.environ.get("ANTON_SCRATCHPAD_SESSION_PATH", "")
+
+# The agent's workspace root, set by `local.py` ONLY when a real workspace is bound
+# (ENG-1366). Used to resolve agent-authored modules while the snapshot is loaded —
+# see `_workspace_on_syspath`. Empty means "no workspace", and every use is inert.
+WORKSPACE_PATH = os.environ.get("ANTON_SCRATCHPAD_WORKSPACE_PATH", "")
 
 # Snapshot size ceiling. Persisting after every cell costs nothing while it always
 # fails; once it works, a namespace holding large frames would be rewritten on each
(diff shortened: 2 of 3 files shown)


alecantu7 (commented):

Reviewed at `dcb12c7`, findings fixed in `a704e2e`. Recording the trail so the human reviewer sees what was already caught rather than re-deriving it. Deep-equivalent single-context review (no agent fan-out available in this session); `/security-review` run separately and clean.

---

### Finding 1 — silent namespace loss on rollback · **critical** · confirmed by execution · introduced by this PR · **fixed**

The v2 snapshot was originally a **dict**. An anton predating this format accepts *any* dict in the session file as the namespace itself (`scratchpad_boot.py` on `origin/staging`, the `isinstance(ns, dict)` branch), so it loaded the envelope **as** the agent's variables.

Executed against the real staging loader, with a snapshot this PR writes:

```
HAS master_df: False
HAS rates: False
junk names: ['__anton_snapshot__', 'values']
--- LOGS ---   ''            <- nothing reported
--- ERROR ---  None
```

Every real variable gone, junk names bound, and **completely silent** — which is precisely the failure class ENG-1124 was filed to end ("silently a no-op for ~2 months while reporting as enabled"). A release rollback, or an older desktop build resuming a conversation whose snapshot a newer build wrote, is enough to reach it.

**Refutation attempted:** checked whether the old loader validates shape beyond `isinstance(ns, dict)` (it does not); whether `_snapshot_payload` would strip the envelope keys (it never sees them); and whether anything else reads the snapshot content cross-repo — only `manager.py:79` does, and it reads `st_mtime` only, so it is content-agnostic. The finding survived.

**What I had missed:** the PR body claimed "backwards compatible" on the strength of new-reads-old. I had never tested old-reads-new.

**Fix (`a704e2e`):** the envelope is now a **tuple** (`scratchpad_boot.py:327`), which trips the old loader's `isinstance(ns, dict)` guard so it starts fresh *and* reports:

```
junk names: []
--- LOGS --- 'Failed to load scratchpad session; starting fresh. ... TypeError'
```

A snapshot from a *newer* anton is also now refused explicitly (`scratchpad_boot.py:222`) rather than guessed at, so forward skew degrades honestly too.

**Test:** `test_the_snapshot_is_not_a_bare_dict` — mutation-verified, fails when the envelope is reverted to a dict:
`AssertionError: where True = isinstance({'__anton_snapshot__': 2, 'values': {...}}, dict)`

---

### Finding 2 — test passed for the wrong reason · **medium** · confirmed by reading · introduced by this PR · **fixed**

`test_an_unresolvable_value_drops_only_itself` asserted:

```python
assert "c" in (cell.logs or "")
```

The letter `c` occurs throughout the note's own prose ("**c**ould not be rebuilt"), so this passed even if nothing had been named at all — it proved the note existed, not that it identified the casualty.

**Fix (`a704e2e`):** parses the rendered name list and asserts the exact set (`tests/test_scratchpad.py:1564`):

```python
named = logs.split("now undefined: ", 1)[1].split(".", 1)[0]
assert sorted(n.strip() for n in named.split(",")) == ["c", "campaign_engine"], named
assert "ModuleNotFoundError" in logs, logs
```

---

### Unverified risks (not fixed, stated deliberately)

| Risk | Status |
|---|---|
| `sys.path.remove` drops only the first occurrence, so a module that appends the same workspace path *during* the load could leave one copy beyond the load window | Judged unlikely; confirming needs a snapshot whose value's module mutates `sys.path` at import time |
| Cross-key aliasing no longer survives a turn (`df2 is df` -> `False`) | Known, accepted; inherent to per-value framing |
| Peak memory during a dump is ~snapshot-size higher (values encoded before writing) | Bounded by the 100 MB cap; measured namespaces ~12 MB, 0 cap hits in 399 turns |
| ENG-1366 Done-when #4 (re-measure to ~0%) | Blocked by ENG-1392, independent of this PR |

### What was reviewed

- Full execution paths for both causes, not just the diff: `local.py` env composition -> subprocess spawn -> `_load_namespace` -> `_decode_values`, and the dump path through `_encode_values` / `_write_snapshot` / `_CappedWriter`.
- **Repros executed:** old-reads-new (dict vs tuple envelope) against a real `origin/staging` worktree; the four dill reference shapes (module / instance / class / function); `sys.path` prepend-vs-append stdlib shadowing; per-value vs single-stream size inflation across 7 namespace shapes.
- **Tests run:** `tests/test_scratchpad.py` 96 passed; full suite 1900 passed / 28 skipped. `test_chat_scratchpad.py` and `test_cloud_turn_process.py` fail identically with and without this change on clean `staging` (verified by stashing) — pre-existing.
- **Cross-repo grep** for other readers of the snapshot file: only `manager.py:79`, `st_mtime` only.
- Base verified fresh against `gh api` before reviewing; base is `staging`, correct.
- CI green at `a704e2e`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed all three changed files in full: the `local.py` workspace-path export, the `scratchpad_boot.py` load/dump rewrite, and the five new regression tests. The one genuinely dangerous trap here — a dict envelope being silently mistaken for a namespace by a rolled-back older anton — was already found in the self-review (commit 2) and fixed by making the envelope a tuple, with a mutation-verified test pinning it. I could not confirm any blocking, major, or minor defect. No actionable findings.

### Validation

- Read the full `scratchpad_boot.py` (`_load_namespace`, `_decode_values`, `_encode_values`, `_write_snapshot`, `_dump_namespace`, `_workspace_on_syspath`) and `local.py` (`_session_snapshot_path`, `start` env setup) at head `a704e2e`.
- Confirmed all `dill.loads`/`dill.load` calls run inside the `_workspace_on_syspath()` context, covering both the per-value decode and the legacy single-stream path.
- Confirmed `_SESSION_EXCLUDE_ALWAYS` still drops `__builtins__` and `_anton_explainability_queries`, so the envelope carries only real variables.
- Confirmed every caller of the changed signatures (`_write_snapshot`, `_encode_values`, `_dump_namespace`) was updated consistently — no stale `payload`-dict call sites remain.
- Traced the fail-closed paths: `_TooBig` → discard tmp → "NOT saved" note; unresolvable value → dropped + named on `logs`, never `error`; newer-version envelope → refused → fresh + reported.
- CI at this head: `run-tests` pass, CodeQL pass, Snyk code/license/security all pass.

### Notes (not findings — disclosed and measured in the PR)

- Per-value framing shifts the effective 100 MB ceiling onto the summed individual pickles (~1.03× a single memoized stream), and peak dump memory rises by ~snapshot size. Measured at 0 cap hits across 399 turns / 12 users / 3 weeks, so no cap pressure today.
- Cross-key aliasing (`df2 = df`) no longer survives a turn boundary.

Both are inherent to the format change, measured, and stated plainly — not defects.

Outside the diff, worth keeping in view: anton is vendored by cowork-server pinned to `main`, so this merge ships nothing to users without a cowork-server bump + desktop release, and ENG-1366's load-failure metric won't move on this merge alone.

alecantu7:
## Addendum: the exact semantics of the aliasing loss

The PR body describes this as `df2 is df` becoming `False`, which understates it. Measured the real restored behaviour, since an `is` check is not what anyone actually relies on:

| Shape | Single stream (before) | Per-key (after) |
|---|---|---|
| `b = a`; mutate `a`, read `b` | `b.n=99`, `is` True | **`b.n=1`**, `is` False |
| Same dict nested in two containers | `y.v=99` | **`y.v=1`** |
| Two objects with a back-reference | `c.peer.name=RENAMED` | **`c.peer.name=p`** |
| Sharing inside **one** name (control) | `g.b.v=99` | `g.b.v=99` (preserved) |

**The rule is broader than top-level aliasing:** any object reachable from **two or more top-level names** restores as independent copies. Sharing contained within a single name is fully preserved — which covers most real object graphs (a dict of frames, a list of nodes, a nested config), and is why the realistic-namespace inflation came out at 1.03x.

**The effect that matters is silent mutation divergence, not identity.** Code that mutates through one name and reads through the other diverges after a restore, with no error and no log line. It also only appears *after* a turn boundary, so it presents as "worked, then stopped working". Equivalent to the agent having written `b = copy.deepcopy(a)`.

**Why it is not fixed rather than documented.** Preserving cross-name sharing and isolating per-value failures are mutually exclusive: objects that share state must occupy the same pickle stream to stay shared, and the same stream means the same failure fate — which is the all-or-nothing behaviour this PR exists to remove. Grouping co-referencing names into one blob would restore aliasing but let one bad object take its whole group down, reintroducing the original bug in miniature.

The trade is a measured 8%-of-cells total namespace loss against an unmeasured-but-near-zero sharing loss.

**Considered and rejected:** a prompt-contract line telling the agent about it. Unlike ENG-1124's "state does not survive a user message" — which is actionable, the response being "checkpoint to disk" — this one is not: the agent would have to notice it had aliased a mutable, predict a cross-turn mutate-then-read, and restructure around it. It would cost system-prompt space and attention on every request for a case measured at effectively zero. Recorded here for humans instead; revisit if it is ever observed biting.

**Also rejected:** detecting top-level aliases at dump time and noting them on `logs`. Free to compute, but it only catches row 1 of the table above — absence of a warning would then read as "sharing preserved", which would be false.

PR #330, opened 2026-08-10
ENG-1390: make the hidden rule-retrieval LLM call countable and its outcomes separable

alecantu7:
Closes Done-when 1 and 2 of [ENG-1390](https://linear.app/mindsdb/issue/ENG-1390). Done-when 3 and 4 were answered by measurement — [posted on the ticket](https://linear.app/mindsdb/issue/ENG-1390).

## The problem

`Cortex._retrieve_relevant_rules` (`cortex.py`) makes an **LLM call during prompt assembly** through the same `llm.code` entry point as the completion verifier and every other coding call. Two things were true and neither was knowable:

1. **It couldn't be counted** — no marker distinguished it in a trace.
2. **Its failure was indistinguishable from success** — `filtered_when = when_engrams` is reached three different ways (no exact match, empty response, exception), and all three are byte-identical in their effect on the prompt. A count of invocations could never separate a working filter from a broken one.

## Measured first, and the ticket's step order turned out to be backwards

The ticket assumed we had to instrument before we could count. We didn't: `_RULES_RETRIEVAL_PROMPT` is its own fingerprint, so the call was countable **retroactively from existing payloads** — no code change, no release, no week of waiting.

6,000 sampled GENERATIONs, 2026-07-16 → 08-06, zero redacted in-window, probe validity confirmed by a control:

| | |
|---|---|
| Rule-retrieval calls | **106 / 6,000 = 1.77% of all LLM calls** |
| Distinct users | 4 |
| Input tokens | median **4,258**, max 14,759 |
| **Latency added before the turn starts** | median **5.5s**, p90 **23.9s** |
| Recorded cost | $1.47 / 106 calls (5 unpriced, $0) |

So the answer is **"it fires"**, not the "rarely" the gates suggested. And the heaviest user is the same one as ENG-1366 — their "slow and expensive" profile has two mechanisms behind it, not one.

**The ticket's three-outcome taxonomy was incomplete.** Measured shares: 55.7% filtered, **30.2% dropped-all** (model returns `NONE`, every conditional rule discarded — a fourth outcome the ticket never named), 14.2% kept-all via no-exact-match. Plus **8% of calls ended at exactly 4,096 output tokens**, and cross-tabbing shows **8 of those 9 were classified as successful filters** — truncation masquerading as a working filter, a third silent-failure path.

## What this PR does

**A `rule-retrieval` trace tag**, via a new `tagged_trace()` context manager on the existing `TraceContext` — whose docstring already invites caller-supplied annotations, so this is the intended extension point rather than a new mechanism. Tags ride `Langfuse-Tags` and are one of the few dimensions the Langfuse metrics API can **group by**, so the call stays countable *without reading payloads* — which matters while ENG-1392 is redacting ~66% of them.

Deliberately a no-op rather than an error when there's nothing to annotate: no turn in flight (the CLI `turn()` path never installs a context — it's the only `set_trace_context` in the codebase), or a provider that ignores trace context (a BYOK user on direct Anthropic has no gateway trace to tag). Callers annotate unconditionally without branching.

**A `rule_retrieval` analytics event** carrying the call's shape: outcome, rule counts, block size, `stop_reason`, tokens, duration. Works for every user and provider, and is **the only way to observe the exception outcome at all** — a failed call leaves no observation, so the retroactive method is blind to it.

Both sinks, because neither alone is sufficient — the tag covers "distinguishable in a trace" (Done-when 1), the event covers "outcomes separable" (Done-when 2) and reaches non-MindsHub users.

**`stop_reason`, not `output_tokens == max_tokens`.** The exact-cap heuristic has known false positives, and the provider reason is authoritative. This is the field that separates a real filter from one that ran out of room.

## Deliberately not done

Per the ticket's out-of-scope list: **no change** to the filtering behaviour, the 6000/1000 thresholds, the silent fallback, or the exact-string-equality comparison that appears to cause it. Every branch returns exactly what it returned before; the only difference is that the branches are now distinguishable.

Also **not** added: a `rules` role in `TurnCost`/`EVENT_ROLES`. It would give per-turn token attribution and follows the precedent `router` sets, but the analytics event already carries the tokens, and per-turn cost attribution belongs with the cost question — which is the follow-up ticket's subject, not this one's.

## Security

Counts, sizes and enums only — never rule text, never the user message, both of which are user content. A test asserts neither can reach the sink and **pins the exact property set**, so a future field addition has to be deliberate. Every `extra` kwarg becomes a queryable PostHog property, so these names are a published schema; that's stated in the code comment.

Analytics failures are swallowed to `logger.debug` — observability must never break prompt assembly, which sits on the latency-critical path.

## Tests

11 new, in `TestRuleRetrievalObservability` (9 in the original commit, 2 added by the self-review), covering all five outcomes (filtered / dropped-all / kept-all-no-match / kept-all-empty / error), the trace tag plus that it does not leak past the call it annotates, the security property, and that the gates still short-circuit without a call.

**Each was watched failing on uninstrumented code.** The capture fixture uses `raising=False` deliberately, so code without the instrumentation fails with *"expected exactly one rule_retrieval event, got 0 — zero means the call is not reporting its outcome at all"* rather than erroring on a missing attribute: an error says the fixture broke, a failure says the property is absent.

Two of the eleven (`test_gates_still_short_circuit_without_an_llm_call` and `test_the_tag_reaches_the_actual_langfuse_header`) pass on uninstrumented code by construction — they are a behaviour-preservation guard and a coverage-gap closer respectively, not regression tests.

Worth noting: `_retrieve_relevant_rules` had **no test coverage at all** before this, which is ENG-1384's theme.

## Suite state

`1907 passed, 28 skipped`. `tests/test_datasource.py` has **2 failures that reproduce on clean `origin/staging`** (verified by checking out the base) — unrelated to this change, but staging is currently red on them and someone should look.

### DIFF
--- a/tests/test_cortex.py
+++ b/tests/test_cortex.py
@@ -240,3 +240,213 @@ class TestMaybeUpdateIdentity:
         await cortex.maybe_update_identity("Hi, I'm Jorge")
         assert (g / "profile.md").exists()
         assert "Name: Jorge" in (g / "profile.md").read_text()
+
+
+# ─────────────────────────────────────────────────────────────────────────────
+# ENG-1390 — the hidden LLM call in prompt assembly
+# ─────────────────────────────────────────────────────────────────────────────
+
+
+def _when_engrams(n: int, filler: int = 500) -> list[Engram]:
+    """Enough conditional rules to clear every gate.
+
+    Four must all hold before the call happens: a user message + an attached LLM,
+    formatted rules over `_RULES_BUDGET_CHARS` (6000), at least one `when` rule,
+    and a conditional block of at least 1000 chars. 12 x ~545 chars clears the
+    6000 budget; `filler=5` deliberately does not (see the short-circuit test).
+    """
+    return [
+        Engram(
+            text=f"When condition {i:03d} applies, do the thing {'x' * filler}",
+            kind="when",
+            scope="global",
+            confidence="high",
+        )
+        for i in range(n)
+    ]
+
+
+class _FakeLLM:
+    """Stands in for LLMClient.code, recording the trace tags it was called under."""
+
+    def __init__(self, content: str = "", raises: bool = False, stop_reason: str | None = None):
+        self._content = content
+        self._raises = raises
+        self._stop_reason = stop_reason
+        self.calls = 0
+        self.tags_seen: tuple[str, ...] = ()
+
+    async def code(self, **kwargs):
+        from anton.core.llm.provider import LLMResponse, Usage
+        from anton.core.llm.tracing import get_trace_context
+
+        self.calls += 1
+        ctx = get_trace_context()
+        self.tags_seen = ctx.tags if ctx else ()
+        if self._raises:
+            raise RuntimeError("provider exploded")
+        return LLMResponse(
+            content=self._content,
+            usage=Usage(input_tokens=4258, output_tokens=252),
+            stop_reason=self._stop_reason,
+        )
+
+
+@pytest.fixture()
+def captured_events(monkeypatch):
+    """Capture the analytics events the rule-retrieval path emits."""
+    events: list[dict] = []
+    import anton.core.memory.cortex as cortex_mod
+
+    # raising=False on purpose: on code without the instrumentation the hook does
+    # not exist, and we want these tests to FAIL with "no event was emitted"
+    # rather than ERROR on a missing attribute. An error says the fixture broke;
+    # a failure says the property under test is absent, which is the point.
+    monkeypatch.setattr(
+        cortex_mod, "_emit_rule_retrieval", lambda **shape: events.append(shape), raising=False
+    )
+    return events
+
+
+def _one_event(events: list[dict]) -> dict:
+    """The single rule-retrieval event, with a failure message worth reading.
+
+    Indexing `events[0]` directly fails as a bare IndexError, which tells a
+    future reader nothing about what broke.
+    """
+    assert len(events) == 1, (
+        f"expected exactly one rule_retrieval event, got {len(events)}: {events}. "
+        "Zero means the call is not reporting its outcome at all — which is the "
+        "defect ENG-1390 exists to close."
+    )
+    return events[0]
+
+
+class TestRuleRetrievalObservability:
+    """ENG-1390 — the call must be countable, and its outcomes distinguishable.
+
+    Before this, `filtered_when = when_engrams` was reached three different ways
+    (no exact match, empty response, exception) and all three were byte-identical
+    in their effect on the prompt — so a count of invocations could never
+    separate a working filter from a broken one. Measured in production at 1.77%
+    of all LLM calls before any of this existed.
+    """
+
+    async def _run(self, cortex, llm, rules):
+        cortex._llm = llm
+        return await cortex._retrieve_relevant_rules(rules, "please do the august campaign")
+
+    async def test_filtered_outcome_is_reported(self, cortex, captured_events):
+        rules = _when_engrams(12)
+        # Echo back exactly two rules, verbatim, as the prompt demands.
+        echo = "\n".join(f"- {rules[i].text}" for i in (0, 1))
+        llm = _FakeLLM(content=echo)
+        out = await self._run(cortex, llm, rules)
+
+        assert llm.calls == 1
+        assert len(out) == 2, "behaviour must be unchanged: only the echoed rules survive"
+        ev = _one_event(captured_events)
+        assert ev["outcome"] == "filtered"
+        assert ev["when_rules"] == 12 and ev["kept_rules"] == 2
+
+    async def test_dropped_all_is_its_own_outcome(self, cortex, captured_events):
+        """`NONE` discards every conditional rule — measured at 30% and absent
+        from the ticket's original three-outcome taxonomy."""
+        rules = _when_engrams(12)
+        out = await self._run(cortex, _FakeLLM(content="NONE"), rules)
+        assert out == [], "no mandatory rules present, so nothing should remain"
+        ev = _one_event(captured_events)
+        assert ev["outcome"] == "dropped_all"
+        assert ev["kept_rules"] == 0
+
+    async def test_no_exact_match_is_distinguishable_from_keeping_everything(
+        self, cortex, captured_events
+    ):
+        """The silent fallback. The model answered, but nothing matched verbatim —
+        so every rule is kept, which is indistinguishable in the PROMPT from the
+        filter deciding all rules were relevant. Only the event separates them."""
+        rules = _when_engrams(12)
+        # Plausible model output that fails exact string equality.
+        llm = _FakeLLM(content="- When condition 001 applies, do the thing (reworded)")
+        out = await self._run(cortex, llm, rules)
+
+        assert len(out) == 12, "behaviour must be unchanged: everything is kept"
+        ev = _one_event(captured_events)
+        assert ev["outcome"] == "kept_all_no_match"
+        assert ev["kept_rules"] == 12
+
+    async def test_empty_response_is_distinguishable(self, cortex, captured_events):
+        rules = _when_engrams(12)
+        out = await self._run(cortex, _FakeLLM(content="   "), rules)
+        assert len(out) == 12
+        assert _one_event(captured_events)["outcome"] == "kept_all_empty"
+
+    async def test_exception_is_reported_at_all(self, cortex, captured_events):
+        """The one outcome no trace could ever show: a failed call leaves no
+        observation to count, so retroactive measurement is blind to it."""
+        rules = _when_engrams(12)
+        out = await self._run(cortex, _FakeLLM(raises=True), rules)
+        assert len(out) == 12, "the permissive fallback must still apply"
+        assert _one_event(captured_events)["outcome"] == "error"
+
+    async def test_truncation_is_reported_via_stop_reason(self, cortex, captured_events):
+        """A verbatim echo cut at the ceiling is accepted as 'the relevant rules',
+        so truncation masquerades as a successful filter — 8 of 9 truncated calls
+        in production were classified that way. `stop_reason` is what separates
+        them; output_tokens == max_tokens is a heuristic with false positives."""
+        rules = _when_engrams(12)
+        echo = "\n".join(f"- {rules[i].text}" for i in (0, 1))
+        llm = _FakeLLM(content=echo, stop_reason="max_tokens")
+        await self._run(cortex, llm, rules)
+        ev = _one_event(captured_events)
+        assert ev["outcome"] == "filtered"
+        assert ev["stop_reason"] == "max_tokens", "a truncated filter must be separable"
+
+    async def test_the_call_is_tagged_so_it_is_isolatable_in_a_trace(self, cortex):
+        """Done-when #1: distinguishable from every other `llm.code` call."""
+        from anton.core.llm.tracing import (
+            TraceContext,
+            reset_trace_context,
+            set_trace_context,
+        )
+
+        rules = _when_engrams(12)
+        llm = _FakeLLM(content="NONE")
+        token = set_trace_context(TraceContext(session_id="conv-1", turn_id=3, harness="anton"))
+        try:
+            await self._run(cortex, llm, rules)
+        finally:
+            reset_trace_context(token)
+
+        assert "rule-retrieval" in llm.tags_seen, llm.tags_seen
+        # …and the tag must not leak past the call it annotates.
+        from anton.core.llm.tracing import get_trace_context
+
+        assert get_trace_context() is None
+
+    async def test_no_rule_text_or_user_message_reaches_analytics(self, cortex, captured_events):
+        """The ticket's security note: shape only — counts, sizes, enums."""
+        rules = _when_engrams(12)
+        await self._run(cortex, _FakeLLM(content="NONE"), rules)
+        ev = _one_event(captured_events)
+        blob = repr(ev)
+        assert "august campaign" not in blob, "user message must never be sent"
+        assert "do the thing" not in blob, "rule text must never be sent"
+        assert set(ev) == {
+            "outcome",
+            "when_rules",
+            "kept_rules",
+            "rules_chars",
+            "stop_reason",
+            "input_tokens",
+            "output_tokens",
+            "duration_ms",
+        }
+
+    async def test_gates_still_short_circuit_without_an_llm_call(self, cortex, captured_events):
+        """Below the budget the call must not happen — and must not be reported."""
+        llm = _FakeLLM(content="NONE")
+        small = _when_engrams(1, filler=5)
+        out = await self._run(cortex, llm, small)
+        assert llm.calls == 0 and captured_events == []
+        assert out == small
--- a/tests/test_cortex.py
+++ b/tests/test_cortex.py
@@ -424,6 +424,61 @@ class TestRuleRetrievalObservability:
 
         assert get_trace_context() is None
 
+    async def test_a_user_abort_is_not_reported_as_a_filter_failure(self, cortex, captured_events):
+        """`CancelledError` is a BaseException, so `except Exception` misses it —
+        but the `finally` still fires. Without a distinct branch every user
+        pressing STOP (or an abandoned SSE stream) was counted as `error`,
+        inflating the failure rate in the one metric this instrumentation exists
+        to produce. Found by adversarial self-review, reproduced before fixing."""
+        import asyncio
+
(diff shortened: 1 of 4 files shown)


alecantu7 (commented):

Reviewed at `94ec48b`, findings fixed in `0cdd3cb`. Recording the trail so the human reviewer sees what was already caught. Deep-equivalent single-context review (no agent fan-out available in this session).

**Triage signals:** async ContextVar manipulation; a new published analytics schema consumed by PostHog. ~160 non-test logic lines, 3 files, 1 subsystem.

---

### Finding 1 — a user abort was reported as a filter failure · **medium** · confirmed by execution · introduced by this PR · **fixed**

`asyncio.CancelledError` is a `BaseException`, so `except Exception` does not catch it — but the `finally` that emits the event still fires. Reproduced in a scratch worktree by cancelling a live task mid-call:

```
turn cancelled (as a user abort would)
events emitted: [{'outcome': 'error', 'when_rules': 12, 'kept_rules': 12, ...}]
>>> outcome reported = 'error'
>>> a user pressing STOP is counted as a FILTER FAILURE
```

Every user pressing STOP, and every abandoned SSE stream, would land in the books as a filter failure — **inflating the one metric this PR exists to produce.** For a measurement change that is the worst defect class available: it does not break the product, it breaks the number we would then make decisions on.

**Refutation attempted.** I tried to kill this by showing cancellation is unreachable during prompt assembly. It is not — `session.py:3054` documents that an abandoned generator already broke this exact area once:

> `# Emit BEFORE reset_trace_context: on an abandoned generator the finalizer runs in a copied context ... which used to abort this finally and drop the turn's books entirely (#309 review).`

The finding survived.

**Fix:** a distinct `except asyncio.CancelledError` branch reporting `outcome="cancelled"` and re-raising — cancellation is not ours to swallow. `kept_rules` is meaningless on that path (no decision was reached), which is why consumers must key on `outcome` first; noted in the code.

**Test:** `test_a_user_abort_is_not_reported_as_a_filter_failure` — mutation-verified, fails on the pre-fix code with *"a cancelled turn must be distinguishable from a filter that failed"*.

---

### Finding 2 — `tagged_trace` omitted a guard its sibling learned the hard way · **low** · hypothesis · introduced by this PR · **fixed**

`finally: _trace_ctx.reset(token)` had no `ValueError` guard for the copied-context case. `ChatSession.turn_stream` carries exactly that guard, added in #309 after an unguarded reset raised, aborted a `finally`, and dropped a whole turn's books.

Labelled a **hypothesis**, not confirmed: forcing a GC-finalized await chain to prove it is hard, and I did not manage it. But the fix is one line, the precedent is documented three files away, and new code should not ship missing a guard the codebase already paid for once.

**No test** — the scenario is what made it a hypothesis. Stated rather than papered over.

---

### Finding 3 — the tag test asserted the ContextVar, not the wire · **low** · **fixed**

Done-when #1 is about a **trace**, so the tag has to survive `_build_trace_headers` and its sanitiser (which drops commas and control chars) to actually count. The original test stopped at the ContextVar.

Verified by hand against the real builder during review:

```
WITHOUT tag: anton
WITH tag:    anton,rule-retrieval
tag present and intact: True
AFTER block: anton
no context -> None
```

Now pinned by `test_the_tag_reaches_the_actual_langfuse_header`, so an allowlist or a rename in that builder cannot break the ticket's requirement with every other test still green. **This test passes on pre-review code by construction** — it closes a coverage gap rather than fixing a defect, and is reported as such.

---

### Refuted during review — recorded so it is not re-raised

I suspected the module-level `AntonSettings` cache could ignore a per-project analytics opt-out that the sibling caller (`session.py`, which constructs fresh each turn) honors. **Dead:** `Workspace.apply_env_to_process` only sets keys *absent* from `os.environ`, so the value is sticky process-wide and a fresh construction reads the same thing. The cache is equivalent, not weaker.

---

### What was reviewed

- Full path, not just the diff: harness env → `_retrieve_relevant_rules` → `tagged_trace` → `OpenAIProvider._build_trace_headers` → wire, plus the analytics sink and its CI/opt-out gates.
- **Two repros executed** in a scratch worktree: real Langfuse header construction, and live task cancellation.
- **Behaviour equivalence traced branch by branch**, including that `early is not None` (rather than truthiness) preserves the original `NONE` → empty-`mandatory` early return.
- Tests: `tests/test_cortex.py` 36 passed; full suite `1727 passed, 28 skipped`.
- `tests/test_datasource.py` has 2 failures that **reproduce on clean `origin/staging`** — pre-existing and unrelated, but staging is red on them.
- `/security-review` not run separately: the diff adds no new input handling, and the one relevant surface (the analytics payload) is counts/sizes/enums only, pinned by a test that asserts the exact property set.

**CI needs a re-check at `0cdd3cb`** — the green run predates these fixes.

alecantu7:
## Correction to this PR's "Suite state" note

I wrote that `tests/test_datasource.py` has 2 failures which "reproduce on clean `origin/staging`" and that "staging is currently red on them and someone should look."

**The second half is wrong. Both tests PASS in CI**, verified in this PR's own `run-tests` log:

```
tests/test_datasource.py::TestConnectionMessagesIncludeLabel::test_save_message_includes_label PASSED [ 32%]
tests/test_datasource.py::TestConnectionMessagesIncludeLabel::test_needs_list_explains_label_field PASSED [ 32%]
```

So staging is **not** red, and nothing is blocking anyone. What is true: they fail consistently on my local machine, and they also fail at `aae099c` — the commit that introduced them (ENG-972 / #300) — so this is not a regression from anything that merged since.

What they assert: that the datasource connect flow reports `Saved as "prod-db"` using the label the user typed, and that its `needs:` list explains the `label` field. Locally the flow instead saves with the auto-derived label (`Saved as "postgresql" (slug "postgresql-...")`), so the typed label never lands.

Ruled out as the cause: colour/ANSI (`NO_COLOR`), terminal width (`COLUMNS`), `PYTHONHASHSEED` 0–5, and `uv run --group dev` (how CI invokes pytest). The `registry` fixture is hermetic — it parses a temp `datasources.md` — so it is not reading my `~/.anton` state either. **I did not isolate it**, and since it is green in CI and unrelated to this PR I stopped rather than keep digging.

Flagging the mistake specifically because a PR-body claim like "staging is red" gets acted on, and mine was not verified against CI before I wrote it — I had only run the suite locally. The same sentence appears in commit `94ec48b`'s message, which I cannot amend post-push; this comment is the correction of record.

tests/test_cortex.py
```diff
@@ -240,3 +240,268 @@ async def test_extracts_identity(self, dirs):
...
+        them; output_tokens == max_tokens is a heuristic with false positives."""
+        rules = _when_engrams(12)
+        echo = "\n".join(f"- {rules[i].text}" for i in (0, 1))
+        llm = _FakeLLM(content=echo, stop_reason="max_tokens")
+        await self._run(cortex, llm, rules)
+        ev = _one_event(captured_events)
+        assert ev["outcome"] == "filtered"
+        assert ev["stop_reason"] == "max_tokens", "a truncated filter must be separable"
+
+    async def test_the_call_is_tagged_so_it_is_isolatable_in_a_trace(self, cortex):
```
pnewsam, line 422: **Nit (test hygiene):** this test doesn't take the `captured_events` fixture, so it exercises the real `_emit_rule_retrieval` → real `send_event`. Harmless in practice (CI-drop + `analytics_enabled` gate + fire-and-forget daemon thread), but in a local non-CI run it can spawn a live analytics GET. Monkeypatching `send_event` here would make it hermetic. Purely optional.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (3 files, +429/−15) at head `0cdd3cb`: the new `tagged_trace()` context manager, the rule-retrieval instrumentation in `_retrieve_relevant_rules`, and the 11 new tests. The change is purely observational — every branch returns exactly what it returned before — so it carries no behavioral merge risk. No blocking or major findings.

### Validation

- Ran the new suite at head: `TestRuleRetrievalObservability` → 11 passed; full `test_cortex.py` + `test_analytics.py` → 42 passed.
- Behavior preservation by inspection: every non-cancel path resolves to the same value as the pre-PR code (`dropped_all`→`mandatory`, the three keep-all paths→`when_engrams`, `filtered`→matched subset). `early` uses a `None` sentinel, so an empty `mandatory` still returns correctly.
- Cancellation: `except asyncio.CancelledError` is listed first and re-raises, so a user STOP still propagates; the only added effect is the `finally` emitting `outcome="cancelled"`. `_emit_rule_retrieval` swallows internally, so it cannot mask the propagating error.
- Sink is safe on the latency path: `send_event` is fire-and-forget (daemon thread, 3s timeout, never raises), gated on `analytics_enabled` + CI-drop.
- Security: the 8-key property set is pinned by test; all values are counts/enums, and `stop_reason` is a provider enum, not user content.
- Tag reaches the `Langfuse-Tags` header, ContextVar scoping is task-local (no cross-turn tagging), and both callers consume the unchanged `list[Engram]` shape.

### Findings

No blocking or major issues. One inline nit below.

**Nit:** the PR body says “9 new” tests; the file adds 11 (`TestRuleRetrievalObservability` has 11 methods). The commit messages reconcile this (9 + 2 from the self-review) — just a stale count in the description.

Note: the disclosed `tests/test_datasource.py` failures reproduce on clean `origin/staging` and are unrelated to this diff — not held against this PR, but staging is red on them.

tests/test_cortex.py
alecantu7: Good catch — addressed in `8f7294d`, with a correction to my own first reading of it.

**What I changed:** an autouse fixture stubs `_emit_rule_retrieval` for every test in the class, so a test added later without asking for `captured_events` still cannot reach the real emit.

**Where I overstated it:** I first concluded this was worse than a nit — that a local suite run would inject synthetic `rule_retrieval` events into the production collector and pollute the very metric this PR creates. I proved the emit path really does issue a live GET when analytics is enabled:

```
GET https://…/zoomInfoCollector?action=rule_retrieval&aid=…
```

But that is the *function* in isolation. Under pytest nothing is sent, because **`tests/conftest.py` already disables analytics suite-wide** — and its comment records why: turn-cost tests began reaching the real sink, so someone already hit this and fixed it globally. I intercepted `urlopen` across the whole class both with and without my new fixture: **zero collector requests either way.**

So your "harmless in practice" was right and my escalation was wrong. Two probes disagreed and the pytest-hosted one was the correct one; I should have checked for an existing guard before asserting the consequence.

**Why keep the fixture anyway:** it is defence in depth rather than a fix. Hermeticity here no longer depends on that global kill switch surviving a future conftest change, and the failure mode if it were ever removed is contaminating this feature's own metric — a bad thing to leave resting on a file none of these tests mention. Docstring says exactly that, so the next reader does not inherit my wrong version.

evals/results/reasoning-sales-dip-01__20260625T174614.json
```diff
@@ -0,0 +1,59 @@
+{
```
pnewsam: **Major (accidental artifacts):** commit `8f7294d` adds 16 `evals/results/*.json` run artifacts unrelated to this PR (this file plus 15 siblings). They're June-dated eval run output (model config, token counts, `elapsed_seconds`), absent from `staging`, ungitignored, and unmentioned in the commit or PR — an accidental `git add`, not intended scope. They make up ~1000 of the +1019 new lines and would newly establish a tracked `evals/results/` directory.

Please `git rm --cached` the 16 files and recommit/force-push, and consider adding `evals/results/` to `.gitignore` so run output can't be swept in again. With those removed, the diff is back to the clean 3-file instrumentation change I approved at `0cdd3cb`.

pnewsam (changes requested):
## Code Review

**Verdict: REQUEST CHANGES** (re-review; reverses my earlier APPROVE at `0cdd3cb`)

This re-reviews the delta since `0cdd3cb`. The new head `8f7294d` does two things: it lands the autouse test-stub fixture I asked for (correct), and it **sweeps in 16 unrelated `evals/results/*.json` run artifacts** (~1000 of the +1019 new lines). The instrumentation code and tests are unchanged and still sound — the stray files are the one thing to fix before merge.

### Validation

- Diffed all three commits: `8f7294d` = the test fixture (17 lines) **+ 16 `added` eval JSONs**. That accounts for the jump from +429/−15 at my approval to +1448/−15 now.
- `evals/` (and `evals/results/`) does **not** exist on `staging` (contents API 404) and is not gitignored — so this PR would newly establish that directory as tracked.
- Sampled `honesty-data-absence-01__20260630T012006.json`: a machine-generated eval run artifact — June-dated, model config, token counts, `elapsed_seconds`. Transient output, unrelated to ENG-1390 (Aug 10), and unmentioned in the commit, PR body, or any review.
- CI green at head: `run-tests`, `build`, `CodeQL`, `scope`, `Analyze` all success.
- The autouse fixture routes `_emit_rule_retrieval` through the capture for every test in the class — exactly addresses the earlier nit, docstring correctly scopes it as defense-in-depth, hermetic.

### Findings

One Major finding inline (the stray eval artifacts). See the inline comment.

No other new issues — the code path, security property set, cancellation handling, and tag-to-wire coverage are unchanged from `0cdd3cb` and remain sound. Two prior non-blockers still stand and don't gate: the body's “9 new tests” is stale (the file has 11), and the `test_datasource.py` failures are pre-existing (already corrected in a PR comment confirming they pass in CI).

Drop the 16 files and this is back to APPROVE.

evals/results/reasoning-sales-dip-01__20260625T174614.json
alecantu7: You're right, and thank you — this was mine. Fixed and force-pushed as `ed80b0b`; the diff is back to the instrumentation change plus one `.gitignore` line.

**Cause:** I used `git add -A`. Those 16 files are untracked local eval output with mtime **Aug 10 15:39 — during this session** — written by a *concurrent session in this shared clone*, not by me and not by anything in the PR. `evals/` isn't tracked on `staging` at all, so `add -A` had nothing to compare against and swept the lot.

**Extent, checked before fixing** — contained to that one commit:

| | |
| -- | -- |
| Introduced by | `8f7294d` only (`--diff-filter=A` confirms) |
| On `origin/staging`? | no — 0 files |
| In the merged #329? | no — 0 files |

So nothing leaked into staging or the ENG-1366 merge, which was my first worry given I used the same `git add -A` there.

**What I did:** `git reset --soft HEAD~1`, unstaged `evals/`, recommitted, `--force-with-lease`. The 16 files are **left untouched on disk** — they belong to whichever session is mid-run, and deleting them to tidy my diff would have destroyed someone else's work in progress.

**Added `evals/results/` to `.gitignore`** as you suggested, so run output can't be swept in again. Flagging that as the one piece of scope in this PR that isn't ENG-1390 — say the word and I'll split it out, though it seems worth having attached to the commit that proves it was needed.

**The real fix is on my side:** `git add -A` in a clone that concurrent sessions write to is unsafe, and reviewing my own diff would have caught 1000 added lines across 16 unrelated files. Explicit pathspecs from here.

pnewsam (approved):
## Code Review

**Verdict: APPROVE** (re-review at `ed80b0b`; clears my earlier REQUEST CHANGES at `8f7294d`)

Re-reviewed the delta since my change request. The 16 stray `evals/results/*.json` artifacts that were my one Major finding are gone — the diff is back to the clean 4-file change (`.gitignore`, `tracing.py`, `cortex.py`, `test_cortex.py`). Both of my prior asks are now satisfied, and CI has caught up. No blocking or major findings remain.

### Validation

- **File list at head is clean:** the accidental run artifacts are fully removed — the diff is 4 files, no `evals/` paths.
- **Instrumentation is untouched since I approved it:** `compare 0cdd3cb...ed80b0b` shows only `.gitignore` (+3) and `tests/test_cortex.py` (+17). `cortex.py` and `tracing.py` are byte-identical to the version I validated at `0cdd3cb`, so the code path, cancellation handling, security property set, and tag-to-wire coverage all still hold.
- **Earlier nit fixed:** the autouse `_stub_the_analytics_emit` fixture routes `_emit_rule_retrieval` through the capture for every test in the class, so a future test added without asking for `captured_events` still can't reach the real emit. Correctly scoped in its docstring as defense-in-depth, not a live-leak fix.
- **`.gitignore` guard is correctly scoped:** ignores `evals/results/` specifically, not all of `evals/`. Nothing is tracked under that path on `staging`, so it removes nothing and only prevents a recurrence.
- **CI:** `build`, `CodeQL`, `Analyze`, `scope`, and all three Snyk checks pass at head.

### Findings

No actionable findings. One carried-over non-blocker, unchanged: the PR body still says "9 new" tests where the file adds 11 (reconciled by the commit trail: 9 + 2 from the self-review). Cosmetic, doesn't gate.

PR #331, opened 2026-08-11
ENG-1357: stop advertising a picker that cannot render

alecantu7:
## Problem

`select_path`'s BROWSE mode cannot work in Cowork. `handle_select_path` returns `picker_unavailable` before it ever reaches the elicitor, because cowork-server injects none. But the tool advertised browse as the right move for an unknown location **and** injected a system-prompt rule closing every alternative:

> "call the `select_path` tool to let them pick it — do **NOT** ask them to paste or type a path, and do not guess. Browse mode (no candidates/pattern) is the right choice when you don't know where it is."

So a user who said *"I have my historical sales data, train a model to forecast next month"* — without attaching the file — left the model with no legitimate move. Picker dead. Asking for a path forbidden by that rule *and* by the harness file-access policy. Guessing forbidden. Attachment never mentioned to it.

It generated a synthetic 36-month dataset, trained three models, and reported `R² 0.920` with a next-month forecast of **$69,442** as her analysis. The completion verifier graded the turn `COMPLETE`. She had no way to detect it from the output, and never came back.

Fabrication was the only door left open. This PR shuts it and opens a real one.

## Changes

**1. Registration is gated on elicitor capability** (`session.py`), mirroring the `ASK_USER_TOOL` idiom immediately above it:

```python
if self.elicitor is not None and "path" in self.elicitor.supported_kinds:
    self.tool_registry.register_tool(SELECT_PATH_TOOL)
else:
    self.tool_registry.register_tool(SELECT_PATH_TOOL_PICK_ONLY)
```

This swaps the **injected prompt**, not just the description — `ToolDef.prompt` is collected from the *registered* defs by `ChatSystemPromptBuilder._build_tool_prompts_section`, so the rule steering the model at a dead picker disappears with it. That's the half that actually caused the incident.

**2. `picker_unavailable` is a documented status** in both definitions, with a per-mode remedy — previously the model received a status the description never mentioned, carrying advice in prose.

**3. The browse remedy names attachment** instead of "ask the user for the path in plain text" — which the harness policy forbids regardless, since a browse-mode file is by definition one the agent could not find in the project.

## Two corrections to the ticket

Both from reading the handler rather than the write-up:

**PICK mode degrades better than the ticket implies.** Without an elicitor: one match auto-resolves, zero returns `no_matches`, and **≥2 returns `picker_unavailable` with the candidate list**. "Ask which of these you meant" is legitimate there — those are in-project paths the agent already found, not a path the user must type from memory. So only the BROWSE remedy was wrong. Pick's is untouched, and `test_ambiguous_pick_still_asks_which_candidate` fails if someone rewrites it wholesale.

**The `no_matches` message was also lying** — not mentioned in the ticket. It advised *"omit candidates/pattern to let the user browse"*, which on a pathless host lands straight back on `picker_unavailable`. Now conditional on `can_pick`.

This is a **narrowing, not a removal**. The ticket offers "drop the tool or describe it as pick-only"; dropping it would cost the auto-resolve and disambiguation paths, which work fine. Same tool name in both variants, so `test_ask_user.py`'s `assert "select_path" in names` still holds.

## Verification

Mutation-verified, each change independently:

| Mutation | Result |
| -- | -- |
| revert the registration gate | **2 fail** |
| restore "ask the user for the path in plain text" | **1 fail** |
| remove `picker_unavailable` from the status docs | **1 fail** |

`test_select_path.py`: **25 passed** (16 pre-existing + 9 new). Full suite: **1939 passed**. The 2 failures in `test_datasource.py::TestConnectionMessagesIncludeLabel` are pre-existing — confirmed identical with these changes stashed.

One process note, since it nearly cost me a false claim: my first attempt at the third mutation **passed**, because `picker_unavailable` appears twice in that description and I had only removed one occurrence. The test was sound; the mutation wasn't. Redone properly, it fails as it should.

## Scope

Ticket items 2 and 3 of Phase 1. **Item 4 deliberately excluded** — the "never substitute synthetic data" prompt rule and its regression eval. A prompt rule without the eval is unverifiable, and shipping half would make the item look handled; it needs its own ticket.

Companion PR **cowork-server#292** covers item 1 (tell the agent attachment exists when nothing is attached). **Independent, not stacked** — either can merge first. Together they close Phase 1: the agent stops being offered a tool that can't work, and starts being told the route that can.

Phase 2 — build the GUI picker or retire browse mode from Cowork entirely — remains a product call on ENG-1357. Worth noting this PR is effectively "retire it, code left in place", so it buys time on that decision without user-facing harm.

## Security

No credential, injection, path-traversal or authz surface introduced. All three changes are prompt/description text plus one registration branch; no change to which paths the tool will resolve or read. The `_browse_start_dir` project-root default and the harness file-access policy are untouched — this PR narrows what the agent is told it can do, never widens it.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -77,6 +77,7 @@ from anton.core.tools.tool_defs import (
     RECALL_TOOL,
     SCRATCHPAD_TOOL,
     SELECT_PATH_TOOL,
+    SELECT_PATH_TOOL_PICK_ONLY,
     UPDATE_ARTIFACT_METADATA_TOOL,
     ToolDef,
 )
@@ -1351,10 +1352,25 @@ class ChatSession:
 
         self.tool_registry.register_tool(scratchpad_tool)
         self.tool_registry.register_tool(READ_IMAGE_TOOL)
-        # Interactive file/folder disambiguation — always available; degrades
-        # to picker_unavailable (and still auto-resolves a single candidate)
-        # when no elicitor supports path questions.
-        self.tool_registry.register_tool(SELECT_PATH_TOOL)
+        # Interactive file/folder disambiguation. Registered on every host, but
+        # in one of two shapes — because "degrades to picker_unavailable" is
+        # not a neutral degradation when the definition also injects a
+        # system-prompt rule telling the model to use the picker INSTEAD of the
+        # alternatives.
+        #
+        # Without a path elicitor, BROWSE mode cannot run at all, so a user who
+        # says "analyse my sales data" without attaching it left the model with
+        # no legitimate move: picker dead, asking for a path forbidden by the
+        # prompt AND by the host's file-access policy, guessing forbidden. It
+        # fabricated the data and reported a forecast from it (ENG-1357).
+        #
+        # PICK mode does still work here — one match auto-resolves, and ≥2
+        # comes back with the candidate list to ask about — so this is a
+        # narrowing, not a removal.
+        if self.elicitor is not None and "path" in self.elicitor.supported_kinds:
+            self.tool_registry.register_tool(SELECT_PATH_TOOL)
+        else:
+            self.tool_registry.register_tool(SELECT_PATH_TOOL_PICK_ONLY)
 
         # Multiple-choice questions — only when some host can actually render
         # them. Without this the model would ask into a void. The cowork kill
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -729,6 +729,14 @@ class ChatSession:
         # Host-injected first; a console-backed session (the real CLI,
         # chat.py passes console=console) falls back to CLIElicitor so it
         # always gets a working elicitor.
+        #
+        # Inject through `config.elicitor` — NEVER by setting
+        # `session.elicitor` after construction. Registration below picks the
+        # select_path variant from this value (full vs pick-only), so a late
+        # setter leaves the model told there is no file browser while browse
+        # would actually work at runtime: the capability exists and the agent
+        # never learns it. That failure is silent in both directions, which is
+        # what makes it worth a comment rather than a docstring.
         self.elicitor: Elicitor | None = config.elicitor
         if self.elicitor is None and config.console is not None:
             from anton.core.interaction.cli import CLIElicitor
--- a/anton/core/tools/tool_defs.py
+++ b/anton/core/tools/tool_defs.py
@@ -558,6 +558,24 @@ SELECT_PATH_TOOL_PICK_ONLY = replace(
         "conversation; this host has no file browser, so do not ask them to "
         "navigate to it and do not ask them to type or paste a path."
     ),
+    # `replace()` copies input_schema by reference, so the schema must be
+    # overridden too — otherwise the model is told two different things by the
+    # two channels it reads: prose saying "there is no BROWSE mode" and a
+    # function-calling schema still offering `start_dir`, whose own description
+    # reads "BROWSE mode only". Dropping it is the point of the variant.
+    #
+    # Rebuilt rather than mutated, for the same reason the ToolDef itself is
+    # copied: `SELECT_PATH_TOOL.input_schema` is a module-level dict shared by
+    # every session in the process. The inner property dicts are shared but
+    # never written, so a one-level rebuild is enough.
+    input_schema={
+        **SELECT_PATH_TOOL.input_schema,
+        "properties": {
+            key: value
+            for key, value in SELECT_PATH_TOOL.input_schema["properties"].items()
+            if key != "start_dir"
+        },
+    },
 )
 
 
--- a/anton/core/tools/tool_defs.py
+++ b/anton/core/tools/tool_defs.py
@@ -13,7 +13,7 @@ from anton.core.tools.tool_handlers import (
     handle_update_artifact_metadata,
 )
 
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Callable, Optional
 
 
@@ -459,8 +459,12 @@ SELECT_PATH_TOOL = ToolDef(
         "immediately with no prompt; zero matches tells you to refine.\n\n"
         'On selection the tool returns {"status":"resolved","path":"<absolute path>"} '
         "— use that path directly and keep going. Other statuses: 'cancelled' (user "
-        "dismissed), 'no_matches', 'invalid'. Never re-ask in plain text after a "
-        "resolved selection."
+        "dismissed), 'no_matches', 'invalid', and 'picker_unavailable' (this host "
+        "cannot render a picker). On 'picker_unavailable' follow the `message` in the "
+        "result: in PICK mode it carries the `candidates` it found, so ask which of "
+        "those the user meant; in BROWSE mode the file is somewhere you cannot reach, "
+        "so ask the user to attach it to the conversation. Never re-ask in plain text "
+        "after a resolved selection."
     ),
     # Injected into the system prompt: bias the model toward the picker over a
     # type-the-path request, which is the whole point of the tool.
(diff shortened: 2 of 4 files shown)


alecantu7 (commented):
## Self-review (adversarial) — head `718fbfd`, base `origin/staging`

Reviewed together with its sibling **cowork-server#292**. **No defects found.** Details below are the evidence for that, not a formality.

### ⚠️ The red `verdict-eval` is inherited, not caused by this PR

It failed on `implied_success_data_never_arrived` with `[INCOMPLETE, STUCK, INCOMPLETE]` — the ENG-1334 flake diagnosed and fixed earlier tonight. Confirmed rather than assumed:

- the fixture at this head is `expected="INCOMPLETE"` — the pre-fix single-valued form
- `git merge-base --is-ancestor ba01dc5 <this head>` → **not an ancestor**

**Merge anton#328 first, then rebase this**, and it goes green. Nothing to fix here.

### What I verified rather than assumed

**The state-gated claim holds workspace-wide.** Per the ENG-597 rule, I grepped `elicitor =` across `anton`, `cowork-server`, `cowork`, `anton_services`, `mdb-ai` and `cowork_evals` at `origin/staging`. **Only readers exist** — `getattr(session, "elicitor", None)` in three places. Nothing outside anton ever assigns it, so the ticket's central claim is true and the gate cannot be defeated by a host that sets it late today.

**The registration gate and the runtime gate cannot disagree.** This is where I expected the bug — a registration-time check drifting from the runtime one. They are the *identical* predicate on the same attribute:

```python
session.py:1370       if self.elicitor is not None and "path" in self.elicitor.supported_kinds
tool_handlers.py:819  can_pick = elicitor is not None and "path" in getattr(elicitor, "supported_kinds", ())
```

Ordering is deliberate too — `self.elicitor` resolves at `session.py:732`, above registration, and the CLI's `console → CLIElicitor` fallback is set there, so the standalone CLI keeps working browse.

**`replace()` is safe.** `SELECT_PATH_TOOL_PICK_ONLY` is a dataclass copy, so anything identity-comparing the original would silently take the wrong branch. Grepped: the only references to `SELECT_PATH_TOOL` are the import and the registration. No comparisons.

**Tests, mutation-verified.** 25 select_path tests; full suite **1902 passed, 28 skipped**. Reverting the gate to the unconditional `register_tool(SELECT_PATH_TOOL)` fails two *named* tests — `test_pathless_host_gets_the_pick_only_definition` and `test_no_elicitor_at_all_gets_the_pick_only_definition`.

**Safe in either order relative to #292.** This PR alone: browse is no longer advertised and the handler messages name attachment. #292 alone: the agent still sees the full definition, but the harness now names attachment, so the dead end has an exit. Neither depends on the other to be correct — which matters, because cowork-server pins anton as `branch = "main"` while this targets `staging`, so #292 will run against an anton without this change until the next staging→main release.

### Two notes, neither a finding

**The ticket's `selection.py` Done-when is already satisfied — by a refactor, not by this PR.** `anton/core/interaction/selection.py` no longer exists on staging (now `elicit.py` / `cli.py` / `emitter.py`), and the false *"the cowork-server harness injects a streaming picker"* claim is gone from the tree entirely. Worth striking from the Done-when so nobody hunts for it. Same drift applies to the ticket's `tool_handlers.py:686` / `_resolve_selection_elicitor` citations — written against `origin/main` `b34dfd9`, same mechanism, moved lines.

**A forward-looking trap for phase 2.** `self.elicitor` resolves once in `__init__`. If a future cowork picker is wired by setting `session.elicitor` *after* construction, registration will already have chosen `PICK_ONLY` — browse would work at runtime while the model was never told it exists. Phase 2 must inject via `config.elicitor`. The comment says "Resolved once, here" but not why a late setter breaks it; one clause would save someone an afternoon.

### Merge order across the ticket

```
cowork-server#291  →  cowork-server#292      (#291 also restores staging publishing — ENG-1402)
anton#328          →  anton#331
```

alecantu7:
Addressed the one actionable note from the self-review — pushed `d0bd019`.

**"A forward-looking trap for phase 2"** — the late-setter hazard is now stated at the assignment itself (`session.py:727`), not left implicit:

> Inject through `config.elicitor` — NEVER by setting `session.elicitor` after construction. Registration below picks the select_path variant from this value (full vs pick-only), so a late setter leaves the model told there is no file browser while browse would actually work at runtime […] silent in both directions.

Comment only, no behaviour change. 25 select_path tests still pass.

**On the `verdict-eval` diagnosis — confirmed, with one correction worth recording.** `#328` is indeed not an ancestor of this head, and the failure is the `implied_success_data_never_arrived` label flake, so nothing here needs fixing. But the eval **did execute**: `1 failed, 10 passed in 75.05s`. That is 11 real tests, not ENG-1334's ~13-second silent skip — so the credential now exists and the gate is armed. All three sampled verdicts give the same reasoning (*"claimed success, the only tool call failed, the answer is fabricated"*); the judge just chose `STUCK` over `INCOMPLETE` on one sample, and the fixture demands one exact label three times.

Which makes this a live instance of ENG-1334's *second* theme rather than its first: the gate is now unreliable in the red direction. `run-tests=SUCCESS` is the signal that actually pertains to this PR.

The other two notes I agree with and have not acted on, deliberately:

* **The `selection.py` Done-when is already satisfied** by the `elicit.py` refactor, and the ticket's `tool_handlers.py:686` / `_resolve_selection_elicitor` citations are stale against `b34dfd9`. That belongs on ENG-1357 as a description edit, not in this diff.
* **Merge order** stands as `#328 → #331` here and `#291 → #292` in cowork-server.

StpMax (approved):
Approve, with one fix requested before merge.**

## Summary
Gates `select_path` registration on elicitor capability (`session.py`), mirroring the existing `ASK_USER_TOOL` pattern, and introduces `SELECT_PATH_TOOL_PICK_ONLY` as a `dataclasses.replace()` copy with browse mode removed from both the tool description and the injected system prompt. `handle_select_path`'s `picker_unavailable` and `no_matches` messages are updated to point at attachment instead of "ask for the path in plain text," conditioned correctly on the same `can_pick` predicate used at registration time. This directly closes the trap described in the ticket: the model is no longer told to use a picker that structurally cannot render in Cowork, and no longer steered toward a remedy (typed path) that the harness policy forbids anyway.

## Correctness check
- The registration-time gate (`session.py`) and the runtime gate (`tool_handlers.py`'s `can_pick`) are the identical predicate (`elicitor is not None and "path" in elicitor.supported_kinds`) evaluated against the same `session.elicitor`, resolved once in `__init__` before registration — so there's no window where the two can disagree, and the late-setter hazard (setting `session.elicitor` post-construction) is now called out at the assignment site.
- `SELECT_PATH_TOOL_PICK_ONLY` is built via `dataclasses.replace()`, not a mutation, so the module-level `SELECT_PATH_TOOL` singleton stays pristine across sessions — confirmed by the diff and by `test_module_singletons_are_not_mutated`.
- Message correctness for both modes checks out: browse-mode `picker_unavailable`/`no_matches` now name attachment; pick-mode's "ask which of these you meant" for ≥2 ambiguous candidates is untouched and correctly left untouched — those are project-internal paths already found, not something the user must type from memory, so the harness's file-access policy doesn't apply there.

## Bug: `SELECT_PATH_TOOL_PICK_ONLY`'s `input_schema` still advertises a BROWSE-only parameter
`replace(SELECT_PATH_TOOL, description=..., prompt=...)` only overrides `description` and `prompt` — `input_schema` is inherited unchanged from `SELECT_PATH_TOOL`. That schema still contains:

```python
"start_dir": {
    "type": "string",
    "description": "BROWSE mode only: directory to open the picker at "
    "(absolute, or relative to the project root). Defaults to the project root.",
},
```

So the pick-only tool definition tells the model, in prose, "there is no BROWSE mode" — while its actual function-calling schema exposes a parameter explicitly documented as "BROWSE mode only." That's a real inconsistency between the two channels the model sees (system-prompt description vs. tool parameter schema), on a host that is exactly the one this PR is trying to stop confusing.

It's not a crash risk — if a model passes `start_dir` alone with no `candidates`/`pattern`, `handle_select_path` still falls into the browse branch, `can_pick` is still false, and it still correctly returns `picker_unavailable` pointing at attachment. So the *handler* is safe either way. But it undercuts the fix's stated goal ("don't advertise a tool that can't work") and isn't covered by the new tests, which only assert against `tool.description` / `tool.prompt` strings, never `tool.input_schema`.

**Suggested fix:** also override `input_schema` in the `replace()` call to drop (or repurpose) `start_dir` for the pick-only variant, and add a test asserting it's absent — the same way `test_pathless_host_gets_the_pick_only_definition` already asserts on `description`/`prompt`.

## Minor nit
`handle_select_path`'s docstring (not user/model-facing, just an internal comment) still describes browse mode unconditionally as available ("the user navigates a picker to locate it... Use this instead of asking the user to type or paste a path") without mentioning the host-gated behavior added in this PR. Harmless, but slightly stale now that browse is conditional.

## Process notes
- The `verdict-eval` red check is diagnosed convincingly as an inherited flake (ENG-1334), confirmed via `git merge-base --is-ancestor` rather than assumed — good practice, not just asserted.
- The self-review's claim that "only readers exist" for `session.elicitor` workspace-wide (grepped across 6 repos) is a reasonable way to confirm the phase-2 late-setter trap doesn't already exist somewhere unexpected.

alecantu7:
@StpMax thanks — good catch, and you found the one place my evidence didn't reach. Both points fixed in `45da76f`.

## The `input_schema` bug

You're right, and the diagnosis is exact: `replace()` copies `input_schema` by **reference**, so overriding only `description` and `prompt` left the pick-only variant advertising `start_dir` — description "BROWSE mode only" — while its prose says there is no browse mode.

What makes it a fair hit rather than a nit: I was deliberately careful that `replace()` not mutate the singleton's *description*, and then missed that `input_schema` is shared the same way. And my new tests asserted on `description` and `prompt` and **never on `input_schema`** — the channel the change actually altered. Asserting on the convenient channel instead of the changed one is precisely the failure mode I'd flagged in another review this week.

Fixed by overriding it, rebuilt one level rather than popped:

```python
input_schema={
    **SELECT_PATH_TOOL.input_schema,
    "properties": {
        key: value
        for key, value in SELECT_PATH_TOOL.input_schema["properties"].items()
        if key != "start_dir"
    },
},
```

Rebuilt for the same reason the ToolDef itself is copied — that dict is module-level and shared by every session in the process, so `pop` would leak across sessions exactly as an in-place description edit would. Inner property dicts stay shared but are never written.

**Tests now cover the schema channel in three places**, since one assertion would have been the same mistake in a new spot:

- `test_pathless_host_gets_the_pick_only_definition` — `start_dir` absent, and the four pick parameters still present
- `test_host_with_a_path_elicitor_keeps_browse_mode` — `start_dir` present there, and only there
- `test_module_singletons_are_not_mutated` — the full definition still has it, and the two `properties` dicts are distinct objects

Mutation-verified: removing the override reproduces your bug (`start_dir present: True`) and fails **2** tests, including the singleton one.

## The docstring nit

Also fixed. `handle_select_path`'s docstring described browse as unconditionally available; it now states the host gate and names `SELECT_PATH_TOOL_PICK_ONLY`. You're right that it went stale the moment the gate landed.

## State

`45da76f`, rebased on `staging` (`cdb05d7`, so #328 is in). Full suite **1955 passed**; the 2 `test_datasource.py::TestConnectionMessagesIncludeLabel` failures are pre-existing and fail identically with my changes stashed. `verdict-eval` was green on the previous head with **11/11 executed in 89s** — a real pass, not the silent skip, which #328 now makes visible.

Your approval predates these two commits. The changes are exactly what you asked for and are test-only plus a schema override, but since the schema override is the part you identified, flag it if you'd rather look again before this merges — happy to wait.

PR #332, opened 2026-08-11
ENG-1390: prefix the rule-retrieval analytics action so the collector relays it

alecantu7:
Follow-up to [ENG-1390](https://linear.app/mindsdb/issue/ENG-1390) (merged in #330). Two-line behaviour change plus two tests.

## The bug

`rule_retrieval` shipped and produced **nothing** in PostHog project 355390. The collector lambda relays an action only if it starts with `anton_` or `ds_connect` — a **prefix rule, not a list** — and drops everything else **while still returning HTTP 200**. `send_event` is fire-and-forget behind a bare `except`, so there is no client-side signal whatsoever.

Measured against the live collector:

```
action=rule_retrieval        -> posthogResult: false   (silently dropped)
action=anton_rule_retrieval  -> posthogResult: true
```

Renamed to `anton_rule_retrieval`. It also matches every other action on this channel — `anton_started`, `anton_query`, `anton_first_query`, `ds_connect_*` — so the unprefixed name was off-convention as well as broken.

## Scope: this fixes ONE of two gates

The lambda **also** allowlists property *names*: `action`, `aid`, `engine`, `llm_provider`, `has_mdb_key`. So after this, the event arrives but carries **none of its data** until [ENG-1355](https://linear.app/mindsdb/issue/ENG-1355) lands.

Verified by probing a *passing* action name with the real payload:

| Sent | Arrived |
|---|---|
| `outcome`, `when_rules`, `kept_rules`, `stop_reason`, `duration_ms` | **all dropped** |
| `llm_provider` | ✅ (pre-allowlisted) |

So this PR is **necessary, not sufficient**. Merging it does not make ENG-1390's Done-when #2 pass — that needs ENG-1355, and specifically the *generic pass-through* option rather than adding `turn_completed`'s 28 names, since `rule_retrieval`'s properties aren't among those either.

## Tests

Two, deliberately in their **own class**. `TestRuleRetrievalObservability` autouse-stubs `_emit_rule_retrieval` — which is exactly what these must not have stubbed, since they assert what reaches `send_event`. I hit that while writing them: the first version silently exercised the stub and failed with `KeyError: 'action'`.

- `test_the_action_name_carries_the_prefix_the_collector_requires` — pins the constant.
- `test_the_emitted_action_is_the_prefixed_one` — pins the name that actually goes out, so a refactor reintroducing a literal is caught.

Both **mutation-verified** against the old name:

```
AssertionError: 'rule_retrieval' would be silently dropped by the collector
AssertionError: {'action': 'rule_retrieval', 'props': {...}}
```

Suite: `1745 passed, 28 skipped`.

## Why a test rather than just a fixed string

The constraint lives in a lambda whose code is **in no repo in the workspace** — ENG-1355 documents that. So nothing in this codebase can enforce it structurally, and a test with the measured `posthogResult` values in its docstring is the only place the constraint can be recorded where the next person will trip over it.

## Note on `turn_completed`

`turn_completed` fails the same gate (`posthogResult: false`) and would be fixed the same cheap way. **Deliberately not touched here:** ENG-1355 is in progress and its Done-when queries name the literal string `turn_completed`, so renaming it unilaterally would invalidate acceptance criteria someone is actively working against. Worth a conversation rather than a surprise — flagged on the ticket.

### DIFF
--- a/anton/core/memory/cortex.py
+++ b/anton/core/memory/cortex.py
@@ -62,6 +62,20 @@ logger = logging.getLogger(__name__)
 _RULE_RETRIEVAL_TAG = "rule-retrieval"
 _RULE_RETRIEVAL_MAX_TOKENS = 4096
 
+# The analytics ACTION NAME, and the `anton_` prefix is load-bearing, not style.
+# The collector lambda relays an event only if its action starts with `anton_` or
+# `ds_connect` — a prefix rule, not a list — and drops everything else while still
+# returning HTTP 200, so the client cannot tell. Measured 2026-08-10:
+#
+#     action=rule_retrieval        -> posthogResult: false   (silently dropped)
+#     action=anton_rule_retrieval  -> posthogResult: true
+#
+# The unprefixed name is why this event produced nothing in project 355390 after
+# shipping. Any new action added here must carry the prefix, or it disappears.
+# (Separately, the lambda also allowlists property NAMES, so the properties below
+# still do not arrive until ENG-1355 lands — the prefix fixes one of two gates.)
+_RULE_RETRIEVAL_ACTION = "anton_rule_retrieval"
+
 # Built once: `AntonSettings()` reads the environment, and while this call site
 # is rare it sits on the latency-critical prompt-assembly path.
 _analytics_settings = None
@@ -88,11 +102,11 @@ def _emit_rule_retrieval(**shape) -> None:
 
         send_event(
             _analytics_settings,
-            "rule_retrieval",
+            _RULE_RETRIEVAL_ACTION,
             **{key: str(value) for key, value in shape.items()},
         )
     except Exception:
-        logger.debug("rule_retrieval analytics event failed", exc_info=True)
+        logger.debug("%s analytics event failed", _RULE_RETRIEVAL_ACTION, exc_info=True)
 
 
 # ─────────────────────────────────────────────────────────────────────────────
--- a/tests/test_cortex.py
+++ b/tests/test_cortex.py
@@ -522,3 +522,42 @@ class TestRuleRetrievalObservability:
         out = await self._run(cortex, llm, small)
         assert llm.calls == 0 and captured_events == []
         assert out == small
+
+class TestRuleRetrievalActionName:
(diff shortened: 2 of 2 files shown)


alecantu7:
Closing in favour of #345, which makes this unnecessary rather than merely redundant.

**What this PR did.** Renamed the action to `anton_rule_retrieval` so it would clear the collector lambda's `anton_` prefix filter. That was the right fix for the problem as framed, and the measured evidence behind it still stands:

```
action=rule_retrieval        -> posthogResult: false   (silently dropped)
action=anton_rule_retrieval  -> posthogResult: true
```

**Why it is being closed.** #345 routes `rule_retrieval` past the collector entirely, direct to PostHog MindsHub main (424726). With no collector in the path there is no prefix filter to satisfy, so the rename has nothing left to do.

Three things make closing better than merging:

1. **The prefix would be wrong at the destination.** Every other event in 424726 — `first_query`, `agent_session_started`, `boot_screen_resolved`, `app_installed` — carries no prefix. `anton_rule_retrieval` would be the only prefixed name there, and the prefix encodes a constraint from a lambda that is no longer in the path.

2. **The test would block the correct name.** `test_the_action_name_carries_the_prefix_the_collector_requires` asserts `_RULE_RETRIEVAL_ACTION.startswith("anton_")`. Once the collector is out of the path that assertion is pinning a dead constraint, and it would fail the eventual rename to the right name.

3. **It never delivered the payload anyway.** This PR's own comment says so: *"the lambda also allowlists property NAMES, so the properties below still do not arrive until ENG-1355 lands — the prefix fixes one of two gates."* The nine properties (`outcome`, `when_rules`, `kept_rules`, `rules_chars`, `stop_reason`, `input`/`output_tokens`, `duration_ms`) are the entire signal; a bare event count with none of them duplicates a frequency measurement the Langfuse trace tag from #330 already provides. #345 delivers all nine.

**Nothing is lost by closing.** The case-sensitivity evidence is recorded on ENG-1355 (comment of 2026-08-11), and the two-gate explanation now lives in `anton/analytics.py`'s module docstring on #345, where the next person to add an event will actually read it.

Ticket: ENG-1495.

PR #334, opened 2026-08-11
 ENG-1359: don't register a web_search fallback with no usable credential

ea-rus:
## Problem
When a provider has no native web search, anton registers `WEB_SEARCH_FALLBACK_TOOL` unconditionally — even when the session has no Exa/Brave key configured. The model calls the tool anyway and gets a guaranteed error every time, with copy telling the user to run `anton setup search`, a CLI command a Cowork desktop or web user does not have. Observed in production: the agent frequently reacts to the dead tool by fabricating the data it could not retrieve and shipping it as research.

This is the anton half of ENG-1359. The cowork-server half (mindsdb/cowork-server#294) stops constructing the MindsHub provider with a flavor that reports no native web tools; this half stops offering a tool that cannot succeed regardless of flavor.

## Fix
- `anton/core/tools/web_tools.py`: extracted `has_search_credential(settings)`, shared by the registration gate and the handler so the two can't drift.
- `anton/core/session.py`: `WEB_SEARCH_FALLBACK_TOOL` is registered only when `has_search_credential` is true. `web_fetch` is unaffected — its fallback is credential-free stdlib HTTP.
- Reworded `_NO_PROVIDER_MSG` so it no longer presents the CLI command as the only route. Still correct for standalone anton CLI users; no longer misleading for a host-app user who can't run it (and who, after this change, won't reach the tool at all).

## Tests
`tests/test_web_tools.py`:
- `test_fallback_web_search_not_registered_without_credential` (new) — with no credential, `web_search` is absent from the built tool list while `web_fetch` is still present.
- `test_fallback_toolDefs_registered_when_provider_lacks_native` (updated) — now configures a credential, since one is required for registration.

Mutation-checked: the new test kills a `has_search_credential` → `return True` mutant, the updated sibling kills `return False`. Full run: 155 passed, 17 skipped across the web-tools, session and provider suites.

Fixes https://linear.app/mindsdb/issue/ENG-1359/web-search-is-broken-for-every-cowork-user-the-provider-flavor-is

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1389,9 +1389,18 @@ class ChatSession:
         # appear in the registry; the model uses the provider's server-side
         # web tools instead and Anton's dispatch loop never sees a ``tool_use``
         # for them. See ``anton/core/tools/web_tools.py`` for the handlers.
+        #
+        # ``web_search`` additionally needs a usable Exa/Brave credential — a
+        # model offered a tool that can only fail will call it anyway, so skip
+        # registration outright rather than let it dispatch to a guaranteed
+        # error (ENG-1359).
         if "web_search" in self._fallback_web_tools:
-            from anton.core.tools.web_tools import WEB_SEARCH_FALLBACK_TOOL
-            self.tool_registry.register_tool(WEB_SEARCH_FALLBACK_TOOL)
+            from anton.core.tools.web_tools import (
+                WEB_SEARCH_FALLBACK_TOOL,
+                has_search_credential,
+            )
+            if has_search_credential(self._settings):
+                self.tool_registry.register_tool(WEB_SEARCH_FALLBACK_TOOL)
         if "web_fetch" in self._fallback_web_tools:
             from anton.core.tools.web_tools import WEB_FETCH_FALLBACK_TOOL
             self.tool_registry.register_tool(WEB_FETCH_FALLBACK_TOOL)
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -1458,7 +1458,7 @@ class ChatSession:
         # ``web_search`` additionally needs a usable Exa/Brave credential — a
         # model offered a tool that can only fail will call it anyway, so skip
         # registration outright rather than let it dispatch to a guaranteed
-        # error (ENG-1359).
+        # error.
         if "web_search" in self._fallback_web_tools:
             from anton.core.tools.web_tools import (
                 WEB_SEARCH_FALLBACK_TOOL,
--- a/anton/core/tools/web_tools.py
+++ b/anton/core/tools/web_tools.py
@@ -446,6 +446,21 @@ _NO_PROVIDER_MSG = (
 )
 
 
+def has_search_credential(settings: object) -> bool:
+    """Whether ``settings`` has a usable Exa/Brave key for the fallback handler.
+
+    Shared by the registration gate (``ChatSession.__init__`` skips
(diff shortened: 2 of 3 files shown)


alecantu7 (commented):
## Adversarial review — head `df0d024e`, base `origin/staging` `d70e8803`

Reviewed together with its sibling **cowork-server#294**. **No defects found.**

### What I checked and it held

**The gate and the handler share one predicate.** `has_search_credential` is used by both the registration gate (`session.py:1467`) and `handle_web_search_fallback`, so they cannot drift into disagreeing about whether a credential is usable. That drift — a registration-time check diverging from the runtime one — is the exact failure class I went looking for first, and it is designed out rather than merely absent.

**Ordering is right.** `self._settings` is assigned at `session.py:651`, well above the gate at `:1467`, so the gate cannot read a not-yet-populated field.

**Nothing downstream assumes the tool was registered.** Grepped `_fallback_web_tools` across `anton/` — the only reads are the two lines that compute it. No prompt or system text promises `web_search` anywhere (every hit is CLI setup flow), so skipping registration creates no registry/prompt mismatch where the model is told about a tool it cannot see.

**`web_fetch` stays registered unconditionally, correctly** — its fallback is stdlib HTTP with no credential, so it has nothing to gate on.

**Mutation:** removing the gate turns **`test_fallback_web_search_not_registered_without_credential`** red — a named test, not an incidental one.

**Suites:** `tests/test_web_tools.py` 39 passed; full suite **1932 passed, 28 skipped**.

**Security pass** (manual — `/security-review` cannot target a scratch worktree from this session): the change only *reduces* surface. It removes a tool that would have sent user queries to a third-party Exa/Brave endpoint, reads settings via `getattr` with no injection surface, adds no logging, and the reworded `_NO_PROVIDER_MSG` carries no secrets.

### On the ticket's copy requirement

ENG-1359's Done-when asks that the fallback message *"names a route a desktop/web user can actually take."* The new copy takes the ticket's second option — saying plainly that search is unavailable — and scopes the `anton setup search` advice to the standalone CLI where it is true. That satisfies the requirement as written; worth confirming deliberately rather than reading it as a miss.

### Merge order

**cowork-server#294 is the actual fix; this is the guard.** Setting the flavor makes anton route `web_search` natively and never reach the fallback, so #294 alone closes the user-visible bug. This PR stops the model being offered a tool that can only fail, which matters on any configuration with no native support and no credential. Either order is safe.

Note that cowork-server pins anton as `branch = "main"`, so this change reaches Cowork only after the next staging→main release — no problem, since #294 does not depend on it.

### One thing for the ticket, not this PR

ENG-1359's Done-when includes *"The same holds for BYOK OpenAI,"* and cowork-server#294 deliberately does not do that — for three Responses-API gaps in anton which I verified are all real (no `response.incomplete` handler; `_translate_user_blocks_to_responses` drops `tool_result` images; no Langfuse headers on that path). Details on (https://github.com/mindsdb/cowork-server/pull/294) **So the ticket should not close on these two PRs** — either amend that item or split it into a follow-up.

alecantu7 (approved):
Approving — see the review above.

No defects. `has_search_credential` being shared by the registration gate and the handler is the right shape: the two cannot drift into disagreeing, which is the failure class this kind of gate usually ships with. Mutation-verified — removing the gate fails `test_fallback_web_search_not_registered_without_credential` by name. Full suite 1932 passed.

Note for sequencing: cowork-server#294 is the fix that closes the user-visible bug; this is the guard. Either order is safe.

PR #335, opened 2026-08-11
feat: generate_artifact — one-tool web artifact pipeline (ENG-968, ENG-969)

StpMax:
## Summary

One tool, `generate_artifact`, takes a web artifact (`html-app`, `fullstack-stateless-app`, `fullstack-stateful-app`) from a user request to verified, running code. It replaces the hand-written scratchpad generation as the primary path for these three types; `create_artifact` registers the folder, `generate_artifact` does the rest. The separate `generate_prd` tool from the first revision of this branch was merged into it: the two paid for their boundary twice (gathering in one, re-inspection of the scratchpads in the other).

The pipeline is a deterministic FSM in five phases on one shared history:

- **A — gathering**: scratchpad / web_search / web_fetch / ask_user until the data the artifact needs is verified; `finish_gathering` fixes the artifact type, constraints and assumptions.
- **B — brief**: a short proposal shown to the user for accept / cancel / revise (loops on revise).
- **C — PRD**: `prd.md` written from the agreed brief only.
- **D — specs**: `spec.md` (implementation notes, no retelling of the PRD) and, for fullstack, `openapi.json` — the only place request/response shapes live; the document's shape and its paths (exactly those `spec.md` lists) are checked, one corrective retry.
- **E — code**: backend and frontend generated in parallel, each in a bounded write loop, then deterministic verifiers (routing under `/api/*`, health endpoint, secrets, dependency manifest, `state_manifest.json` for stateful; CSS/URL/CDN policy, viewport, stable `id`s for the frontend), then launch + health/route probe for fullstack. Backend and frontend are held to the same `openapi.json`: a `fetch()` to an undefined path or a missing operation fails the step. When `ANTON_HTML_LINT_BROWSER` points at a browser, the page is also loaded once — the `html-app` page from disk, the fullstack page from its running backend after the launch (console errors, missing files, 4xx/5xx from its own origin fail the step).

Design points that shaped the branch:

- **Shared prefix, one compression point.** Phases A–D continue one message list under one system prompt; step rules travel in the step message, not in per-node system prompts, so the prefix cache holds. The history is dropped once, after the specs are on disk; phase E and every cold start rebuild their context from state (`spec.md`, `data_notes`, `web_notes`, journal).
- **File bodies travel as text**, between markers, plus a `write_file(path)` call in the same reply: a large tool-call argument is not streamed incrementally (measured), which cost dropped connections and silent minutes. `write_file` reports characters/lines added; parts are appended with `mode="a"`.
- **Checkpoints** (`discovery.json`) with stage-based re-entry, and a **spend guard** across all phases that winds the loops down and asks the user instead of failing.
- **Prompts** for html-app, fullstack frontend and backend share one skeleton (task → inputs → workflow → output protocol → size rules → verifier contract → rules → tools); the backend gets only the data-source catalog entries the artifact declared. Tailwind via CDN is recommended for styling (hand-written CSS allowed; offline pages use CSS only), ECharts for charts, chart type chosen by purpose, units and formatters on every number. The two built-in skills for the hand-built path mirror these contracts (test-pinned).
- **Attachments.** `attachments` names the files the user attached: data files are read by the gathering step, other assets are copied into the artifact before generation and referenced by name.
- **Result** carries `url`/`port` for a running fullstack app (the agent names that URL as the entry point — `static/index.html` from disk cannot reach `/api/*`), plus `warnings` and `checks_skipped`: a check that could not run is relayed to the user, never presented as passed.
- **Progress**: every step announces itself as a `ToolProgress` line with `step N of M` (and `attempt K` on a retry), and the live tail of the reply being streamed is shown under it in the CLI footer while a long step runs.

## Manual testing: `generate_artifact` (this branch) vs `staging` (skill-based)

Same prompts run on both branches, using the LLM call log (`history.txt`) and, for this branch, the pipeline trace (`trace.json`). Token totals include cache creation + cache read (the gateway bills these 1:1 against quota), and exclude session-startup memory-compaction calls that are unrelated to the task.

| Prompt | Branch | Tokens (in+out+cache) | Time | Outcome |
|---|---|---|---|---|
| Simple HTML artifact (ASCII cow) | `staging` | ≈104,251 | ~36s | completed; no verifier exists on this path |
| Simple HTML artifact (ASCII cow) | this branch | ≈147,160 | ~71s (pipeline: ~54s) | completed; browser check passed, no findings |
| Fullstack dashboard (Postgres `home_rentals`, 5s polling) | `staging` | ≈697,105 | ~4m 08s | completed; verified only via a manual `httpx` smoke test, no browser check |
| Fullstack dashboard (Postgres `home_rentals`, 5s polling) | this branch | ≈282,065 | ~4m 36s (pipeline: ~3m 46s) | completed; backend + static frontend + live app all verified via headless browser, no findings |
| Article summary (Habr → HTML) | `staging` | ≈341,712 | ~2m 27s | completed; no verifier exists on this path |
| Article summary (Habr → HTML) | this branch | ≈281,814 | ~2m 41s (pipeline: ~1m 47s) | completed; browser check passed, no findings |
| Complex multiplayer game (Tic-Tac-Toe/Tetris hybrid, rooms, real-time polling, full ruleset) | `staging` | ≈1,235,332 | ~6m 46s (paused) | did not finish; self-paused near the ~1.2M token turn ceiling with only the game engine and a bare backend done, frontend not started |
| Complex multiplayer game (Tic-Tac-Toe/Tetris hybrid, rooms, real-time polling, full ruleset) | this branch | ≈1,247,299 | ~9m 29s (pipeline: 5m 32s, then paused) | pipeline delivered a complete, verified app (backend + frontend + live app all passed automated checks); the turn then self-paused near the same ~1.2M ceiling during additional manual playtesting on top of the already-finished result |

**Takeaways**
- On trivial single-file artifacts, this branch costs somewhat more (structured PRD/spec steps + real browser verification add overhead) but gains an actual automated check — `staging`'s skill path has no verifier at all.
- On non-trivial builds (fullstack app, content-heavy summary), this branch is both faster and cheaper in tokens than `staging`, which burns many more iteration rounds (and a much larger cached context) without any structured intermediate steps.
- On a genuinely complex build, both branches hit the same ~1.2M-token turn ceiling and self-paused — but this branch reached a fully working, automatically verified app before running into it, while `staging` was still stuck on the backend with no frontend at all.

## Resolves

- [ENG-968: Artifact code generation tool](https://linear.app/mindsdb/issue/ENG-968/artifact-code-generation-tool)
- [ENG-969: PRD generation tool](https://linear.app/mindsdb/issue/ENG-969/prd-generation-tool) — folded into `generate_artifact` as phases A–C
- ENG-970 (generation progress), ENG-1116 (output cap), ENG-1986 (gateway script-tag escaping, removed)

## Test plan

- [x] `pytest tests` — 3944 passed, 41 skipped, no failures (at `3ea33a3d`, with `ANTON_HTML_LINT_BROWSER`)
- [x] `generate_artifact` unit and contract tests: FSM prompts and their locks against the verifier messages, `_run_loop` and the file-body protocol, orchestrator retries and budgets, backend/frontend verifiers, spec/api-spec budget and shape checks, progress counter and live tail, tool-handler wrapping, checkpoint re-entry
- [x] 23 live runs on the branch (html-app pages and dashboards, fullstack stateless and stateful apps, pages built from a web article); each run's trace was reviewed and fed back into the prompts and verifiers — see the commit messages on this branch
- [x] Eight findings of a code review of the branch fixed
- [ ] Changes after `a7ba6f37` (API contract on both sides, attachments, fullstack browser gate, result fields) are unit-tested only — a live run on the two reference requests is pending


anton/core/llm/openai.py
```diff
@@ -4,6 +4,7 @@
 import json
 import logging
 import os
+import re
```
ea-rus: Import added but not used?

anton/publisher.py
```diff
@@ -38,24 +39,14 @@
...
-    ".anton_state.db",
-    ".anton_state.db-wal",
-    ".anton_state.db-shm",
-    _STATE_SNAPSHOT,
-}
+# Names inside an artifact folder that are housekeeping — never bundled. The
+# store's definition, so what the agent hides and what the bundle omits cannot
+# drift. `_zip_fullstack` is allowlist-based, so root files are not bundled
+# anyway; the set guards `_zip_html` and any future change to that allowlist.
+_FULLSTACK_EXCLUDED = _HOUSEKEEPING_FILES | _HOUSEKEEPING_DIRS
```
ea-rus: Public-publish leak (html-app). store.py excludes prd.md, spec.md, discovery.json (and, for fullstack, openapi.json) from the artifact's file listing (GENERATION_INPUT_FILES folded into _EXCLUDED_FROM_FILES), but publisher.py's _zip_html (directory branch) still filters only the old, unmodified _BUNDLE_SKIP_NAMES = {".published.json", ".revisions"} — it was never updated to also check GENERATION_INPUT_FILES. Publishing an html-app ships the PRD and discovery state into the public bundle. (backend.log does not apply here — html-app has no backend; _zip_fullstack correctly excludes it via _FULLSTACK_EXCLUDED = _HOUSEKEEPING_FILES | _HOUSEKEEPING_DIRS, which the PR did update.) Confirmed independently by 3 separate passes.

anton/publish_access.py
```diff
@@ -196,7 +195,7 @@ def _user_files(folder: Path) -> list[Path]:
                 continue
             rel = p.relative_to(folder)
             top = rel.parts[0] if rel.parts else ""
-            if top in _HOUSEKEEPING_FILES:
+            if top in _HOUSEKEEPING_FILES or top in _HOUSEKEEPING_DIRS:
```
ea-rus: Same root cause, second consequence. publish_access.py:198 (_user_files) also fails to exclude these files. When _pick_primary's primary_hint doesn't resolve, its fallback to files[0] can pick an internal file (discovery.json, prd.md) as the artifact's "primary" entry point instead of real UI content.
Fix is one line: fold GENERATION_INPUT_FILES into the set both publisher.py and publish_access.py import, not just store.py's local one.

anton/core/llm/openai.py
StpMax: Fixed in 94456db3.

anton/publisher.py
StpMax: Fixed in 94456db3.

anton/publish_access.py
StpMax: Fixed in 94456db3.

anton/core/tools/generate_artifact/attachments.py
```diff
@@ -0,0 +1,206 @@
...
+            )
+        except (KeyError, TypeError, ValueError):
+            return None
+
+
+def _kind(path: Path) -> str:
+    return KIND_DATA if path.suffix.lower() in DATA_SUFFIXES else KIND_ASSET
+
+
+def resolve_attachments(paths) -> tuple[list[Attachment], list[str]]:
```
ZoranPandovski: resolve_attachments doesn't check that the path is under the known upload directories (.cowork/files/<uuid>/, .anton/uploads/). A path like .anton/.env has no recognized data suffix, so it'd get classified as a plain asset and copied straight into the published artifact bundle, making it publicly readable. Worth rejecting anything outside the expected upload roots here.

anton/core/tools/registry.py
```diff
@@ -134,7 +134,8 @@ async def dispatch_tool(
                 if emitter is not None:
                     await emitter.emit(
                         StreamTaskProgress(
-                            phase="tool_progress", message=item.text, id=tool_call_id,
+                            phase="tool_peek" if item.kind == "peek" else "tool_progress",
```
ZoranPandovski, line 137: This adds a new progress phase, tool_peek, that reaches the cloud JSONL wire through session.emitter, but anton/cloud_turn/contract.py still only lists continuation and handback as phases. Can you add tool_peek there too, so scratchpad-controller and cowork-server know it exists?

anton/core/tools/generate_artifact/discovery/prd.py
```diff
@@ -0,0 +1,121 @@
...
+    # The type is settled BEFORE the step runs, not after: the step's own
+    # progress line is the first one that carries `step N of M`, and `M`
+    # depends on the type; the metadata write below stays where it was.
+    registered_type = state.artifact_type
+    final_type = state.final_artifact_type or registered_type
+    state.settle_artifact_type(final_type)
+
+    # An empty reply raises inside `plan_step`: writing an empty prd.md and
+    # reporting `prd_written` would be a silent lie about what happened.
+    full_prd, _ = await sub_tools.plan_step(
```
ZoranPandovski, line 91: write_prd still sees the full phase A-D message history here, so "PRD from the agreed brief only" is really just the prose instruction above, not something structural. A live run already leaked an article's text through this path once (fixed in c906cb00), and the regression test for that only checks the instruction string, not what write_prd actually outputs. Might be worth a test that exercises write_prd directly and asserts source text doesn't end up in the PRD.

ZoranPandovski (commented):
Few comments

anton/core/tools/generate_artifact/attachments.py
StpMax: Fixed in f0c71871. `resolve_attachments` now checks the resolved path (so a symlink can't point elsewhere) and accepts only: a cowork upload by its `<COWORK_HOME>/files/<uuid>/<name>` layout; a file under `.anton/uploads/` (CLI) or `attachments/` (cloud turn) inside the workspace; or a plain workspace file with no dot-component in its relative path. Everything else — `.anton/.env`, `.anton/secrets.json`, anything outside the workspace — is dropped with the reason in the trace, for data files as well as assets. Without a workspace only the cowork rule applies. Tool description tells the agent what is refused. 7 tests added.

anton/core/tools/registry.py
StpMax: Fixed in 99fc6ae3. `tool_peek` is now relayed only to a host that declared `ChatSessionConfig.live_tool_peek` (the CLI does; cowork-server and the cloud pod get the default False), so it never reaches the JSONL wire. `cloud_turn/contract.py` now documents `tool_progress`, `tool_done` and `tool_peek`, the last as kept off the wire and what a consumer should do if it meets one anyway. A defensive `tool_peek` branch in cowork-server's `stream_formatter` would be a separate follow-up there.

PR #337, opened 2026-08-11
Pin local dev to Python 3.12 (matches CI/prod)

tino097:
Fixes [ENG-1445](https://linear.app/mindsdb/issue/ENG-1445/anton-local-dev-python-is-unpinned-314-breaks-dill-function-pickling)

- No `.python-version` file existed, so `uv sync` on a dev machine picks whichever interpreter is newest/installed — on this machine that was 3.14.
- CI (`tests.yml`: `uv python install 3.12`) and prod (`Dockerfile`: `FROM python:3.12-slim`) both pin 3.12, so this never showed up there.
- On 3.14, `dill==0.3.8` fails to pickle plain functions (`TypeError: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'` — a private CPython pickle-internals signature change in 3.14 that predates dill 0.3.8, released Jan 2024). This made `tests/test_scratchpad.py::TestSessionPersistence::test_namespace_survives_restart` fail locally, looking like a scratchpad-snapshot bug when it's actually an interpreter-version mismatch.
- Verified: identical repro pickles/unpickles fine on 3.11, 3.12, and 3.13 — only 3.14 fails.
- Fix: add `.python-version` pinning 3.12, so local dev matches CI/prod by default. With it, full suite: 1975 passed, 0 failed, 28 skipped.
- Also adds a `.gitignore` entry for `.superpowers/`, the scratch workspace used by the subagent-driven-development skill (ledgers, task briefs, review packages) — untracked but not previously ignored.


alecantu7 (approved):
Reviewed at `e2fb525b`, base `origin/staging` `8c9a3ef8` (verified fresh against the API).

No defects — the diagnosis holds and I reproduced it from scratch rather than taking it on trust. Two notes below: one wording fix, and one observation that I think is bigger than the ticket's "follow-up worth considering" framing.

## The diagnosis is right, and the mechanism is narrower than the description says

The body says 3.14 breaks "plain functions". My first repro — `dill.dumps({'style_title': f})` — **passed** on 3.14, so I went to the real call site (`scratchpad_boot.py:310`, `dill.dumps(value)` per value) and probed the shapes that path actually sees:

```
python 3.12.12 dill 0.3.8
  module-level func:     OK
  exec-defined func:     OK
  whole exec namespace:  OK

python 3.14.3 dill 0.3.8
  module-level func:     OK          <- pickled by REFERENCE, unaffected
  exec-defined func:     FAIL TypeError: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'
  whole exec namespace:  FAIL TypeError: _Pickler._batch_setitems() missing 1 required positional argument: 'obj'
```

So the trigger is **functions pickled by value** — dill walks `__globals__` through `save_module_dict` into `_batch_setitems`, whose signature changed in 3.14. A module-level function survives because it is saved by reference.

That is exactly the agent-written-helper case, which is why it presents as ENG-1366. And the symptom chain closes cleanly: `_encode_values` catches `Exception` per value and **silently drops the key** (`scratchpad_boot.py:311-317`), so the failure surfaces as `NameError: style_title` after a restart rather than as an error — which is precisely why it was expensive to root-cause. Worth folding both details into ENG-1445: "functions defined in a cell" is the reproducible statement, and the silent drop is the reason it cost investigation time.

## Claims verified

| claim | result |
| -- | -- |
| CI pins 3.12 | yes — and it is **four** workflows, not one: `tests.yml`, `verifier-eval.yml`, `tests_e2e_release.yml`, and `release.yml` (setup-python 3.12) |
| Prod pins 3.12 | yes — `Dockerfile:5` `FROM python:3.12-slim` |
| The pin resolves correctly | yes — `uv run python -V` gives **CPython 3.12.12** |
| The named test passes | yes — `test_namespace_survives_restart`, 1 passed |
| Full suite green | yes — **1970 passed, 28 skipped** (body says 1975; a 5-test drift from staging moving since, not a discrepancy worth chasing) |
| The `.gitignore` entry is safe | yes — zero tracked files match, `.superpowers/` is directory-anchored, and a path like `src/superpowers.py` is **not** ignored |

Independent corroboration worth recording: reviewing #339 earlier today, its body listed `test_namespace_survives_restart` as a known local failure — and it **did not reproduce** for me (128 passed). My machine resolves 3.12. Same divergence, observed from the other side before either of us knew the cause. This PR is the right fix.

## One wording fix

The body says the CPython change *"predates dill 0.3.8, released Jan 2024"*. It **post**dates it — 3.14 shipped well after Jan 2024, which is exactly why a January-2024 dill cannot handle it. ENG-1445 has it right ("postdates"); only the PR body is inverted. One word.

## The follow-up is understated

The ticket calls upgrading dill *"forward-looking … once the runtime actually moves past 3.12"*. But `pyproject.toml:11` is `requires-python = ">=3.11"` with **no upper bound**, so `uv tool install anton-agent` on a 3.14 machine installs happily today and then silently drops every agent-written function from scratchpad snapshots. No error, just a `NameError` later.

The desktop app is already safe, for an unrelated reason — cowork's `uv-paths.ts:14-16` sets `PYTHON_RANGE = ">=3.12,<3.14"` with the comment *"PyO3 (used by pywinpty on Windows) doesn't support 3.14 yet."* So the dev-only framing holds for the desktop. The standalone anton CLI path has no such guard.

Capping `requires-python` at `<3.14` is the cheap version and turns silent snapshot data loss into an install-time refusal; moving to dill 0.4.x is the real fix. Also worth reconciling, as the ticket already notes: `scratchpad_boot.py` carries "Verified with dill 0.4.1" comments while `pyproject.toml:31` pins `dill==0.3.8`.

Neither is a reason to hold this PR — pinning local dev to match CI and prod is correct and self-contained.

PR #339, opened 2026-08-11
ENG-1324: surface the liveness heartbeat and pin the total-budget stop's lesson

tino097:
## What

Two consequences of the liveness heartbeat that shipped in #312 (ENG-578) — same mechanism, one ticket: **surface the heartbeat to the user**, and **test the stop path that heartbeat made load-bearing**.

## Why

#312 correctly keeps a deliberately quiet cell alive by heartbeating past the inactivity window. But the heartbeat is unconditional — it proves the process is alive, not that it's progressing. A wedged-but-beating cell (deadlock, `while True: pass`, a stuck native call) beats happily and falls through to the **total** timer, sized for how long legitimate work should take, not how long to tolerate a hang. That timer is now the only thing that ends a hang, and today it runs completely silently — up to 60 minutes on the throttled-batch workload #312 exists to enable.

## How

- `anton/core/backends/local.py`: in `_read_result`'s `HEARTBEAT_MARKER` branch, yield a rate-limited `"still running — Xm elapsed of ~Ym budget"` notice once silence passes a threshold (60s), then no more often than every 60s. Reuses the existing `phase="scratchpad"` progress channel already rendered by the UI — no new event type. The notice does **not** extend `current_inactivity`; only an explicit `progress()` call may widen that window. A chatty cell never reaches this branch (it ships `STDOUT_CHUNK_MARKER` instead), so it can never get a notice.
- `tests/test_scratchpad_watchdog_contracts.py`:
  - `TestQuietCellNotice` — a quiet cell gets a rate-limited notice; a chatty cell never does.
  - `TestBeatingWedgeTotalBudgetStop` — the missing behavioural test: a beating-but-wedged cell (some real output, then a tight infinite loop) is killed only by `cell_total_max`, with the total-budget message, and `detect_kill_loop` classifies it `"heavy"` ("make cells smaller"), never `"liveness"` ("reset and retry unchanged") — pinning down the exact lesson-mismatch the ticket flags for durable ACC memory.

## Done when

- [x] A quiet cell past the threshold emits a periodic "still running — Xm of ~Ym budget" notice; it does not extend the inactivity window.
- [x] A chatty cell emits no such notices.
- [x] A beating-but-wedged cell driven to `cell_total_max` is killed, with the total-budget message, and the lesson it produces is asserted.
- [x] No new event type or renderer change needed.

## Checks run

- `uv run pytest tests/test_scratchpad_watchdog_contracts.py -v` — 32 passed (repeated 3x, stable, no timing flakiness).
- `uv run pytest tests/test_scratchpad.py tests/test_scratchpad_watchdog_contracts.py -q` — 1 pre-existing failure (`TestSessionPersistence::test_namespace_survives_restart`, unrelated `style_title` NameError), 127 passed.
- `uv run pytest -q` (full suite) — same pre-existing failure, 1972 passed, 28 skipped. Confirmed via `git stash` that the failure reproduces identically on unmodified `staging`.

Closes ENG-1324.


alecantu7 (commented):
Reviewed at head `c572ec3b`, base `origin/staging` `59195780` (verified fresh against the API, not the local clone).

The mechanism is right, the cross-repo "no renderer change" claim checks out end-to-end, and three of the four new assertions are load-bearing under mutation. One real defect: the notice fires on the wrong clock.

---

## Finding 1 — the gate measures total runtime, not silence · medium · confirmed by execution

`anton/core/backends/local.py:920-927`:

```python
if (
    elapsed >= _QUIET_NOTICE_AFTER
    and elapsed - last_notice >= _QUIET_NOTICE_EVERY
):
```

`elapsed` is `monotonic() - start` — time since the **cell began**, not since it last said anything. So the first condition asks *"has this cell been running 60s?"*, not *"has it been quiet for 60s?"*

That contradicts the constant's own comment three lines above:

> `# First notice no earlier than this many seconds of silence`

and the PR body (*"once silence passes a threshold (60s)"*), and Done-when #2 (*"A chatty cell emits no such notices"*).

**Executed.** A cell chatty for ~1.2s then quiet for only ~0.25s, threshold set to 0.5s:

```
AssertionError: notice fired after only ~0.25s of silence:
  ['still running — 0m elapsed of ~0m budget']
```

**Why it matters in production.** `_heartbeat_loop` (`scratchpad_boot.py:847-860`) only substitutes a salvage chunk on ticks that *have* new output — the gaps between prints still emit bare beats. So the motivating workload, the throttled batch printing `sent 12/50` every ~30s against a 10s beat, reaches this branch constantly. Past 60s it earns a notice roughly every 60s — about **25 of them across a healthy 25-minute run**.

The ticket's stated goal is *"any way for the user to tell 'working' from 'hung'."* A line that prints in both states can't serve that.

**Refutation attempted (all failed to kill it):** checked whether the heartbeat suppresses beats for intermittently-chatty cells — it doesn't, the chunk substitution is per-tick; whether the 60s rate limit makes the volume tolerable — ~50 notices on a 50-minute cell; whether the new chatty test catches it — see finding 2.

**Smallest fix** — key it to last output instead of start:

```python
last_output = 0.0                       # init beside last_notice
...
if (
    elapsed - last_output >= _QUIET_NOTICE_AFTER
    and elapsed - last_notice >= _QUIET_NOTICE_EVERY
):
```

setting `last_output = elapsed` in the `STDOUT_CHUNK_MARKER` and `PROGRESS_MARKER` branches. The probe above is exactly the test that fails before and passes after.

## Finding 2 — the chatty-cell test cannot fail · low · confirmed by mutation

`test_chatty_cell_never_gets_a_quiet_notice` prints every 0.02s against a 0.3s heartbeat, so every tick carries new output and the bare-heartbeat branch is never reached. It re-asserts the #312 `STDOUT_CHUNK`/`HEARTBEAT` split — pre-existing behaviour — rather than anything this PR added.

**Evidence:** under M1 (notice on *every* beat, rate limit removed) and M2 (notice never fires), this test **passed both times**. It is blind to the gate in either direction.

**Fix:** have the cell print at an interval *longer* than the heartbeat — which is also the case finding 1 breaks, so one change covers both.

---

## What held up

| claim | verdict |
| -- | -- |
| "No new event type or renderer change needed" | ✅ traced all three repos — `session.py:3512` wraps the yielded string as `StreamTaskProgress(phase="scratchpad")`, the same path `progress()` already uses; cowork-server `stream_formatter.py:385` renders it as `thought_progress` with `PHASE_LABELS["scratchpad"] = "Running code"`. The user sees **"Running code: still running — 1m elapsed of ~2m budget"** |
| Throttling won't eat it | ✅ `PROGRESS_THROTTLE = 0.25s` against a 60s cadence |
| Does not extend `current_inactivity` | ✅ yield only; `PROGRESS_MARKER` remains the sole widener |
| The other new assertions are load-bearing | ✅ M1 (drop the rate limit) → fails; M2 (never notify) → fails; M4 (put "liveness" into the total-budget message) → fails. Each mutation confirmed applied before its result was trusted |
| No timing flakiness | ✅ 5 consecutive runs, 3 passed each |
| Suites | ✅ `test_scratchpad_watchdog_contracts.py` + `test_scratchpad.py` = **128 passed**. The PR body's one claimed pre-existing failure (`TestSessionPersistence::test_namespace_survives_restart`) did **not** reproduce — staging has moved since |

**Security:** nothing to flag — no new input surface, and the message carries two timing floats, no user data.

alecantu7 (commented):
## Re-review — head `cf852189`

Re-verified rather than assumed: both findings from my pass at `c572ec3b` reproduce identically at the new head. The only commit since is the staging merge, so this is the same code plus a fresher base.

**The merge itself is clean, and I checked because it conflicted in exactly the two files this PR touches.** A hand-resolved conflict is where staging work gets silently reverted, so:

```
git diff 741802f4 HEAD
  anton/core/backends/local.py                 +23 / -0
  tests/test_scratchpad_watchdog_contracts.py +100 / -0
```

Zero deletions against the merged staging commit, and the additions are exactly this PR's own. Both changes that caused the conflict — `[ENG-1275]` auto-install budget (`8c9a3ef8`) and `[12484]` uv discovery (`f673c192`) — survive intact. Nothing to fix there.

---

### Finding 1 — still keyed to total runtime rather than silence · medium · re-executed

`anton/core/backends/local.py:996` is unchanged:

```python
if (
    elapsed >= _QUIET_NOTICE_AFTER
    and elapsed - last_notice >= _QUIET_NOTICE_EVERY
):
```

`elapsed` is `monotonic() - start`, so the first clause asks *"has this cell been running 60s?"*, not *"has it been quiet for 60s?"* Same probe as last time, re-run at `cf852189` — a cell chatty for ~1.2s then quiet for only ~0.25s, threshold 0.5s:

```
AssertionError: notice fired after only ~0.25s of silence:
  ['still running — 0m elapsed of ~0m budget']
```

This is the one I'd hold the PR on, and the reason is narrow: `Done when` #1 asks for a notice "the user can see", and #2 asks that a chatty cell emit none. As written, the throttled batch printing `sent 12/50` every ~30s against a 10s beat reaches this branch constantly and earns a notice roughly every 60s — about 25 of them across a healthy 25-minute run. A line that prints identically whether the cell is working or wedged can't do the job the ticket describes.

Fix, unchanged from last pass — key it to last output rather than to start:

```python
last_output = 0.0                       # init beside last_notice
...
if (
    elapsed - last_output >= _QUIET_NOTICE_AFTER
    and elapsed - last_notice >= _QUIET_NOTICE_EVERY
):
```

setting `last_output = elapsed` in the `STDOUT_CHUNK_MARKER` and `PROGRESS_MARKER` branches. The probe above is exactly the test that fails before and passes after.

### Finding 2 — the chatty test still can't fail, in either direction · low · re-mutated

Mutated the gate two ways at this head and ran both new tests:

| Mutation | `test_quiet_cell_gets_rate_limited_notice` | `test_chatty_cell_never_gets_a_quiet_notice` |
|---|---|---|
| notice on **every** beat (`if True`) | FAILED | **passed** |
| notice **never** fires (`if False`) | FAILED | **passed** |

The chatty test prints every 0.02s against a 0.3s heartbeat, so every tick carries new output and the bare-heartbeat branch is never reached — it re-asserts the ENG-578 `STDOUT_CHUNK`/`HEARTBEAT` split, which is pre-existing. Having it print at an interval *longer* than the heartbeat fixes it, and that is also the case finding 1 breaks, so one change covers both.

### Finding 3 — new, and created by the merge rather than by you · low

`[ENG-1275]`'s auto-install span was **not** in this PR's original base, so it could not have been considered.

During an install the worker runs `subprocess.run(..., capture_output=True)` (`scratchpad_boot.py:1209-1218`), so nothing reaches the wire between `INSTALL_START_MARKER` and `INSTALL_END_MARKER` — only heartbeats. The notice branch sits above all install-aware handling and never consults `installing`, so it fires throughout. Meanwhile `total_timeout` is temporarily inflated (`local.py:1039`, `max(total, now + install_budget + _INSTALL_GRACE)` → `now + 150s` on defaults), so the budget the notice quotes is not the cell's:

```
Installing pandas...
still running — 2m elapsed of ~4m budget      <- install-span budget
…INSTALL_END…
still running — 3m elapsed of ~2m budget      <- restored, install time excluded
```

Deliberately low rather than medium: `yield f"Installing {installing}..."` already fires at install start so the user has context, and after `INSTALL_END` the arithmetic is honest again (`pre_install_total + install duration`, so the elapsed/budget ratio is preserved). What it costs is the property the ticket leans on — *"12m of ~25m reads as on-track, 44m of ~50m reads as stuck"* — since a denominator that moves mid-cell isn't a yardstick. One line: name the install when `installing` is set, the way the three other messages in this function already do.

---

### Two process notes, neither blocking

- **Merged rather than rebased.** The repo convention is rebase + `--force-with-lease`; this is a merge commit with hand-resolved conflicts in both PR files. The resolution is clean, so I would not redo it — flagging it because a hand resolution is the kind of thing nobody re-reads.
- **Already behind again.** `origin/staging` moved to `5eb775e0` (anton#345) about ten minutes after this merge landed, so the base needs another pass before merge regardless.

### Where I land

**Changes requested**, on finding 1 alone. Findings 2 and 3 are small and I am happy for either to follow.

To be explicit about why finding 1 is not a nit: as written the PR satisfies the letter of `Done when` #1 — a notice does appear — while defeating what it is for. The notice fires in the working state and the hung state alike, so a user still cannot tell them apart, which is the entire problem this ticket was opened to solve. Keying it to last output is a handful of lines and makes the feature do the thing.

Everything else from my first pass still holds: the no-renderer-change claim traces end-to-end, throttling won't eat the notice, `current_inactivity` is correctly not extended, and the total-budget stop test's other assertions are load-bearing under mutation.

alecantu7 (approved):
## Approving — head `98417231`

All three findings from the last pass are fixed, and I verified each by execution rather than from the commit message.

**Finding 1 — the notice is keyed to silence now.** The probe that failed at `cf852189`, re-run unchanged at this head — a cell chatty for ~1.2s then quiet for only ~0.25s against a 0.5s threshold:

```
1 passed in 2.56s
```

Full watchdog suite: **40 passed**, including the new install test.

**Finding 2 — the chatty test can now fail, which is the point.** The old one printed faster than the heartbeat and never reached the gated branch, so it passed under every mutation. The rewrite (0.4s cadence, 0.2s heartbeat, 0.7s threshold) bites:

| Mutation | old chatty test | new chatty test |
| -- | -- | -- |
| notice on **every** beat (`if True`) | passed | **FAILED** |
| stop tracking `last_output` on `STDOUT_CHUNK` | n/a | **FAILED** |

**Finding 3 — the install is named, and pinned.** Removing the `installing` branch fails `test_quiet_notice_during_install_names_the_install`.

---

### One gap left, not worth another round

`last_output = elapsed` is written in **two** branches; only the `STDOUT_CHUNK` one is covered.

```
mutation: drop `last_output = elapsed` from the PROGRESS_MARKER branch
  -> 3 passed        (survived)
```

Worth a line because that branch is load-bearing for the workload anton explicitly asks for. `tool_defs.py` tells the model *"Call progress(message) to narrate long phases"*, and `progress()` writes `PROGRESS_MARKER` straight to `_real_stdout` (`scratchpad_boot.py:881-883`) — it never emits a `STDOUT_CHUNK`. So a throttled batch that narrates with `progress("sent 12/50")` rather than `print()` keeps `last_output` fresh **only** through the untested branch, and losing it reopens finding 1 for exactly the workload ENG-578 exists to enable. `scratchpad_boot.py:470` emits on the same path internally (`Waiting for LLM… (Ns)`).

One test closes it: a cell calling `progress()` on a cadence longer than the heartbeat but shorter than the notice threshold, asserting no notice. It fails under that mutation and passes today. Happy for it to be a follow-up.

### Note on the base

`staging` has moved to `b7221ab0` since this branch merged `741802f4` — #345, #348 and #336 have all landed. Worth refreshing before merge; #336 in particular touches `local.py` in the same region.

PR #342, opened 2026-08-12
ENG-695 Recover truncated tool call

ea-rus:
# Never dispatch a tool call the output cap cut open (ENG-695)

## Problem

Anton's planning calls run with `max_tokens = 8192` (`anton/core/settings.py:27`). When a round hits that cap while the model is still writing a tool call's JSON arguments, we ran the handler on the unfinished arguments anyway.

Four of the five ENG-695 sample traces are this shape: output landed on exactly 8192 with the cut inside the arguments — `{"action": "exec", "name": "eng-578-eng-clinic-brief", ..., "estimated_execution_time_seconds": 15` — unterminated, `code` never emitted. Replaying that payload through `safe_parse_tool_input` shows why nothing caught it: the repair pass closes the open brace and returns a **syntactically valid dict with `parse_error: None`**. So the truncation recovery skipped it (its gate was `not llm_response.tool_calls and looks_truncated(...)`, and a partial call makes `tool_calls` non-empty) and the dispatcher's `parse_error` short-circuit skipped it too.

Here the damage was contained by luck — `prepare_scratchpad_exec` rejects an empty `code`, costing one wasted round (~125s, ~$0.28). A tool whose repaired arguments look *complete* would have acted on partial content: the repair can close an open string, it cannot invent the value that never arrived.

## Fix

An unfinished tool call is never dispatched, and no path leaves one unaccounted for.

1. **`ToolCall.repaired`** (`core/llm/provider.py`) — `safe_parse_tool_input` returns `(parsed, parse_error, repaired)`, so "parsed only after being patched up" stops being indistinguishable from "parsed cleanly". `repaired` means the body genuinely ended mid-value: the salvage branch that merely drops trailing junk after a complete object reports `False`, or a relay leaking a stop sentinel would deadlock the turn in re-emit requests.
2. **Both truncation gates** (`core/session.py`, pre-loop and in-loop) read `looks_truncated(...) and not usable_tool_call(...)`. Usable means tool calls exist and none is damaged; one damaged call among intact ones makes the whole round unusable, since dispatching the intact half runs part of what the model was still in the middle of asking for. The in-loop gate matters most in practice — continuation rounds carry the longest history, so the output budget runs out first there.
3. **Every tool loop refuses damaged calls**, via one shared builder (`damaged_tool_call_result`): the streaming loop, the non-streaming `turn()`, and `agentic_loop` in the scratchpad subprocess. Each answers with an `is_error` tool_result asking for a re-emit with smaller arguments, which keeps the recovery inside the tool_use / tool_result protocol. `turn()` and the subprocess have no truncation retry of their own, so there the refusal *is* the recovery.
4. **Both structured-output paths** (`LLMClient.generate_object*` and its sync twin) check `repaired` before validation. This is the silent case: a repaired dict whose cut-off field is optional or `default_factory` validates cleanly, so the caller recorded "no lessons" / a half-sentence verdict and never raised. `parse_error` deliberately stays with the validation branch, which classifies it only when the budget ran out.
5. **The Responses API transport** (`core/llm/openai.py`, streaming and not) parses through the shared helper. It previously either raised `JSONDecodeError` out of the generator or swallowed it into `input={}`, which is indistinguishable from a call sent with no arguments — so `usable_tool_call`'s guarantee had been transport-specific.
6. **Refused calls retire their UI step** (`_discarded_tool_call_events`). `StreamToolUseStart` is emitted while arguments are still streaming, so a step is open before anything knows the call is unusable, and only the markers the dispatch path emits retire one. Both markers are now sent without the execution, following that path's contracts exactly: `tool_done` carries the tool's **name** as its message, because that is how the CLI finds the activity to close (`chat_ui` matches `act.name == message`), while `scratchpad_done` carries prose, as it does on the dispatch path. For a scratchpad cell the tool **result** is sent too — `ok=False` on the marker is not enough, since consumers derive a cell's outcome from its result text and cowork's `classify_cell_status` needs a recognised marker (`[error] exec failed — …` → `cell_status: "error"`; verified by running these events through cowork-server's real formatter). Marker first, result second, matching the dispatch path: a consumer that marks the step complete on the marker has to see the failing result after it.
7. **The retry's nudge names what was lost.** A round can lose both halves at once — paragraphs already shown to the user, then a tool call cut open — so the nudges are composed from one lead and per-loss tails and both instructions travel. Telling the model only to re-issue the call leaves it free to rewrite the answer, putting two copies in history.

## Behavior changes

- Cap-truncated round with a damaged tool call: retried once at double budget instead of dispatched.
- `parse_error` in a cap-truncated round: takes that retry rather than the dispatcher's re-ask at the budget that just ran out.
- `repaired` call in a round that did **not** hit the cap: no longer dispatched, comes back as an error tool_result — one extra round-trip where we previously ran silently on patched-up arguments.
- Structured output on a repaired call: raises `StructuredOutputError` (retryable when the cap was the cause) instead of returning a half-built object. Callers that already log-and-fall-back, like the memory subsystem, now log instead of silently recording nothing.
- A discarded scratchpad cell renders as failed rather than complete, and its CLI activity closes.
- Rounds inside their budget, and cap-truncated rounds with intact tool calls, are untouched.

## Tests

Across `test_session_truncation_recovery.py`, `test_verifier_truncation.py`, `test_openai_provider.py` and `test_provider.py`, each verified to fail without its own fix: repaired and unparseable calls at the cap; a damaged call alongside an intact one; a cut-open call in a *continuation* round (this one caught that the gate is duplicated and only the first copy had been fixed); a repaired call below the cap; the non-streaming `turn()`; the subprocess loop; both Responses-API paths; the retirement events — phase, `message` contract per phase, the failing result and its ordering — and the combined nudge. Full suite: 2011 passed, 28 skipped.

## Known limitation

For a generic (non-scratchpad) tool, cowork-server renders the retirement marker as a progress line rather than `thought.tool_call.end`, because that branch is keyed on `progress_tool_ids` (`stream_formatter.py:321`) and a refused call never enters it. That is the same line cowork already shows when any non-streaming tool completes normally, so it is not a new artifact; the CLI closes such activities correctly via the name match. One line in the formatter would cover it.

## Not in scope

The 197s / 16,384-token outlier in the same ticket is a different client: the Hermes CLI sent `max_tokens: null` *and* `tools: null` with a text tool protocol in its system prompt, so the model role-played the entire agent loop — 20 fabricated tool calls, 21 invented search results. No Anton involvement.

cowork-server needs no change for the paragraphing: its round-break condition is deliberately narrower than the new gate, because this retry re-issues the call rather than finishing the sentence, so the break is correct. Per-call budgets also stay as they are — lowering the main loop's 8192 would produce more of this failure, not less.

Fixes https://linear.app/mindsdb/issue/ENG-695


StpMax (approved):
### 1. `anton/core/llm/anthropic.py:237` — non-streaming `complete()` never sets `repaired`/`parse_error`, so the new guarantee is still transport-specific

This is the same gap the PR explicitly closes for OpenAI. `openai.py:977` (non-streaming `complete()`) was changed in this PR to route through `safe_parse_tool_input`, but the Anthropic non-streaming path was not:

```python
# anthropic.py:237 — non-streaming complete()
ToolCall(id=block.id, name=block.name, input=block.input)
```

versus the streaming path in the same file:

```python
# anthropic.py:361 — stream()
parsed_input, parse_error, repaired = safe_parse_tool_input(raw_json)
ToolCall(..., parse_error=parse_error, repaired=repaired)
```

The path is live: `client.py:288` (`_generate_object_with`, which is exactly where the new `any(tc.repaired for tc in response.tool_calls)` guard sits at `client.py:314`), `client.py:110/198/223` (`plan` / coding / router tiers), and `scratchpad_boot.py:629` (`agentic_loop`). On all of them `repaired` keeps its dataclass default of `False`, so `usable_tool_call()` reports the round as usable, `damaged_tool_call_result()` returns `None`, and the structured-output guard never fires. A half-written dict whose missing field is optional or defaulted validates cleanly and is returned as the model's answer — the exact silent case point 4 of the PR description is meant to eliminate.

One caveat that makes this not a verbatim copy of the OpenAI fix: the Anthropic SDK hands back `block.input` already parsed as a dict, so there is no raw JSON string to feed to `safe_parse_tool_input`. The only signal available on that path is `response.stop_reason`, which is already in scope at `anthropic.py:242`. There is also no test covering Anthropic `complete()` with truncated arguments.

Either close this path, or state in the PR why it is out of scope (for example, if non-streaming Anthropic is not used for tool calls in production).

### 2. `anton/core/session.py:3989` — the truncation gate is copy-pasted rather than extracted

The condition `looks_truncated(...) and not usable_tool_call(...)`, together with the `response = None` / `async for event in self._recover_truncated_stream(...)` recovery block, is duplicated at `session.py:3452` and `session.py:3989`.

The PR description already records what this costs: one of the tests "caught that the gate is duplicated and only the first copy had been fixed". Nothing structural prevents that from happening again on the next change to the condition. Extracting the gate and its recovery block into a single method would remove the failure mode instead of relying on a test to catch it each time.

PR #343, opened 2026-08-12
ENG-1453: Stop the anton test suite from opening real browser windows

alecantu7:
## What

Adds one autouse fixture to `tests/conftest.py` so no test in the anton suite can open a real browser window, plus a regression test that fails if that guard is ever removed. Test-only — no shipped code path changes.

Closes ENG-1453.

## Why

Running the suite locally pops **three browser windows** on the MindsHub signup page. Nothing user-facing is broken; it's a papercut on every local `pytest` run, plus a moment of "did I just trigger something real?"

### Root cause

`_setup_minds` (`anton/cli.py`) asks three yes/no questions through the same `Confirm.ask`:

| Line | Question | What "no" does |
|---|---|---|
| `cli.py:709` | "Do you have a MindsHub API key?" | **opens the signup page in a browser** |
| `cli.py:770` | "Continue without SSL verification?" | skips the insecure retry |
| `cli.py:824` | "Try again?" | gives up, raises `_SetupRetry` |

Three tests in `tests/test_openai_setup.py` want to answer **no** to "Try again?" so setup terminates instead of looping, and do it by replacing `Confirm.ask` wholesale:

```python
monkeypatch.setattr(cli.Confirm, "ask", lambda *a, **kw: False)   # "# no retry"
```

The stub is prompt-blind, so it also answers "no, I don't have an API key" to the *first* prompt — which is exactly the branch that helpfully opens the signup page. Correct product behaviour, wrong context.

The three: `TestSetupErrorTextIsMarkupSafe::test_route_in_error_text_does_not_crash_setup`, `TestSetupErrorTextIsMarkupSafe::test_json_pointer_in_error_text_does_not_crash_setup`, `test_setup_minds_skips_no_ssl_retry_on_http_error`.

## How

```python
@pytest.fixture(autouse=True)
def _no_browser_windows(monkeypatch):
    import webbrowser
    for name in ("open", "open_new", "open_new_tab"):
        monkeypatch.setattr(webbrowser, name, lambda *a, **kw: True)
```

Setup still *decides* to open a browser; the function it calls does nothing. Same *intent* as the suite-wide analytics kill directly above it in that file (ENG-1288) — the suite is not allowed to touch the world outside the process — but deliberately **not** the same mechanism, and the difference is worth stating: the analytics kill is a module-level `os.environ.setdefault(...)`, so it is in force during collection and is inherited by subprocesses; an autouse `monkeypatch` fixture is per-test and parent-process only. That weaker reach is exactly why the seventh call site below is out of scope.

Plus a regression test for the guard itself (`tests/test_suite_guards.py`). Without it the guard is deletable in silence: neuter the fixture and three real browser calls return while CI stays green, because nothing asserts on it — the same rot the guard exists to prevent.

**Why the guard and not the three stubs.** Fixing the stubs fixes today. `webbrowser.open` is reachable from six in-process places — `/publish` and artifact preview (`tools.py:472`, `:613`), `/publish` and `/share` in chat (`chat.py:412`, `:473`, `:780`), and setup (`cli.py:715`) — and any blunt prompt stub can fall into one. Three tests already do; nothing stops the fourth.

**Known limit — there is a seventh call site this guard structurally cannot cover.** `anton/demo_data/nvda_btc_scratchpad_backup.py:497` calls `webbrowser.open` at **module level** (its step 12, "OPEN IN BROWSER"), and `_agent_zero` (`chat.py:937`) runs that file in a real scratchpad cell — i.e. a **subprocess**. An autouse fixture patches this process only, so nothing in `conftest.py` can reach it. Harmless today: nothing imports the module (it is referenced by path), pytest does not collect it, and no test exercises `_agent_zero`. Recorded here and on the ticket rather than papered over, because "fix the class, not the instance" is this PR's whole framing and this is where the class ends. Covering it would mean an env-var guard the subprocess inherits — out of scope, and not worth it until a test actually goes near that path.

**It masks no assertion.** The three `tests/test_publish_*.py` files that already `patch("webbrowser.open")` use it purely to suppress the side effect — checked each: none binds the mock or asserts it was called. Their local patches are now redundant but harmless, and still take precedence for any future test that *does* want to assert.

## Evidence

Measured, not reasoned. Ran the suite with a plugin that swapped `webbrowser.open`/`open_new`/`open_new_tab` for a stack-recorder, so calls were counted instead of executed.

**Before** (1945 passed, 28 skipped, `-m "not slow"`):

```
============================== BROWSER OPEN TRAP ===============================
>>> test_openai_setup.py::TestSetupErrorTextIsMarkupSafe::test_route_in_error_text_does_not_crash_setup
    -> webbrowser.open('https://auth.mindshub.ai/auth/realms/mindsdb/protocol/openid-connect/registrations?...')
       anton/cli.py, line 715, in _setup_minds
>>> test_openai_setup.py::TestSetupErrorTextIsMarkupSafe::test_json_pointer_in_error_text_does_not_crash_setup
>>> test_openai_setup.py::test_setup_minds_skips_no_ssl_retry_on_http_error

TOTAL webbrowser calls: 3
```

Exactly three — one per window, all from `_setup_minds`. No other test reaches any of the other in-process call sites.

Re-measured on the **full** suite including slow tests (`1997 passed / 28 skipped`): still **0** calls with the guard. Nothing hides behind the `slow` marker.

**After**, on this branch off `origin/staging`, same trap still installed:

```
============================== BROWSER OPEN TRAP ===============================
no webbrowser.open* calls
56 passed in 2.29s
```

That's the mutation check: same tests, 3 calls before the guard, 0 after, all still green.

**The guard's own regression test, mutation-verified both directions** (interpreter pinned to 3.11.14, no trap plugin loaded — the trap patches `webbrowser.open` itself and would fool a provenance assertion):

```
fixture neutered  -> 3 failed   tests/test_suite_guards.py
fixture restored  -> 3 passed
full suite        -> 1999 passed, 28 skipped
```

## Security pass

Reviewed per convention 8. The change is a test-only fixture that *removes* an outbound side effect; it adds no input handling, no logging, no new endpoint, no deserialization. Two things checked specifically:

- **Does it weaken a security-relevant assertion?** No — nothing in the suite asserts on `webbrowser.open`, so no test's meaning changes.
- **Does it hide a real problem?** The signup URL being opened is a hardcoded public Keycloak registration endpoint with no secrets in it, and it is correct behaviour for a real user. Nothing is being suppressed except the window.

## Scope

Deliberately **not** in this PR:

- Any change to `anton/cli.py`. A real user answering "no" should still get the signup page.
- **`os.environ.setdefault("BROWSER", "/usr/bin/true")` beside the analytics kill**, which would extend coverage to subprocesses and collection time. Declined, on two grounds. It buys nothing measurable today — the only subprocess call site is the demo-data script, which pytest does not collect and no test exercises, so there is no path it closes. And it is not portable: `/usr/bin/true` is POSIX-only, this repo ships `install.ps1`, and `BROWSER` leaks into *every* subprocess the suite spawns, including scratchpad cells. If a test ever does reach `_agent_zero`, the right form is that line guarded by `os.name == "posix"` — noted on ENG-1453 so the option isn't lost.
- The prompt-aware stub (`lambda prompt, *a, **kw: "API key" in str(prompt)`) that would fix the three tests' *intent* — they currently traverse a branch unrelated to what they assert. Tidiness, not a defect; worth doing next time that file is touched.
- ~~Three pre-existing failures seen during the investigation~~ — **withdrawn, they were never this PR's neighbours.** The original full-suite run happened in a clone sitting on an unrelated feature branch (`alejandrocantu/eng-1390-…`), and I attributed its failures to "the suite" rather than to that branch. Isolated since, with the interpreter held constant at 3.11.14: on clean `origin/staging` all three **pass**; on eng-1390 two of them still fail (`test_scratchpad_exec_via_chat` and `test_scratchpad_in_streaming_path`, both `Process exited unexpectedly`), and the third only fails in a full run, so it is order-dependent. Branch-local, not staging, and nothing to do with this PR. Flagged to the eng-1390 author separately.

## How to verify

```bash
uv run pytest -q tests/test_openai_setup.py
```

Before this change, three browser tabs open on the MindsHub signup page. After, none — and the tests pass either way, which is why this went unnoticed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -75,6 +75,24 @@ def make_llm_response():
 
 
 @pytest.fixture(autouse=True)
+def _no_browser_windows(monkeypatch):
+    """No test may pop a real browser window.
+
+    Every /publish, /share and `anton setup` path ends in ``webbrowser.open``,
+    and a stubbed prompt is all it takes to reach one: three tests in
+    test_openai_setup.py answer ``Confirm.ask`` with a blanket ``False``, which
+    the first prompt in ``_setup_minds`` reads as "no, I don't have an API key"
+    and so opens the MindsHub signup page — three windows on every local run.
+    Same posture as the analytics kill above: the suite never touches the
+    desktop. Tests that assert on the call still patch it themselves.
+    """
+    import webbrowser
+
+    for name in ("open", "open_new", "open_new_tab"):
+        monkeypatch.setattr(webbrowser, name, lambda *a, **kw: True)
+
+
+@pytest.fixture(autouse=True)
 def _no_builtin_skills(tmp_path_factory, monkeypatch):
     """Point the built-in skills root at an empty dir for all tests.
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -83,8 +83,18 @@ def _no_browser_windows(monkeypatch):
     test_openai_setup.py answer ``Confirm.ask`` with a blanket ``False``, which
     the first prompt in ``_setup_minds`` reads as "no, I don't have an API key"
     and so opens the MindsHub signup page — three windows on every local run.
-    Same posture as the analytics kill above: the suite never touches the
-    desktop. Tests that assert on the call still patch it themselves.
+
+    Same *intent* as the analytics kill above — the suite never touches the
+    world outside the process — but not the same mechanism, which matters for
+    what it reaches: that one is a module-level env var, so it is live during
+    collection and inherited by subprocesses, while this is per-test and
+    parent-process only. Hence the one call site it cannot cover, the
+    module-level ``webbrowser.open`` in demo_data/nvda_btc_scratchpad_backup.py
+    that ``_agent_zero`` runs in a scratchpad subprocess (ENG-1453). No test
+    goes near it today.
+
+    Tests that assert on the call still patch it themselves. Deleting this
+    fixture is caught by tests/test_suite_guards.py.
     """
     import webbrowser
 
--- a/tests/test_suite_guards.py
+++ b/tests/test_suite_guards.py
@@ -1,0 +1,36 @@
+"""The suite's own guards must stay armed.
+
+conftest.py disarms things the suite is never allowed to do, and a disarmed
+side effect is invisible when it works — which is exactly how it rots. Delete
+the browser guard and three real `webbrowser.open` calls come back on every
+local run while CI stays green, because no assertion anywhere depends on it
+(ENG-1453). These tests are the thing that fails instead.
+"""
+from __future__ import annotations
+
+import webbrowser
+
+import pytest
+
+
+@pytest.mark.parametrize("name", ["open", "open_new", "open_new_tab"])
+def test_browser_guard_is_installed(name):
+    """`_no_browser_windows` must have replaced the stdlib entry point.
+
+    Checked by provenance rather than behaviour: actually calling the real
+    function is the failure mode under test, so the assertion has to hold
+    before any call happens.
+    """
+    fn = getattr(webbrowser, name)
+    assert fn.__module__ != "webbrowser", (
+        f"the _no_browser_windows guard in tests/conftest.py is not covering "
+        f"webbrowser.{name} — a local run will open real browser windows"
+    )
+    assert callable(fn)
+
+
+# Deliberately NOT asserted by calling it: `webbrowser.open(...) is True` holds
+# whether the guard is installed or not — the real function also returns True,
+# after opening a window. Mutation-verified: with the guard neutered that
+# assertion passes *and* opens a browser, making it both a blind test and an
+# instance of the very problem this file exists to catch.


alecantu7:
## Self-review (adversarial) — head `0485f68d`, base `origin/staging` `8c9a3ef8`

**No defects in the change.** The fix is verified three independent ways. Two corrections to this description, one of which turns out to confirm a different ticket.

### The fix, re-measured rather than trusted

Rebuilt the trap plugin from scratch and ran it on both branches:

```
BEFORE (clean origin/staging, tests/test_openai_setup.py)
  open(https://auth.mindshub.ai/...)  from test_openai_setup.py:344
  open(https://auth.mindshub.ai/...)  from test_openai_setup.py:344
  open(https://auth.mindshub.ai/...)  from test_openai_setup.py:507
  TOTAL webbrowser calls: 3        47 passed

AFTER (PR branch, same trap)
  no webbrowser.open* calls
  TOTAL webbrowser calls: 0        47 passed
```

**The evidence gap in the description is now closed.** The "exactly three" figure came from `-m "not slow"`. I ran the **entire** suite including slow tests: **0 calls across 1997 passed / 28 skipped**. Nothing hides behind the slow marker.

**Mutation:** neutering the fixture to iterate an empty tuple returns it to **3 calls**. The fixture is the cause, not an incidental effect of the diff. Confirmed the mutation applied before trusting the result.

### Completeness checks

Three things would silently defeat this fix. All three hold:

- **Every call site uses module-attribute style** (`webbrowser.open(...)`), never `from webbrowser import open` — verified at all of them. A by-name import would not be reached by the monkeypatch.
- **No `webbrowser.get()`** anywhere in `anton/` or `tests/` — no browser-controller path routes around the module attributes.
- **"It masks no assertion" is true** — all six `patch("webbrowser.open")` sites across the three files, checked individually: none binds the mock (`as x`), none asserts on it.

---

## Correction 1 — there is a seventh call site, and this guard structurally cannot cover it

`anton/demo_data/nvda_btc_scratchpad_backup.py:497` has a **module-level** `webbrowser.open(f"file://{...}")`, and `chat.py:937` runs that file *"in a real scratchpad cell"* — a **subprocess**. An autouse fixture patches the parent process only, so nothing here can reach it.

Practically harmless today: nothing imports the module (it is referenced by path), no test exercises that path, and pytest does not collect it. But the Done-when says *"a future test reaching any of the six `webbrowser.open` sites cannot open one either"* — there are **seven**, and the seventh is out of process. Worth one sentence on ENG-1453, since the whole framing of this ticket is "fix the class, not the instance", and this is where the class ends.

## Correction 2 — the three "pre-existing failures" do not reproduce, and the real one belongs to ENG-1445

The Scope section lists three unrelated failures as out of scope. Run in isolation on this branch: **3 passed**. They did not appear in the full-suite run either.

What *did* fail in the full run is a test not on that list — `tests/test_scratchpad.py::TestSessionPersistence::test_namespace_survives_restart` — and the cause is exact:

```
python this worktree resolved: Python 3.14.3
.python-version present? NO — anton#337 not merged yet
```

That is **ENG-1445 reproducing live** in a fresh worktree: `uv` picks 3.14, `dill==0.3.8` cannot pickle the exec-defined helper, `_encode_values` silently drops the key, `NameError` after restart. Independent confirmation that #337 is the right fix, and a concrete reason to land it soon — every new worktree in this repo currently rolls the dice on interpreter version, and this is exactly the misdiagnosis ENG-1445 was written to prevent.

So the Scope list wants correcting: those three pass, and the one failure actually present is ENG-1445's, not this PR's.

---

### What was reviewed

`tests/conftest.py` (whole diff); all seven `webbrowser` references in `anton/`; all six test-side patches; pytest config and collection scope; the demo-data subprocess path via `chat.py:930-950`. Two worktrees (clean `origin/staging` and PR head), trap plugin written fresh, three runs plus one mutation.

Security pass agrees with the description: the fixture removes an outbound side effect and adds no input handling, logging, endpoint or deserialization. Nothing is suppressed except the window.

The 18 lines are right as written — both corrections are description/ticket edits, not code changes.

alecantu7:
## Response to the self-review — one correction accepted, one rejected on evidence

Both were description-level, so no code changed. Head is still `0485f68d`. Description and ENG-1453 updated for the accepted one.

### Correction 1 — accepted, and it's a real gap in the framing

Verified independently: `anton/demo_data/nvda_btc_scratchpad_backup.py:497` is a module-level `webbrowser.open(f"file://{...}")` under its own `── 12. OPEN IN BROWSER ──` header, and `_agent_zero` (`chat.py:929-937`) resolves that file by path and runs it in a real scratchpad cell. A subprocess, so no autouse fixture in `conftest.py` can reach it.

My "six call sites" was wrong — there are **seven**, and the seventh is out of process. That matters here specifically because the PR argues for a category fix over an instance fix, so the boundary of the category is part of the claim. Both the description and ENG-1453 now state the limit explicitly, along with what covering it would take (an env-var guard the subprocess inherits) and why that's not worth doing until a test goes near `_agent_zero`.

The three completeness checks in the review — module-attribute style at every site, no `webbrowser.get()`, no bound mocks — match what I found. Agreed on all three.

### Correction 2 — rejected. The three failures are real; the review changed two variables at once

The review reports "run in isolation on this branch: **3 passed**" and concludes the Scope list "wants correcting". That run differed from mine in **two** ways — base branch *and* interpreter (3.14 in a fresh worktree vs 3.11 in the original clone) — so a pass there doesn't isolate anything.

Held the interpreter constant at **3.11.14** and varied only the branch:

| Base | Interpreter | `test_scratchpad_exec_via_chat` | `test_scratchpad_in_streaming_path` |
|---|---|---|---|
| clean `origin/staging` | 3.11.14 | pass | pass |
| `alejandrocantu/eng-1390-…` | 3.11.14 | **FAIL** `Process exited unexpectedly` | **FAIL** |

So the failures reproduce, deterministically, and the cause is the **eng-1390 branch** — not the suite, not the interpreter, not this PR. (The third test, `test_the_workspace_is_not_left_on_syspath`, passes in isolation on both and only fails in a full run, so it is order-dependent.)

The right conclusion is narrower than either version: those failures were never this PR's neighbours, because the original full-suite run happened in a clone parked on an unrelated feature branch. That's my error — I attributed branch-local breakage to "the suite". The Scope bullet is now struck through with the isolation table rather than silently deleted, and the eng-1390 breakage has been raised with its author instead of filed here.

### On ENG-1445

The review's incidental finding stands and is useful: its worktree had no `.python-version`, `uv` picked 3.14, and `test_namespace_survives_restart` failed on dill pickling. Under 3.11 on clean staging that test passes. Independent confirmation of the ENG-1445 diagnosis and of why anton#337 should land — every fresh worktree of this repo currently rolls the interpreter dice, which is exactly the confound that produced correction 2. Noted on ENG-1453 as the methodology trap, not as scope.

### Net

No code changes. The 18 lines stand as written. One counted-wrong claim fixed, one withdrawn claim replaced with a measured attribution. All 10 checks green including `run-tests`.

Still wants a human reviewer — I am the author here, so nothing above should count as approval.

alecantu7:
## Second self-review pass — head `0485f68d` (unchanged)

Read the response above first. **The method critique is right and I accept it** — my "3 passed" run varied base branch *and* interpreter, so it isolated nothing. The controlled table does.

One note on the label though: we landed in the same place by different routes. The Scope bullet was struck because those failures were branch-local to `eng-1390` rather than the suite's — which is the correction. My conclusion ("the Scope list wants correcting") was right; my evidence for it was not. "Rejected" undersells the correction that was then made.

Now the inversions I skipped the first time round, which is what a second pass is for.

## Finding A — the guard has no regression net

I ran this mutation in the first pass and reported the number without drawing the conclusion from it. Neutering the fixture to iterate an empty tuple:

```
TOTAL webbrowser calls: 3        47 passed
```

**Three real browser calls come back and the suite stays green.** Nothing in CI can tell the guard was deleted.

For a PR whose central argument is *"fixing the stubs fixes today; the guard fixes the category"*, the guard itself is the one thing with nothing protecting it. Same shape as the enforcement gap on ENG-1282 — a control that can be silenced without anything failing.

Written and mutation-verified rather than proposed blind:

```python
def test_browser_guard_is_active():
    assert webbrowser.open.__module__ != "webbrowser", "guard not installed"
    assert webbrowser.open("https://example.invalid/probe") is True
```

```
with the guard in place:   1 passed
fixture neutered:          1 failed   <- FAILED test_browser_guard_is_active
```

Four lines, and it is the difference between a fix and a fix that stays fixed.

## Finding B — "same posture as the analytics kill" is not accurate, and the difference is exactly the seventh call site

The analytics kill is **not** a fixture. It is module-level, at conftest import:

```python
os.environ.setdefault("ANTON_ANALYTICS_ENABLED", "false")
```

Environment variables are inherited by subprocesses and are already in force during collection. An autouse `monkeypatch` fixture is neither — per-test, parent process only. So the precedent this PR cites uses the *stronger* mechanism and the new guard uses a strictly weaker one, in the same file, seventy lines apart.

That difference **is** the seventh-call-site gap. The response above names the remedy — "an env-var guard the subprocess inherits" — and defers it as not worth doing yet. It is cheaper than that: one line beside the analytics kill. Verified on this machine rather than assumed:

```
BROWSER unset          -> handler: MacOSXOSAScript
BROWSER=/usr/bin/true  -> handler: GenericBrowser
                          open() -> True, and no window appears
```

Keeping **both** is the right shape: the fixture stays (explicit, hermetic, in-process) and `os.environ.setdefault("BROWSER", "/usr/bin/true")` covers subprocesses and collection time.

Worth noting the ticket's "Considered and rejected" entry rules out `BROWSER` *set in CI* — correctly, since that misses local runs, which is where the problem lives. Set in `conftest.py`, that objection does not apply.

One honest caveat against it: `/usr/bin/true` is POSIX-only. anton's CI and Dockerfile are Linux and dev is macOS, so it holds here, but it is not portable — a fair reason to decline.

## Scope re-checked

`tests/e2e/conftest.py` exists but declares no autouse fixture and never touches `webbrowser`, so the root conftest cascades into it correctly. No test files outside `tests/`. No test asserts on the guard today — the three `test_publish_*.py` files patch it but, as established in the first pass, none binds or asserts on the mock.

## Net

Neither finding blocks, and the 18 lines are still right as written. Finding A is the one I would take: four lines, mutation-verified, and it closes the hole in the PR's own argument. Finding B is a one-line addition plus a wording fix in the description, and reasonable to decline on portability grounds.

Still wants a human reviewer — I am the author, so nothing here counts as approval.

alecantu7:
## Response to the second pass — Finding A taken (with a defect in its own test), Finding B split

Head is now `1769749`. One new commit: `tests/test_suite_guards.py`.

On the label quibble: fair. "Rejected" applied to the evidence, not the conclusion, and the conclusion did get acted on — the Scope bullet is struck through either way. Not worth more words than that.

### Finding A — accepted. This was the real hole, and it was in my argument, not the code

The PR argues "the guard fixes the category" while the guard itself had nothing holding it in place. Reproduced the mutation independently: neuter the fixture to iterate an empty tuple and **3 real `webbrowser.open` calls return with the suite fully green**. Nothing in CI could tell the guard had been deleted. That is the same rot the guard exists to prevent, one level up.

**But the proposed test is half wrong, and the wrong half is instructive.** This line:

```python
assert webbrowser.open("https://example.invalid/probe") is True
```

passed **with the guard neutered**, because the real `webbrowser.open` also returns `True` — after opening a window. So that assertion is (a) blind as a regression detector and (b) an instance of the exact bug this file exists to catch. It opened a Safari window on my machine while I was verifying it, which is how I noticed.

What landed asserts by **provenance** instead — the entry point is no longer the stdlib function — parametrized across all three patched names, and never calls it:

```python
@pytest.mark.parametrize("name", ["open", "open_new", "open_new_tab"])
def test_browser_guard_is_installed(name):
    fn = getattr(webbrowser, name)
    assert fn.__module__ != "webbrowser", (...)
    assert callable(fn)
```

Mutation-verified in both directions, interpreter pinned to 3.11.14:

```
fixture neutered  -> 3 failed
fixture restored  -> 3 passed
full suite        -> 1999 passed, 28 skipped
```

One methodology note worth recording, since it nearly produced a false green: **the trap plugin cannot be loaded while verifying this test.** The trap patches `webbrowser.open` itself, so its own replacement satisfies the provenance assertion — my first mutation run reported "3 passed" with the guard neutered purely because `-p browser_trap` was on the command line. Re-ran without it to get the real answer. A measurement instrument that looks like the thing being measured.

### Finding B — wording accepted, env var declined

**The wording critique is correct and I've fixed it.** "Same posture as the analytics kill" was wrong: the analytics kill is a module-level `os.environ.setdefault(...)`, so it is in force at collection and inherited by subprocesses; an autouse `monkeypatch` fixture is per-test and parent-process only. Strictly weaker reach, seventy lines apart in the same file. Description and ticket now say "same *intent*, deliberately not the same mechanism", and name that weaker reach as the reason the seventh call site is out of scope.

**The env var itself I'm declining**, on two grounds rather than the portability one alone:

1. **It closes no path that exists today.** The only subprocess call site is the demo-data script — not collected by pytest, not imported, and no test exercises `_agent_zero`. Adding a guard for a path nothing reaches means the guard has no regression test either, which is precisely Finding A's complaint.
2. **Portability is worse than "not portable".** `/usr/bin/true` is POSIX-only, this repo ships `install.ps1`, and `BROWSER` set in `conftest.py` leaks into *every* subprocess the suite spawns — including scratchpad cells running agent-authored code. Broad blast radius for zero present coverage.

Recorded on ENG-1453 as the named fix **if** a test ever reaches `_agent_zero`: that line guarded by `os.name == "posix"`. Better in the ticket than in the tree.

### Net

`tests/conftest.py` unchanged from `0485f68d`. One new test file, four lines of assertion, mutation-verified. Two description corrections. CI re-running on the new head.

Still wants a human reviewer — I'm the author, and two rounds of self-review don't substitute for one.

tests/conftest.py
```diff
@@ -74,6 +74,24 @@ def _factory(
...
+@pytest.fixture(autouse=True)
+def _no_browser_windows(monkeypatch):
+    """No test may pop a real browser window.
+
+    Every /publish, /share and `anton setup` path ends in ``webbrowser.open``,
+    and a stubbed prompt is all it takes to reach one: three tests in
+    test_openai_setup.py answer ``Confirm.ask`` with a blanket ``False``, which
+    the first prompt in ``_setup_minds`` reads as "no, I don't have an API key"
+    and so opens the MindsHub signup page — three windows on every local run.
+    Same posture as the analytics kill above: the suite never touches the
```
pnewsam: **Nit (optional, no failure mode):** the docstring still reads "Same posture as the analytics kill above," but your own Finding B established this framing is imprecise — the analytics kill is a module-level `os.environ.setdefault(...)` (in force at collection, inherited by subprocesses) while this is a per-test, parent-process-only `monkeypatch` fixture. You corrected the PR body and ticket to "same *intent*, deliberately not the same mechanism"; the in-code comment lags that correction. Purely cosmetic — the code is right as written.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (both files), all seven `webbrowser` call sites in `anton/`, the six test-side patches, the analytics-kill precedent, and the root-cause stubs in `test_openai_setup.py`. Ran the tests and the mutation check in an isolated worktree at head `1769749`. Test-only change that removes an outbound side effect and adds a regression net for it; every structural claim in the description holds up under independent checking. No actionable defects — one optional doc nit inline.

### Validation

- **Guard test**: `pytest tests/test_suite_guards.py` -> 3 passed.
- **Mutation, both directions** (independently reproduced): fixture neutered to iterate `()` -> 3 failed; restored -> the 47 root-cause tests in `test_openai_setup.py` pass with the guard active.
- **No conflict with existing patches**: `test_publish_tool_access.py` + `test_publish_chat_access.py` + `test_publish_api_key.py` -> 9 passed; their own `mock.patch("webbrowser.open")` coexists with the autouse fixture (monkeypatch restores cleanly).
- **Call sites verified by grep**: 6 in-process + 1 module-level 7th (`demo_data/nvda_btc_scratchpad_backup.py:497`); all module-attribute style, no `from webbrowser import open`, no `webbrowser.get()`.
- **"Masks no assertion" verified**: none of the six test-side `webbrowser.open` patches bind (`as x`) or assert on the mock.
- **CI**: all checks green on head `1769749`.
- *Limitation*: the fresh worktree had no `.python-version`, so `uv` selected 3.12.10 rather than 3.11.14; the behavior under test is interpreter-agnostic, so this doesn't affect confidence.

tests/conftest.py
alecantu7: Fixed in `47dab9a` — good catch, and it was the one place the correction hadn't reached. The docstring now states the intent/mechanism split and, since the reach difference is the whole reason the seventh call site is out of scope, names that consequence in-place rather than leaving it only in the PR body:

```
Same *intent* as the analytics kill above — the suite never touches the
world outside the process — but not the same mechanism, which matters for
what it reaches: that one is a module-level env var, so it is live during
collection and inherited by subprocesses, while this is per-test and
parent-process only. Hence the one call site it cannot cover, the
module-level `webbrowser.open` in demo_data/nvda_btc_scratchpad_backup.py
that `_agent_zero` runs in a scratchpad subprocess (ENG-1453). No test
goes near it today.
```

Also added a pointer to `tests/test_suite_guards.py`, so anyone reading the fixture knows deleting it fails a test rather than going silent.

Comment-only, no behaviour change; `tests/test_suite_guards.py` + `tests/test_openai_setup.py` still green (50 passed).

PR #344, opened 2026-08-12
ENG-1286: bound a turn's token spend and ask the user before continuing

alecantu7:
Closes the spend-ceiling half of [ENG-1286](https://linear.app/mindsdb/issue/ENG-1286). The thrash-breaker half is split to [ENG-1492](https://linear.app/mindsdb/issue/ENG-1492) (instrumentation first) and is **not** in this PR.

## What this does

A turn that spends **1,250,000 raw tokens** stops, tells the user what it accomplished and what remains, and asks whether to continue. Reads ENG-1288's `TurnCost.total_tokens`; no new plumbing.

Before this, nothing anywhere counted what a single task had spent. The Kiranam session (ENG-836) burned ~3.64M tokens in ~50 minutes for no working result — **73% of a 5M monthly allowance in one sitting** — and breached no existing bound: never 60 RPM, never 20 concurrent, TPM only intermittently.

## Three decisions worth reviewing, with the evidence

**1. Raw tokens, not cost-weighted.** The instinct is to discount cache reads since they bill at ~a tenth. That is wrong *for this control*: cache reads draw the user's included-token allowance at full 1:1 weight — `auth/entitlements/services/included_usage.py:88-92` sums `input + output + cached_input`, and its docstring says so deliberately (*"leaving them out would let a long cached prefix buy far more than the granted allowance"*). The cheap rate applies only to overage dollars. This ceiling protects the **user's allowance**, so it counts the unit that allowance drains in. The TPM limiter's identical weighting *is* a defect (ENG-1132) because it guards money — same arithmetic, opposite verdict, because they protect different things.

**2. 1.25M, not 1.5M.** Kiranam's worst single turn was **1,480,766** tokens. 1.5M is the number you reach for instinctively and it misses the motivating case by 20k.

**3. The reserve is adaptive — `ceiling - max(200k, 2 × peak_context_tokens)`.** Two calls land after the last passing check, not one: the call that carries the total over the gate (it is only read *between* calls) and then the hand-back. Simulated over the same 30 days of production:

| reserve | trips | turns still **exceeding** 1.25M | worst |
|---|---|---|---|
| flat 200k | 12.3% | **15.6%** of trips | 1,704,923 |
| adaptive (this PR) | 12.7% | **1.2%** of trips | 1,546,740 |

The flat reserve breaks the bound for 1 in 6 trips. The adaptive one costs 0.4pp more trips. Residual, stated rather than hidden: the reserve estimates from the peak seen *so far*, so a call suddenly exceeding twice the previous peak can still overshoot.

## Placement

Checked **per tool round** and at the continuation gate, not the continuation gate alone. The measured runaway shape is 13–26 *consecutive* `scratchpad` calls inside one tool loop, reaching the round cap with **zero** continuations — a continuation-gate-only check never sees it. `test_trips_inside_one_tool_loop_with_no_continuation` is the regression guard.

Verification is skipped after a trip (same as the round cap): a verdict call would spend more of the budget just declared exhausted, and an INCOMPLETE verdict would force a continuation straight past the ceiling.

## Measured impact — read this before approving

This is a **visible product behaviour**, not a silent backstop. Over 30 days: **19.4% of external turns** would trip (351/1,812), 7.1% of staff turns, **12.7% overall ≈ 17 turns/day**. It fires at a median of **16 LLM calls** into the turn (p10 6, p90 25) — comparable to where the 25-round cap already sits.

The position taken: a turn spending a quarter of someone's monthly allowance should say so. If you disagree, the number is one constant.

## The path to continue

- Continuing is already free — a new user message is a new turn with a fresh `TurnCost`. Consent renews per turn rather than latching; a permanent "yes" would restore the runaway.
- The message mirrors the round cap's wording (`Do NOT retry automatically — wait for the user's response`) so users meet one behaviour, not two.
- Persisted via `_stream_handback_diagnosis`, so history holds what the user actually read (ENG-1155 property).

**Deviations from the ticket, flagged deliberately.**

*Two, not one.* The second was raised in review and is declared here rather than left for QA to file as a miss:

**(a) No resume injection.** ENG-1286's *How* says the resume should carry the existing *"Do not repeat work already done"* wording, and *Done when* has *"resumes without redoing completed work"*. The **budget** half is implemented and correct — the next turn gets a fresh `TurnCost`. The **resume** half has no mechanism: that wording exists only inside the verifier-INCOMPLETE branch, which a ceiling exit never reaches. Left as-is because the same ticket bullet says to reuse the round cap's shape rather than invent a second one, and the round cap plus the other three hand-back paths have no resume injection either. Adding one here alone would make the ceiling behave differently from every other stop.

**(b) The message quotes tokens, not allowance.** The ticket says the message should express cost *in allowance terms*. It does not. The turn knows its token count but not the plan behind it, and BYOK users have no MindsHub allowance at all — quoting "a quarter of your month" to them would be a confident lie. Wiring `check_minds_token_limits()` into the session is a follow-up. Reject this call if you'd rather have the allowance framing gated on provider.

## Testing

`tests/test_spend_ceiling.py` — 10 tests driving the real `turn_stream`/`turn`. The fake LLM invokes `usage_listener` exactly as `LLMClient` does; a stub that skipped it would leave `total_tokens` at 0 and every test would pass by never reaching the gate.

**Mutation-verified** — each of these was applied and the intended test failed:
- disable the per-round gate → the no-continuation placement test fails *(**corrected 2026-08-13:** as originally written this claim was FALSE. That test passed under the mutation; two others failed instead, and I inferred which one from a ``1 failed`` count without checking. The fixture tripped on the very first call, so its ``rounds``/``continuations`` assertions were unfalsifiable. Fixture re-sized; the claim now holds and is re-verified.)*
- let the verifier run after a trip → the skip test fails
- flat reserve instead of adaptive → the bound and scaling tests fail
- count only `input_tokens` (discount cache) → the raw-unit tests fail

Full suite: **2,009 passed**, 2 failures in `test_datasource.py` that reproduce on unmodified `origin/staging` (verified by stashing this diff) — pre-existing, not from this change.

## Security

Reviewed per convention 8. No new input parsing, endpoints, or authz surface. The one new user-visible string interpolates an integer from our own accounting — no user input reaches it. New log lines carry token counts and config values only, never conversation content or the verifier's `reason` (matching the existing verifier-logging convention). No path handling, deserialization, or credential access added.

## Merge order

Independent and safe to merge alone — the ceiling works standalone at its default. The follow-up cowork-server PR exposing **"Max Tokens per Step"** as a per-user setting (beside "Max Steps per Task" and "Max Auto-Continues") should merge **after** this one, since it overlays a field that must exist in anton first.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2211,7 +2211,8 @@ class ChatSession:
         fire after real work had happened, on the reasoning that every call site
         sits inside the tool loop; that is true of LLM calls and false of tool
         dispatches, which happen later in the same iteration. The per-round call
-        site carries the progress condition explicitly — see `_spend_ceiling_ok`.
+        site carries the progress condition explicitly — see
+        `_spend_ceiling_stops_the_tool_loop`.
         """
         if self._max_turn_tokens <= 0 or self._turn_cost is None:
             return False
@@ -2788,7 +2789,14 @@ class ChatSession:
             # the cap mark above: public API, and the books are wired here too.
             # This path runs no completion verifier, so there is no continuation
             # gate to also check — the per-round check is the whole gate here.
-            if self._spend_ceiling_reached():
+            #
+            # Uses the same `_spend_ceiling_stops_the_tool_loop` as the streaming
+            # loop, NOT the bare predicate: the "never end a turn that dispatched
+            # no tools" guarantee is stated unconditionally ("at any ceiling"),
+            # and CLI/host callers have no lower bound on `max_turn_tokens` at
+            # all, so this path is exactly where a tiny ceiling would break the
+            # loop on round 1 having run nothing (#344 review).
+            if self._spend_ceiling_stops_the_tool_loop():
                 if self._turn_cost is not None:
                     self._turn_cost.ended_by = "spend_ceiling"
                 logger.info(
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -292,6 +292,21 @@ _TRANSIENT_VERDICT_ERRORS: tuple[type[BaseException], ...] = (
 # verdict" can never fire, since a latched session makes no verdict calls.
 _VERIFIER_LATCH_REPROBE_TURNS = 10
 
+# Floor for the tokens held back from the spend ceiling (ENG-1286).
+#
+# TWO calls land after the last check that passed, not one: the call that
+# carried the total over the gate (the gate is only read between calls, so the
+# crossing is never observed mid-flight), and then the hand-back diagnosis,
+# which is a full-history planning call. So the reserve has to cover both or the
+# turn overshoots the ceiling it promised — the actual per-call cost on these
+# turns is ~200k, which a flat 200k reserve under-covers by half.
+#
+# The live reserve is therefore `max(this, 2 x peak_context_tokens)`: the
+# turn's own largest call is the best available estimate of what the next two
+# will cost, and ENG-1288 already tracks it. This floor only applies before any
+# call has reported usage.
+_SPEND_CEILING_RESERVE = 200_000
+
 # Appended to the verifier system prompt. Shortens the preamble on narrating
 # models but is not sufficient alone (0/3 at 256 with it), so it pairs with the
 # budgets above. Tool name comes off the schema class so it can't go stale.
@@ -669,6 +684,10 @@ class ChatSession:
         self._max_tool_rounds = s.max_tool_rounds
         self._max_continuations = s.max_continuations
         self._verify_min_tool_rounds = s.verify_min_tool_rounds
+        # Per-turn raw-token ceiling (ENG-1286). getattr so a host passing an
+        # older settings object doesn't break — absent means "no ceiling",
+        # matching the pre-ENG-1286 behaviour rather than silently applying one.
+        self._max_turn_tokens = getattr(s, "max_turn_tokens", 0)
         # Latch for a verifier that fails the same hard way every turn — e.g.
         # kimi-K3 rejecting forced `tool_choice` with a 400 (ENG-1095), which
         # fails on every verdict call until the gateway fix lands. Without the
@@ -2175,6 +2194,71 @@ class ChatSession:
             # Reporting must never affect the turn that just ran.
             pass
 
+    def _spend_ceiling_reached(self) -> bool:
+        """True when this turn has spent enough that it must stop and ask.
+
+        Reads `TurnCost.total_tokens` — raw tokens, the unit the user's included
+        allowance drains in (see `CoreSettings.max_turn_tokens` for why cache
+        reads are NOT discounted here). Gated at `ceiling - reserve` so the
+        hand-back's own call fits inside the promised ceiling.
+
+        Returns False when the ceiling is disabled (0) or the books are closed,
+        so a host on older settings, or any path that runs outside a turn, keeps
+        exactly its pre-ENG-1286 behaviour.
+
+        Note this can only ever be consulted AFTER at least one LLM call has been
+        made — every call site sits inside the tool loop, which is only entered
+        once a planning response exists. So the gate cannot produce a turn that
+        made no call and said nothing.
+        """
+        if self._max_turn_tokens <= 0 or self._turn_cost is None:
+            return False
+        return self._turn_cost.total_tokens >= self._spend_ceiling_gate()
+
+    def _spend_ceiling_gate(self) -> int:
+        """The token total at which the turn must stop starting new work.
+
+        Sits a reserve below the ceiling because two more calls are still coming
+        after the last check that passed — the one that crosses the gate, and
+        the hand-back. Sized from this turn's own `peak_context_tokens` rather
+        than a constant: a turn carrying 190k of context per call overshoots a
+        flat 200k reserve by roughly a full call, which is precisely the bound
+        the ceiling exists to make true.
+
+        Floored at 1 so a pathological ceiling smaller than the reserve still
+        gates on something positive instead of tripping before any work.
+        """
+        peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
+        reserve = max(_SPEND_CEILING_RESERVE, 2 * peak)
+        return max(self._max_turn_tokens - reserve, 1)
+
+    def _spend_ceiling_notice(self) -> str:
+        """The SYSTEM injection for a ceiling stop.
+
+        Mirrors the 25-round cap's pause-and-ask wording deliberately: the user
+        already meets that shape when a long turn stops, and two different
+        "I stopped, shall I go on?" behaviours would read as two different
+        products. `Do NOT retry automatically` is what keeps this a question
+        rather than a speed bump the model drives straight through.
+
+        Deliberately does NOT quote a share of the user's allowance. The turn
+        knows its token count but not the plan behind it — BYOK users have no
+        MindsHub allowance at all — and inventing "you've used a quarter of your
+        month" for them would be a confident lie. The model is told what it
+        spent and asked to be honest that it was a lot.
+        """
+        spent = self._turn_cost.total_tokens if self._turn_cost is not None else 0
+        return (
+            f"SYSTEM: This turn has used {spent:,} tokens, which is a large amount "
+            "of work for a single request and counts against the user's plan. "
+            "Pause here. Summarize what you have accomplished so far and what "
+            "remains. Be straightforward that this has already taken a lot of "
+            "work, so the user can decide whether it is worth continuing. "
+            "If you are on a good track and can finish, say so and ask if they'd "
+            "like you to continue. "
+            "Do NOT retry automatically — wait for the user's response."
+        )
+
     async def _stream_handback_diagnosis(self, *, system: str, label: str):
         """Stream a hand-back diagnosis and persist exactly what the user read.
 
@@ -2652,6 +2736,28 @@ class ChatSession:
                             "Do NOT retry automatically — wait for the user's response."
                         ),
                     }
+                )
+                response = await self.plan_with_recovery(system=system)
+                break
+
+            # Spend ceiling (ENG-1286), mirroring turn_stream. Same reasoning as
+            # the cap mark above: public API, and the books are wired here too.
+            # This path runs no completion verifier, so there is no continuation
+            # gate to also check — the per-round check is the whole gate here.
+            if self._spend_ceiling_reached():
+                if self._turn_cost is not None:
+                    self._turn_cost.ended_by = "spend_ceiling"
+                logger.info(
+                    "spend ceiling reached (turn): tokens=%d ceiling=%d "
+                    "tool_round=%d — pausing to ask the user",
+                    self._turn_cost.total_tokens if self._turn_cost else 0,
+                    self._max_turn_tokens, tool_round,
+                )
+                self._append_history(
+                    {"role": "assistant", "content": response.content or ""}
+                )
+                self._append_history(
+                    {"role": "user", "content": self._spend_ceiling_notice()}
                 )
                 response = await self.plan_with_recovery(system=system)
                 break
@@ -3412,6 +3518,12 @@ class ChatSession:
         # task isn't actually done yet.
         continuation = 0
         _max_rounds_hit = False
+        # Set when the per-turn spend ceiling stopped the turn (ENG-1286). Like
+        # `_max_rounds_hit` it also suppresses verification: the turn is already
+        # ending on an honest "I stopped, shall I continue?", and a verdict call
+        # would both cost more of the budget we just declared spent and risk
+        # forcing a continuation past it.
+        _spend_ceiling_hit = False
         # Set per verification-loop iteration (see below): tells the post-loop
         # fallback the current reply is already in history, so it must not append
         # it a second time (ENG-1155 double-append).
@@ -3476,6 +3588,41 @@ class ChatSession:
                     _reply_persisted = True
                     async for event in self._stream_handback_diagnosis(
                         system=system, label="max-tool-rounds"
+                    ):
+                        yield event
+                    break
+
+                # Spend ceiling (ENG-1286). Checked per round rather than only at
+                # the continuation gate: the measured runaway shape is 13-26
+                # CONSECUTIVE scratchpad calls inside one tool loop, which reaches
+                # the round cap having triggered no continuation at all — a
+                # continuation-gate-only check never sees it. Ordered after the
+                # round cap so a turn that breaches both still reports
+                # `round_cap`, leaving that path's behaviour untouched.
+                if self._spend_ceiling_reached():
+                    _spend_ceiling_hit = True
+                    if self._turn_cost is not None:
+                        self._turn_cost.ended_by = "spend_ceiling"
+                    logger.info(
+                        "spend ceiling reached: tokens=%d ceiling=%d reserve=%d "
+                        "tool_round=%d continuation=%d — pausing to ask the user",
+                        self._turn_cost.total_tokens if self._turn_cost else 0,
+                        self._max_turn_tokens, _SPEND_CEILING_RESERVE,
+                        tool_round, continuation,
+                    )
+                    self._append_history(
+                        {"role": "assistant", "content": llm_response.content or ""}
+                    )
+                    self._append_history(
+                        {"role": "user", "content": self._spend_ceiling_notice()}
+                    )
+                    # Same ENG-1155 capture the other four hand-back sites need:
+                    # the reply above is already in history, so without this the
+                    # post-loop fallback appends it again and the message the user
+                    # actually read is lost.
+                    _reply_persisted = True
+                    async for event in self._stream_handback_diagnosis(
+                        system=system, label="spend-ceiling"
                     ):
                         yield event
                     break
@@ -3960,7 +4107,7 @@ class ChatSession:
             # Skip when too few tool rounds were used (pure Q&A always skips at
             # tool_round==0; raising verify_min_tool_rounds also skips trivial
             # single-round turns) or when we hit the max-rounds hard stop.
-            if tool_round < self._verify_min_tool_rounds or _max_rounds_hit:
+            if tool_round < self._verify_min_tool_rounds or _max_rounds_hit or _spend_ceiling_hit:
                 break
 
             # Append the assistant's final text so the verifier can see it.
@@ -4236,7 +4383,35 @@ class ChatSession:
                     yield event
                 break
 
-            # INCOMPLETE — continue working
+            # INCOMPLETE — continue working, unless the budget for this turn is
+            # already gone. Checked here as well as per-round because a
+            # continuation re-enters the tool loop with a full history: the round
+            # counter resets to 0, so without this gate a turn can spend the
+            # ceiling several times over across continuations and each pass looks
+            # cheap to the per-round check (ENG-1286).
+            if self._spend_ceiling_reached():
+                _spend_ceiling_hit = True
+                if self._turn_cost is not None:
+                    self._turn_cost.ended_by = "spend_ceiling"
+                logger.info(
+                    "spend ceiling reached at the continuation gate: tokens=%d "
+                    "ceiling=%d continuation=%d — pausing to ask the user",
+                    self._turn_cost.total_tokens if self._turn_cost else 0,
+                    self._max_turn_tokens, continuation,
+                )
+                self._append_history(
+                    {"role": "user", "content": self._spend_ceiling_notice()}
+                )
+                # `_reply_persisted` is already True here — the verification block
+                # appends the assistant reply before requesting a verdict — so the
+                # post-loop fallback is already suppressed and only the diagnosis
+                # needs capturing.
+                async for event in self._stream_handback_diagnosis(
+                    system=system, label="spend-ceiling"
+                ):
+                    yield event
+                break
+
             continuation += 1
             if self._turn_cost is not None:
                 self._turn_cost.continuations = continuation
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2206,30 +2206,74 @@ class ChatSession:
         so a host on older settings, or any path that runs outside a turn, keeps
         exactly its pre-ENG-1286 behaviour.
 
-        Note this can only ever be consulted AFTER at least one LLM call has been
-        made — every call site sits inside the tool loop, which is only entered
-        once a planning response exists. So the gate cannot produce a turn that
-        made no call and said nothing.
+        This is pure arithmetic and says nothing about whether the turn has done
+        any WORK yet. An earlier version of this docstring claimed it could only
+        fire after real work had happened, on the reasoning that every call site
+        sits inside the tool loop; that is true of LLM calls and false of tool
+        dispatches, which happen later in the same iteration. The per-round call
+        site carries the progress condition explicitly — see `_spend_ceiling_ok`.
         """
         if self._max_turn_tokens <= 0 or self._turn_cost is None:
             return False
         return self._turn_cost.total_tokens >= self._spend_ceiling_gate()
 
+    def _spend_ceiling_stops_the_tool_loop(self) -> bool:
+        """Should the tool loop stop here on the spend ceiling?
+
+        The ceiling, plus the one thing it must never do: end a turn that has
+        dispatched no tools at all. The gate is checked at the TOP of a round,
+        before dispatch, so on the first round a small ceiling would break the
+        loop having run nothing — the user gets a "this used a lot of tokens"
+        message for a task that never started.
+
+        `TurnCost.rounds` is turn-wide (it accumulates across verifier-forced
+        continuations rather than resetting with the loop-local counter), so
+        `> 1` means at least one round has completed its dispatch — including
+        rounds from an earlier continuation, which is why a continuation cannot
+        hand out a fresh free round.
+
+        Belt and braces with the reserve cap in `_spend_ceiling_gate`: that cap
+        keeps the gate proportional to the ceiling, this keeps the guarantee true
+        even for a ceiling smaller than a single call — a shape the settings
+        floor is supposed to prevent, but which anton's own `max_turn_tokens`
+        has no bounds on for CLI and host callers.
+        """
+        if not self._spend_ceiling_reached():
+            return False
+        rounds = self._turn_cost.rounds if self._turn_cost is not None else 0
+        return rounds > 1
+
     def _spend_ceiling_gate(self) -> int:
         """The token total at which the turn must stop starting new work.
 
         Sits a reserve below the ceiling because two more calls are still coming
-        after the last check that passed — the one that crosses the gate, and
-        the hand-back. Sized from this turn's own `peak_context_tokens` rather
-        than a constant: a turn carrying 190k of context per call overshoots a
-        flat 200k reserve by roughly a full call, which is precisely the bound
-        the ceiling exists to make true.
+        after the last check that passed — the one that crosses the gate (the
+        total is only read *between* calls, so the crossing is never observed
+        mid-flight), and the hand-back. Sized from this turn's own
+        `peak_context_tokens` rather than a constant: a turn carrying 190k of
+        context per call overshoots a flat 200k reserve by roughly a full call.
 
-        Floored at 1 so a pathological ceiling smaller than the reserve still
-        gates on something positive instead of tripping before any work.
+        **The reserve is capped at half the ceiling**, and that cap is the whole
+        reason small ceilings are usable. Without it the reserve bears no
+        relation to the ceiling's size, so any ceiling at or below the reserve
+        drove this to the `max(..., 1)` floor — the gate then tripped on the
+        first check, before any tool had been dispatched, and the turn still
+        paid for two calls. Measured at a 100k ceiling with 190k contexts: zero
+        tools run, 400k spent, 300% over the number the user asked for. The cap
+        binds only below ~2x the reserve; at the 1.25M default it changes
+        nothing.
+
+        The bound this produces is close, not exact: the reserve is derived from
+        `peak_context_tokens` (input + cache_read + cache_creation) but compared
+        against `total_tokens`, which also counts output, so a turn can land
+        over the ceiling by roughly two output budgets. See
+        `test_total_is_bounded_by_the_ceiling` for the asserted form.
         """
         peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
-        reserve = max(_SPEND_CEILING_RESERVE, 2 * peak)
+        reserve = min(
+            max(_SPEND_CEILING_RESERVE, 2 * peak),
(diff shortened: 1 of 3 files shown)


alecantu7 (commented):

Reviewed at `79a36e9` against `origin/staging` `f4c7ae3` (verified against the GitHub API, not the local ref). Both findings below are **already addressed in `79a36e9`** — posting them so the trail of what was caught pre-review is visible.

### 1 — A test that could not fail · medium · confirmed by execution

`tests/test_spend_ceiling.py::test_verification_is_skipped_after_a_ceiling_trip` guarded with a raising side-effect:

```python
session._llm.generate_object_code = AsyncMock(
    side_effect=AssertionError("verifier must not run after a ceiling trip"))
```

The verdict site catches bare `Exception` and converts anything raised there into a "hard verdict failure" that continues down the diagnosis path — so the assertion was swallowed by the code under test.

**Proof:** with `_spend_ceiling_hit` removed from the verification-skip condition, the test still passed. The behaviour was covered incidentally (another test's `ended_by` assertion caught the mutation), but the test named for it proved nothing.

**Fix:** assert on call count. Re-mutated after the fix — now fails with `the completion verifier ran 1x after the ceiling stopped the turn`.

### 2 — Misleading dead fixture code · low · confirmed

The fixture built a `MagicMock` settings object, populated seven fields, and never passed it to `ChatSessionConfig`. It read as though those values were in effect when the session was resolving a real `CoreSettings`. Removed.

---

### Findings I killed at the refutation gate

Recorded so a reviewer doesn't spend time re-deriving them:

- **Adaptive gate degenerating at large contexts.** `gate = max(ceiling - 2*peak, 1)` collapses to 1 when peak context exceeds half the ceiling, which would stop a big-context turn after one call. **Killed by data:** 0 of 4,049 turns over 30 days have peak context above 625k, and 0 of 513 trips fire on the first check.
- **Silent disablement via the `getattr` default (the ENG-597 class).** `getattr(s, "max_turn_tokens", 0)` fails *open*: a host passing a settings object without the field gets no ceiling, and every test would still pass. **Checked:** `anton/config/settings.py:31` is `class AntonSettings(CoreSettings)` and `cowork-server .../harness.py:804` passes exactly that object, so the ceiling is live on the desktop path. Grepped all 13 repos in the workspace — no other writer of this setting exists.

### Open risk a reviewer should weigh

`ended_by="spend_ceiling"` is a **new value in an enum the ENG-1355 collector relays**. That collector filters by allowlist, which is why `turn_completed` produced nothing for weeks while its log line looked healthy. If it validates `ended_by` values rather than passing the string through, ceiling trips will be invisible in PostHog — silently, and in the reassuring direction. Not verifiable from this repo (the lambda is in no local clone); worth a probe once ENG-1355 lands.

Also stated rather than hidden: the 1.25M bound holds in tests and for ~98.8% of simulated production trips. A call that suddenly exceeds twice the previous peak can still overshoot (worst simulated: 1,546,740).

### Evidence

Full anton suite: **2,009 passed**. The 2 `test_datasource.py` failures reproduce on unmodified `origin/staging` (verified by stashing this diff) — pre-existing, not from this change. 5 targeted mutations run, each caught by its intended test after the fix.

alecantu7:
## Adversarial self-review, deep mode — head `2b676da1`, base `origin/staging` `f4c7ae35`

26 agents across five failure classes, every finding adversarially refuted before reporting. 19 of 20 survived; deduped to 11 across the three PRs, **6 of them here**. One high.

**Nothing below blocks the default behaviour. Finding 1 blocks shipping the new setting.**

---

## 1 — HIGH · the reserve can exceed the whole ceiling, so the bottom of the setting's advertised range dispatches zero tools

`anton/core/session.py:2232-2233` — introduced. Four independent finders reported this; all four verifiers confirmed, three rated it high.

```python
reserve = max(_SPEND_CEILING_RESERVE, 2 * peak)
return max(self._max_turn_tokens - reserve, 1)
```

The reserve bears no relation to the ceiling's size, so for any ceiling <= 200,001 the gate collapses to the `max(..., 1)` floor. `_spend_ceiling_reached()` is consulted at `session.py:3602`, **before** the tool-dispatch block at `:3640`, so the loop breaks having dispatched nothing.

**The band is reachable through the supported UI, and all three repos agree on it:**

```
cowork  settingsTransform.js:603   min: 100_000
cs      user_settings.py:524       ge=100_000
cs      app_settings.py:601        ge=100_000   <- COWORK_DEFAULT_MAX_TURN_TOKENS, so one operator env var bricks an org
anton   CoreSettings.max_turn_tokens          no bounds at all
```

The renderer *snaps into* the broken band — `clampBudgetValue('0', spec) === '100000'` is an assertion in this PR's own sibling test — and the field's subtitle invites it: "lower it to cap what any single request can cost".

**Executed**, driving the real `turn_stream` and counting `tool_registry.dispatch_tool.call_count`:

```
ceiling=  100,000  per_call=200,000 | tools_run=0  rounds=1  ended_by=spend_ceiling  total=400,000  (+300%)
ceiling=  200,000  per_call=200,000 | tools_run=0  rounds=1                          total=400,000
ceiling=  250,000  @190k context     | tools_run=0                                   total=388,000  (+55%)
ceiling=1,250,000  (default)         | tools_run=17 rounds=18                        <- healthy
```

So the same input produces **both** failure modes the brief names: zero output *and* material ceiling violation. Three consecutive turns reproduce identically — "continue" never escapes, because a fresh `TurnCost` trips the same way.

**Why the earlier refutation didn't cover this.** The self-review killed "gate degenerates when peak > half the ceiling" with *"0 of 4,049 turns over 30d have peak > 625k"*. That is correct **for the 1.25M default**. The sibling PRs make the ceiling user-settable down to 100,000, which moves the degeneracy threshold from peak > 625,000 to peak > 50,000 — a different input class that the production data cannot speak to, because the knob does not exist in production yet.

**Zero coverage, and the floor is decorative.** `tests/test_spend_ceiling.py` uses only `CEILING = 1_000_000` and `ceiling=0`. The candidate fix leaves the suite at **11 passed, unchanged**. Removing the `max(..., 1)` floor entirely *also* leaves 11 passed with byte-identical behaviour — the gate is never consulted at `total_tokens == 0`.

**The shape was spotted and the mitigation mis-sized.** `test_agent_budget_settings.py:198` rejects `"50000"` with `# below ge: would stop nearly every turn immediately`. The floor chosen, 100,000, is below the 200,000 reserve — so it is itself 100% degenerate.

**Fix — both parts, the first verified insufficient alone:**

1. `session.py:2232` → `reserve = min(max(_SPEND_CEILING_RESERVE, 2 * peak), self._max_turn_tokens // 2)`. Guarantees `gate >= ceiling // 2`, keeps the invariant inside anton for the unbounded CLI path, leaves the 1.25M default bit-identical, suite stays at 11 passed.
2. Raise the floor in the three places that must stay in lockstep (`user_settings.py:524`, `app_settings.py:601`, `settingsTransform.js:603`). The sweep shows 400,000 still yields 0-4 rounds; **~600,000 is the lowest defensible floor** — the first value where a 190k-context turn survives its first call.
3. Add a test asserting `dispatch_tool.call_count > 0` at the minimum allowed ceiling, and correct the docstring at `:2228-2229`, which asserts the safety property the floor does not provide.

---

## 2 — MEDIUM · nothing verifies `max_turn_tokens` ever reaches `ChatSession`

`session.py:690` — introduced (the *fail-silent* variant is; the siblings at `:684-686` use direct attribute access and fail loud).

`getattr(s, "max_turn_tokens", 0)` is deliberately fail-open: absent -> 0 -> `_spend_ceiling_reached()` returns False on its first branch. No exception, no log, no `ended_by`.

**Executed.** Mutating the key to `"max_turn_tokenz"`: **11 passed**, full suite **2,012 passed / 28 skipped** — byte-identical to baseline. `grep -rn "_max_turn_tokens" tests/` returns exactly one line, the post-construction poke at `test_spend_ceiling.py:100`. The cross-repo assertion that would catch the sibling default-drift mutation (`test_agent_budget_settings.py:165`) is **currently skipped** ("pinned anton predates ENG-1286") and never constructs a `ChatSession`.

Feature verified ON today: a bare `ChatSession` prints `_max_turn_tokens = 1250000`.

**Fix.** One assertion: `ChatSession(ChatSessionConfig(...))._max_turn_tokens == CoreSettings().max_turn_tokens == 1_250_000`. Kills both mutations.

## 3 — MEDIUM · two of the three ceiling call sites have zero coverage; both delete clean

`session.py:4396` (continuation gate) and `:2747` (non-streaming `turn()`) — introduced. Four reports merged.

The per-round check lives inside `while llm_response.tool_calls:` (`:3547`), so a planning call returning a **text** reply exits the loop without that call's tokens being checked. No shipped test constructs that shape — all 11 trip at tool_round 1 on the per-round check.

**Executed.** Both sites replaced with `if False:` -> full suite **2,012 passed, 28 skipped**, identical to baseline. Instrumenting `:4396` directly: the predicate is evaluated **16 times across all 2,040 tests and is `False` every time** — the branch body never executes. Both sites are live, not dead: a text-reply continuation with an INCOMPLETE verdict gives `ended_by=spend_ceiling` clean and `ended_by=completed, continuations=1` mutated.

**Two corrections worth having:**

- **The comment defending the gate at `:4391-4395` is factually wrong.** It says "the round counter resets to 0, so without this gate a turn can spend the ceiling several times over ... each pass looks cheap to the per-round check". `TurnCost` is constructed once per turn (`:2660`, `:3087`) and `add()` accumulates (`turn_cost.py:141-149`), so the per-round check sees the full running total — executed: with `:4396` deleted, `ended_by=spend_ceiling, tokens_total=1,200,000`. The gate's **real** value is the no-tool continuation, where the turn otherwise ends `ended_by=completed` with no hand-back notice at all. As written, a future reader could delete this gate on the strength of disproving its stated rationale.
- The non-streaming `turn()` site has **no in-tree caller** — `grep -rn "\.turn("` across anton, cowork-server, cowork_evals and anton_services returns only tests. Public-API tidiness, not risk.

**Fix.** Two tests on the existing `_session` helper. (a) One tool round, then a text reply sized so the gate is crossed only after the loop exits, with an INCOMPLETE verdict; assert `ended_by == "spend_ceiling"`. Do **not** assert `continuations == 0` alone — it also passes in the tool-using shape. (b) Mirror `test_turn_cost_terminals.py:293`; assert through the patched `send_event` kwargs, since `session._turn_cost` is None after `turn()` returns.

## 4 — MEDIUM · the named placement guard doesn't guard it, and one PR-body mutation claim doesn't reproduce

`tests/test_spend_ceiling.py:143` — introduced.

With `PER_CALL = 500_000`, `_usage()` splits into quarters so `context_tokens = 375,000`, `reserve = max(200k, 750k) = 750,000`, gate = 250,000 — **the first LLM call breaches it**. Every ceiling test in the file therefore trips at tool_round 1 after one call, making `continuations == 0` and `rounds < 25` structurally unfalsifiable.

**Executed.** The body claims *"disable the per-round gate -> the no-continuation placement test fails"*. Mutating `:3602` to `if False:` and running that test alone -> **1 passed**. Whole file -> 2 failed, and neither is the placement test. Under the mutation the turn spent `tokens_total=7,000,000` against a 1,000,000 fixture ceiling with all three assertions holding. **The other three mutation claims in the body do reproduce.**

The module comment at line 31 ("enough that two calls clear (CEILING − reserve) and one does not") is arithmetically correct only for the *flat* 200k reserve this PR replaced — the fixture was never re-sized.

The placement is not unguarded: `test_total_is_bounded_by_the_ceiling` catches it incidentally. The defect is that the named guard isn't the one working, two of its three assertions are inert, and one documented verification claim is false.

**Fix.** Size the fixture so the trip needs several rounds and assert `rounds > 1`, or capture the per-round site's distinct log marker. Correct the line-31 comment.

## 5 — LOW · "bounded by the ceiling" is approximate, and the test pinning it sits at margin zero

`session.py:2232` + `tests/test_spend_ceiling.py:130` — introduced.

The reserve is sized from `peak_context_tokens` (`input + cache_read + cache_creation`, output excluded) but compared against `total_tokens`, which includes output. Residual is exactly `2 x output − gate slack`.

At the shipped 1.25M ceiling with output pinned to `max_tokens = 8192`: **2 of 15 realistic shapes overshoot, worst +67,344 (+5.4%)** — two verifiers derived that independently. Pinned to the token: at c0=150k, last passing check 949,152 vs gate 950,000, overshoot 15,536 = 2x8192 − 848. The shipped test passes only because its fixture lands on exactly 1,000,000 (margin 0); parametrising `per_call` over [500k, 170k, 150k, 130k] gives 3 failures. The assertion is not vacuous — loosening the reserve does redden it — it is knife-edge.

No user-visible number is contradicted (the notice quotes actual spend). But the test name, its docstring, and `_spend_ceiling_gate`'s docstring ("precisely the bound the ceiling exists to make true") all state it absolutely.

**Fix.** Assert `<= CEILING + 2 * session._turn_max_tokens()` and drop "precisely" from the docstring.

## 6 — LOW · the ENG-1155 double-append guard is load-bearing and untested

`session.py:3623` — introduced.

`_reply_persisted = True` is the **sole** writer on this path: the verification block's own assignment at `:4123` is unreachable because `:4114` breaks on `_spend_ceiling_hit` first — the verifier-skip feature is precisely what makes this guard load-bearing. Deleting the line survives the whole repo (2,012 passed). Probe with an empty scripted diagnosis: guard present -> reply once; guard deleted -> index [3] is a second `assistant` copy.

**Fix.** One variant of `test_handback_asks_the_user_and_is_persisted_as_streamed` with `content=""`; mutation-verify by deleting `:3623`.

## 7 — LOW · the log line reports the constant as `reserve`, not the reserve applied

`session.py:3610` — introduced. At the default with 200k/call context: `reserve=200000` when the reserve actually applied was 400,000 and the gate 850,000. Within one line, `ceiling=` is the applied value and `reserve=` is a floor. Nothing parses these logs and `peak_context_tokens` rides `turn_completed`, so the signal is recoverable — misleading, not lost. Log `gate=self._spend_ceiling_gate()` instead.

---

## Deviation to declare

ENG-1286's *How* says the resume injection should carry the existing "Do not repeat work already done" wording, and *Done when* has "resumes without redoing completed work". The **budget** half is implemented and correct (executed: turn 2 books 2,000 tokens on a fresh `TurnCost`). The **resume** half has no mechanism — `grep -n "repeat work"` returns one hit at `:4434`, inside the verifier-INCOMPLETE branch, unreachable from the ceiling exit.

In fairness: the same ticket bullet says "reuse the round cap's ask-permission shape, do not invent a second one", and the round cap plus three other hand-back paths have no resume injection either. The body already declares one deviation, so the convention is being followed — this bullet just got missed. Declare it, or set a flag beside `ended_by` and inject the existing wording once at the top of the next `turn_stream`.

## Verdict on the bound

Real at the default, soft by roughly two output budgets. Genuine on the per-round path (mutating the gate reddens 3 tests); unverified on the continuation gate and non-streaming `turn()`; and the wiring carrying the value from settings into the session is itself untested. **Below the default, in the 100k-400k band the sibling PRs open up, enforcement inverts** — the turn stops before any tool runs and can exceed the ceiling by up to 300% while producing nothing.

## Process warning

All three shared worktrees were mutated concurrently by parallel sessions during this review — `pr-344` carried foreign `MUT1`/`MUT2` markers that appeared and vanished mid-run. Every result above was re-derived in private copies. Two traps for whoever re-runs it: a `cp -R` copy's `.venv/bin/pytest` execs an **absolute** interpreter path back into the original tree, so mutations silently appear to survive — use `uv run python -m pytest` and prove the loaded module path with `inspect.getsource`; and re-confirm each mutation is still applied *after* a long suite run, not only before.

alecantu7:
## Review findings addressed — `1c6e8ea`

Thanks, this was a good catch list. Independently reproduced the HIGH before fixing it, and it's worse than reported at the top of the band: at 190k contexts the zero-tools band reaches **400,000**, not 200,001.

| ceiling | tools (before) | tools (after) |
|---|---|---|
| 100,000 | **0** — 400k spent, +300% | 1 |
| 250,000 | **0** | 1 |
| 400,000 | **0** | 1 |
| 750,000 (new floor) | 1 | 2 — 760k, +1% |
| 1,250,000 (default) | 5 | 5 — unchanged |

### The fix is two parts, because the proposed one is insufficient alone

Capping `reserve` at `ceiling // 2` keeps the gate proportional, but at a 100k ceiling the first 190k call still crosses it with nothing dispatched. So the loop **also refuses to trip before at least one tool round has run**, keyed on `TurnCost.rounds` (turn-wide, so a continuation can't hand out a fresh free round).

That makes *"a ceiling stop never ends a turn that did nothing"* structural rather than a property of whatever floor the settings enforce — which matters because anton's own `max_turn_tokens` has no lower bound for CLI and host callers. The parametrised test covers down to a ceiling of **1**.

Floor raised to **750,000** in all three repos (not the suggested 600,000 — measured, 600k buys a single tool call; 750k gives several and sits just above the p75 external turn at 736k).

### On the refutation that failed

Worth stating plainly: I killed this exact finding in my own self-review with *"0 of 4,049 turns have peak context above 625k."* That was true — **for the 1.25M default only**. The sibling PRs then moved the degeneracy threshold from peak >625k to peak >50k and I never revisited the refutation. Evidence I invalidated myself, and didn't notice.

### Everything else

- **Finding 2** — wiring assertion added; the `max_turn_tokenz` typo mutation now fails.
- **Finding 3** — both uncovered sites now covered (`if False:` on either fails its test). The continuation-gate comment is rewritten: you were right that `TurnCost` accumulates, so the stated rationale was false. Real reason recorded — a text-reply continuation never enters the tool loop.
- **Finding 4** — confirmed, and the PR body has been corrected. Fixture re-sized (`PER_CALL` 500k → 100k) so a trip needs ~8 rounds; disabling the per-round gate now fails the placement test, which it did not before.
- **Finding 5** — assertion loosened to `CEILING + 2 * output`, docstring no longer claims "precisely".
- **Finding 6** — covered, deleting `_reply_persisted = True` now fails.
- **Finding 7** — logs `gate=` (applied) instead of the reserve constant.
- **Deviation** — accepted as stated; the resume-injection bullet is not implemented, matching the round cap and the other three hand-back paths.

**Every new test mutation-verified** — 7 mutations, each caught by its intended test. Full suite **2,019 passed**, same 2 pre-existing `test_datasource.py` failures.

### One correction to the sibling report

The pre-existing `_build_chat_session` blocker on cowork-server (`skills_extra_roots`) was real but is **already fixed** — andrew reverted it in `6175c3b` today. It only appeared live because my branch was based on stale staging; rebased now. The other half, `unlock_skill`, was never present (0 occurrences at my base and on staging).

### Process note, confirmed the hard way

My first probe today was invalid: another session had switched the shared anton clone to `eng-1495`, so I measured a tree without the feature and got the opposite result. Everything above was re-derived in a private worktree at the PR head with the module path proven via `inspect.getsource`.

alecantu7:
## Second pass — head `1c6e8ea4` (was `2b676da1` at the deep review)

**The high finding is closed, and the fix is stronger than the one I proposed.**

### The zero-tools failure is gone — re-measured with my own sweep, not the new tests

Driving the real `turn_stream` and counting `tool_registry.dispatch_tool.call_count`, 14 cells:

```
                                    before    after
ceiling=  100,000  per_call=200,000 |   0    →   1
ceiling=  200,000  per_call=200,000 |   0    →   1
ceiling=  400,000  @190k context     |   0    →   1
ceiling=  750,000  (the new floor)   |   -    →  13 rounds @40k/call, 2 @200k
ceiling=        1  (absurd)          |   0    →   1
ceiling=1,250,000  (default)         |  17    →  25 @40k, 4 @200k
```

**Zero `tools_run=0` anywhere**, including `ceiling=1` — a value the settings floor prevents, so the invariant now holds even for the CLI and host callers that `max_turn_tokens` has no bounds on.

### Two mechanisms, not one

I suggested clamping the reserve to `ceiling // 2`. What landed is that **plus** `_spend_ceiling_stops_the_tool_loop()` gating on `TurnCost.rounds > 1`, so "never end a turn that dispatched no tools" is guaranteed by construction rather than falling out of arithmetic. Both halves mutation-verified independently:

```
remove the reserve cap        → FAILED test_gate_stays_proportional_to_the_ceiling
remove the rounds>1 guarantee → 3 failed, incl. test_a_ceiling_stop_never_dispatches_zero_tools[1]
```

Using `TurnCost.rounds` rather than the loop-local counter is the right choice and the docstring says why — it accumulates across verifier-forced continuations, so a continuation can't hand out a fresh free round.

### The wrong docstring was corrected in place, not deleted

The old text claimed the gate *"cannot produce a turn that made no call and said nothing"*. The replacement says explicitly that this was true of LLM calls and **false of tool dispatches, which happen later in the same iteration**. Better than removing it — the next reader learns the trap instead of rediscovering it.

Same for the bound: `_spend_ceiling_gate`'s docstring now states the bound is *"close, not exact"* and names the mechanism (reserve derived from `peak_context_tokens`, compared against `total_tokens`, so a turn can land over by roughly two output budgets). That closes the finding about "precisely the bound the ceiling exists to make true".

### The other findings

Five of the seven now carry tests:

| finding | test |
| -- | -- |
| gate degenerates at low ceilings | `test_a_ceiling_stop_never_dispatches_zero_tools` (parametrized), `test_gate_stays_proportional_to_the_ceiling` |
| nothing verifies the setting reaches `ChatSession` | `test_the_setting_actually_reaches_the_session` |
| continuation gate uncovered | `test_ceiling_trips_at_the_continuation_gate` |
| non-streaming `turn()` uncovered | `test_ceiling_applies_to_the_non_streaming_turn` |
| ENG-1155 double-append guard untested | `test_empty_diagnosis_does_not_double_append_the_reply` |

Suite: **2,021 passed, 28 skipped** (was 2,009 before the new tests).

### Still open, neither blocking

- The per-round log line still reports `_SPEND_CEILING_RESERVE` rather than the reserve actually applied. At the default with 200k contexts it prints `reserve=200000` when the real reserve was 400,000 and the real gate 850,000 — within one line, `ceiling=` is the applied value and `reserve=` is a floor. `gate=self._spend_ceiling_gate()` is the number that decided the stop.
- The resume-injection deviation still wants declaring in the body: the ticket's *"Do not repeat work already done"* wording has no mechanism on the ceiling exit. The budget half resets correctly; only the injection is absent, and the round cap has the same gap, so this is a declared deviation rather than a defect.

### A note on my own method, since it nearly produced a false finding

Three of my mutation patterns failed to match multi-line constructs. Each time the `assert` caught it — but the first run printed **"20 passed"** before the assert fired, which reads exactly like *"the reserve cap has no coverage"*. A no-op mutation and a surviving one are indistinguishable from the exit code alone. Anyone re-running these should confirm the edit applied by grepping the changed line back, not by trusting the suite result.

alecantu7:
## Re: second pass — two of the open items were already closed at the head you reviewed

Both remaining anton items are addressed; one of them already was.

**The log line — already fixed in `1c6e8ea`, the head under review.** At `session.py:3651`:

```python
"spend ceiling reached: tokens=%d ceiling=%d gate=%d "
"tool_round=%d continuation=%d — pausing to ask the user",
self._turn_cost.total_tokens if self._turn_cost else 0,
self._max_turn_tokens, self._spend_ceiling_gate(),
```

It logs the applied gate, not the constant. Worth flagging as a stale read rather than letting it look like a live gap.

**The resume-injection deviation — now declared in the PR body**, as deviation (a) of two, with the reasoning: the budget half is implemented and correct, the resume half has no mechanism, and it is left absent because the same ticket bullet says to reuse the round cap's shape rather than invent a second one — the round cap and the other three hand-back paths have no resume injection either. Adding one here alone would make the ceiling the odd stop out.

## On the `skills_extra_roots` item, carried on cowork-server#301

Checked again this pass against both refs rather than the working tree:

```
git show origin/alejandrocantu/eng-1286-max-tokens-per-task:.../harness.py | grep -c 'skills_extra_roots\|unlock_skill'  -> 0
git show origin/staging:.../harness.py                                     | grep -c 'skills_extra_roots\|unlock_skill'  -> 0
```

Zero on both. andrew reverted it in `6175c3b` ("revert eng 764") and my branch is rebased onto `e4cbdb1`. `unlock_skill` was never present at all — 0 at my original base too. This one is closed by someone else's work, not open.

## And a thank-you for the method note

The warning about no-op mutations printing a clean pass is the most reusable thing in either pass — a mutation that fails to apply and a mutation that survives are indistinguishable from the exit code. I hit the same class earlier in this PR from the other direction: I read `1 failed` and *inferred* which test it was, which is how a false claim reached the PR body. Confirming the edit landed, and confirming *which* test moved, are two separate checks and both are load-bearing.

## Status

- **anton#344** — 8 findings closed, 2,021 passed.
- **cowork-server#301** — floor at 750k, overlay extracted and covered, both mutations now fail.
- **cowork#636** — floor at 750k; also corrected the *pre-existing* `422` in the comment above `BUDGET_FIELDS` (`404f8023`). Verified rather than assumed: the settings route catches `ValidationError` and raises `HTTPException(400)` at `settings.py:98,114,123,550`. There is no 422 path on it.

The one open item nobody has closed is the `SettingsView` render test — genuinely pre-existing debt, since the `maxContinuations` field and the `hasBudgetSettings` gate mutate clean too. Not taking it in this PR.

anton/core/session.py
```diff
@@ -2656,6 +2784,28 @@ async def turn(self, user_input: str | list[dict]) -> str:
                 response = await self.plan_with_recovery(system=system)
                 break
 
+            # Spend ceiling (ENG-1286), mirroring turn_stream. Same reasoning as
+            # the cap mark above: public API, and the books are wired here too.
+            # This path runs no completion verifier, so there is no continuation
+            # gate to also check — the per-round check is the whole gate here.
+            if self._spend_ceiling_reached():
```
pnewsam: **Minor:** this non-streaming `turn()` path gates on bare `_spend_ceiling_reached()`, while `turn_stream()` uses `_spend_ceiling_stops_the_tool_loop()` (the `rounds > 1` guard). So `turn()` lacks the "never end a turn that dispatched no tools" guarantee that *Done when* states holds "at any ceiling": at a small `max_turn_tokens` (CLI/host callers have no lower bound — the streaming test parametrizes down to `1` for exactly this), the first planning call can cross the gate and this breaks on round 1 having dispatched nothing, then still pays for the hand-back.

Blast radius is near-zero — `turn()` has no in-tree production caller and the 750k product floor keeps it safe in practice — and it is untested for `turn()`. Suggest calling `_spend_ceiling_stops_the_tool_loop()` here too, for parity with the invariant.

anton/core/session.py
```diff
@@ -2175,6 +2194,115 @@ def _emit_turn_cost(
...
+        Returns False when the ceiling is disabled (0) or the books are closed,
+        so a host on older settings, or any path that runs outside a turn, keeps
+        exactly its pre-ENG-1286 behaviour.
+
+        This is pure arithmetic and says nothing about whether the turn has done
+        any WORK yet. An earlier version of this docstring claimed it could only
+        fire after real work had happened, on the reasoning that every call site
+        sits inside the tool loop; that is true of LLM calls and false of tool
+        dispatches, which happen later in the same iteration. The per-round call
+        site carries the progress condition explicitly — see `_spend_ceiling_ok`.
```
pnewsam: **Nit:** dangling reference — `_spend_ceiling_ok` does not exist. The per-round call site is `_spend_ceiling_stops_the_tool_loop`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the per-turn spend-ceiling against ENG-1286: the gate, the adaptive reserve and its proportional cap, per-round + continuation-gate placement, verifier-skip after a trip, ENG-1155 persistence, and the tests. The core is sound and the ticket deviations are declared. Two non-blocking notes inline, neither gating.

### Validation
- Read the full diff and inspected both tool-loops (`turn_stream` and `turn`) plus the four `_spend_ceiling_*` helpers on the head commit.
- Confirmed `turn()` has no in-tree production caller — every production path (`chat.py`, `cloud_turn`, `goal.py`, `local_runtime.py`) uses `turn_stream`.
- Did not run the suite.

alecantu7:
## Both notes addressed — `0aba519`

**Minor (turn() parity) — fixed, and you were right that the invariant was stated more broadly than the code delivered.** `turn()` now calls `_spend_ceiling_stops_the_tool_loop()` like the streaming loop. Its test is parametrized down to a ceiling of `1` to match the streaming one; reverting the gate to the bare predicate fails both small-ceiling cases:

```
ceiling=100,000 ended turn() having dispatched no tools
ceiling=1       ended turn() having dispatched no tools
```

Your framing of the blast radius was the useful part: no in-tree caller and the 750k floor make it safe in practice, but **CLI and host callers have no lower bound on `max_turn_tokens` at all** — which is precisely the population the guarantee was written for. The floor is a product bound; it can't carry an invariant anton states unconditionally.

**Nit (dangling `_spend_ceiling_ok`) — fixed.** That name never existed; the call site is `_spend_ceiling_stops_the_tool_loop`.

One thing I left deliberately, worth stating so it doesn't read as an oversight: **the continuation gate keeps the bare predicate.** A text-only turn never enters the tool loop, so `rounds` stays 0 and the guard would stop that gate from ever firing — letting a turn continue *past* the ceiling, which is the worse failure. The guarantee belongs to the tool loop, not to every call site.

Suite: **2,021 passed**, same 2 pre-existing `test_datasource.py` failures.

---

Separately, thank you for **#636**. That was a real blocking defect and a good catch: `spec.unlimited` was a leftover from the 0-means-unlimited draft I built and then replaced, and `String(undefined)` is valid JS so nothing threw. Fixed, with the component test you asked for.

The root cause is worth recording because it's the trap you named in your own method note, from the other direction: the patch that replaced that block **carried no assertion that the pattern matched**, while every other replacement in the same script did. The whitespace differed, the edit silently no-opped, and a green suite is indistinguishable from an applied one. `SettingsView` having no coverage of that path is what let it reach you.

pnewsam:
Re-reviewed at `0aba5198` — both my inline notes are addressed. My approval stands; posting confirmation.

**Minor (turn() parity) — fixed and covered.** `turn()` now calls `_spend_ceiling_stops_the_tool_loop()` rather than the bare predicate, so the "never end a turn that dispatched no tools" guarantee holds on the non-streaming path too — which is the one that matters here, since CLI/host callers have no lower bound on `max_turn_tokens` and the 750k product floor can't carry an invariant anton states unconditionally. `test_ceiling_applies_to_the_non_streaming_turn` is parametrized down to a ceiling of `1`.

I mutation-verified it rather than trusting the green suite: reverting that one call site back to `_spend_ceiling_reached()` reddens `[100000]` and `[1]` with `ceiling=… ended turn() having dispatched no tools` — so the test is load-bearing, not incidental. `test_spend_ceiling.py`: **22 passed**.

Agreed with leaving the **continuation gate** on the bare predicate — a text-only turn never enters the tool loop, so `rounds` stays 0 and the `rounds > 1` guard would wrongly stop that gate from ever firing, letting a turn run past the ceiling. The guarantee belongs to the tool loop.

**Nit (dangling `_spend_ceiling_ok`) — fixed.** The docstring now points at `_spend_ceiling_stops_the_tool_loop`; the bad name is gone from the tree.

PR #346, opened 2026-08-13
Fix memory compaction: use indexes instead of strings

ea-rus:
Fix memory compaction fail error:
```
memory-compaction: forced _CompactionResult call failed (TRUNCATED, budget=4096, output_tokens=4096) retrying=True
```
Use line indexes instead of strings in output, should reduce chance of token budget overflow

Fixes https://linear.app/mindsdb/issue/ENG-1523/memory-compaction-is-permanently-broken-for-any-file-large-enough-to


alecantu7 (commented):
Reviewed at `ca46582d`, base `origin/staging` `3f74f06e` (verified fresh against the API).

Filed the missing ticket for this while reviewing: **[ENG-1523](https://linear.app/mindsdb/issue/ENG-1523)**. It carries the measurements below and the two findings, so the PR has something to close.

## This fixes the reported failure, and the arithmetic checks out

Measured against the actual file that produced the error on the reporting machine, `~/.cowork/memory/rules.md` (78 entries):

| | before | after |
| -- | -- | -- |
| output tokens the model must emit | **5,272** | **~234** |
| vs rung 1 (4,096) | **exceeded** -> `TRUNCATED` | fits, ~3,800 to spare |
| vs rung 2 (8,192) | 5,272 fits, but +~2,920 narration -> **exceeded** | fits |

**23x reduction.** That is the root cause, not a symptom: `kept: list[str]` forced the model to echo every surviving entry verbatim including its `<!-- ... -->` comment, so the response scaled with the file — and the file that most needs compacting is the one whose response cannot fit. The code comment on staging already said so (*"Compaction echoes the kept entries back, so its output scales with the file"*); this is the change that stops it being true.

Worth recording explicitly because it justifies a decision in the diff: **rung 2 failing proves narration is material on its own** (~2,920 tokens on this file). So keeping the ladder rather than deleting it is correct, and the comment saying so is accurate.

Three things this gets right beyond the headline:

- `kept = [e for i, e in enumerate(entries, 1) if i in keep]` copies from the source list, never from the response. An invented index costs at most the entry it failed to name; duplicates collapse into the set. **The model can drop an entry but can never reword or corrupt one** — for a file the agent reads as behavioural rules that is a safety property, not just a side effect.
- `if not kept: return` — an empty answer leaves the file alone rather than emptying it.
- The select-only tradeoff is called out honestly in the docstring. It loses no information (the prompt says keep both when in doubt); it only forgoes tidying.

**Tests:** 8 new cases, precise, and mostly load-bearing under mutation:

```
model order instead of file order  -> 2 failed  (file-order guarantee is genuinely pinned)
0-indexed prompt vs 1-indexed pick -> 1 failed  (off-by-one is pinned)
empty answer -> keep everything    -> 44 passed (see finding 2)
```

Full suite: **2,009 passed, 28 skipped**.

---

## Finding 1 — medium · this turns compaction back on, and compaction duplicates entries

`anton/core/memory/cortex.py:643-650` — **pre-existing**, but dormant only because compaction has been dead since 2026-08-04.

The three section filters overlap:

```python
always = [e for e in kept if "always" in e.lower() or not any(k in e.lower() for k in ("never","when","if "))]
never  = [e for e in kept if "never" in e.lower()]
when_rules = [e for e in kept if "when" in e.lower() or "if " in e.lower()]
```

An entry containing both "when" and "always" lands in two lists, is written twice, and is read back next pass as two entries.

**Executed against the real file:**

```
entries in           : 78     (always=30  never=3  when=51)
entries written out  : 84
DUPLICATED           : 6      (8% growth per pass)

  x2: "- When reading email bodies... always encode/decode with errors=..."
  x2: "- When debugging a secret that looks malformed, never print it..."

compounding: 78 -> 84 -> 90 -> 96
```

The file has **0 exact duplicates today** — precisely because compaction has not run. The first successful compaction after this merges introduces 6.

**Refutation attempted.** Net effect is still strongly positive: the same file has 30 pairs with >50% word overlap, so compaction should prune far more than the 6 it re-adds. That is why this is medium and not a blocker. But it is a leak in the exact mechanism being repaired, on a ticket about unbounded growth, and it is one line.

**Fix.** Make the buckets mutually exclusive — e.g. `when_rules = [e for e in kept if e not in never and (...)]` — or dedupe `lines` before writing. Worth folding into this PR rather than a follow-up, since the whole point is that the file stops growing.

## Finding 2 — low · `test_empty_answer_leaves_the_file_untouched` cannot detect its own behaviour

Replacing `if not kept: return` with `kept = entries` leaves the suite at **44 passed**, mutation confirmed applied.

The test asserts `hc._lessons_path.read_text() == before`, and for a *lessons* file the full rewrite is byte-identical — so "returned early" and "rewrote everything" are indistinguishable to it.

It matters because the rewrite is **not** idempotent for `rules.md`; that is finding 1. The guard is genuinely load-bearing on the rules path, and the test that names it is blind there.

**Fix.** Assert on a write spy, or run the same case with `kind="rules"`.

## Finding 3 — low · nothing exercises the rules path

All eight new tests use `_lessons_path`. The only `rules` fixture in the file (`test_cortex.py:99`) holds a single entry. So the section bucketing — the one non-trivial branch in the rebuild, and the branch this PR simplified — has no coverage from the new tests.

---

## What this does not fix, stated so the ticket does not close early

- **The ratchet is still one-way.** Any *other* compaction failure still leaves the file untouched, so it still grows and still fails again. This moves the threshold ~23x further out; it does not remove the mechanism. Recorded as out-of-scope on ENG-1523.
- **The truncation log still cannot tell you which cause you hit.** `raise_unusable_tool_call` computes *"did not return a tool call"* (narrated past the budget) versus *"returned an unusable tool call"* (budget ran out inside the JSON), but `generate_with_truncation_retry` logs only `TRUNCATED`, deliberately excluding the exception message for content-safety. That ambiguity is exactly what made this take a round-trip to diagnose — you could not tell from the log whether to shrink the output or buy more room. Four lines to carry the boolean rather than the message; it is on ENG-1523's Done-when.

---

**Recommended: APPROVE**, with finding 1 worth folding in first. Without it, the first thing the restored compaction does is duplicate six rules — a small regression, but on the one file this PR exists to stop growing.

alecantu7 (commented):
## Re-review — head `d2fb6675`, base `origin/staging` `741802f` (verified fresh against the API)

**All three findings from my `ca46582d` pass are closed**, and two of them are closed more thoroughly than I asked. Re-measured rather than re-read.

### Finding 1 — duplication: fixed by construction

Entries now carry the heading they were read under, so each is written back to exactly one section. Re-ran the same experiment against the same real file, `~/.cowork/memory/rules.md` (on a copy), with a model that keeps every index:

| | `ca46582d` | `d2fb6675` |
| -- | -- | -- |
| entries in | 78 | 78 |
| entries written out | **84** | **78** |
| duplicated | **6** | **0** |
| second pass | 90 — compounding | 78 — **idempotent** |

`get_rules()` after compaction returns `always 48 / never 3 / when 27` — byte-for-byte the same distribution the original file parses as, so the round trip preserves `kind` and not just text. `_RULE_SECTIONS` matches `save_rules`'s writer at `hippocampus.py:487` exactly, same three names, same order.

### Finding 2 — the blind test: fixed on both paths

Applied the exact mutation from that finding (`if not kept: return` → keep everything):

```
FAILED test_empty_answer_leaves_the_file_untouched        <- the lessons test that was blind
FAILED test_empty_answer_does_not_rewrite_the_rules_file  <- the one I asked for
```

The lessons test can now see its own behaviour, which I had not expected to get for free.

### Finding 3 — no rules coverage: fixed

Seven new tests exercise `kind="rules"`: original-heading preservation, idempotency, re-parse as Engrams, emptied sections keeping their heading, entries above the first heading, entries under an unknown heading.

Three further mutations, each caught by its own tests and nothing else:

```
delete the `if not kept: return` guard        -> 2 failed
bucket by keyword instead of by heading       -> 5 failed
collapse the verdict back to one TRUNCATED    -> 2 failed
```

The truncation-diagnosis gap I listed as out-of-scope was pulled in too. I grepped `cowork-server`, `cowork`, `cowork_evals`, `anton_services` and `mdb-ai` for consumers of the old `TRUNCATED` / `NO_TOOL_CALL` log strings — there are none, so the three-way split breaks no downstream reader.

---

## Four new items, none blocking

### 1 — nothing logs how much compaction deleted, and indices made over-pruning free · medium

`_compact_file` has three `logger` calls and none on the success path. A compaction that keeps 18 of 78 rules rewrites the file and says nothing. `_encode_with_lock` doesn't log either, and there's no wrapper log in `compact_all` — checked all three.

This is pre-existing, but the mechanism that kept it harmless is what this PR removes. With `kept: list[str]` the model had to retype every survivor, so dropping 60 entries cost it output tokens and showed up in the response size. With `keep: list[int]`, dropping 60 entries is a *shorter* answer than keeping them. The economic pressure against silent over-pruning is gone, and nothing replaced it.

One line, and it matches the principle already stated at `cortex.py:647` — *"never silently: an unlogged swallow made a dead memory subsystem indistinguishable from a working one"*:

```python
logger.info("memory-compaction: kept %d of %d entries", len(kept), len(entries))
```

Worth folding into this PR rather than a follow-up, for the same reason as the section fix: the whole point is that this file stops changing size without anyone knowing.

### 2 — an entry above the first heading becomes a binding rule · low

Deliberate and tested (`test_entries_above_the_first_heading_are_not_lost`), and I agree keeping them beats dropping them. But "not lost" has a second half worth stating, because `get_rules` requires `current_kind` (`hippocampus.py:443`) and therefore **ignores** those entries today:

```
BEFORE compaction: orphan visible to the agent?  False    (8 rules)
AFTER  compaction: ('always', 'NOTE to self: ignore the stagi…')   (9 rules)
```

So a hand-written note in a file users do edit goes from inert to a rule the agent follows. Right behaviour, but it should be recorded on the ticket rather than found later.

### 3 — `anton/README.md:834` still advertises merging · low

> `compact_all()` | LLM-assisted deduplication **+ merge** on all oversized files.

This PR removes merging outright — `merged` and `pruned` are deleted and the new docstring says compaction selects rather than rewrites. Same class of defect as the docstring that let ENG-1495 ship dead for 16 days, in the file the repo treats as its deep dive.

### 4 — the size regression test never runs the rules path · low

`test_the_answer_fits_the_budget_at_the_size_that_broke_it` builds its 78 entries with `encode_lesson`. ENG-1523 asks for both a real-size regression test and a `kind="rules"` test; both exist, but as different tests, so the section logic is never exercised at scale. The section code is size-independent so this is completeness, not risk — and I covered it by hand against the real 78-entry file above.

---

## One finding I withdrew, because the invariant is worth knowing

I had this as a defect and reproduced it before checking the callers:

```
exception message : LLM returned an unusable tool call …
truncated=False   reached_tool_call=True   ->  LOG VERDICT: NO_TOOL_CALL
```

It doesn't survive. Every real caller makes that state unreachable — `client.py:309` and `scratchpad_boot.py:579` fire only when `not response.tool_calls`, and `client.py:319` and `scratchpad_boot.py:589` fire only inside `if looks_truncated(...)`. The only way I could produce it was by calling `raise_unusable_tool_call` directly.

Recording it because the invariant lives in four call sites rather than in `truncation_verdict` itself, so a fifth added without the `looks_truncated` guard would log a contradiction silently. Optional one-liner, not a defect here.

---

**Recommended: APPROVE**, with item 1 worth folding in first. It is the difference between "compaction works again" and "compaction ran, and we would not know if it ate half the file."

Two notes unrelated to the diff: `staging` moved to `741802f` under this PR (#350, #351) and neither commit touches anything here — I checked that #350's `cloud_turn/__main__.py` changes leave `main()` and the exit path alone. And this merges cleanly against anton#345, verified with a real merge.

alecantu7 (commented):
## Re-review of everything after `d2fb6675` — validated at `e064c83a`

No blocking findings. The incremental commit is sound, and the implementation held up under a mutation battery: **9 of 9 mutations were caught**, including the exact one ENG-1523 called out as previously undetected. Full suite green locally on this head (2195 passed, 28 skipped, 3m09s), on a base verified fresh against `origin/staging` = `6be5d34f`.

| Mutation | Caught by |
| --- | --- |
| empty-answer guard → keep everything | `test_empty_answer_leaves_the_file_untouched` |
| log reports total instead of survivors | `test_a_rewrite_reports_how_much_it_deleted` |
| log carries entry text | same |
| rules buckets made overlapping again | `test_keeping_everything_is_idempotent` |
| unknown heading no longer falls back | `test_entries_under_an_unknown_heading_are_not_lost` |
| `reached_tool_call` hardcoded False | `test_payload_overflow_is_logged_as_inside_call` |
| off-by-one in the numbering **shown** | `test_numbering_shown_matches_the_numbering_applied` |
| off-by-one in the numbering **applied** | 11 tests |
| `kind == "rules"` inverted | 5 tests |

`_with_hand_written_note` is the right fix for the weak `test_empty_answer_leaves_the_file_untouched` — replacing `if not kept: return` with `kept = list(zip(sections, entries))` now fails, where before the byte-identical lessons rewrite left the suite green. Both tests added in `fab8d99d` are load-bearing.

Three things for the record, none of them requesting a code change here.

---

### 1. Land #342 (ENG-695) first — it closes the one residual hole in this PR's premise

A truncation that the repair pass can *fix* is still accepted as a complete answer, and with an index schema that means compaction deletes every entry the model hadn't yet named. Executed against this head:

```
raw = '{"keep": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1'
safe_parse_tool_input -> parse_error = None       # repair balanced the bracket
unwrap_structured_response -> keep=[1..12, 1]     # validates: 13 of an intended 78
looks_truncated(response, 4096) -> True           # evidence exists; the success path never asks
```

`client.py:311` only consults `looks_truncated` inside the `except` branch, so a repaired dict that validates never reaches it. `_compact_file` then writes 13 of 78 entries through `_encode_with_lock(mode="write")` — atomic temp+rename, no `.bak` — and the only trace is one INFO line reading `kept 13 of 78`.

**This is pre-existing, not introduced.** I ran the same probe against the old `kept: list[str]` schema: it also returned `parse_error=None`, validated, and then wrote the truncated model string `"- rule two <!-- ts"` into the file verbatim — strictly worse. What changed is the *shape*: a bracket-balanceable int array is the most repairable payload there is, so the residual truncation now lands in the silent branch instead of the loud one the bug report showed. Overall risk still drops, because 234 output tokens vs 5,272 makes truncation ~23× less likely to begin with.

The fix is already written — #342 adds `ToolCall.repaired` and `if not response.tool_calls or any(tc.repaired for tc in response.tool_calls)` → raise, so this becomes a `StructuredOutputError`, the ladder retries, and a final failure leaves the file alone. So: **merge order, not new code.**

Worth flagging to whoever reviews #342: its body describes the memory consequence as "recorded no lessons" — a silently empty result. That was accurate before this PR. After this one lands, the same path means *deleted the entries it hadn't named*, unrecoverably. Same fix covers both, but the severity in that description understates the post-#346 case.

One coverage note that explains why this wasn't spotted: `_truncated_mid_json` in `tests/test_memory_truncation_retry.py` models only the `parse_error`-was-set variant (step 3 of the salvage cascade). The repair-succeeded variant (step 2) — the silent one — has no fixture.

### 2. This PR conflicts with #342 in `session.py`

```
git merge-tree --write-tree --name-only <pr-346> <pr-342>
CONFLICT (content): Merge conflict in anton/core/session.py
```

Both rewrite the verifier truncation log at `session.py:4582`. It resolves trivially either way, but per convention 3 the merge order belongs in both PR bodies rather than in someone's head. #342 first gets you finding 1 for free and leaves this PR the smaller rebase.

Checked the other overlapping PR too: **#354** (`fix(memory): block persistent prompt injection`) also touches `cortex.py` and `test_cortex.py` but merges **clean** against this and never goes near `_compact_file` — it filters unsafe engrams at read time, complementary to this PR removing the write-side sink.

### 3. One `Done when` item met in spirit, not in wording

> A memory file with 78+ entries compacts successfully at the default budget, **verified against a real file rather than a small fixture**.

`test_the_answer_fits_the_budget_at_the_size_that_broke_it` uses a *synthetic* 78-entry file (`"detail " * 40`), and `len(llm.prompt) > 20_000` correctly guards the fixture from shrinking. But the budget assertion measures the **fake's** serialized answer at ~4 chars/token: it proves the schema no longer scales with the file, which is the important half — it does not prove a real model's narration *plus* answer fits rung 1, and rung 2's failure in the original report shows narration is material on its own. Post-fix verification against the reporter's actual `rules.md` isn't evidenced in the PR. Flagging so it can be accepted deliberately rather than read as covered.

---

### Not findings, for the record

- **This PR closes a persistent prompt-injection sink.** The old `kept: list[str]` wrote model-authored text into `rules.md`, which the agent reads as binding behavioural gates — an injected instruction could persist itself across sessions. Index selection removes it entirely; entries are copied from disk (`test_survivors_are_byte_identical_including_metadata`). The new INFO log is counts + `path.name` only, and the exact-message assertion means entry text can't start riding along.
- **Over-pruning has no floor and no recovery.** Even with a complete answer, `keep=[1]` on 78 entries is valid, deletes 77 user rules, and the INFO line is the only signal. The code comment names the incentive honestly. A `.bak` before the rewrite would make it recoverable. Not this PR's job — noting it since the log is observability-only.
- **Read-modify-write race, pre-existing and unchanged:** `_compact_file` reads, awaits the LLM for seconds, then overwrites, while `maybe_vacuum` runs it as a background task and `encode_rule` → `save_rules` is itself a full-file rewrite. `_encode_with_lock` locks each write but not the pair, so a `/memorize` in that window is silently discarded. Own ticket if it matters.

Verified `kind` is passed as `"rules"` by both `compact_all` and `vacuum`; grepped the whole workspace for external consumers of the changed API (`StructuredOutputError`, `truncation_verdict`, `_CompactionResult`) — none outside anton, and both log sites use the shared helper.

alecantu7 (approved):
Approving at `aa42c943`.

**#342 is confirmed in the tree, which is the precondition for this being safe.** Its merge commit `be259c47` is an ancestor of this head, and the code itself is present — `any(tc.repaired …)` in `client.py`, `repaired: bool = False` on `ToolCall`, `usable_tool_call` in `structured.py`. Verified as code in the tree rather than a commit in the history, because that distinction is the whole point here.

**Why that matters, and why it should stay in the record:** this PR makes the compaction answer a bracket-balanceable int array, which is the most repairable payload there is. Without #342's guard, a truncated-but-repaired answer parses, validates, and silently deletes every entry the model had not yet named. #342 landing first is what makes this a narrowing rather than a new silent-loss path — so if it were ever reverted, this PR's blast radius changes with it.

The `fixture` commit answers my last round's coverage note and goes past it. I had only observed that `_truncated_mid_json` models the loud variant (`parse_error` set) while the silent one had no fixture. What landed adds the silent twin, a ladder-level guard, and — which I had not asked for — the consequence at the call site: ten lessons, the model names two, assert the file is byte-unchanged and the failure logged. The `_repaired_mid_json` docstring also names the vacuous-pass trap itself ("`payload` must be VALID for the schema under test, or the ladder raises on validation instead"), and the payloads are in fact schema-valid, so the tests can only pass because the guard exists.

Both are load-bearing, mutation-verified:

```
MUTATION: if not response.tool_calls or any(tc.repaired …)  →  if not response.tool_calls:
FAILED test_a_repaired_payload_is_a_truncation_not_an_answer
FAILED test_a_repaired_compaction_answer_does_not_delete_the_rest
2 failed, 14 passed
```

Everything else from my earlier rounds is closed and re-verified on this head: the `session.py` conflict resolved keeping both sides, `merge-tree` against fresh `staging` clean, full suite **2215 passed / 28 skipped**, and all eight checks green including `run-tests` and `verdict-eval`.

**One deliberate deviation, recorded rather than silently accepted.** ENG-1523's `Done when` says the 78-entry case should be *"verified against a real file rather than a small fixture"*. `tests/test_cortex.py:667` is still a synthetic `range(78)`. The intent behind that wording — don't test at the 8-entry minimum, test at the size that actually broke it — is met, and `len(llm.prompt) > 20_000` guards the fixture from shrinking. Approving on that basis; flagging it so QA accepts it deliberately rather than filing a false miss.

The read-modify-write race I raised earlier stays out of scope: pre-existing, unchanged here, and worth its own ticket rather than this PR paying for it.

Merge is yours.

PR #347, opened 2026-08-13
[12484] scratchpad uv discovery

tino097:
Fixes #https://github.com/mindsdb/mindshub/issues/12484

- _find_uv() now also checks Homebrew MacPorts and Linuxbrew
- venv.create()'s stdlib fallback now passes symlinks=(not win32) 
- A failed venv verification now reports the real cause 
- Tests added

### DIFF
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -338,21 +338,28 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
 
         uv = self._find_uv()
         if uv:
-            _sp.run(
-                [
-                    uv,
-                    "venv",
-                    self._venv_dir,
-                    "--python",
-                    sys.executable,
-                    "--system-site-packages",
-                    "--seed",
-                    "--quiet",
-                ],
-                check=True,
-                capture_output=True,
-                timeout=30,
-            )
+            try:
+                _sp.run(
+                    [
+                        uv,
+                        "venv",
+                        self._venv_dir,
+                        "--python",
+                        sys.executable,
+                        "--system-site-packages",
+                        "--seed",
+                        "--quiet",
+                    ],
+                    check=True,
+                    capture_output=True,
+                    timeout=30,
+                )
+            except (_sp.CalledProcessError, _sp.TimeoutExpired) as exc:
+                # CalledProcessError/TimeoutExpired.__str__ omits the captured
+                # stderr, so uv's actual reason (bad --python, disk full) was
+                # never seen — only "returned non-zero exit status N".
+                stderr = (exc.stderr or b"").decode("utf-8", errors="replace").strip()
+                raise RuntimeError(f"uv venv failed: {stderr}" if stderr else str(exc)) from exc
         else:
             # symlinks=False is venv.create()'s own default on every platform
             # (only the `python -m venv` CLI defaults it per-OS); a copied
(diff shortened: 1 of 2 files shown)


anton/core/backends/local.py
```diff
@@ -396,8 +408,15 @@ def _verify_venv_python(self) -> bool:
...
                 timeout=5,
             )
-            return result.returncode == 0 and "ok" in result.stdout.decode("utf-8", errors="replace")
-        except Exception:
+            ok = result.returncode == 0 and "ok" in result.stdout.decode("utf-8", errors="replace")
+            if ok:
+                self._last_verify_error = None
+            else:
+                stderr = result.stderr.decode("utf-8", errors="replace").strip()
+                self._last_verify_error = f"exit {result.returncode}" + (f": {stderr}" if stderr else "")
```
pnewsam, line 425: **Nit:** the two early returns above (`_venv_python is None` at :399 and `not os.path.exists(...)` at :401) don't update `_last_verify_error`, unlike this branch. If a retry ever reaches the `raise RuntimeError(...{detail})` in `_ensure_venv` via one of those early returns, the message can carry a *previous* iteration's reason. The trigger — create succeeds yet yields no interpreter, after a prior iteration already set an error — is essentially unreachable, so this is purely cosmetic. Set the field in those two branches too if you want it airtight.

pnewsam (commented):
## Code Review

**Verdict: COMMENT**

Reviewed the full diff against `anton/core/backends/local.py` and the new `tests/test_local_venv_provisioning.py`, plus the surrounding `_ensure_venv`/`_create_venv`/`_verify_venv_python` context. Three tightly-scoped fixes: widen `_find_uv()` fallbacks, force `symlinks=True` on POSIX in the stdlib venv path, and surface the real verify failure. Each change is correct, guarded per-platform, and directly tested. One cosmetic nit inline; nothing merge-blocking.

### Validation

- Confirmed `sys` and `venv` are imported, so `symlinks=sys.platform != "win32"` and the win32 branch resolve. The value is `True` on macOS/Linux (matches the `python -m venv` CLI's own POSIX default, fixing the copied-binary `@rpath` crash) and `False` on Windows (avoids the symlink-elevation requirement).
- Traced `_last_verify_error`: `_verify_venv_python()` sets it on both the success and the failure/exception branches, so the detail appended in `_ensure_venv` is fresh in the normal failure path.
- CI green: `run-tests` passed, CodeQL/Snyk clean.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Re-reviewed at head `ccad7d0`, focusing on the 3 commits added since my prior COMMENT review (`3cc136d`): a `cleanup`, a `staging` merge, and `clear last_verify_error`. The only substantive delta is that `_verify_venv_python()` now sets `self._last_verify_error = None` on its two early-return branches (`_venv_python is None` and interpreter missing), with 3 new tests pinning that behavior. The three original fixes remain correct and unchanged. No actionable findings.

### Validation

- `_verify_venv_python()` now assigns `_last_verify_error` on every return path (both early returns, the ok/not-ok split, and the exception handler). So the `detail` appended in `_ensure_venv`'s failure message can never carry a previous attempt's error into an attempt that failed for an unrelated reason — closing the early-return gap left open before.
- Detail wiring confirmed sound: in the retry loop `_verify_venv_python()` runs immediately before `RuntimeError(...{detail})`, so `_last_verify_error` always reflects that same call.
- New tests are accurate: the two `test_verify_clears_a_stale_error_*` cases cover exactly the branches the commit touched; the fake-python exit-code tests exercise the stderr capture path.
- CI green at head: `run-tests`, `Analyze (python)`, CodeQL, and Snyk all pass.

Strictly an improvement over the version last reviewed.

PR #345, opened 2026-08-13
ENG-1495: send turn_completed and rule_retrieval straight to PostHog instead of through the collector

alecantu7:
Closes ENG-1495. Related: ENG-1288 (shipped the counter), ENG-1355 (the collector fix, independent of this), ENG-1286 (downstream), ENG-1459 (desktop-vs-web split).

## The problem

ENG-1288 shipped per-turn cost accounting. The `turn_cost` log line works. **The PostHog sink has produced nothing, ever.** Measured 2026-08-10 in project 355390:

```
events matching %turn% or %cost%, all time  →  0
control — all events, any type               →  12 types, ~27k events
```

The control is what makes that evidence rather than an absence: the project is healthy and anton is definitely reaching it. The collector filters twice — actions must begin `anton_`/`ds_connect`, and only `action`, `aid`, `engine`, `llm_provider`, `has_mdb_key` survive — and it returns **HTTP 200 while dropping**, so `send_event`'s fire-and-forget path never learns.

This blocks **Phase 0 criteria 3 and 6**, and ENG-1286's spend ceiling was deliberately deferred so it could be derived from real distribution rather than guessed.

## What this does

Routes **two** events direct to PostHog **MindsHub main (424726)** — `turn_completed` and `rule_retrieval`. `_POSTHOG_EVENTS` is the register; anything absent keeps the collector path, so the ~105k/30d `ds_connect_*` volume stays exactly where it is.

`rule_retrieval` (ENG-1390) shipped without the `anton_` prefix, so the collector dropped it on the action filter just like `turn_completed` — and its eight properties (`outcome`, `when_rules`, `kept_rules`, `rules_chars`, `stop_reason`, `input_tokens`, `output_tokens`, `duration_ms`) would have been cut to nothing by the property allowlist even if the name had passed. Adding it is one register entry, because the first commit already built the path.

Note it does not by itself close **Phase 0 criterion 5**, which asks for the call's frequency. That arrives one release later — ENG-1390's instrumentation and this PR's working sink ship in the same train, so there is no production data yet by construction, not by fault.

[...]

PR #348, opened 2026-08-13
ENG-1492: classify tool failures into a comparable root-cause key

alecantu7:
Measurement only — **nothing trips, nothing hands back, the existing per-tool error streak is untouched.** This is the half of ENG-1286's thrash breaker that has to come first, on the same reasoning that made its spend ceiling defensible: [ENG-1288](https://linear.app/mindsdb/issue/ENG-1288) counted, then the ceiling picked a number from data instead of from one incident.

Ticket: [ENG-1492](https://linear.app/mindsdb/issue/ENG-1492). The control that consumes this is [ENG-1531](https://linear.app/mindsdb/issue/ENG-1531), deliberately unbuilt.

## Why now

[ENG-1276](https://linear.app/mindsdb/issue/ENG-1276) built `ToolOutcome.reason` and **nothing ever read it** — both dispatch sites took `.content` and `.ok` and dropped it on the floor. The registry docstring has been saying this ticket's name for two months.

And the thresholds ENG-1531 needs cannot be measured from Langfuse: the only place a turn's failures are visible is the verifier's input, which keeps the **last 12 tool events at a 400-char cap** with `[error]` appended *last* — so any failure on a cell that printed output is truncated away. Counts read from there are lower bounds with a bias nobody can correct for.

## The tiers are the safety argument, not the thresholds

| tier | examples | can trip? |
|---|---|---|
| `self_inflicted` | NameError, SyntaxError, AttributeError, UnicodeEncodeError, PyCompileError | **never** |
| `transient` | 429/5xx, connection resets, timeouts, watchdog kills | **never** (v1) |
| `external_wall` | missing_dependency, permission_denied, auth_missing, connection_refused… | yes |

That is **probe H encoded structurally**. ENG-1286's original argument was that self-repair escapes because the error changes between attempts — probabilistic, and it fails exactly when an agent retries the same broken line. It matters here specifically because the 2026-08-12 harvest found the heaviest turns **dominated** by self-inflicted errors (UnicodeEncodeError 11, AttributeError 8, NameError 8, SyntaxError 8), so a breaker on guessed thresholds would land squarely on them.

`transient` does the same job for probe G, and closes a real gap: [ENG-673](https://linear.app/mindsdb/issue/ENG-673)'s `classify_transient` lives in the LLM clients and is **unreachable from a tool outcome**. Tiering by exception type buys transient exclusion without building tool-level transient typing first.

## Two-level key, because the motivating case needs it

`pyodbc` → `apt` → `.deb` → `pymssql` are **four identifiers and one class**. An exact-signature match sees four unrelated problems and never fires — which is what happened. `test_the_kiranam_ping_pong_collapses_to_one_class` pins both rungs: the class rung counts 4, the exact rung stays below it.

Status codes are read **before** the exception type: an `HTTPError` is transient or a wall depending on its status, which its class name cannot tell you.

## Occurrences, never a streak

There is no reset API at all, and its absence is the design — asserted by a test. The existing per-tool streak resets on success, and interleaved false successes (a 404 HTML page that parsed as a valid archive) are precisely why it never counted to five through ENG-836. `test_interleaved_successes_do_not_reset_the_count` drives the real loop with `WALL, OK, WALL, OK, WALL` and expects 3.

## Writer inventory, done twice

Convention 9. **All 14 handler sentinels in the tree are mapped**, and ambiguous ones (`artifact_not_found`, `unknown_datasource`, `launch_failed`) resolve **away from tripping** — a false wall interrupts a working agent, a missed one is caught later by ENG-1286's ceiling, and those costs are not symmetric.

`test_every_sentinel_reason_is_mapped` greps the tree and fails on any new unmapped sentinel, so the inventory maintains itself rather than going stale the first time someone adds a handler. Verified by deleting a mapping — it fails with the missing name.

`root_cause_reason_coverage` reports the live share in production, because a key computed over a third of failures would set thresholds wrong in the safe-looking direction.

## Where it lands

On the existing `turn_completed` event **and** appended to the `turn_cost` log line.

**Corrected after review discussion — the destination is [ENG-1495](https://linear.app/mindsdb/issue/ENG-1495), not [ENG-1355](https://linear.app/mindsdb/issue/ENG-1355).** The `zoomInfoCollector` lambda is being retired, so waiting on its allowlist fix was the wrong framing. anton#345 routes `turn_completed` **straight to the PostHog Capture API** into **MindsHub main (424726)**, and `_posthog_body` passes every `extra` kwarg through as a property — so these **eleven** fields arrive with no further work once it lands (ten, plus `root_cause_classify_errors` added in `1b86a2b`). `_posthog_body` is a pass-through minus three keys (`_` cache-buster, `action`, `timestamp`) with no allowlist, so a new property needs nothing beyond being passed as a kwarg.

Verified rather than assumed, since ENG-1495 carries a correction about exactly this: **this PR and anton#345 merge cleanly** (real merge in a throwaway worktree, not `merge-tree` eyeballing — it reports "changed in both" for `session.py` because both touch it, which is not a conflict).

Until #345 lands, `turn_completed` is dropped at the collector's *action-name* filter — probed live: `action=turn_completed` → `posthogResult: false`, `action=anton_turn_completed` → `true`. So **absence in PostHog is not evidence the classification is broken**, and the deliverable distribution is readable from the `turn_cost` log line regardless of merge order.

Reuses `_normalise_error_signature` as a **pure function**, deliberately not the ACC subsystem it lives in: `_acc_observe` returns early when `ANTON_ACC_MODE == "off"`, and a signal that vanishes when a memory feature flag flips is convention 9's state-gated trap.

## Testing

53 new tests across `tests/test_root_cause.py` (pure logic) and `tests/test_root_cause_wiring.py` (the real `turn_stream`/`turn`, since "the value exists and nobody reads it" is the defect being fixed).

**Mutation-verified** — each applied, each caught by its intended test:
- drop `.reason` at the dispatch sites (the original defect) → 5 failed
- promote `self_inflicted` to trip-eligible → 16 failed
- reset the ledger on success (the streak bug) → 1 failed
- remove the class rung → 4 failed
- delete a sentinel mapping → the inventory test names it

**No behaviour change, asserted:** `test_classification_never_changes_the_turn` runs the same turn with classification on and off and compares `ended_by`, `rounds` and the history roles. `test_a_raising_classifier_cannot_break_a_turn` patches the classifier to raise and expects `ended_by == "completed"`.

Full suite: **2,074 passed** (2,021 before), with the 2 pre-existing `test_datasource.py` failures that reproduce on clean `origin/staging`.

## Security

Reviewed per convention 8. Classification reads error *text* that already flows through the tool loop and emits only counts, tier names and a class name — never identifiers from conversation content beyond a module/path/host already present in an error the model can see. No new endpoint, credential path or deserialization. The recorder is guarded whole so a reporting failure can never cost a user their turn.

## What happens next

This ships and does nothing visible. After ≥1 week of traffic the distribution of `root_cause_max_exact` / `root_cause_max_class` goes on the ticket — and it may well say the wall-repeat population is too small to justify ENG-1531 at all, which is a valid and useful outcome.

### DIFF
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -304,10 +304,28 @@ def classify(reason: str, result_text: str = "") -> RootCause:
     # it is visible as noise in the ENG-1492 distribution rather than hiding a
     # wall — and this is measurement-only until ENG-1531 reads any of it.
     #
+    # SECOND, OPPOSITE INTERACTION — this branch also runs BEFORE the
+    # `_WALL_TYPES` lookup below, so an ENUMERATED wall type whose path or host
+    # happens to contain a bare 3-digit 4xx/5xx token is demoted to transient
+    # and never reaches `root_cause_wall` or any trip rung. Measured:
+    #
+    #     FileNotFoundError … '/data/500/report.csv'  -> transient http_500
+    #     ConnectionRefusedError … db.internal:503    -> transient http_503
+    #     …'/data/report.csv'                         -> external_wall  (control)
+    #     …db.internal:5432                           -> external_wall  (control,
+    #                                                    4 digits, `\b` blocks it)
+    #
+    # So the status branch errs BOTH ways: it inflates walls from non-enumerated
+    # types (above) and deflates walls from enumerated ones (here). Both are
+    # tolerable while nothing consumes the tier, and the motivating dependency
+    # walls carry no such tokens (`pyodbc`/`pymssql` are unaffected).
+    #
     # For ENG-1531: do not trip on `_STATUS_WALLS` classes without either an
     # HTTP-shape precondition on the reason or a check that the exception type
-    # was enumerated. Frequency here is UNMEASURED; the combinatorial floor is
-    # 3/900 of uniformly distributed 3-digit tokens.
+    # was enumerated — that single precondition fixes both directions at once.
+    # And when sizing thresholds from the ENG-1492 distribution, know that walls
+    # with numeric ports or paths read LOW. Frequency here is UNMEASURED; the
+    # combinatorial floor is 3/900 of uniformly distributed 3-digit tokens.
     status = _STATUS_RE.search(reason)
     if status:
         code = status.group(1)
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -400,6 +400,27 @@ class RootCauseLedger:
     tiers: Counter = field(default_factory=Counter)
     failures: int = 0
     with_reason: int = 0
+    #: Times the recorder's guard swallowed an exception instead of booking a
+    #: failure. Emitted so a BROKEN instrument is distinguishable from a QUIET
+    #: one — without it the two are byte-identical:
+    #:
+    #:     3 wall failures, classifier raising -> failures=0 … wall=0
+    #:     3 successes, genuinely clean turn   -> failures=0 … wall=0
+    #:
+    #: That ambiguity matters because "the wall-repeat population is too small
+    #: to justify the control" is one of ENG-1492's own sanctioned conclusions,
+    #: so a silent instrument failure reads as a legitimate answer and would
+    #: cancel ENG-1531 on the strength of a bug. Same defect `reason_coverage`
+    #: was built to fix, one level down: a metric must not exclude its own
+    #: failures from its own denominator.
+    #:
+    #: Deliberately NOT counted in `failures` — that would corrupt the very
+    #: distribution this exists to protect. It is a separate signal.
+    classify_errors: int = 0
+
+    def note_error(self) -> None:
+        """Record that the guard caught something. Must never raise."""
+        self.classify_errors += 1
 
     def add(self, rc: RootCause) -> None:
         self.failures += 1
@@ -453,6 +474,7 @@ class RootCauseLedger:
         """
         return {
             "root_cause_failures": self.failures,
+            "root_cause_classify_errors": self.classify_errors,
             "root_cause_distinct": len(self.exact),
             "root_cause_max_exact": self.max_exact,
             "root_cause_max_class": self.max_class,
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -455,6 +455,19 @@ class RootCauseLedger:
         The live form of the writer inventory: a key computed over a third of
         failures would set thresholds wrong in the safe-looking direction, and
         this is the number that says whether that is happening.
+
+        **Read this together with `classify_errors`.** A classification that
+        RAISED reaches neither counter, so it leaves this ratio entirely — on
+        partial failure the number reads high:
+
+            3 classified (2 with reason) + 2 raised
+              -> reported 0.667, true 2/5 = 0.4
+
+        The systematic case is safe rather than flattering (if every
+        classification raises, `failures` is 0 and this reads 0.0, not 1.0), and
+        `classify_errors` is non-zero in both cases — so a dashboard should
+        treat any `classify_errors > 0` as "this ratio is an over-estimate"
+        rather than trusting it at face value.
         """
         return (self.with_reason / self.failures) if self.failures else 0.0
 
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -53,6 +53,22 @@ TIER_UNCLASSIFIED = "unclassified"
 #: ignored — see the module docstring for why that is structural, not a default.
 TRIP_ELIGIBLE_TIERS = frozenset({TIER_WALL})
 
+#: Wall classes that count on the EXACT rung only, never the coarse one.
+#:
+#: `missing_file` is genuinely ambiguous — a wrong path the agent invented is
+#: self-inflicted, a missing binary is a wall — and the two are indistinguishable
+#: from the error alone. The exact rung tells them apart by construction: the
+#: same path three times is a wall, three different paths is an agent exploring.
+#: Counting it on the coarse rung merges those, and worse, drowns real walls:
+#: measured, five file probes plus the four-attempt ENG-836 dependency wall in
+#: one turn emitted `max_class=5, top_class="missing_file"` — the wall this
+#: whole ticket exists to see, masked by exploration.
+#:
+#: The cost, stated because ENG-1492 already accepted it: a genuine wall spread
+#: over several paths (one missing ODBC install surfacing as three absent files)
+#: scores 1 on both rungs and is invisible.
+_EXACT_ONLY_CLASSES = frozenset({"missing_file"})
+
 #: The agent's own bugs. It can fix these by writing better code, so repetition
 #: is iteration, not a wall — however many times it repeats.
 _SELF_INFLICTED = frozenset({
@@ -242,9 +258,30 @@ def classify(reason: str, result_text: str = "") -> RootCause:
 
     exc = _exception_name(reason)
 
-    # Status codes are checked BEFORE the exception type: an HTTPError carrying
-    # a 429 is transient no matter what its class name suggests, and one
-    # carrying a 403 is a wall. The type alone cannot tell those apart.
+    # The explicit type tables win over the status heuristic, and the ORDER here
+    # is the safety property — not a style choice.
+    #
+    # The status check used to run first, on the reasoning that an HTTPError
+    # carrying 429 is transient whatever its class name says. True, but it
+    # searched `\b(4\d\d|5\d\d)\b` against the WHOLE reason with no HTTP-shape
+    # precondition, so any self-inflicted exception whose message happened to
+    # contain a bare 401/403/404 was promoted to `external_wall`, trip-eligible.
+    # Measured on real raised exceptions:
+    #
+    #     {200:..,404:..}[403]        -> KeyError: 403          -> wall, trip=True
+    #     assert 404 == 200           -> AssertionError         -> wall, trip=True
+    #     SyntaxError … (line 404)    -> wall, trip=True
+    #     …the same error at line 405 -> self_inflicted, trip=False
+    #
+    # That is exactly the "agent retries the same broken line" case this tier
+    # exists to exclude, so the guarantee in the module docstring was false for
+    # 8 specific integers. `HTTPError` is in neither table, so it still falls
+    # through to the status branch — which keeps the legitimate case working.
+    if exc in _SELF_INFLICTED:
+        return RootCause(TIER_SELF, exc, "")
+    if exc in _TRANSIENT:
+        return RootCause(TIER_TRANSIENT, exc, "")
+
     status = _STATUS_RE.search(reason)
     if status:
         code = status.group(1)
@@ -253,11 +290,6 @@ def classify(reason: str, result_text: str = "") -> RootCause:
         wall_cls = _STATUS_WALLS.get(code)
         if wall_cls:
             return RootCause(TIER_WALL, wall_cls, _identifier_for(wall_cls, reason))
-
-    if exc in _SELF_INFLICTED:
-        return RootCause(TIER_SELF, exc, "")
-    if exc in _TRANSIENT:
-        return RootCause(TIER_TRANSIENT, exc, "")
 
     wall_cls = _WALL_TYPES.get(exc)
     if wall_cls:
@@ -336,7 +368,8 @@ class RootCauseLedger:
             # trip" actually happens.
             return
         self.exact[rc.key] += 1
-        self.classes[rc.cls] += 1
+        if rc.cls not in _EXACT_ONLY_CLASSES:
+            self.classes[rc.cls] += 1
 
     @property
     def max_exact(self) -> int:
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -137,6 +137,21 @@ _STATUS_TRANSIENT = frozenset({"429", "500", "502", "503", "504"})
 
 _EXC_LINE_RE = re.compile(r"\b([A-Z][A-Za-z0-9_]*(?:Error|Exception|Warning))\b")
 
+#: Hard cap on what any regex here is allowed to see.
+#:
+#: `_HOSTPORT_RE` is O(n^2) on input containing no colon — `[\w.-]+` matches
+#: greedily from every start position and then fails. Measured: **1,051 ms** on
+#: 20,000 characters. Nothing reaches it that long today (the scratchpad handler
+#: caps its reason at 160 chars and every other producer emits a short sentinel
+#: or an exception name), but that guarantee lives in another file and would be
+#: silently lost the first time a handler passes a long reason. Capping here
+#: makes it local, and covers every pattern in this module rather than the one
+#: that happens to be quadratic.
+#:
+#: 500 is well past the longest real reason: a traceback's last line arrives
+#: pre-truncated to 160.
+_MAX_REASON_CHARS = 500
+
 
 @dataclass(frozen=True)
 class RootCause:
@@ -214,9 +229,10 @@ def classify(reason: str, result_text: str = "") -> RootCause:
     recorded as `from_reason=False` and lands in `unclassified`, which is not
     trip-eligible.
     """
-    reason = (reason or "").strip()
+    # Bounded before any regex sees it — see `_MAX_REASON_CHARS`.
+    reason = (reason or "").strip()[:_MAX_REASON_CHARS]
     if not reason:
-        sig = _normalise_error_signature(result_text or "")
+        sig = _normalise_error_signature((result_text or "")[:_MAX_REASON_CHARS])
         return RootCause(TIER_UNCLASSIFIED, "unclassified", sig[:80], from_reason=False)
 
     sentinel = _SENTINEL_REASONS.get(reason)
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -279,9 +279,28 @@ class RootCauseLedger:
     valid archive) are precisely why it never counted to five through the
     ENG-836 runaway.
 
-    Session-scoped because the control that will consume this fires at most once
-    per root cause per session, so the counts the thresholds are read off must be
-    the session's, not one turn's.
+    **Lifetime is one ChatSession, and that means different things per host —
+    check before reading the numbers as "per conversation".**
+
+    | host | ChatSession built | ledger effectively |
+    |------|-------------------|--------------------|
+    | CLI (`chat.py`) | once, then loops `turn_stream` | spans the conversation |
+    | Cowork (`anton_harness`) | inside `stream_response()`, i.e. per HTTP turn | **per turn** |
+
+    So on the primary product this resets every turn. That was NOT the original
+    intent — the design note said session-scoped, on the reasoning that
+    ENG-1531 fires once per root cause per session — and it is recorded here
+    because the difference is invisible from this file.
+
+    What it still measures correctly: a wall hit repeatedly **within one turn**,
+    which is the ENG-836 shape (all four pyodbc/apt/.deb/pymssql attempts were
+    in a single turn of 29 calls). What it misses: a wall that persists across
+    turns because the user said "try again", where each turn starts from zero.
+
+    Consequence for ENG-1531: its "at most once per root cause per session"
+    cannot be built on this object alone — it needs cowork-server to hold the
+    ledger per conversation, or to accept per-turn scope deliberately. Sizing
+    thresholds from this data is still valid; claiming session coverage is not.
     """
 
     exact: Counter = field(default_factory=Counter)
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -291,6 +291,23 @@ def classify(reason: str, result_text: str = "") -> RootCause:
     if exc in _TRANSIENT:
         return RootCause(TIER_TRANSIENT, exc, "")
 
+    # RESIDUAL, by construction — the guarantee above is scoped to types IN the
+    # table, and this branch is what a non-enumerated type falls through to. An
+    # agent bug raising a type nobody enumerated (`RuntimeError`, `LookupError`,
+    # a bare `Exception`) whose message happens to carry 401/403/404 still mints
+    # a trip-eligible `external_wall` — `RuntimeError: record 404 not found` is
+    # the realistic shape.
+    #
+    # Left as-is deliberately: widening `_SELF_INFLICTED` to catch it would have
+    # to include `Exception`, which would swallow every genuine wall raised as a
+    # generic type. The direction is the safe one — it INFLATES wall counts, so
+    # it is visible as noise in the ENG-1492 distribution rather than hiding a
+    # wall — and this is measurement-only until ENG-1531 reads any of it.
+    #
+    # For ENG-1531: do not trip on `_STATUS_WALLS` classes without either an
+    # HTTP-shape precondition on the reason or a check that the exception type
+    # was enumerated. Frequency here is UNMEASURED; the combinatorial floor is
+    # 3/900 of uniformly distributed 3-digit tokens.
     status = _STATUS_RE.search(reason)
     if status:
         code = status.group(1)
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -363,14 +363,14 @@ class RootCauseLedger:
     def event_fields(self) -> dict:
         """Flat properties for the turn's analytics event and log line."""
         return {
-            "rc_failures": self.failures,
-            "rc_distinct": len(self.exact),
-            "rc_max_exact": self.max_exact,
-            "rc_max_class": self.max_class,
-            "rc_top_class": self.top_class,
-            "rc_reason_coverage": round(self.reason_coverage, 3),
-            "rc_self_inflicted": self.tiers.get(TIER_SELF, 0),
-            "rc_transient": self.tiers.get(TIER_TRANSIENT, 0),
-            "rc_wall": self.tiers.get(TIER_WALL, 0),
-            "rc_unclassified": self.tiers.get(TIER_UNCLASSIFIED, 0),
+            "root_cause_failures": self.failures,
+            "root_cause_distinct": len(self.exact),
+            "root_cause_max_exact": self.max_exact,
+            "root_cause_max_class": self.max_class,
+            "root_cause_top_class": self.top_class,
+            "root_cause_reason_coverage": round(self.reason_coverage, 3),
+            "root_cause_self_inflicted": self.tiers.get(TIER_SELF, 0),
+            "root_cause_transient": self.tiers.get(TIER_TRANSIENT, 0),
+            "root_cause_wall": self.tiers.get(TIER_WALL, 0),
+            "root_cause_unclassified": self.tiers.get(TIER_UNCLASSIFIED, 0),
         }
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -140,7 +140,16 @@ _SENTINEL_REASONS = {
 # matters here — the opposite error merely groups two walls together, which is
 # visible in the data rather than silent.
 _MODULE_RE = re.compile(r"No module named ['\"]([\w.]+)['\"]")
-_IMPORT_NAME_RE = re.compile(r"cannot import name ['\"](\w+)['\"]")
+# Captures the MODULE, not the symbol. Keying on the symbol split one broken
+# package into as many keys as symbols the agent tried to import from it —
+# three failures on `pandas` scoring max_exact=1, distinct=3 — and made
+# `cannot import name 'read_excel' from 'pandas'` a different wall from
+# `No module named 'pandas'`. The `.split(".")[0]` at the use site is a no-op on
+# a symbol, which is the evidence the module was always intended.
+_IMPORT_NAME_RE = re.compile(
+    r"cannot import name ['\"]\w+['\"] from "
+    r"(?:partially initialized module )?['\"]([\w.]+)['\"]"
+)
 _PATH_RE = re.compile(r"['\"]([^'\"]{1,120})['\"]")
 _WINERR_PATH_RE = re.compile(r"\[(?:Errno|WinError) \d+\][^:]*: ['\"]?([^'\"]{1,120})")
 _HOSTPORT_RE = re.compile(r"([\w.-]+):(\d{2,5})")
@@ -394,7 +403,19 @@ class RootCauseLedger:
         return (self.with_reason / self.failures) if self.failures else 0.0
 
     def event_fields(self) -> dict:
-        """Flat properties for the turn's analytics event and log line."""
+        """Flat properties for the turn's analytics event and log line.
+
+        **Emitted per turn; the VALUES are cumulative for the ledger's lifetime**
+        (see the class docstring — one turn under Cowork, the whole conversation
+        under the CLI). The names read as per-turn and are not.
+
+        This matters for ENG-1492's actual deliverable, the distribution that
+        sets ENG-1286/ENG-1531's thresholds: on a host where the ledger spans
+        turns, a 10-turn session emits 10 events whose counts only climb, so a
+        naive per-EVENT percentile is dominated by long sessions and reads high.
+        Take the LAST event per session, or compute per-session maxima — never
+        a percentile over the raw event stream.
+        """
         return {
             "root_cause_failures": self.failures,
             "root_cause_distinct": len(self.exact),
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -1,0 +1,341 @@
+"""Root-cause classification for tool failures (ENG-1492).
+
+Turns ENG-1276's per-failure ``ToolOutcome.reason`` into a **comparable key**,
+so "is this the same wall as last time?" becomes a lookup instead of a guess.
+This module only measures. Nothing here trips, hands back, or touches the
+existing per-tool error streak — the control that consumes it is ENG-1531, and
+it is deliberately not built until this has reported real numbers.
+
+Why a key at all
+----------------
+The ENG-836 runaway ping-ponged ``pyodbc`` -> ``apt`` -> ``.deb`` -> ``pymssql``:
+four approaches, four different error texts, **one wall** (no ODBC driver, no
+root). Matching normalised error *text* would see four unrelated problems, which
+is why the key has two levels — a ``class`` that collapses all four, and an
+``identifier`` that keeps them apart when that matters.
+
+The tiers are the safety argument
+---------------------------------
+Not the thresholds. A breaker keyed on repetition is one bad classification away
+from interrupting an agent that is productively fixing its own bug, and the
+2026-08-12 harvest showed the heaviest turns are *dominated* by exactly that:
+UnicodeEncodeError, NameError, SyntaxError, AttributeError, PyCompileError. So
+``self_inflicted`` failures are excluded **structurally** — they are counted, and
+they can never become trip-eligible however often they repeat.
+
+ENG-1286's original argument was that self-repair escapes because the error
+changes on each attempt. That is probabilistic and fails exactly when an agent
+retries the same broken line. This does not.
+
+``transient`` is the same move for ENG-836's probe G (a 429 twice is not a wall),
+and it closes a real gap: ENG-673's ``classify_transient`` lives in the LLM
+clients and is unreachable from a tool outcome, so tiering by exception type buys
+transient exclusion without building tool-level transient typing first.
+"""
+
+from __future__ import annotations
+
+import re
+from collections import Counter
+from dataclasses import dataclass, field
+
+# Reused as a PURE FUNCTION. Deliberately not the ACC subsystem it lives in:
+# `_acc_observe` returns early when ANTON_ACC_MODE == "off", and a signal that
+# disappears when a memory feature flag flips is convention 9's state-gated trap.
+from anton.core.memory.acc import _normalise_error_signature
+
+TIER_SELF = "self_inflicted"
+TIER_TRANSIENT = "transient"
+TIER_WALL = "external_wall"
+TIER_UNCLASSIFIED = "unclassified"
+
+#: Only this tier may ever contribute to a trip. Everything else is measured and
+#: ignored — see the module docstring for why that is structural, not a default.
+TRIP_ELIGIBLE_TIERS = frozenset({TIER_WALL})
+
+#: The agent's own bugs. It can fix these by writing better code, so repetition
+#: is iteration, not a wall — however many times it repeats.
+_SELF_INFLICTED = frozenset({
+    "NameError", "SyntaxError", "IndentationError", "TabError", "TypeError",
+    "AttributeError", "KeyError", "IndexError", "ValueError", "UnboundLocalError",
+    "ZeroDivisionError", "PyCompileError", "UnicodeEncodeError",
+    "UnicodeDecodeError", "UnicodeError", "AssertionError", "StopIteration",
+    "RecursionError", "NotImplementedError",
+})
+
+#: Retryable by nature. Two of these is a provider blip, not a blocked task.
+_TRANSIENT = frozenset({
+    "ConnectionResetError", "ConnectionAbortedError", "BrokenPipeError",
+    "TimeoutError", "ReadTimeout", "ReadTimeoutError", "ConnectTimeout",
+    "SSLError", "ChunkedEncodingError", "IncompleteRead", "RemoteDisconnected",
+})
+
+#: Exception type -> wall class. The identifier is pulled separately below.
+_WALL_TYPES = {
+    "ModuleNotFoundError": "missing_dependency",
+    "ImportError": "missing_dependency",
+    "PermissionError": "permission_denied",
+    "FileNotFoundError": "missing_file",
+    "NotADirectoryError": "missing_file",
+    "IsADirectoryError": "missing_file",
+    "ConnectionRefusedError": "connection_refused",
+    "OperationalError": "db_unavailable",
+    "InterfaceError": "db_unavailable",
+    "DatabaseError": "db_unavailable",
+    "MemoryError": "resource_exhausted",
+    "OSError": "resource_exhausted",
+}
+
+#: Sentinel reasons handlers emit instead of an exception name. These are the
+#: handler's own verdict, so they are the most trustworthy input this has —
+#: every one is enumerated from the tree by `test_every_sentinel_reason_is_mapped`,
+#: which fails when a new one is added without a decision here.
+#:
+#: **Ambiguous sentinels are deliberately NOT trip-eligible.** A false wall
+#: interrupts a working agent; a missed wall only means the spend ceiling catches
+#: it later. The costs are not symmetric, so ambiguity resolves away from the
+#: tier that can trip.
+_SENTINEL_REASONS = {
+    # The agent passed bad or missing arguments — it can fix these itself.
+    "scratchpad_empty_code": (TIER_SELF, "empty_code"),
+    "scratchpad_missing_name": (TIER_SELF, "missing_name"),
+    "missing_name": (TIER_SELF, "missing_argument"),
+    "missing_slug": (TIER_SELF, "missing_argument"),
+    "missing_description": (TIER_SELF, "missing_argument"),
+    "invalid_type": (TIER_SELF, "invalid_argument"),
+    "invalid_port": (TIER_SELF, "invalid_argument"),
+    "invalid_health_timeout": (TIER_SELF, "invalid_argument"),
+    "invalid_datasources": (TIER_SELF, "invalid_argument"),
+    # The agent named something that does not exist. Ambiguous — it could be a
+    # genuinely absent resource — but the agent chose the identifier and can
+    # list the real ones, so it resolves to the non-tripping side.
+    "artifact_not_found": (TIER_SELF, "unknown_resource"),
+    "unknown_datasource": (TIER_SELF, "unknown_resource"),
+    # Genuine walls: the environment is missing something the agent cannot add.
+    "package_install_failed": (TIER_WALL, "missing_dependency"),
+    "store_unavailable": (TIER_WALL, "service_unavailable"),
+    # Could be environment or config and the sentinel does not say which, so it
+    # stays out of every trip rung until something distinguishes them.
+    "launch_failed": (TIER_UNCLASSIFIED, "unclassified"),
+}
+
+# Identifier extraction, per class. Kept narrow on purpose: a wrong identifier
+# splits one wall into several and hides it, which is the failure mode that
+# matters here — the opposite error merely groups two walls together, which is
+# visible in the data rather than silent.
+_MODULE_RE = re.compile(r"No module named ['\"]([\w.]+)['\"]")
+_IMPORT_NAME_RE = re.compile(r"cannot import name ['\"](\w+)['\"]")
+_PATH_RE = re.compile(r"['\"]([^'\"]{1,120})['\"]")
+_WINERR_PATH_RE = re.compile(r"\[(?:Errno|WinError) \d+\][^:]*: ['\"]?([^'\"]{1,120})")
+_HOSTPORT_RE = re.compile(r"([\w.-]+):(\d{2,5})")
+_STATUS_RE = re.compile(r"\b(4\d\d|5\d\d)\b")
+
+#: HTTP statuses that are walls rather than blips, mapped to their class.
+_STATUS_WALLS = {"401": "auth_missing", "403": "permission_denied", "404": "missing_file"}
+#: …and the ones that are always transient, whatever else the text says.
+_STATUS_TRANSIENT = frozenset({"429", "500", "502", "503", "504"})
+
+_EXC_LINE_RE = re.compile(r"\b([A-Z][A-Za-z0-9_]*(?:Error|Exception|Warning))\b")
+
+
+@dataclass(frozen=True)
+class RootCause:
+    """One failure, classified.
+
+    ``key`` is the exact rung (``class:identifier``); ``cls`` alone is the
+    coarse rung that collapses the ENG-836 ping-pong. ``trip_eligible`` is the
+    whole safety contract in one boolean — a consumer must never re-derive it
+    from the tier, or the exclusion stops being structural.
+    """
+
+    tier: str
+    cls: str
+    identifier: str
+    #: False when nothing usable was available and the key came from normalised
+    #: result text. Counted separately so the coverage share is measurable
+    #: rather than assumed (convention 9's writer-inventory, as a live metric).
+    from_reason: bool = True
+
+    @property
+    def key(self) -> str:
+        return f"{self.cls}:{self.identifier}" if self.identifier else self.cls
+
+    @property
+    def trip_eligible(self) -> bool:
+        return self.tier in TRIP_ELIGIBLE_TIERS
+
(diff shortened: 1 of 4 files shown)


alecantu7:
## Risk pass — found a 6.5-second stall, now fixed (`bc2f0f1`)

Asked to assess whether this PR is risky, so I probed the surface rather than restating that it's measurement-only. One real finding.

### `_HOSTPORT_RE` is quadratic, and a long reason could stall a turn

`([\w.-]+):(\d{2,5})` on colon-free input backtracks from every start position. Measured **end-to-end through `classify()`** with inputs that actually route to `_identifier_for` — an exception name mapping to a host-bearing wall class, then a long colon-free tail:

| input | before | after |
|---|---|---|
| `ConnectionRefusedError: ` + 50k | **6,565 ms** | 0.7 ms |
| `OperationalError: …` + 50k | **6,577 ms** | 0.6 ms |
| `HTTPError: 401 ` + 50k | **6,772 ms** | 0.6 ms |

**Not reachable today** — the scratchpad handler truncates its reason to 160 chars and every other producer emits a short sentinel or an exception name. But that guarantee lives in a different file, 200 lines from the regex, and disappears the first time a handler passes something long. `classify()` now bounds its own input, which makes the property local and covers all seven patterns rather than the one that happens to be quadratic.

### My first guard test for it was worthless

It used `"A" * 200_000`, which never gets past the exception-type lookup and so never reaches `_identifier_for` — it passed with the cap removed and proved nothing.

That is the **second time in this PR** a test of mine passed for the wrong reason (the first was `test_the_ledger_spans_turns_within_a_session`, which only held because it reused one `ChatSession`, the CLI shape rather than the deployed one). Same root cause both times: asserting the *shape* of a failure without constructing an input that reaches the code under test. Both now fail under mutation — this one with `classify took 6767 ms`.

### The rest of the risk surface, checked rather than asserted

- **Imports** — no cycle (`import anton.core.root_cause, anton.core.session` clean).
- **`tool_reason` bound before every read** — verified by AST walk, not by eye: assigned at 6 sites, read at 1, first assignment precedes it.
- **Throughput** — 0.8 µs per classification on realistic input; 12,000 in 9 ms.
- **Memory** — 5,000 distinct identifiers produce 4,995 in-memory entries and still exactly **10** emitted property names. No identifier is ever emitted, so PostHog's permanent property namespace is untouched.
- **Class vocabulary is closed** — fuzzed 20,000 random inputs, **zero** produced a class outside the 50-value set. Novel failures land in `unclassified`, which is not trip-eligible, so they are safe by default.

### Known gap, not fixed here

Multimodal tool results `continue` before the recorder, so their failures are not classified. Consistent with the existing note at that branch about the nudge/breaker text also being skipped there; worth a line in ENG-1531 rather than a change in a measurement-only PR.

Full suite: **2,075 passed**, same 2 pre-existing `test_datasource.py` failures that reproduce on clean `origin/staging`.

alecantu7 (commented):
## Adversarial self-review — head `3d86aa61`, base `origin/staging` `2aae3db2`

Deep pass. 12 distinct defects after dedup and refutation. Nothing here can break a turn — that isolation is verified by trace below — but three of the twelve mean the numbers this PR exists to emit are wrong in exactly the populations the ticket cares about.

---

## 1 — The status-code pre-check preempts the `self_inflicted` tier · blocking

`anton/core/root_cause.py:248`. `_STATUS_RE = re.compile(r"\b(4\d\d|5\d\d)\b")` (line 131) is searched against the **whole reason string** with no HTTP-shape precondition, and its branch `return`s before the `_SELF_INFLICTED` check at line 257. Any self-inflicted exception whose message contains a bare **401/403/404** becomes `external_wall`, `trip_eligible=True`; 429/500/502/503/504 make it `transient`. All 19 self-inflicted types are affected.

Executed with real raised exceptions through the verbatim producer expression:

```
{200:..,404:..}[403]          -> KeyError: 403           -> wall / permission_denied / trip=True
(b"x"*404+b"\xff").decode()   -> UnicodeDecodeError@404  -> wall / missing_file      / trip=True
assert 404 == 200             -> AssertionError          -> wall / missing_file      / trip=True
```

**One-byte control:** the identical decode error at position **405** → `self_inflicted`, trip=False.

**End-to-end:** five retries of one broken line through the real `turn_stream` emit `root_cause_max_exact=5, max_class=5, wall=5, self_inflicted=0` — the ENG-1286 trip shape, produced by the failure class this tier exists to exclude.

The trigger is realistic because `tool_handlers.py:539` sets `reason=error.splitlines()[-1][:160]` — the traceback's last line, which is exactly where a bare number lives.

**Coverage is zero, and green by luck.** Hoisting the check → 38/38 still pass, which proves the current ordering is pinned by nothing. But changing one existing fixture from `(detected at line 250)` to `line 404` turns `test_self_inflicted_failures_can_never_trip` **red**: `assert 'external_wall' == 'self_inflicted'`. Every fixture in `SELF_INFLICTED_REASONS` (`tests/test_root_cause.py:70-79`) happens to use out-of-range numbers.

**Fix (verified — 38/38 green, hijack table empty):** move

```python
if exc in _SELF_INFLICTED:
    return RootCause(TIER_SELF, exc, "")
```

above the status branch. `HTTPError` is not in `_SELF_INFLICTED`, so `test_a_status_code_beats_the_exception_name` stays green. Add `"SyntaxError: … (detected at line 404)"` to `SELF_INFLICTED_REASONS`.

Blocking not on blast radius — this is measurement-only — but because it falsifies the module docstring's central claim, which is the reason the tier exists at all.

*Scope correction, for accuracy:* three of my first six example strings were **refuted** as producer-unrealistic (py3.12 puts the line number in the discarded `File "…"` frame; `\b404\b` doesn't match `404abc`). The real surface is 8 exact integers, not "any 4xx/5xx".

## 2 — `missing_file` counts on the class rung, against the ticket's explicit spec · medium

`root_cause.py:339` — `self.classes[rc.cls] += 1` is unconditional. ENG-1492 §4, verbatim:

> It counts on the **exact** rung only, never the class rung: the same path missing three times is a wall, three different paths is the agent exploring.

**Executed:** 5 file probes + the 4-identifier ENG-836 pyodbc wall in one turn, through the real `turn_stream` with `send_event` captured → ledger `{'missing_dependency': 4, 'missing_file': 5}`, **emitted** `max_class=5, top_class='missing_file', max_exact=1`. The PR body's own centrepiece signal ("the class rung counts 4") is invisible in the event, and `event_fields()` emits no per-class breakdown, so it is unrecoverable per-turn.

This is an undeclared deviation — the PR body never mentions it, and the only record is a test docstring (`tests/test_root_cause.py:164`) reassigning the requirement to ENG-1531. ENG-1492 has zero comments; no amendment exists.

**Fix:** `if rc.cls not in _EXACT_ONLY_CLASSES: self.classes[rc.cls] += 1`. Applied → 38/38 green (nothing pins current behaviour either way), and the mixed case becomes `max_class=4 / top_class='missing_dependency'`.

**Cost worth stating on the ticket:** it also zeroes a genuine multi-path wall (3 paths of one missing ODBC install). The ticket already made that call. If that call now looks wrong, amend the ticket — leaving the reversal in a docstring is what makes it undiscoverable.

## 3 — `reason_coverage` is structurally pinned at 1.0 · medium

`session.py:955` `if tool_ok is not False: return`, against `registry.py:142-144` returning `ToolOutcome(content=result)` with `ok: bool | None = None`. Handlers that signal failure by **returning an error string** are dropped entirely — not counted, not `unclassified`, not in the coverage denominator.

AST-checked: all 16 `SideEffectResult.failed()` calls pass `reason=`, and the only reason-less `.to_outcome()` calls are `success=True`. So `from_reason=False` is **unreachable in production**. Executed a mixed turn (1 migrated + 3 unmigrated failures): true coverage 0.25, **emitted 1.0**, `failures=1`, `wall=1`. The metric that exists to expose under-coverage reports perfect coverage by excluding the uncovered population from its own denominator.

The real instances are worse than the obvious ones:

- `scratchpad action:"install"` returns a bare `str` (`tool_handlers.py:580-583` → `local.py:1063`), so **a pyodbc build failure — the canonical ENG-836 wall — is invisible**, while the same wall via exec-with-packages *is* counted (`scratchpad.py:248-251` sets `reason="package_install_failed"`).
- cowork-server injects six plain-string ToolDefs into the deployed session (`harness.py:929-938`, no `ToolOutcome` anywhere in its `tools.py`), including `connect_new_datasource`.

(`web_fetch`/`web_search` are *not* examples — both are provider-native on Anthropic and OpenAI/MINDS-passthrough, so anton never dispatches them on the deployed product.)

**This is a scoping decision, not a one-liner.** The tempting one-liner — `_apply_error_tracking(..., ok=None)` — was tested and refuted: on `"Fetch returned HTTP 403 …"` it leaves the streak at 0, because none of its five markers match. Minimum: name it in the PR body as a known gap (as the multimodal gap already is) and state the coverage number is over migrated handlers only. Better: have `_record_root_cause` accept `ok=None` failures as `unclassified` when a legacy error verdict exists.

---

## Tests that pass for the wrong reason

Each mutation was confirmed **applied** before its result was trusted.

| # | Test | Mutation | Result |
|---|---|---|---|
| 4 | `test_a_raising_classifier_cannot_break_a_turn` | delete the try/except at `session.py:957-962` | **38 passed** |
| 5 | 4 of 5 `tool_reason` writer sites | `tool_reason = ""` at `session.py:3888` | **38 passed** |
| 6 | `test_the_ledger_spans_turns_on_a_reused_session` | make the ledger turn-scoped | **38 passed** |

### 4 — the consequence isn't a dead turn, it's a hijacked prompt · medium

`tests/test_root_cause_wiring.py:219`. `grep -rn "classify_root_cause" tests/` = one hit, line 222, out of 2,105 collected tests.

Histories dumped both ways. Unguarded, `history[2]` = tool_result `"[interrupted by error — tool call did not complete]"` plus `"SYSTEM: An error interrupted execution: boom … Adjust your approach"`. Guarded, it's the real `ModuleNotFoundError` tool_result.

Path traced: the raise leaves `_stream_and_handle_tools` into the pre-existing broad `except Exception as _agent_exc:` at `session.py:3245`, matches no re-raise guard, hits `_seal_dangling_tool_uses("interrupted by error")` (3317) and the recovery note (3339), then retries — so `ended_by` stays `"completed"`, which is the test's only assertion. The model loses the tool output **and** is told to adjust its approach, because of an internal reporting bug.

**Fix (verified: 2 passed on HEAD, 1 failed under the mutation, each assertion load-bearing independently)** — with `blob = json.dumps(session._history)`:

```python
assert "interrupted by error" not in blob
assert "An error interrupted execution" not in blob
assert "No module named 'pyodbc'" in blob
```

### 5 — four of five `tool_reason` writer sites have zero coverage, including the dominant one · medium

`session.py:3888`. Coverage run (`--source=anton.core.session` over the wiring tests): lines 3827, 3888, 3956, 4065 **never executed**; only 4011 (generic dispatch) is. Every wiring test issues `{"action":"view"}`, so the `exec` branch is never entered — but `session.py:3821` gates on `action == "exec"`, which is where every ENG-836 / 2026-08-12-harvest failure arrives on the deployed streaming path.

The regression would be silent and self-flattering: under the mutation a real 3-failure wall turn emits `wall 3→0`, `reason_coverage 1.0→0.0`, `unclassified 0→3`. The code is correct today — verified by writing the missing test: real multi-line tracebacks through `{"action":"exec"}` give `wall=3, max_exact=3, top_class='missing_dependency', coverage=1.0` on HEAD, and go red (`assert 0 == 3`) under the mutation.

**Fix:** one wiring test driving `{"action":"exec"}` with a `Cell` whose `error` is a ~20-line traceback; a second case with a raising handler covers 4065.

### 6 — the turn-spanning test passes for the wrong reason · medium

`tests/test_root_cause_wiring.py:162`. Turn 2 installs a plan that returns a tool call forever plus `dispatch_tool = AsyncMock(return_value=WALL)`, so it runs to the 25-round cap: turn 1 contributes 1, turn 2 contributes 25, against assertions of `>= 2`. Wiping the ledger between turns is invisible.

This is the replacement written for self-review item 2 — but `git show a17c180` shows the diff is a **rename plus a docstring**; the body and both `>= 2` assertions are unchanged. That closure is incomplete.

**Fix (verified both directions):** deterministic turn 2 (one failing tool call, then a text answer) plus exact equality — `assert failures == 2 and max_exact == 2`. Passes on HEAD, fails under the mutation.

---

## Lower severity

**7 — `_IMPORT_NAME_RE` keys on the symbol, not the module · low.** `root_cause.py:127` captures `X` from `cannot import name 'X' from 'Y'`, so three symbols from one broken pandas give `max_exact=1, distinct=3`. The `.split(".")[0]` on the next line is a no-op on a symbol — evidence the module was intended. Zero coverage (`grep "cannot import name" tests/` → nothing). Also easier to hit than it looks: `No module named 'pandas'` and `cannot import name 'read_excel' from 'pandas'` don't group. Verified fix, 38/38 green:

```python
r"cannot import name ['\"]\w+['\"] from (?:partially initialized module )?['\"]([\w.]+)['\"]"
```

**8 — `test_event_fields_are_flat_and_complete` doesn't test completeness · low.** Three of ten properties can be deleted with 38 green: `root_cause_distinct`, `root_cause_max_exact`, `root_cause_unclassified`. Stronger: reverse-renaming any of those three to `rc_*` — the exact refactor `3d86aa6` just performed forward — passes the **full 2,077-test suite**. Both sinks iterate rather than index (`session.py:2172`, `2190-2196`), so a dropped key is silently absent. Fix: one `assert set(f) == {…}`.

**9 — counters are session-cumulative, not per-turn · low.** `_root_causes` is assigned once (`session.py:702`) and never reset. Two-arm probe: as-merged `[(1,5,5),(2,10,10),(3,15,15)]`; with a per-turn reset `[(1,5,5),(2,5,5),(3,5,5)]`. ENG-1492's `Done when` says per-turn, §5 says session-scoped — **the ticket contradicts itself**, so this needs a decision rather than a patch. Bounded in practice: cowork desktop *and* channels both build a session per HTTP turn (`harness.py:342→389→941`), `local_runtime` has **zero** usages in cowork-server staging, and the slices separate via the `harness` property. So the cumulative slice is CLI/embedded only.

**10 — `test_the_ledger_never_resets_on_success` asserts method *names* · low.** An uncalled no-op `def clear()` turns it red; a real reset called from `_record_root_cause` on every success leaves it green (only `test_interleaved_successes_do_not_reset_the_count` fires). A matched pair proved it is never the sole detector of any real reset defect, and it will trip on the natural ENG-1531 refactor. Delete it or replace it with the behaviour.

**11 — `test_classification_never_changes_the_turn` compares role sequences only · low.** Injecting `{'role':'user','content':'ROOT_CAUSE_SIDE_EFFECT'}` from `_record_root_cause` passes the **full 2,077-test suite** — history bundling merges the message into the adjacent tool-result block, so roles stay byte-identical while the content sent to the LLM changes. Note the obvious fix does **not** work: `a._history == b._history` is `False` on clean HEAD, 5/5 runs, because index 6 embeds `<coroutine object … at 0x10c2fd040>`. Working fix, validated both ways:

```python
[(role, re.sub(r"0x[0-9a-f]+", "0xADDR", json.dumps(content, default=str))) for ...]
```

Caveat (hypothesis, not observed): index 0 carries a minute-resolution timestamp; normalise that too or it flakes on a minute boundary.

**12 — the writer-inventory regex sees one spelling · low.** `tests/test_root_cause.py:272` matches only `reason="[a-z_]+"`. Decoys: double-quoted lowercase → caught; single-quoted, digit-containing (`wall_v2`), and constant-referenced → all **missed**. There is no ruff/black/pre-commit config anywhere in the repo, and single-quoted kwargs already exist in this tree (`anton/tools.py`, `tool_defs.py`, `backends/local.py`). The current inventory is accurate; the guarantee is one naming choice from breaking. Fix: `reason=["']([A-Za-z0-9_]+)["']`, or ast-walk the `reason` keywords (precedent at `test_verifier_truncation.py:248`).

---

## The three claims this PR rests on

**(a) "Self-inflicted failures can never trip" — not structural today.** The tier *table* is structural in shape (a frozenset lookup, not a heuristic), but it is unreachable for any reason containing a bare 401/403/404 because the status branch returns first — which is precisely the "agent retries the same broken line" case lines 26-28 claim to have solved. After finding 1's fix it becomes structural **over the enumerated set**, which is a real guarantee. Two residual leaks are worth stating rather than papering over: an agent's circular import → `ImportError: cannot import name 'helper' from partially initialized module 'mycode'` → wall/trip=True; and an agent's wrong path repeated → `missing_file` on the exact rung (the documented, defensible call). The docstring should read *"failures whose exception type is in this table can never trip."*

**(b) "No behaviour change" — true today, contingent on a guard nothing pins.** Confirmed by trace, not assertion: `_record_root_cause` (`session.py:940-962`) returns early on `tool_ok is not False`, wraps the classify call in a bare `except Exception`, and only calls `ledger.add`; the ledger has exactly three references workspace-wide (702 construct, 958 add, 2150 `event_fields`); nothing anywhere reads `trip_eligible` to change control flow; the only outputs are the `turn_cost` log line (2172) and `turn_completed` properties (2198). No turn can break. Two caveats: the guard is the *only* reason that holds (finding 4), and `classify()` raises on every non-str input (bytes → TypeError, everything else → `has no attribute 'strip'`) while `reason or ""` does not coerce. All six current writers pass strings, so nothing is live — but it is one careless `reason=exc` away.

**(c) The sink — none of the ten properties arrive today, verified against production.** PostHog shows zero `turn_completed` events in either project over 30 days; only `anton_`-prefixed rows land, carrying five allowlisted names. Driving the real `send_event` with the real `event_fields()` in three trees:

| tree | destination | result |
| -- | -- | -- |
| #348 alone | zoomInfo collector | properties stripped |
| #348 + #333 | `collect.mindshub.ai` | project 355390 per its docstring only — a 14-day query grouped by `properties.source` returned zero rows |
| #348 + #345 | `us.i.posthog.com/capture/` | project **424726**, no allowlist, all ten as JSON numbers (`"root_cause_reason_coverage": 1.0`) |

So this PR is inert on its own and has no ordering constraint of its own — but **the measurement window should not start until the sink is settled**, and #345 is the only executed-verified sink. Two things make the "splits mid-window" worry smaller than it sounds: #345 merged into #333 **conflicts** (`anton/analytics.py` and `tests/test_analytics.py`), so a human resolves the sink module by hand; and #345 ships `test_turn_completed_goes_to_posthog_not_the_collector` pinning the route in CI. Still worth a line in both PR bodies — and if both #333 and #345 land, the ENG-1492 dashboard must union 355390 + 424726.

---

## Not verified

- **Production frequency of the status-token collision.** No Langfuse pull; only the combinatorial floor — 3/900 (0.33%) of uniformly distributed 3-digit tokens hit 401/403/404, 8/900 (0.89%) the full set. The structural claim in (a) is refuted regardless of rate, but the *pollution magnitude* in the deliverable is unmeasured.
- **The actual `max_exact` / `max_class` distribution** — the ticket's deliverable; needs ≥1 week of real traffic through a working sink, which does not exist yet.
- **#333's collector destination** — zero observed rows carrying its stamped `source`.
- **Anything on a real deployed host.** All results here are unit- or in-repo-e2e; the per-turn-session claim for cowork was read from `origin/staging` source, not observed live.
- **Multimodal tool results** (`continue` before the recorder → unclassified) — known, disclosed, deferred to ENG-1531; not re-verified here.
- **The minute-boundary flake** in finding 11 — reasoned from a timestamp at history index 0, never observed.

*Method note:* several runs found the shared worktree concurrently mutated by sibling agents (a foreign `MUTATION M10` in `session.py`, a foreign `missing_file` exclusion in the ledger). Every number above was re-run in an isolated copy with module identity checked (`__file__` + md5) before the result was trusted; two contaminated early runs were discarded.

alecantu7:
## Three findings fixed, all mutation-verified — `HEAD`

Finding 1 is a genuine catch and it falsified the module's central claim. Reproduced independently before fixing, including the one-byte control:

```
{200:..,404:..}[403]          -> KeyError: 403          -> external_wall  trip=True
assert 404 == 200             -> AssertionError         -> external_wall  trip=True
SyntaxError … (line 404)      -> external_wall  trip=True
CONTROL — the same at line 405 -> self_inflicted trip=False
```

**Fixed** by moving the type tables above the status branch. `HTTPError` is in neither table, so it still falls through and the legitimate 429-vs-403 distinction is untouched — verified: `429 → transient`, `403 → wall`, `401 → wall`, `KeyError: 403 → self_inflicted`.

You were right that coverage was green by luck: every fixture in `SELF_INFLICTED_REASONS` used out-of-range numbers. Five now use the real ones (`line 404`, `KeyError: 403`, `404 != 200`, `'401'`, `at 500`), and reverting the ordering reddens 5 tests.

### 2 — `missing_file` on the class rung · fixed

Confirmed against ENG-1492 §4. Added `_EXACT_ONLY_CLASSES`; the mixed case now emits `max_class=4, top_class="missing_dependency"` instead of `5 / "missing_file"`.

Your point about *where* the deviation was recorded is the part I'd most want to keep: it lived in a test docstring that reassigned the requirement to ENG-1531, which nobody would ever find. The cost you flagged (a genuine wall spread over several paths now scores 1 on both rungs) is in the constant's comment rather than left implicit.

### 3 — `reason_coverage` pinned at 1.0 · fixed

Verified: 1 migrated + 3 unmigrated failures reported `coverage 1.0, failures 1`. `ok is None` failures are now counted as `unclassified` — never trip-eligible, since the text is prose the model can influence.

I took the "better" option rather than the PR-body note. The legacy verdict is **shared with `_apply_error_tracking` via one constant** rather than reimplemented, so the two cannot drift about what a failure is. Its false-positive direction (`"0 records failed"` counts) is inherited deliberately and pinned by a test, so the inheritance is visible if it ever changes.

Your refutation of the tempting one-liner saved me from it.

---

## Not yet done — findings 4-12

The three test-quality ones (4, 5, 6) and the six lower-severity items are **not** addressed. None affects the correctness of what ships; all affect whether a future regression is caught. Finding 6 is a fair hit — `a17c180` renamed the test and its docstring and left both `>= 2` assertions, so it is still blind to a turn-scoped ledger.

Suite: **2,083 passed**, same 2 pre-existing `test_datasource.py` failures.

## A method note back

Worth recording alongside yours: mid-fix I ran a mutate → test → restore → verify sequence as compound commands, and the restore raced the verification. I twice concluded the fix had not worked when it had, and only resolved it by clearing `__pycache__` and re-running each step in isolation. Same family as your "confirm the mutation is applied before trusting the result" — the inverse direction, confirming the *restore* landed.

Five tests in this PR have now passed for the wrong reason (two I found, three you did). The common thread in every one: asserting the shape of a failure without constructing an input that reaches the code under test.

alecantu7:
All 12 review findings addressed — `badf75f` (1–3) and `57b481e` (4–12).

**Every fix was mutation-verified**: mutation confirmed applied, the test watched *failing* on the mutant, then restored. That mattered — one fix turned out to have no test behind it at all.

| # | Finding | Disposition | Mutation that proves it |
|---|---|---|---|
| 1–3 | ReDoS bound; status-code check preempting `self_inflicted`; `rc_*` naming | fixed in `badf75f` | — |
| 4 | Raising-classifier test too weak | now asserts on what reached the model | guard removed → **fails** |
| 5 | Scratchpad exec path uncovered | new test w/ real traceback + raising-tool test | reason blanked → **fails** |
| 6 | Turn-spanning test nondeterministic | deterministic, exact equality | ledger reset per turn → **fails** |
| 7 | Import regex captured the symbol | **had zero coverage** — new test added | symbol regex restored → **fails** |
| 8 | `event_fields` subset assertion | exact set equality, all 10 names | field deleted → **fails** |
| 9 | Per-turn vs session scope | no code change — see below | — |
| 10 | "Never resets" asserted method *names* | behavioural assertion | reset on success → **fails** |
| 11 | No-behaviour-change compared roles only | compares normalised content | injection on failure path → **fails** |
| 12 | Writer-inventory regex too narrow | widened | planted `reason='wall_v2'` → **fails** |

**Finding 7 is the one worth flagging.** The regex was fixed earlier in this branch, but the mutation reverting it **survived a full green run** — nothing tested it. Four attempts against a broken `pandas` were scoring `max_exact=1, distinct=4` instead of the one wall they are, which is exactly the signal this ticket exists to produce. Now pinned across all four shapes (two symbols, a submodule, a circular import).

**Finding 4 was worse than it looked.** Probing the path directly: a bare `raise` inside `_record_root_cause` **fails no test in the file** — the tool loop's broad `except` swallows it. But it seals the tool_use as `[interrupted by error]` and injects a SYSTEM "adjust your approach" note, so the turn still ends `completed` while the model has lost the tool output to an internal reporting bug. `ended_by` could never have caught that; the test now asserts on history content.

**Finding 9 — no code change, deliberately.** The ticket isn't self-contradictory: §5 specifies a session-scoped ledger, §6 specifies emitting per turn. The real hazard is that the *field names* read as per-turn while the values are cumulative — so a naive percentile over the raw event stream is dominated by long sessions and reads high. Since that percentile **is** this ticket's deliverable (it sets ENG-1286/ENG-1531's thresholds), I documented it at `event_fields`, the point of consumption, with the correct aggregation spelled out: take the last event per session, or compute per-session maxima. Happy to change the field names instead if you'd rather make it structural.

Suite: **2,086 passed**. The 2 `test_datasource.py` failures are pre-existing — verified failing with this branch's changes stashed out entirely.

alecantu7 (commented):
## Fourth pass — head `57b481e6`

Two commits since the review at `3d86aa61`: `badf75f4` (findings 1–3) and `57b481e6` (findings 4–12). **All twelve are fixed**, and I verified them independently rather than reading the commit messages — ten mutations, each confirmed *applied* before its result was trusted.

| Mutation | Reverts | Result |
| -- | -- | -- |
| M4 — hoist the status check back above the type tables | F1 | **5 failed** |
| M1 — remove the `_EXACT_ONLY_CLASSES` guard | F2 | 1 failed |
| M2 — `_legacy_looks_like_failure` → `return False` | F3 | 2 failed |
| M5 — delete the try/except around the classify call | F4 | 1 failed |
| M8 — blank the exec-path `tool_reason` | F5 | 1 failed |
| M7 — reset the ledger per turn | F6 | 1 failed |
| M3 — regex captures the symbol again | F7 | 1 failed |
| M6 — drop `root_cause_distinct` from the emitted set | F8 | 1 failed |
| M9 — add an **uncalled** `clear()` | F10 | **49 passed** ✓ correct |
| M10 — plant `reason='wall_v2'` in a real writer | F12 | 1 failed |

M9 is the one worth pausing on. An uncalled `clear()` used to turn the old test red — a harmless addition failing the suite while a *real* reset called on every success sailed through. It no longer reddens, which is the correct direction: the replacement asserts behaviour, so it stays green on the no-op and would catch the reset that matters.

Two of my own runs were no-ops before I caught them: zsh doesn't word-split an unquoted `$T`, so `pytest $T` received one bogus path and printed *"no tests ran"* — which reads exactly like "the mutation survived" if you skim the tail. Worth knowing, since it is the same class of trap as a `sed` anchor that silently doesn't match.

**Full suite: 2088 passed, 28 skipped, 0 failed** (153 s). The two `test_datasource.py` failures the commit message calls pre-existing do not reproduce here — same environment-specific pattern as the earlier "7 pre-existing" claim, so both numbers are worth treating as local rather than universal.

---

## The three that mattered

### F1 — the hijack table is empty, and the controls survive

```
IndexError: index 404 is out of bounds…    -> self_inflicted   trip=False
KeyError: 403                              -> self_inflicted   trip=False
AssertionError: assert 404 == 200          -> self_inflicted   trip=False
UnicodeDecodeError … position 404          -> self_inflicted   trip=False
SyntaxError … (detected at line 404)       -> self_inflicted   trip=False
--- controls ---
HTTPError: 429 Too Many Requests           -> transient        http_429
HTTPError: 403 Forbidden                   -> external_wall    permission_denied  trip=True
requests…HTTPError: 500 Server Error       -> transient        http_500
```

`HTTPError` sitting in neither table is what preserves the legitimate status routing, and the comment now records the ordering as a **safety property rather than a style choice** — which is the part that stops a future tidy-up from reintroducing it. Switching five `SELF_INFLICTED_REASONS` fixtures to in-range integers is what makes M4 redden five tests where it previously reddened none.

### F2 — the mixed case reports the right rung

Five file probes plus the four-identifier pyodbc wall through the real ledger:

```
before: max_class=5  top_class='missing_file'         max_exact=1
after:  max_class=4  top_class='missing_dependency'   max_exact=1
```

### F3 — you took the stronger option, not the one-line disclosure I suggested

`ok=None` failures now enter as `unclassified` behind a legacy-failure gate, so the metric can finally do the job it exists for:

```
mixed turn (1 migrated + 3 unmigrated failures):
  reason_coverage = 0.25        <- was a false 1.0
  failures = 4   unclassified = 3   wall = 1
```

And the instance I named as the real one is captured: `install_packages` returns `"Install failed (exit N):…"` (matches `failed`) and `"Install timed out after Ns."` (matches `timed out`), so the canonical ENG-836 pyodbc build failure is now visible instead of dropped.

**One residual, worth a doc line rather than a change.** `_LEGACY_FAILURE_MARKERS` is the same five-marker list, and it still misses plain-prose failures. The one that matters is cowork-server's `connect_new_datasource`, which returns `"Could not connect to the datasource."` and matches nothing:

```
captured=True   Install failed (exit 1): … fatal error: sql.h
captured=True   Install timed out after 300s.
captured=False  Could not connect to the datasource.
captured=False  Successfully installed pandas-2.2.0      <- correctly not captured
```

That is now *undercounting which surfaces as low coverage* rather than a silent 1.0 — the important difference — but the markers list is the boundary of the metric, and naming it in the docstring costs nothing and saves the next reader from rediscovering it.

---

## F9 — documented rather than changed, and that is the right call

`root_cause.py:408-417`, at the point of consumption. Better than what I asked for. The ticket genuinely is self-consistent (§5 session-scoped ledger, §6 emitted per turn); the hazard was never the storage scope but that the **field names read as per-turn while the values are cumulative**, so a naive per-event percentile is dominated by long sessions and reads high — and that percentile is this ticket's actual deliverable. The note states the correct aggregation. The harness table at `:345` also now records that Cowork builds a session per HTTP turn, which is what bounds the cumulative slice to CLI/embedded.

## F7, for the record

All four import shapes now collapse to one key, which is the behaviour the `.split(".")[0]` always implied:

```
No module named 'pandas'                                   -> missing_dependency:pandas
cannot import name 'read_excel' from 'pandas'              -> missing_dependency:pandas
No module named 'pandas.io.parsers'                        -> missing_dependency:pandas
cannot import name 'helper' from partially initialized …   -> missing_dependency:pandas
max_exact=4  distinct=1
```

---

## Not re-verified

F11's normalised-content comparison — read, not mutated, so the claim that it now catches an injected message rests on your verification rather than mine. Same for the classifier-input bound in `65fbc842` from the risk pass, which predates my review and I have never checked.

Nothing here is blocking.

*Method note: all mutation work ran in an isolated copy with module identity checked (`__file__` + md5 `3f36100836210a32b57985416b204541`) before any result was trusted, and every mutation was reverted with a clean-baseline re-run. Note there is another worktree on this branch at the same head under `~/.claude/jobs/`, so the verification above is stamped at `57b481e6` specifically.*

anton/core/root_cause.py
```diff
@@ -0,0 +1,430 @@
...
+    # That is exactly the "agent retries the same broken line" case this tier
+    # exists to exclude, so the guarantee in the module docstring was false for
+    # 8 specific integers. `HTTPError` is in neither table, so it still falls
+    # through to the status branch — which keeps the legitimate case working.
+    if exc in _SELF_INFLICTED:
+        return RootCause(TIER_SELF, exc, "")
+    if exc in _TRANSIENT:
+        return RootCause(TIER_TRANSIENT, exc, "")
+
+    status = _STATUS_RE.search(reason)
```
pnewsam, line 329: **Minor:** the status heuristic can still mint a *trip-eligible* `external_wall` from a self-inflicted bug when the exception type is not in `_SELF_INFLICTED`. Hoisting the type tables above this branch closes the enumerated set, but a non-enumerated agent error — `RuntimeError`, `LookupError`, a bare `Exception` — carrying a bare `401/403/404` in its message (e.g. `RuntimeError: record 404 not found`) lands here as `external_wall`/`missing_file`, `trip_eligible=True`.

Not a docstring violation (the claim is correctly scoped to types in the table), measurement-only, and the direction inflates wall counts — visible in the data before ENG-1531 consumes any of it. Frequency is unmeasured, as you already flagged. Worth one line in the module comment naming the non-enumerated exception surface alongside the existing status-collision note; no code change needed pre-merge.

anton/core/session.py
```diff
@@ -598,6 +611,7 @@ def _build_verify_request(
 
 
 @dataclass
+
```
pnewsam: **Nit:** accidental blank line inserted between `@dataclass` and `class ChatSessionConfig:` (valid Python, harmless). There's a matching extra blank line after `_legacy_looks_like_failure`. Trim if you touch the file again.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the complete diff — the new `root_cause.py` classifier, its two wiring points in `session.py`, and both test files. This is measurement-only, and I confirmed from source that it is structurally incapable of altering a turn, so it clears both the "does it break anything" and "are the numbers trustworthy" bars. No blocking or major findings. Two small inline notes below.

### Validation

- **No behaviour change, verified from source:** `_record_root_cause` wraps both branches in `try/except`, only mutates `self._root_causes`, returns `None`, and never touches `result_text`. The ledger is read nowhere for control flow — only `event_fields()` (guarded) feeds the log line and `send_event`. Nothing reads `trip_eligible`.
- **No double-counting:** `turn()` runs its own dispatch loop and does not delegate to `turn_stream`→`_stream_and_handle_tools`. The two record sites are mutually exclusive paths, one record per tool — consistent with `test_the_non_streaming_turn_records_too` expecting `failures==2`.
- **Log-format arity:** 15 `%`-specifiers against 15 args in the `turn_cost` line — matches.
- **No kwarg collision:** every key in `**_root_cause_fields` is `root_cause_*`-prefixed; none clashes with the explicit `send_event` kwargs.
- Parsed the stray `@dataclass`-then-blank-line edit with `ast` — valid Python, not a syntax break.

### Note (not a defect)

`event_fields()` emits session-cumulative values under per-turn-reading names. This is documented at the point of consumption with the correct aggregation ("take the last event per session"), which is the right call — flagging only that the ENG-1492 dashboard consumer must honor it, since a naive per-event percentile over a host where the ledger spans turns reads high.

The extensive self-review (12 findings + one risk finding, all mutation-verified at this head) holds up against independent re-derivation of the load-bearing invariants.

alecantu7:
Thanks @pnewsam — both notes addressed in `02239ea`, **documentation and whitespace only, no logic change**. Approval is stamped at `57b481e`; this repo doesn't dismiss on push, so it still reads `APPROVED` at `02239ea` (verified via the API, not assumed).

**The non-enumerated exception surface — you're right, and I reproduced it:**

```
RuntimeError: record 404 not found  -> external_wall / missing_file       trip=True
LookupError: 403 no such entry      -> external_wall / permission_denied  trip=True
Exception: got 401 back             -> external_wall / auth_missing       trip=True
--- the enumerated case still holds ---
KeyError: 403                       -> self_inflicted                     trip=False
RuntimeError: boom                  -> unclassified                       trip=False
```

Documented rather than fixed, for the reason you implied: the only widening that catches it is putting `Exception` in `_SELF_INFLICTED`, which would swallow every genuine wall raised as a generic type. The comment records why it's left, that the direction is the safe one (inflates wall counts → visible as noise rather than a hidden wall), and — the part I care most about surviving — a forward constraint for ENG-1531: *don't trip on a `_STATUS_WALLS` class without either an HTTP-shape precondition on the reason or a check that the type was enumerated.* Frequency stays unmeasured; the combinatorial floor is 3/900.

**Blank lines** — both trimmed (`@dataclass`/`ChatSessionConfig`, and the doubled one after `_legacy_looks_like_failure`).

**One more residual I documented while in there**, from my own fourth pass rather than yours: `_LEGACY_FAILURE_MARKERS` is the boundary of the coverage metric. A legacy handler reporting failure in plain prose matches nothing and isn't counted — cowork-server's `connect_new_datasource` returns `"Could not connect to the datasource."`:

```
captured=True   Install failed (exit 1): … fatal error: sql.h
captured=True   Install timed out after 300s.
captured=False  Could not connect to the datasource.
captured=False  Successfully installed pandas-2.2.0   <- correctly skipped
```

Safe direction — it surfaces as *low* `reason_coverage` rather than the false 1.0 the gate exists to prevent — but the denominator is itself substring-defined, and that shouldn't have to be rediscovered by whoever reads the first distribution.

On the `event_fields` aggregation note: agreed, and it's the one thing I'd want carried into the dashboard rather than the code — last event per session, or per-session maxima, never a percentile over the raw event stream.

**Suite: 2,086 passed.** On the two `test_datasource.py` failures — you're right to treat them as local. They reproduce here with this branch's changes stashed out entirely, so they're not from this PR, but since they don't reproduce for you they're environment-specific rather than a broken baseline. I'll file that separately rather than carry it here.

Not merging — that's yours.

anton/core/root_cause.py
alecantu7: Confirmed and documented in `02239ea` — comment only, no logic change. Reproduced exactly as you described:

```
RuntimeError: record 404 not found  -> external_wall / missing_file       trip=True
LookupError: 403 no such entry      -> external_wall / permission_denied  trip=True
Exception: got 401 back             -> external_wall / auth_missing       trip=True
--- enumerated case still holds ---
KeyError: 403                       -> self_inflicted                     trip=False
```

Left as behaviour for the reason your note implies: the only widening that catches it is putting `Exception` in `_SELF_INFLICTED`, which would swallow every genuine wall raised as a generic type. The comment records that, notes the direction is the safe one (inflates wall counts → visible as noise rather than a hidden wall), and carries a forward constraint to ENG-1531: do not trip on a `_STATUS_WALLS` class without either an HTTP-shape precondition on the reason or a check that the exception type was enumerated. Frequency still unmeasured; combinatorial floor 3/900.

anton/core/session.py
alecantu7: Both trimmed in `02239ea` — the blank line between `@dataclass` and `ChatSessionConfig`, and the doubled blank after `_legacy_looks_like_failure`.

anton/core/root_cause.py
```diff
@@ -0,0 +1,447 @@
...
+    # to include `Exception`, which would swallow every genuine wall raised as a
+    # generic type. The direction is the safe one — it INFLATES wall counts, so
+    # it is visible as noise in the ENG-1492 distribution rather than hiding a
+    # wall — and this is measurement-only until ENG-1531 reads any of it.
+    #
+    # For ENG-1531: do not trip on `_STATUS_WALLS` classes without either an
+    # HTTP-shape precondition on the reason or a check that the exception type
+    # was enumerated. Frequency here is UNMEASURED; the combinatorial floor is
+    # 3/900 of uniformly distributed 3-digit tokens.
+    status = _STATUS_RE.search(reason)
```
pnewsam, line 329: **Minor (measurement fidelity, non-blocking):** the `_STATUS_RE` branch runs before the `_WALL_TYPES` lookup, so an *enumerated* wall type whose text or path incidentally contains a bare 3-digit 4xx/5xx token gets reclassified. Trigger: `FileNotFoundError: [Errno 2] No such file or directory: '/data/500/report.csv'` → `_STATUS_RE` matches `500` → `TIER_TRANSIENT, http_500`, so a genuine `missing_file` wall is recorded as a transient blip and never reaches `root_cause_wall` or a trip rung (a DB/connection wall on a port like `:503`/`:429` hits the same path). This only ever *undercounts* walls — it cannot manufacture a false wall, since `_SELF_INFLICTED`/`_TRANSIENT` are matched first — so it is in the direction ENG-1492 already accepts, and the motivating dependency walls (`pyodbc`, `pymssql`) carry no such tokens. The forward-constraint note at lines 313–316 covers the *residual* (non-enumerated) types but not this enumerated-type-through-status interaction; worth one line there so ENG-1531's sizing knows walls with numeric ports/paths read low.

anton/core/session.py
```diff
@@ -924,6 +961,52 @@ def last_compaction(self) -> dict | None:
...
+            # verdict available. Dropping these entirely was the defect —
+            # `reason_coverage` then reported 1.0 by excluding the uncovered
+            # population from its own denominator, i.e. the metric that exists
+            # to expose under-coverage could not. Measured: a turn with 1
+            # migrated and 3 unmigrated failures reported coverage 1.0.
+            #
+            # Counted as `unclassified`, which is not trip-eligible — the text
+            # is prose the model can influence, which is the ENG-1276 defect one
+            # level up, so it must never create a wall.
+            if not _legacy_looks_like_failure(result_text or ""):
```
pnewsam: **Nit (non-blocking):** this `_legacy_looks_like_failure(result_text or "")` call sits *outside* the `try/except` that guards the rest of the method, which slightly undercuts the “guarded whole” wording in the docstring. In practice `result_text` is only ever `str` or `list` (both safe for the `m in text` membership test), so I could not construct a reachable crash — flagging it as an observation, not a defect. If you want the invariant literally airtight, move the predicate inside the guard (or guard the whole body).

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff: `root_cause.py` (new, 447 lines), the `session.py` wiring (+121/-1), and both test files. The measurement-only invariant holds — `_record_root_cause` reads the tool result without mutating it, the ledger is write-only with respect to turn control flow (its consumer, ENG-1531, is deliberately unbuilt), and every classify/emit path is guarded so a reporting failure can't fail a turn. No blocking or major issues. Two non-blocking notes inline, both in the safe (undercount) direction.

### Validation

- Verified the module-top import chain resolves: `_normalise_error_signature` (`acc.py:264`) is pure — no env/`ANTON_ACC_MODE`/global reads in it or `_unwrap_envelope_message` — and there is no `acc → session` cycle. Since `session.py` imports `root_cause` at load time, this was the only credible turn-breaking risk. Refuted.
- Confirmed `ToolOutcome.reason` exists (`registry.py:42`, default `""`), so all five writer sites are valid.
- Traced both dispatch sites (`turn()` and `_stream_and_handle_tools`): `_record_root_cause` runs once per tool, returns `None`, does not feed `result_text` back into the loop, and nothing reads the ledger to alter behavior.
- `turn_cost` log line: 15 placeholders match 15 args after the added `%s`.
- `send_event(**_root_cause_fields, …)`: all keys are `root_cause_*`; no collision with the explicit kwargs; empty-dict fallback spreads harmlessly.
- All regex inputs are bounded to `_MAX_REASON_CHARS` (500) before any pattern runs, containing the documented `_HOSTPORT_RE` quadratic.
- Did not run the suite (needs the branch in a worktree with the test env); the author reports 2,086 passing with per-finding mutation verification, and I confirmed the import-time preconditions that are the only path to a turn break.

anton/core/root_cause.py
alecantu7: Confirmed and documented in `a39d229`. Reproduced both directions:

```
FileNotFoundError … '/data/500/report.csv'  -> transient http_500
ConnectionRefusedError … db.internal:503     -> transient http_503
--- controls ---
FileNotFoundError … '/data/report.csv'      -> external_wall / missing_file
ConnectionRefusedError … db.internal:5432    -> external_wall / connection_refused
```

The `:5432` control is the neat part — four digits, so `\b` blocks it; `:503` is exactly three and does not.

What your note made me see is that the status branch errs in **both** directions: it inflates walls from non-enumerated types (the earlier note) and deflates them from enumerated ones (this one). I had documented only the first, which made the comment read as if the bias had a single sign. Both are now recorded together, with the point that the single ENG-1531 precondition already noted — HTTP-shape or enumerated-type check — closes both at once, and with the sizing consequence you asked for: **walls with numeric ports or paths read LOW in the distribution.**

Documentation only. `pyodbc`/`pymssql` carry no such tokens, so the motivating population is unaffected.

anton/core/session.py
alecantu7: Good catch — fixed rather than left, in `a39d229`.

You are right that no reachable crash exists today (`result_text` is only ever `str` or `list`, both safe for `in`), so this is not a defect. I moved it anyway because "guarded whole" is the claim the entire no-behaviour-change argument rests on, and the escape route is finding 4's damage through a different door: the raise leaves into the tool loop's broad `except`, seals the tool_use as `[interrupted by error]` and injects the "adjust your approach" note. An almost-true invariant in exactly that position is the kind that rots quietly.

Pinned by `test_the_legacy_predicate_is_inside_the_guard`, which feeds a truthy object raising on `in`. Mutation-verified — reverting the predicate back outside the guard fails it with `RuntimeError: boom`.

alecantu7:
@pnewsam heads-up — one more commit, `1b86a2b3`, and it **adds an eleventh event property**, so it changes the shape you validated in both passes. Flagging rather than slipping it in.

**What I found self-reviewing `a39d229`.** The recorder's guard swallows classification failures into `logger.debug`, which production doesn't emit. So a systematically broken classifier produces an event byte-identical to a turn where nothing failed:

```
3 real wall failures, classifier raising:
  failures=0 distinct=0 max_exact=0 max_class=0 top_class="" wall=0
3 successes, genuinely clean turn:
  failures=0 distinct=0 max_exact=0 max_class=0 top_class="" wall=0
```

Ten properties, no field separating them.

**Why I didn't leave it as a note.** ENG-1492's `Done when` names *"the wall-repeat population is too small to justify the control at all"* as a valid outcome. So a silent instrument failure doesn't look like missing data — it looks like **the answer**, and would cancel ENG-1531 on the strength of a bug. It's also the same defect `reason_coverage` exists to fix, one level down: a metric mustn't exclude its own failures from its own denominator.

**The change:** `classify_errors` on the ledger, incremented in both guards, emitted as `root_cause_classify_errors`. Deliberately *not* folded into `failures` — counting a swallowed classification as a tool failure would corrupt the distribution it's meant to protect.

**Mutation-verified three ways, and the third is the interesting one:**

| Mutation | Result |
|---|---|
| `note_error()` → no-op | 1 failed (`assert 0 == 3`) |
| drop the property from `event_fields` | 2 failed |
| remove `note_error()` from the **legacy** guard | **16 passed** ← survived |

The two guards are separate writers, and my new test only drove the migrated path — so the legacy one (the `tool_ok is None` branch `a39d229` had just widened) was unpinned. The guard test grew an assertion for it and now fails `assert 0 == 1`.

**~~One thing for the sink, @pnewsam or whoever picks up ENG-1355: this is an 11th property and needs the same allowlist pass-through as the other ten.~~ — WRONG, corrected below.**

**The 11th property needs nothing.** I routed that ask to a cancelled ticket. ENG-1355 was cancelled on 2026-08-14 and its PRs closed unmerged (anton#333, mindshub_services#186, terraform#162, cowork#611); `collect.mindshub.ai` is absent from `origin/staging`. The sink is **ENG-1495 / anton#345**, whose `_posthog_body` has **no allowlist** — `properties` is every kwarg minus `_`, `action` and `timestamp`. So a new property arrives on the same terms as the other ten, with no registration step anywhere.

My own PR body already carried the corrected framing, which makes the slip worse rather than better — apologies for the noise. ENG-1492's description and the ENG-1495 count comment are both updated.

Suite: 2,088 passed. Re-request review if you'd rather look again before this lands — the event shape is genuinely different from what you approved.

alecantu7:
All six addressed. Four were prose rather than code — and finding 1 was mine, so thank you for catching it before it sent someone to a dead ticket.

| # | Finding | Where | Status |
|---|---|---|---|
| 1 | Routed the 11th-property ask to cancelled ENG-1355 | my PR comment | corrected in place, struck through |
| 2 | PR body said "these ten fields" | PR body line 45 | now eleven, + the no-allowlist detail |
| 3 | Same stale ENG-1355 blockquote | ENG-1492 description | replaced with the corrected sink |
| 4 | "40 distinct properties" now off by one | ENG-1495 comment | new comment: 41 = 27 + 11 + 3 |
| 5 | False `# pragma: no cover` on both guards | `b0e71fc7` | removed |
| 6 | `reason_coverage` skew on partial failure | `b0e71fc7` | documented |

**Finding 1 — verified rather than taken on faith,** since it contradicted something I'd written twice. ENG-1355 is `Canceled` (2026-08-14, canceledAt confirmed via the API); anton#333 is `CLOSED` unmerged; `collect.mindshub.ai` is absent from `origin/staging`; and `_posthog_body` in #345 builds `properties` as every kwarg minus `_`, `action`, `timestamp` — no allowlist anywhere. **So the 11th property needs nothing.** The galling part is that my own PR body already carried the corrected framing, which makes it a copy-paste of a superseded belief rather than a genuine misunderstanding.

**Finding 5 is the one I'd have hated to leave.** The pragmas excluded the exact lines `1b86a2b` added, so a future removal of either `note_error()` would have been invisible to a coverage gate — the same silent hole that commit exists to close, reintroduced one level up. The mutation evidence is what proves the premise: removing `note_error()` from the migrated guard fails `assert 0 == 3`, from the legacy guard `assert 0 == 1`. Both pragmas gone, with a comment saying why they mustn't return. The two unrelated pre-existing ones (`session.py:383`, `:2207`) are untouched.

**Finding 6 — confirmed the arithmetic exactly as you stated it:**

```
3 classified (2 with reason) + 2 raised -> reports 0.667, true 2/5 = 0.4
```

Documented on `reason_coverage` itself rather than the ledger, per your point about where a dashboard author actually looks. I kept your self-refutation in the docstring too, because it's the part that stops someone over-correcting: the *systematic* case reads 0.0, not 1.0 — the safe direction — so this only bites on partial failure, and `classify_errors > 0` flags it either way.

Suite: 2,088 passed. Head is now `b0e71fc7`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Re-reviewed the full diff at head `b0e71fc`, focusing on the two commits that landed after my last pass (`1b86a2b` `classify_errors`, `b0e71fc` pragma removal) plus verification that every previously-flagged finding is closed. All 12 self-review findings and my 4 follow-ups are addressed with mutation-verified tests; CI is green across all 13 checks including `run-tests` and CodeQL/Snyk. This is measurement-only code — nothing trips or hands back — and the remaining biases are documented and forward-constrained to ENG-1531. No actionable findings.

### Validation

- Traced both `_record_root_cause` call sites (`turn()` and streaming): the `ok=None` / `ok=True` / `ok=False` branches are mutually exclusive, guarded whole, and cannot double-count.
- Confirmed the new `classify_errors` counter is load-bearing: `test_a_broken_classifier_is_distinguishable_from_a_quiet_turn` proves a raising classifier now emits `classify_errors=3` vs `0`, where before all ten fields were byte-identical to a clean turn.
- Confirmed `test_event_fields_are_flat_and_complete` asserts the exact 11-key set (now including `root_cause_classify_errors`), so a dropped field reddens it.
- Verified the pragma removal in `b0e71fc` is correct: the two `except` guards in `_record_root_cause` are executed by `test_a_raising_classifier...`, `test_a_raising_tool...`, and `test_the_legacy_predicate_is_inside_the_guard`, so `# pragma: no cover` was rightly false.
- Checked the earlier inline threads: status-heuristic both-direction bias documented (`a39d229`), legacy predicate moved inside the guard and pinned (`a39d229`), blank lines trimmed (`02239ea`). All resolved.
- Verified the `_emit_turn_cost` log-line change balances placeholders and args (`by_role=%s %s` + one appended join arg); `run-tests` exercises this path and is green.

The only things left in the code are the two explicitly-accepted measurement biases (status-token inflation of non-enumerated types; deflation of enumerated walls carrying a bare 3-digit port/path token). Both are documented at the branch, err in directions ENG-1492 already sanctioned, and carry a single forward constraint to ENG-1531. Nothing here consumes the tier, so neither is merge-relevant.

PR #349, opened 2026-08-13
ENG-1537: wait out a velocity rate-limit instead of retrying into it

alecantu7:
Closes part of [ENG-1537](https://linear.app/mindsdb/issue/ENG-1537). **anton half only** — see *Merge order* below.

## The bug

A gateway velocity 429 (`rate_limited`) was classified `session_backoff=False`, which skips the session's backoff path (`session.py:3218`) and falls to the count-based recovery path. That path **re-issues the request twice with no delay**, appending a *"transient service issue… continue as planned"* note to history each time. So when the gateway said *"too many tokens per minute"*, anton immediately sent more — with a larger context each attempt, because the note grows the payload. Then it made one more call to summarize, into the same limiter.

Nothing in anton read `Retry-After`, so the interval the gateway explicitly hands us was never used.

The gateway is unambiguous about this (`minds/inference/errors.py:139`):

> Distinct from `allowance_exhausted`: the caller should slow down and retry after `retry_after` seconds, **not** wait for an allowance reset or add credits.

**Measured:** 239 denied calls / 5 external users / 30d on the `anton:turn` surface — the largest single error family by call volume for external Cowork users. The mechanism explains its own inflation, so user-visible failures are far fewer than 239.

## The fix

A velocity limit is the one failure class where waiting is both necessary *and* sufficient, and the server tells us how long. So it waits.

- **`retry_after_seconds()`** parses the header — integer seconds only. The HTTP-date form is legal but nothing in use emits it, and misreading a date as a number would produce an absurd delay, so it's ignored rather than guessed. Both mappers pass it through.
- **`session_backoff=True`** on the velocity-429 branch routes it to the *existing* budgeted backoff path, which already seals dangling `tool_use`s, sleeps, and resumes the **same step** with completed tool results intact and **no recovery note injected**. That path was written for ENG-673 and documented as deliberately not telling the model to "adjust your approach" — exactly right here.
- **Its own budget:** 90s per turn, separate from the 30s incident budget so neither starves the other, plus a 60s cap on a server hint (incident cap unchanged at 30s). A hint above the cap stops the wait and surfaces the card naming the interval, rather than silently absorbing a multi-minute stall.
- **The user is told why the turn is quiet**, through the existing `StreamTaskProgress` affordance. A silent 90s pause is indistinguishable from a hang — that's how a correct fix gets reported as a freeze. No new event type; the turn stays alive so nobody types "continue".
- **On budget exhaustion** the error carries `code="rate_limited"` for a wait-and-retry card.

### Why `session_backoff=False` was right before and wrong here

The flag means *"should the SESSION spend its budget on this?"* The SDK's own retries fire seconds apart — the right answer for a 5xx that recovers instantly, and useless against a per-minute token ceiling. 5xx and mid-stream behaviour are unchanged.

## Out-of-credits must not wait, and doesn't

`wallet_empty` and `included_allowance_exhausted` share the 429 status but are **permanent for the identical request** — the allowance resets monthly, so a wait could only delay the card the user needs, and making "429 → wait" the blanket rule would regress ENG-1169.

Guarded twice over, both preserved: the wallet branch runs *before* `classify_transient`, and `classify_transient` re-checks the wallet code. New parametrised test asserts both dialects of both denials still fail fast **even when a `Retry-After` is present**.

## Verification

- **Full suite: 2003 passed.** 7 failures, all pre-existing — confirmed by re-running the same files against clean `origin/staging` source and getting the identical 7 (subprocess-spawning scratchpad / cloud-turn tests).
- **Mutation-verified.** All 10 new/changed assertions were watched **failing** against `origin/staging` source before the fix, then passing after. Source was restored by checksum, not by eye.
- Two existing tests asserted the old behaviour and were **updated deliberately, not flipped** — they encoded the defect, and the reasoning is in the diff.

Honest note on the 6 billing-guard assertions: on unfixed code they fail on the new kwarg's signature, so they prove *future* protection rather than a behaviour change. The two that demonstrate new behaviour are the `session_backoff` flip and the header read.

## Security pass

No new user input reaches a sink. The parsed header is numeric-only and clamped, so a hostile or buggy `Retry-After` cannot stall a turn beyond the 60s cap or the 90s budget; unparseable, negative and non-finite values are dropped. Error copy stays curated — no provider internals reach the chat. The added log line records the status and parsed interval, and the body is already `scrub_credentials`-filtered.

## Merge order — do not merge this alone and call ENG-1537 done

This PR makes the wait real, which is the half that matters. Two follow-ups complete the ticket:

1. **cowork-server** — map the gateway's `rate_limited` reason (the one of its five reasons `_map_gateway_reason` doesn't know, which is why a velocity 429 currently shows *"You're out of credits — Top up balance"*), add the code + copy, and split the credits card so an allowance denial reads differently from an empty wallet. Also add `rate_limited` to the progress never-throttle set so the wait message can't be dropped.
2. **cowork** — the `rate_limited` card with a time-gated Retry, and the `PHASE_LABELS` entry for the waiting state.

### Correction — what budget exhaustion actually shows today

An earlier version of this description claimed exhaustion degrades to the provider-overloaded card. **That is wrong, and I verified it by building the real exception chain and asking `friendly_turn_error`:**

```
type=ProviderOverloadedError code='rate_limited'
provider_overloaded_info  -> ('rate_limited', 'sonnet')     <- the right code IS available
friendly_turn_error       -> ('token_limit', "You're out of credits. Add credits to keep working.")
```

The bare-status rule — `if status in (402, 429, 503) and _from_minds_gateway(host)` — fires **before** `provider_overloaded_info`, and the 429 is still in the `__cause__` chain, so it wins. **So on the exhaustion path this PR alone does not fix the message: the user still sees the out-of-credits card.**

Two consequences worth being explicit about:

- **cowork-server's change is required, not cosmetic.** It is what makes the failure honest when the wait doesn't succeed.
- **It should be small.** The exhaustion path already carries `code="rate_limited"` and `provider_overloaded_info` already surfaces it — the fix is to check a positively-identified `rate_limited` (reason header *or* body code) *before* the bare-status rule, rather than adding new plumbing.

**What this PR does deliver on its own:** the wait itself. Most velocity 429s now resolve inside the turn — no instant retries, no context growth, no SYSTEM notes in history, no card at all — so far fewer users reach the wrong message in the first place. The message on the residual is follow-up 1's job.

---

# Review round — what changed after the adversarial pass

A deep review on all three PRs found defects that execution proved and my bodies denied. Fixed in follow-up commits; recording the corrections here because several were claims *in this description*.

**Merge / delivery order is INVERTED** to `anton#349 → cowork#648 → cowork-server#309`. Under the original order the 90s wait ships **silent** — staging's renderer drops the ad-hoc `thought.progress` phase — which is the exact perceived-freeze the design exists to prevent. And landing the server's allowance split first sends a code the renderer has no branch for, dropping a user who just spent their free grant to a buttonless alert at the paid-conversion moment. cowork#648 is provably inert without #309, so the inversion costs nothing. **Enforce it at delivery** (UI/shell release before the PyPI publish), not at merge-to-staging.

**The ENG-1282 tripwire is intra-repo only.** I claimed it proves cross-repo enforcement. It doesn't: both code lists are hand-maintained *within* their own repos and neither reads the other, so it went red because the same-file literal needed updating in the same commit. There is also a hole it cannot see — `provider_overloaded_info` returns anton's `exc.code` verbatim, so any code anton invents reaches the renderer without being a `*_CODE` constant. Two tickets now lean on protection that isn't there.

**Test-count claims in this body are unreliable.** The `cloud_turn_process` subprocess tests are flaky on my machine and pass in CI; the review measured 0 failures where I reported 7. Trust CI, not the numbers I quoted.

### DIFF
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -107,9 +107,20 @@ def _raise_for_status_error(
             error_type=envelope.get("type"),
         ) from exc
 
+    # Body `code` in both dialects — the SDK may deliver the wire envelope
+    # unmodified, unlike the openai client which peels it.
+    _env = body.get("error") if isinstance(body.get("error"), dict) else {}
+    _body_code = body.get("code") or _env.get("code")
+    # ENG-1537: a session wait needs POSITIVE evidence of a velocity limit —
+    # our gateway names it on the reason header and in the body code. Anything
+    # else (a bare 429, a provider quota in an unrecognised dialect) keeps the
+    # old fail-fast path rather than burning the budget on a limit that may not
+    # clear.
+    _velocity = gate_reason == "rate_limited" or _body_code == "rate_limited"
     transient = classify_transient(
         exc.status_code, body, provider=provider, model=model,
         retry_after=retry_after_seconds(exc),
+        velocity_confirmed=_velocity,
     )
     if transient is not None:
         logger.warning(
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -27,6 +27,7 @@ from .provider import (
     Usage,
     classify_404,
     classify_transient,
+    retry_after_seconds,
     compute_context_pressure,
     wallet_denial_code,
     raise_on_empty_response,
@@ -106,11 +107,15 @@ def _raise_for_status_error(
             error_type=envelope.get("type"),
         ) from exc
 
-    transient = classify_transient(exc.status_code, body, provider=provider, model=model)
+    transient = classify_transient(
+        exc.status_code, body, provider=provider, model=model,
+        retry_after=retry_after_seconds(exc),
+    )
     if transient is not None:
         logger.warning(
-            "transient provider error (%s): status=%s body=%s",
-            transient.code, exc.status_code, scrub_credentials(str(exc.body))[:500],
+            "transient provider error (%s): status=%s retry_after=%s body=%s",
+            transient.code, exc.status_code, transient.retry_after,
+            scrub_credentials(str(exc.body))[:500],
         )
         raise transient from exc
 
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -183,9 +183,16 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
     # Retryable provider/infra failures — overload/api_error (incl. the mid-stream
     # HTTP-200 case), 5xx, or a plain 429 — get backed off and retried by the
     # session loop rather than surfacing an instant, misleading failure (ENG-673).
+    # ENG-1537: a session wait needs POSITIVE evidence of a velocity limit —
+    # our gateway names it on the reason header and in the body code. Anything
+    # else (a bare 429, a provider quota in an unrecognised dialect) keeps the
+    # old fail-fast path rather than burning the budget on a limit that may not
+    # clear.
+    _velocity = gate_reason == "rate_limited" or code == "rate_limited"
     transient = classify_transient(
         exc.status_code, body, provider="The model provider", model=model,
         retry_after=retry_after_seconds(exc),
+        velocity_confirmed=_velocity,
     )
     if transient is not None:
         logger.warning(
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -32,6 +32,7 @@ from .provider import (
     Usage,
     classify_404,
     classify_transient,
+    retry_after_seconds,
     compute_context_pressure,
     wallet_denial_code,
     raise_on_empty_response,
@@ -182,11 +183,15 @@ def _raise_for_status_error(exc: "openai.APIStatusError", model: str) -> NoRetur
     # Retryable provider/infra failures — overload/api_error (incl. the mid-stream
     # HTTP-200 case), 5xx, or a plain 429 — get backed off and retried by the
     # session loop rather than surfacing an instant, misleading failure (ENG-673).
-    transient = classify_transient(exc.status_code, body, provider="The model provider", model=model)
+    transient = classify_transient(
+        exc.status_code, body, provider="The model provider", model=model,
+        retry_after=retry_after_seconds(exc),
+    )
     if transient is not None:
         logger.warning(
-            "transient provider error (%s): status=%s body=%s",
-            transient.code, exc.status_code, scrub_credentials(str(exc.body))[:500],
+            "transient provider error (%s): status=%s retry_after=%s body=%s",
+            transient.code, exc.status_code, transient.retry_after,
+            scrub_credentials(str(exc.body))[:500],
         )
         raise transient from exc
 
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -521,8 +521,47 @@ def wallet_denial_code(body: Any) -> str | None:
     return code if isinstance(code, str) and code in _WALLET_DENIAL_CODES else None
 
 
+def retry_after_seconds(exc: BaseException) -> float | None:
+    """Seconds from a response's ``Retry-After`` header, or ``None`` (ENG-1537).
+
+    The MindsHub gateway sends this on the velocity 429 (``rate_limited``,
+    `minds/inference/errors.py`) as integer seconds — the only form we act on.
+    The HTTP-date form is legal but nothing in use emits it, and mis-reading a
+    date as a number would produce an absurd delay, so an unparseable value is
+    treated as absent: the caller then falls back to its own backoff curve.
+
+    Negative and non-finite values are dropped for the same reason. Zero is
+    meaningful ("retry now") and is preserved by the caller's ``> 0`` checks
+    behaving as "no hint", which is the same outcome.
+    """
+    resp = getattr(exc, "response", None)
+    headers = getattr(resp, "headers", None)
+    if headers is None:
+        headers = getattr(exc, "headers", None)
+    if headers is None:
+        return None
+    try:
+        raw = headers.get("retry-after") or headers.get("Retry-After")
+    except Exception:
+        return None
+    if raw is None:
+        return None
+    try:
+        secs = float(str(raw).strip())
+    except (TypeError, ValueError):
+        return None  # HTTP-date form, or junk
+    if secs != secs or secs in (float("inf"), float("-inf")) or secs < 0:
+        return None
+    return secs
+
+
 def classify_transient(
-    status_code: int | None, body: Any, *, provider: str = "", model: str = ""
+    status_code: int | None,
+    body: Any,
+    *,
+    provider: str = "",
+    model: str = "",
+    retry_after: float | None = None,
 ) -> "TransientProviderError | None":
     """Arm A of the transient classifier (see ENG-673): inspect an
     ``APIStatusError``'s status + body and return a ``TransientProviderError`` if
@@ -531,6 +570,10 @@ def classify_transient(
     Shared by both providers so the mapping can't drift. Call this only AFTER the
     permanent classifications (401 / 429-quota / 403 model-gate) have been ruled
     out — this decides between "retryable transient" and "generic unavailable".
+
+    ``retry_after`` is the parsed ``Retry-After`` hint (see
+    :func:`retry_after_seconds`); it is attached to the velocity-429 result so
+    the session waits the interval the server actually named (ENG-1537).
     """
     b = body if isinstance(body, dict) else {}
     # Two body dialects: Anthropic nests the error under `error` ({"error":
@@ -567,9 +610,20 @@ def classify_transient(
             return None
         if wallet_denial_code(b):
             return None
+        # session_backoff=True, unlike every other request-time status here
+        # (ENG-1537). The flag means "should the SESSION spend its budget on
+        # this?", and the SDK's own 2 retries fire seconds apart — the right
+        # answer for a 5xx that recovers instantly, and useless against a
+        # per-minute token ceiling. Leaving it False sent this down the
+        # count-based path, which re-issued the request TWICE with no delay and
+        # a recovery note appended each time: told "too many tokens per
+        # minute", we immediately sent more. This is the one failure class
+        # where waiting is both necessary and sufficient, so it waits — for the
+        # interval the server named, when it named one.
         return TransientProviderError(
             f"{provider or 'The model provider'} is rate-limiting requests.",
-            provider=provider, code="rate_limited", session_backoff=False, model=model,
+            provider=provider, code="rate_limited", session_backoff=True,
+            retry_after=retry_after, model=model,
         )
     return None
 
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -478,12 +478,17 @@ class ProviderOverloadedError(ConnectionError):
 
     def __init__(
         self, message: str, *, provider: str = "", model: str = "",
-        code: str = "provider_overloaded",
+        code: str = "provider_overloaded", retry_after: float | None = None,
     ) -> None:
         super().__init__(message)
         self.provider = provider
         self.model = model
         self.code = code
+        # Seconds the server said to wait, when it said so (ENG-1537). Set only
+        # on the rate-limit exhaustion path, where the wait was skipped or ran
+        # out and the card needs to name the interval rather than say
+        # "a moment".
+        self.retry_after = retry_after
 
 
 # Body ``error.type``/``error.code`` values that mean "provider is momentarily
@@ -562,6 +567,7 @@ def classify_transient(
     provider: str = "",
     model: str = "",
     retry_after: float | None = None,
+    velocity_confirmed: bool = False,
 ) -> "TransientProviderError | None":
     """Arm A of the transient classifier (see ENG-673): inspect an
     ``APIStatusError``'s status + body and return a ``TransientProviderError`` if
@@ -574,6 +580,16 @@ def classify_transient(
     ``retry_after`` is the parsed ``Retry-After`` hint (see
     :func:`retry_after_seconds`); it is attached to the velocity-429 result so
     the session waits the interval the server actually named (ENG-1537).
+
+    ``velocity_confirmed`` says the caller POSITIVELY identified a velocity
+    limit — our gateway's ``rate_limited`` reason header or body code. Only then
+    does the 429 earn a session wait. The absence of billing carriers is not
+    evidence of transience: both fail-fast guards below are string-exact, so a
+    provider whose quota denial uses a different dialect (Gemini sends an
+    INTEGER ``code`` with ``status: RESOURCE_EXHAUSTED``) slips past them and
+    would otherwise spend the whole budget waiting out a daily quota that resets
+    at midnight — then be told it is not a credits problem. Unconfirmed 429s
+    keep the pre-ENG-1537 behaviour: typed, honest, and failed fast.
     """
     b = body if isinstance(body, dict) else {}
     # Two body dialects: Anthropic nests the error under `error` ({"error":
@@ -622,8 +638,10 @@ def classify_transient(
         # interval the server named, when it named one.
         return TransientProviderError(
             f"{provider or 'The model provider'} is rate-limiting requests.",
-            provider=provider, code="rate_limited", session_backoff=True,
-            retry_after=retry_after, model=model,
+            provider=provider, code="rate_limited",
+            session_backoff=velocity_confirmed,
+            retry_after=retry_after if velocity_confirmed else None,
+            model=model,
         )
     return None
 
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -674,6 +674,23 @@ class ChatSession:
     # per-surface / config knob) and injectable in tests.
     _transient_budget_s: float = 30.0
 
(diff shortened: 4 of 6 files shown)


alecantu7 (commented):
## Adversarial self-review — head `99738360`, base `origin/staging` `f673c192`

Deep pass across all three ENG-1537 PRs. The core idea is right, and the ENG-1169 guard is stronger than the ticket claims. Two blocking findings here, four worth fixing, and several body claims that don't survive execution.

---

# Blocking

## 1 — a 1-second `Retry-After` is obeyed verbatim, ~90 times, in one turn

`session.py:1817-1824`. `_transient_backoff_delay` returns `min(float(retry_after), max_delay)` whenever a hint > 0 exists and **ignores `attempt` entirely** — no floor, no escalation. Nothing else bounds the loop: `_retry_count += 1` is at `:3315`, *after* the branch's `continue` at `:3289`, and the ENG-1286 spend ceiling counts tokens a gate-denied request never returns. The only bound is the 90 s budget.

Executed on the real `turn_stream` with a virtual clock:

```
hint=1  -> 91 attempts, 90 sleeps of 1.0s, 90.0s total
hint=2  -> 46      hint=5 -> 19      hint=30 -> 4
no hint -> 8 attempts, jittered 1.75/8.78/15.17/17.15/19.67   <- the healthy curve
origin/staging (session_backoff=False) -> 3 attempts, 0 sleeps
```

**The 1 s hint is the deployed default, not an edge case.** All three velocity controls map to reason `rate_limited` with `retry_after=decision.retry_after` (`auth/entitlements/views/inference_authorize.py:314-327`):

- **concurrency** → `concurrency_retry_after_seconds`, default **1** (`auth/common/app_settings.py:444-448`) — whose own description reads *"slots free on report, not on a clock"*, i.e. an explicitly fake end time, with `in_flight_ttl_seconds=300`;
- **RPM** → `retry = math.ceil(deficit / refill)` (`auth/common/rate_limiter.py:666`) at capacity 60 / refill 1.0 / cost 1.0, so `deficit ∈ (0,1]` → **always exactly 1**;
- **TPM** is the only control that produces large hints.

The only bound anywhere is `max(1, retry_after)` (`rate_limiter.py:655`) — a floor at exactly the worst case. BYOK behaves the same: an `anthropic.APIStatusError(429, headers={'retry-after':'1'})` gives `code='rate_limited' backoff=True retry_after=1.0`.

**Introduced by this PR.** On `origin/staging`, `retry_after` is declared (`provider.py:450/456`) and consumed (`session.py:1799`) but populated by **no construction site** — the verbatim branch was dead code. `provider.py:625` wires it live.

Two honest corrections to my own first read. Magnitude is **~9–30×**, not "135 vs 4": both SDKs already default to 2 internal retries on staging too. And it does **not** deepen the limit — `deny()` runs its refunds (`rate_limiter.py:652-655`). What it does is turn one denial into ~90 requests and ~90 SSE frames, because cowork-server#309 marks this notice never-throttled.

**Fix — executed, full suite green.** Never retry faster than our own curve:

```python
base = (2.0, 10.0, 18.0)
d = base[attempt] if attempt < len(base) else base[-1]
d = d * (0.8 + 0.4 * random.random())
if retry_after is not None and retry_after > 0:
    return min(max(float(retry_after), d), max_delay)
return d
```

hint=1 → 91→**7**; hint=2 → 46→**7**; hint=5 → 19→**7**; hint ≥ 30 and the no-hint path unchanged. Suite with the patch: **2049 passed, 28 skipped, 0 failed (156 s)** — identical to the unpatched baseline. `test_backoff_honors_retry_after` still passes.

## 2 — a hint above the cap does not card immediately, contradicting ticket How #3 and this body

Ticket How #3: *"If `Retry-After` itself exceeds the cap, don't wait: card immediately, naming the wait."* This body: *"A hint above the cap stops the wait and surfaces the card naming the interval."*

No such branch exists. `_rate_limit_max_delay_s` (`session.py:692`) is only ever the clamp argument at `:1824`. Executed:

```
hint=61   -> sleeps [60.0, 30.0], 90.0s, then "Waiting a moment and continuing should work"
hint=3600 -> identical
```

The copy is false for a 3600 s hint, and `ProviderOverloadedError.__init__` (`provider.py:469`) has no `retry_after` param, so anton cannot name the interval even if it wanted to. Cross-repo consequence: cowork-server recovers 3600 from the cause chain, so cowork#648 would gate Retry for an hour underneath copy telling the user to wait a moment.

---

# Worth fixing

## 3 — the whole 85-line `session.py` diff reverts with the suite byte-identically green

**2049 passed / 28 skipped** in both directions, md5-verified source swaps (`a9b7914a` ↔ merge-base `d5344b26`, `grep -c ENG-1537` = 0). Same for the single-line mutation `_rate_limited = False`.

Worth being precise about what that does and doesn't mean, because it moderates the severity: **the wait itself is not at risk.** A probe asserting "waits the interval, resumes the same step, no history note" passes on merge-base too — that behaviour comes from `session_backoff=True` at `provider.py:625`, which *is* covered (flipping it to `False` reddens 3 tests). Instant-retry cannot silently come back.

Genuinely uncovered, each mutation-proven:

1. `_rate_limit_max_delay_s = 60.0` / the `max_delay` passthrough (`:692, :1808, :3256`) — reverting silently clamps a 45 s hint to 30 s, i.e. retrying into a window still open. The one with live behavioural weight.
2. `_rate_limit_budget_s = 90.0` accounted separately from the incident budget (`:686, :3247-3255`).
3. `StreamTaskProgress(phase="rate_limited", …)` (`:3280-3284`) — the real cross-repo coupling. cowork-server's never-throttle set and `PHASE_LABELS`, and cowork's reducer branch, all key on an event nothing in anton pins.
4. `ProviderOverloadedError(code="rate_limited")` on exhaustion (`:3301-3311`) — lower stakes than it looks: cowork-server's `test_exhausted_rate_limit_wait_keeps_its_code_over_the_bare_status_rule` passes with `code="provider_overloaded"` anyway.

Remedy executed: 5 tests in `test_transient_retry.py`'s style → 5 pass on HEAD, **4 fail on merge-base**, suite 2054 passed. Drop the fifth, which isn't load-bearing.

## 4 — the 90 s budget is a wall-clock deadline, not a wait budget

`session.py:3253-3255`. `_rate_limit_deadline = now + 90` is set on the first velocity 429 and never reset, so everything the turn does between failures — tool rounds, verifier continuations, compaction — is charged against it. Executed, varying work-time per attempt at hint=30:

```
work=0s   -> 4 attempts, sleeps [30,30,30]
work=45s  -> 3 attempts, sleeps [30,15]
work=120s -> 2 attempts, sleeps [30]   -> card, wall=270s
```

This is the dominant shape for the cohort this ticket targets: ENG-1122 (cache reads at full weight) is what puts long-context sessions near the ceiling for a whole turn, so denials recur over minutes rather than back to back. The comment's *"90s covers a full window rolling over"* holds only for consecutive failures.

## 5 — a hard BYOK quota in an unrecognised dialect now burns 90 s and is told "this isn't a credits problem"

Both fail-fast guards are string-exact: `etype == "insufficient_quota"`, and `wallet_denial_code(b)` requires `isinstance(code, str)` (`provider.py:520`). Gemini's OpenAI-compat 429 carries `{"error":{"code":429,"status":"RESOURCE_EXHAUSTED"}}` — an **integer** `code` — so both miss.

Executed: `TransientProviderError code=rate_limited session_backoff=True`, costing 8 retried calls over 90 s; then through #309 it reads `("rate_limited", "…this isn't a credits problem")` where staging gave `("provider_overloaded", …)`. Gemini's daily free quota resets at midnight PT, so 90 s can never clear it. **Worsened, not introduced.**

Finding 1's escalating curve does *not* fix this. Gating `session_backoff` on a **positively identified** velocity signal — the `X-MindsHub-Reason` header or our own body code — rather than on the *absence* of billing carriers, does.

## 6 — the shared attempt counter, and one stale claim

- `_transient_attempt` (`session.py:3270`) is a single counter, so a rate-limit wait advances the *incident* curve. Executed: alternating failures → the incident path's first wait was 8.02 s (index 1) instead of ~2 s. Worst case for one turn is 30 s + 90 s = **120 s** of sleep, not 90.
- The anthropic mapper's `retry_after=` passthrough (`anton/core/llm/anthropic.py:112`) has no coverage — deleting it leaves all 87 mapper tests passing. The control (same deletion in `openai.py`) correctly reddens, and the `_anthropic_sdk_error` fixture already exists.

---

# Body corrections

- **"Full suite: 2003 passed. 7 failures, all pre-existing"** — did not reproduce. Two independent runs on this head gave **2049 passed, 28 skipped, 0 failed**. Environment-specific, but the number in the body is not what a reviewer will see.
- **"A hint above the cap stops the wait and surfaces the card naming the interval"** — see finding 2. It doesn't, and it can't name the interval.
- **Ticket How #5** (*"skip the retry-exhausted summarize call when the failure is a rate limit"*) **is delivered**, structurally — on the exhaustion path `plan_stream.await_count == 0`, because the branch raises before reaching `_retry_count`. Worth adding to the body; right now nothing records that the item shipped.
- **"no history note"** is not literally true. When the 429 interrupts a dangling `tool_use`, `_seal_dangling_tool_uses` writes a synthetic `tool_result` row (`"[interrupted by a transient provider error…]"`, executed). `test_turn_transient_retry_does_not_inject_recovery_note` passes **vacuously** — its fake never leaves one dangling. The substantive claim (no *"adjust your approach"* recovery note) does hold.
- The honest note about the 6 billing-guard assertions is right, and understates one thing: the **two 402 rows are vacuous** — `classify_transient` returns `None` for any 402 regardless of body, so they pass with every guard deleted. Only the two 429 rows are load-bearing (confirmed: exactly those went red).

# Verified and correct

The **ENG-1169 guard is real defence in depth**, and stronger than the ticket describes — three layers, all mutation-tested:

1. the mapper wallet branch — disabling it reddens 4 tests;
2. `classify_transient`'s `if wallet_denial_code(b): return None` (`provider.py:612`) — deleting it reddens 3;
3. **unmentioned anywhere:** `session.py:3222` fails fast on `TokenLimitExceeded` *before* the backoff check, and `issubclass(TokenLimitExceeded, TransientProviderError)` is `False` (executed) — so a billing denial structurally cannot reach the wait.

Executed with `Retry-After: 30` deliberately present: `wallet_empty` 402 and `included_allowance_exhausted` 429 both → `TokenLimitExceeded`, no wait, via header carrier, body carrier, and both together.

**"No change to 5xx incident handling" holds** — executed: an `overloaded_error` still uses the 30 s budget, 30 s cap, 3 retries, ~29.5 s wall, identical to pre-PR. The deadlines are separate; only the attempt counter is shared (finding 6).

**CodeQL is not a blocker.** The two alerts at `openai.py:194` / `anthropic.py:118` are the *same rule already dismissed on `refs/heads/staging`* at `:189` / `:113`; this PR reflowed the call and shifted the sink five lines. `mergeStateStatus` is UNSTABLE rather than BLOCKED and the only ruleset (`staging-freeze`) is `"enforcement":"disabled"`. `transient.retry_after` is a float parsed from a header, and the body arg was already `scrub_credentials`-filtered. Worth dismissing on the same grounds rather than leaving the PR red.

# Merge order

**Invert it: this PR → cowork#648 → cowork-server#309.** Reasons are in the other two PRs, but the part that belongs here: under the currently stated order the 90 s wait ships **silent** — staging's renderer drops the ad-hoc `thought.progress` phase (`return state`), so the pause is invisible until #648 lands. That is exactly the *"a silent 90s pause is indistinguishable from a hang"* failure this body warns about. cowork#648 is provably inert without #309, so the inversion costs nothing.

# Not verified

The real production distribution of `Retry-After` — the auth-side arithmetic above is traced and executed, but never observed live. A Langfuse histogram of `Retry-After` on `rate_limited` 429s, split by which control denied, is what decides whether finding 1 is routine or occasional. Also unverified: the **239 denied calls / 5 users / 30d** figure this body repeats — I did not re-measure it — and Gemini's live 429 body, for which no captured fixture exists in `anton/tests/` (finding 5's *mechanism* is certain; its Gemini instantiation rests on the documented shape).

alecantu7 (commented):
## Second pass — head `f37f53d8`

Verified by execution rather than from the commit message. **Everything I raised is fixed**, one of them with a better mechanism than I proposed. Two things left, both small.

### B1 — fixed

The hint is now a floor on the curve rather than a replacement for it. Executed against the real `_transient_backoff_delay` inside the 90 s budget:

```
hint=1   ->  7 attempts   (was 91)        hint=30  -> 3   (unchanged)
hint=2   ->  7            (was 46)        hint=45  -> 2
hint=5   ->  7            (was 19)        no hint  -> 7   (unchanged)
incident branch (30s budget, cap 30): 3 attempts, 29.8s — unchanged
```

The docstring now carries the arithmetic that makes the small hints the *common* case rather than an edge one (RPM's `ceil(deficit / refill)` always landing on 1, concurrency's flat 1 that its own config calls a fake end time). That is the part that stops this being re-simplified later.

### W6 — fixed, and with the better mechanism

`velocity_confirmed` gates the wait on **positive** evidence — the gateway's reason header or body code — rather than on the *absence* of billing carriers. That is the right shape: the absence of a carrier was never evidence of transience, only evidence that our two string-exact guards didn't recognise the dialect.

```
gateway rate_limited (header)         -> code=rate_limited backoff=True  retry_after=30
gateway rate_limited (body code)      -> code=rate_limited backoff=True  retry_after=30
Gemini RESOURCE_EXHAUSTED (int code)  -> code=rate_limited backoff=False retry_after=None
bare 429, no carriers                 -> code=rate_limited backoff=False
```

So a BYOK daily quota that resets at midnight PT now fails fast instead of spending 90 s waiting for something that cannot clear.

### W4 and W5 — fixed

A hint above the cap raises immediately and names the interval, and `ProviderOverloadedError` gained the `retry_after` field that made naming it possible. The 90 s budget became `_rate_limit_slept` — accumulated sleep rather than a wall-clock deadline — so work between denials is no longer charged against the wait budget. Both were things the ticket asked for and the earlier body claimed while they were absent.

### ENG-1169 — still holds

All four dialects still fail fast with a `Retry-After: 30` deliberately present:

```
wallet_empty 402 / nested          -> None (fail fast)
allowance 429 / nested             -> None (fail fast)
```

Mutations: reverting the curve reddens `test_a_hint_never_retries_faster_than_the_curve`; reverting the velocity gate reddens `test_anthropic_mapper_does_not_wait_on_an_unconfirmed_429` and one more. **Full suite: 2057 passed, 28 skipped, 0 failed**, with 8 new tests.

---

## Two left

### 1 — W4's fix has zero coverage, and it is the item most likely to regress

Disabling the above-cap branch:

```python
- if _rate_limited and _hint is not None and _hint > max_delay:
+ if False:
```

leaves **2057 passed** — the whole suite, not just the touched files. `test_exhaustion_carries_the_rate_limited_code_and_the_interval` covers the *budget-exhausted* path, which is a different branch.

Worth pinning specifically because of this item's history: it is ticket How #3, and the PR body asserted it shipped for a whole revision while no such branch existed. An untested branch with a body claiming it works is exactly the state it was in before.

A test driving a confirmed velocity 429 with `retry_after=3600` and asserting the turn raises immediately — zero sleeps, `code="rate_limited"`, `retry_after=3600`, and `"3600s"` in the message — would do it.

### 2 — the shared attempt counter is unaddressed (low)

`_transient_attempt` is still one counter (`session.py:3142`, `:3302`), so a rate-limit wait advances the incident curve and vice versa. Executed with interleaved rate-limit and incident failures:

```
[RL 1.8, INC 10.2, RL 17.1, INC 2.8, RL 18.9, INC budget-out, RL 14.9, …]
total slept in one turn = 100.5s   (rate-limit share 87.6s)
```

Defensible as-is — the budgets do stay separate, which was the point — but the "90s" the comments quote is not the ceiling for a turn that hits both classes. One clause, or a second counter.

---

CodeQL is still red for the two pre-existing alerts already dismissed on `refs/heads/staging`; nothing in this delta touches them.

pnewsam (approved):
## Code Review — anton#349 (ENG-1537, in-turn wait)

**Verdict: APPROVE**

Reviewed the full diff at `8fe7f01`. No blocking or major defects.

Verified:
- The retry loop terminates: `_rate_limit_slept += delay` bounds the 90s budget, and the exhaustion raise is reached once the remaining budget drops below 0.5s.
- Rate-limit and incident paths have independent budgets *and* attempt indices, so an interleaved turn can't jump the incident curve.
- The above-cap immediate-card path (`_hint > max_delay`) raises with `code="rate_limited"` + the interval instead of absorbing a multi-minute stall.
- Both mappers define `gate_reason`/`code` before the new velocity check, and only the velocity-429 branch carries `retry_after` — so the new backoff floor (`min(max(retry_after, curve), max_delay)`) never alters incident behavior, since incidents never carry a hint.

**CI note** — the two CodeQL "clear-text logging of sensitive data" high alerts (`openai.py:201`, `anthropic.py:129`) are **pre-existing sanitized false positives**, not introduced here. The flagged sink `scrub_credentials(str(exc.body))[:500]` is byte-identical to `origin/staging` (verified against merge-base); this PR only added `transient.retry_after` (a float) and a `retry_after=%s` format token on the same line, which re-attributes the old alert to this diff. `scrub_credentials` is a real sanitizer (redacts DS_* secrets, provider keys, `mdb_`/`sk-`/`AIza` shapes) that CodeQL doesn't model. Recommend dismissing both to unblock the merge; no code change warranted.

Delivery-order note: agree with landing/publishing this first — most velocity 429s then never reach a card at all.

PR #353, opened 2026-08-14
ENG-1617: flush in-flight analytics sends at exit so short-lived hosts stop losing events

Dav1dF:
Linear: [ENG-1617](https://linear.app/mindsdb/issue/ENG-1617/analytics-events-race-interpreter-shutdown-so-the-cloud-path-can)

Found by @lucas-koontz reviewing #345. Not introduced there — the hazard is in the module's design and predates it — but ENG-1495 is what makes it consequential, because `turn_completed` is the first event whose loss costs something.

## The problem

`anton/analytics.py` sends every event from a daemon thread. Daemon threads are killed at interpreter shutdown without running to completion, and the module docstring listed that as a *guarantee*. On a long-lived host it is harmless. On `python -m anton.cloud_turn`, which runs one turn and returns, it is total loss.

Measured on 2026-08-14, on CPython 3.14 and 3.12. A daemon thread started immediately before `main()` returns does not finish at any duration:

```
daemon needing   1ms  ->  did NOT finish
daemon needing   5ms  ->  did NOT finish
daemon needing  50ms  ->  did NOT finish
daemon needing 200ms  ->  did NOT finish
```

So this is not a race the cloud path sometimes loses. On that entry point it loses every time, against a ~300ms POST.

Worth stating plainly: #345's end-to-end verification ran from a long-lived CLI process, where the thread always has time. The sink was verified on the host least affected and unverified on the host that produces most turns.

## The fix

Module-owned rather than per-host:

* `_spawn` registers each send before its thread starts, and clears the registration in a `finally`.
* `flush(timeout)` waits for the outstanding ones, with the timeout as a **total** budget rather than per-send, so a turn emitting several events cannot hold exit for a multiple of it.
* `atexit.register(flush)` at import.

The ticket offered a targeted `join()` in `cloud_turn/__main__.py` as the smaller alternative. I took the module-owned version because the smaller one leaves the next short-lived entry point to rediscover this, which is how this one arrived.

## Why one second

The ticket asked for a measured basis rather than a guess. Real POST to `us.i.posthog.com/capture/`, 12 samples:

| | |
| -- | -- |
| cold, DNS + TLS handshake | 351 ms |
| median | 283 ms |
| max | 296 ms |

urllib opens a fresh connection per `urlopen`, so a one-shot process pays the cold number every time. 1s is roughly 3x the median with headroom for a slower network, and stays under `_TIMEOUT`'s 3s, so a send that has genuinely hung costs exit 1s rather than 3s.

It is a ceiling, not a cost. With nothing outstanding, `flush` returns in **0.0002 ms median over 200 calls** — which is the interactive CLI case, where the event went out long before the user quit. That is the ticket's "process exit is not measurably slower for an interactive CLI session".

The probe used a deliberately bogus api_key, so it wrote no real rows into 424726.

## Tests

10 added, 32 in the file. Every new assertion mutation-verified: removing the atexit registration, removing the wait, or dropping the leak guard each fails a specific test and nothing else.

The regression test runs a **real child process** against a real loopback endpoint, because the bug only exists at interpreter shutdown and no in-process stub reproduces it. Its companion, `test_short_lived_process_loses_its_event_without_the_flush`, asserts the pre-fix loss — so if the atexit hook ever stops being the thing doing the work, or CPython changes when it kills daemon threads, that surfaces as a failure instead of a silently vacuous pass.

Also covered: registration happens before the thread starts (no invisible-in-flight window), the registration clears on success, on a raising sender, and when the thread cannot start at all. That last one matters because a stale registration would turn one failed spawn into a permanent 1s tax on every later exit.

Full suite: 2149 passed, 28 skipped, 0 failed. Green on 3.12 as well as 3.14.

## Also changed

`docs/configure/analytics.md` said events "never block Anton". Exit can now wait up to a second, so it says so, including that switching analytics off removes the wait along with the send.

## What this does not close

Verifying that a real cloud turn produces its `turn_completed` row in project 424726 needs the anton staging to main to PyPI to cowork-server `uv.lock` bump to desktop chain to run first — the same constraint #345 wrote up in its Step 0. This PR fixes the mechanism and proves it at the process level; the cloud-path confirmation is a post-release check.

Until it lands, per-turn cost data stays incomplete by an unknown amount on the host producing most turns, which matters for ENG-1286.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -401,11 +401,16 @@ def flush(timeout: float = _FLUSH_TIMEOUT) -> None:
             outstanding = list(_pending)
         for done in outstanding:
             remaining = deadline - time.monotonic()
-            if remaining <= 0:
+            # Both halves of the budget check matter. `remaining <= 0` catches a
+            # budget already spent by earlier sends; `not done.wait(...)` catches
+            # it running out *inside* the wait, which is what happens whenever
+            # there is a single outstanding send — the common case. Checking only
+            # the first left exactly that case exiting the loop silently, so the
+            # loss it exists to record was the one it never recorded.
+            if remaining <= 0 or not done.wait(remaining):
                 logger.debug("analytics flush budget exhausted, %d send(s) abandoned",
                              sum(1 for d in outstanding if not d.is_set()))
                 return
-            done.wait(remaining)
     except Exception:
         logger.debug("analytics flush failed", exc_info=True)
 
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -58,7 +58,9 @@ paths, credentials, hostnames, or email addresses.
 Guarantees:
   • Never blocks the caller.
   • Never raises — all exceptions are silently swallowed.
-  • Daemon threads die automatically when the process exits.
+  • Exit waits briefly for sends already in flight (see ``flush``), then
+    gives up.  Daemon threads are killed at interpreter shutdown, so
+    without that wait a send started just before exit is simply lost.
 
 Machine fingerprint
 ===================
@@ -84,6 +86,7 @@ enough to be a readable query parameter.
 
 from __future__ import annotations
 
+import atexit
 import hashlib
 import json
 import logging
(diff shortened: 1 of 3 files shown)


Dav1dF:

anton/analytics.py
```diff
@@ -301,6 +350,78 @@ def _fire_posthog(url: str, body: bytes) -> None:
...
+            # budget already spent by earlier sends; `not done.wait(...)` catches
+            # it running out *inside* the wait, which is what happens whenever
+            # there is a single outstanding send — the common case. Checking only
+            # the first left exactly that case exiting the loop silently, so the
+            # loss it exists to record was the one it never recorded.
+            if remaining <= 0 or not done.wait(remaining):
+                logger.debug("analytics flush budget exhausted, %d send(s) abandoned",
+                             sum(1 for d in outstanding if not d.is_set()))
+                return
+    except Exception:
```
lucas-koontz, line 414: issue: Ctrl-C during the exit wait prints a traceback. `done.wait()` raises `KeyboardInterrupt` on the main thread, `except Exception` doesn't catch it, and atexit prints whatever escapes a callback. So a user who quits and then gets impatient inside the 1s window sees a stack trace from a module whose docstring says it never raises.

```suggestion
    except (Exception, KeyboardInterrupt):
```

anton/analytics.py
```diff
@@ -301,6 +350,78 @@ def _fire_posthog(url: str, body: bytes) -> None:
...
+        logger.debug("analytics flush failed", exc_info=True)
+
+
+# Runs during interpreter shutdown, while daemon threads are still alive and
+# schedulable — CPython calls atexit callbacks before tearing down the thread
+# state that kills them. Registering at import is what makes this the module's
+# responsibility rather than each host's.
+atexit.register(flush)
```
lucas-koontz, line 422: question: Does the pod's process actually get its second? scratchpad-controller stops reading at the terminal event, and `stream_anton_turn_in_live_pod`'s `finally` sets `stop`, so the exec websocket closes within one `ws.update(timeout=1)` cycle and kubelet kills `python -m anton.cloud_turn`. That teardown and this flush budget are the same order of magnitude. (`scratchpad-controller` `src/scratchpad_controller/live_pod.py:373`)

tests/test_analytics.py
```diff
@@ -511,3 +523,264 @@ def start(self):
...
+    finally:
+        release.set()
+
+
+def test_flush_logs_when_the_only_pending_send_is_abandoned(caplog):
+    """One outstanding send is the common case, and it is the case an earlier
+    version logged nothing for: the budget ran out *inside* `Event.wait`, so the
+    `remaining <= 0` branch was never reached and the loop just ended. The loss
+    this line exists to record was the one it never recorded."""
+    release = threading.Event()
```
lucas-koontz, line 668: optional: This asserts `1 send(s) abandoned`, but the four `flush` tests share the real module-level `_pending`, and the one above releases its thread in a `finally` without waiting for it to clear. If that thread hasn't been scheduled yet the count is 2 and this fails. Reset the set first.

```suggestion
    analytics._pending.clear()
    release = threading.Event()
```

lucas-koontz (approved):
Read the whole diff at `793773f6`. The negative test that pins the pre-fix loss is what makes the regression test evidence rather than decoration, and the 1s number has an actual measurement behind it.

Nothing blocks. One real defect: Ctrl-C during the exit wait escapes the guard and prints a traceback. One question about whether the pod's process survives long enough to use the wait at all, which is the same thing your "what this does not close" section is about. One test-isolation note.

The first Done-when still needs the release chain, and you've said so.

PR #357, opened 2026-08-14
ENG-1632: a deterministic verifier denial latches silently on first occurrence

alecantu7:
## What

The completion verifier's verdict loop gets a typed clause for **deterministic denials** — a wallet 402 / allowance 429 (`TokenLimitExceeded`, code-exact per ENG-1169) or a 404'd/403'd model (`ModelUnavailableError`). A denied verdict now latches on the **first** occurrence and ends the turn quietly on the already-streamed reply: no `SYSTEM: The task-completion check failed to run (internal error)` history injection, no diagnosis stream, no "Something went wrong — checking in with you...".

Linear: [ENG-1632](https://linear.app/mindsdb/issue/ENG-1632) — full root cause, production-trace evidence, and the layer map live on the ticket.

## Why

A denial recurs on every retry by construction, yet it was handled as a generic hard failure: a full-history "internal error" handback on the first turn, latch only at two — and cowork-server rebuilds the ChatSession per message, so every message paid again. In production (14 days): **208 aux-call wallet-402s across 33 users**, each surfaced as an internal error and an apology **for turns whose work had succeeded**. Verified in-session (amdou440): work completes on the free model → verifier 402s on a pinned paid model → the user is told an internal check failed — every turn.

## How

- New `_DENIED_VERDICT_ERRORS` clause, ordered **before** `_TRANSIENT_VERDICT_ERRORS` — the ordering is load-bearing and commented on the constant: `ModelUnavailableError` subclasses `ConnectionError → OSError`, so the transient clause used to swallow it (diagnose every turn, never latch — this PR deliberately changes that behavior for model-404/403s too); `TokenLimitExceeded` fell to the generic "hard" handler.
- Handled via a `"denied"` sentinel in the post-loop `else` (a bare `break` in the except clause only exits the token-budget `for` and falls into the diagnosis block).
- Deliberately **silent, never auto-upgraded** to the planning model: a missing check beats a surprise bill. The existing 10-turn re-probe keeps recovery automatic — top up and the next re-probe restores verification; still-denied re-probes stay silently latched.
- The skip log names the latch reason (new `_verifier_latch_reason`) instead of printing "latched after 0 hard failures" for a denied latch. `_verifier_hard_failures` stays clean, so a later genuine capability failure still gets its one-per-session honest diagnosis.
- Safe by verification: `TokenLimitExceeded` is billing-only (context overflow is `ContextOverflowError`), propagates out of `generate_object_code` unwrapped, is never SDK-retried on 402, and velocity 429s map to `TransientProviderError` — they cannot land in the denied set.

## Coordination (no stacking)

Independent of the cowork-server and cowork PRs; merge in any order. Should ride the **same weekly train** as the cowork-server resolution fix (anton reaches desktop only via the cowork-server bump — ENG-1094).

## Test plan

- New in `tests/test_verifier_failsafe.py`: parametrized over wallet-402 / allowance-429 / model-404 — asserts **on history content and streamed events** (no SYSTEM injection, no diagnosis plan call, no progress line), latch-on-call-one, clean hard counter, single announcement, reason-bearing skip logs; plus the denied re-probe staying latched and silent.
- **Mutation-verified**: all four new tests fail on unfixed `session.py`, pass with the fix.
- Full suite on this base: green except the pre-existing `test_scratchpad.py::TestSessionPersistence::test_namespace_survives_restart`, which fails identically on pristine origin.

## Security pass

Reviewed the diff: the denied log line uses `_safe_error_detail` (type name only — no exception message, which can carry conversation or wallet content, per ENG-1081); no new input surface, no credential handling, nothing user-visible added.

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -109,7 +109,10 @@ _TIMEOUT = 3  # seconds
 # cowork's renderer register (``src/renderer/cowork/lib/analytics.js``), so
 # both clients read the same way.
 #
-#   turn_completed   ended_by, tokens_total, input_tokens, output_tokens,
+#   turn_completed   ended_by, verification_skipped ("true" when the turn
+#                    completed without a verifier verdict — denied/latched,
+#                    ENG-1632 — so honest-stop denominators can exclude it),
+#                    tokens_total, input_tokens, output_tokens,
 #                    cache_read_tokens, cache_creation_tokens, llm_calls,
 #                    rounds, continuations, peak_context_tokens, duration_ms,
 #                    {planning,coding,router}_{model,tokens,calls},
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2286,11 +2286,13 @@ class ChatSession:
         except Exception:  # pragma: no cover - defensive
             _root_cause_fields = {}
         logger.info(
-            "turn_cost session=%s turn=%d ended_by=%s tokens_total=%d "
+            "turn_cost session=%s turn=%d ended_by=%s verification_skipped=%s "
+            "tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
             "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d "
             "by_role=%s %s",
-            self._session_id, turn_index, tc.ended_by, tc.total_tokens,
+            self._session_id, turn_index, tc.ended_by,
+            str(tc.verification_skipped).lower(), tc.total_tokens,
             tc.input_tokens, tc.output_tokens, tc.cache_read_tokens,
             tc.cache_creation_tokens, tc.llm_calls, tc.rounds,
             tc.continuations, tc.peak_context_tokens, tc.duration_ms,
@@ -2332,6 +2334,10 @@ class ChatSession:
                 # classification is broken.
                 **_root_cause_fields,
                 ended_by=tc.ended_by,
+                # A completed-but-unverified turn (denied/latched verifier,
+                # ENG-1632) must be excludable from the honest-stop
+                # denominator without a per-conversation Langfuse hop.
+                verification_skipped=str(tc.verification_skipped).lower(),
                 tokens_total=str(tc.total_tokens),
                 input_tokens=str(tc.input_tokens),
                 output_tokens=str(tc.output_tokens),
@@ -4529,6 +4535,14 @@ class ChatSession:
                         _VERIFIER_LATCH_REPROBE_TURNS, continuation,
                         self._max_continuations, tool_round,
                     )
+                    # Stamp the books: this turn books ended_by="completed"
+                    # but was NOT verified — without the flag it is
+                    # byte-identical in analytics to a verified pass and
+                    # silently joins the honest-stop denominator (ENG-1632
+                    # review). Stamped here, never read from latch state at
+                    # emit (late finalizers would see a later turn's state).
+                    if self._turn_cost is not None:
+                        self._turn_cost.verification_skipped = True
                     break
                 _verifier_log.info(
                     "completion-verifier re-probing after %d skipped verifications",
@@ -4652,6 +4666,9 @@ class ChatSession:
                         "verification this session; turn ends on the "
                         "already-streamed reply"
                     )
+                    # Unverified turn — see the latched-skip stamp above.
+                    if self._turn_cost is not None:
+                        self._turn_cost.verification_skipped = True
                     break
                 if verdict_failure == "hard":
                     self._verifier_hard_failures += 1
@@ -4664,6 +4681,9 @@ class ChatSession:
                         _verifier_log.info(
                             "completion-verifier re-probe failed — staying latched"
                         )
+                        # Unverified turn — see the latched-skip stamp above.
+                        if self._turn_cost is not None:
+                            self._turn_cost.verification_skipped = True
                         break
                     if self._verifier_hard_failures >= 2:
                         # Second hard failure in a row: the first could have been
@@ -4693,6 +4713,9 @@ class ChatSession:
                             "verification this session",
                             self._verifier_hard_failures,
                         )
+                        # Unverified turn — see the latched-skip stamp above.
+                        if self._turn_cost is not None:
+                            self._turn_cost.verification_skipped = True
                         break
                 # The verifier call failed on every budget it was given —
                 # truncated past the last retry, an unusable tool call, or a
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -313,6 +313,29 @@ _TRANSIENT_VERDICT_ERRORS: tuple[type[BaseException], ...] = (
     httpx.TransportError,
 )
 
+# Verdict-call failures that are DETERMINISTIC DENIALS: the provider will give
+# the identical answer on every retry this session — a wallet that can't pay
+# for the verifier's model (TokenLimitExceeded: 402 ``wallet_empty`` / 429
+# ``included_allowance_exhausted``, both code-exact per ENG-1169; velocity 429s
+# map to TransientProviderError and never land here) or a model id that doesn't
+# resolve for this key (ModelUnavailableError: 404 model-not-found / 403
+# model-access-denied). These latch on the FIRST occurrence and stay silent:
+# the turn's work already succeeded, and "the task-completion check failed
+# (internal error)" is a lie when the check was merely priced out (ENG-1632 —
+# 208 aux-call wallet-402s across 33 users in 14 days, every one surfaced to
+# the user as an internal error and an apology).
+#
+# ORDER MATTERS in the verdict loop: this clause must precede
+# _TRANSIENT_VERDICT_ERRORS. ModelUnavailableError subclasses ConnectionError →
+# OSError, so the transient clause would otherwise swallow it as a blip and
+# re-diagnose every turn forever (its pre-ENG-1632 behavior). TokenLimitExceeded
+# is a plain Exception and would fall to the generic "hard" handler instead —
+# wrong the other way: a full-history diagnosis on turn one, latch only at two.
+_DENIED_VERDICT_ERRORS: tuple[type[BaseException], ...] = (
+    TokenLimitExceeded,
+    ModelUnavailableError,
+)
+
 # Turns a latched session skips before spending one verdict call to see whether
 # the cause has gone away (user switched model, gateway fix shipped). Without a
 # re-probe the latch is permanent for the session and "reset on a successful
@@ -761,6 +784,10 @@ class ChatSession:
         self._verifier_hard_failures = 0
         self._verifier_latched = False
         self._verifier_latch_skips = 0
+        # Why the latch is set — "hard" (capability, ENG-1095) or "denied"
+        # (billing/model access, ENG-1632) — so the skip log names the real
+        # cause instead of reporting "0 hard failures" for a denied latch.
+        self._verifier_latch_reason = ""
         self._context_pressure_threshold = s.context_pressure_threshold
         self._max_consecutive_errors = s.max_consecutive_errors
         self._resilience_nudge_at = s.resilience_nudge_at
@@ -4490,11 +4517,14 @@ class ChatSession:
                 self._verifier_latch_skips += 1
                 if self._verifier_latch_skips < _VERIFIER_LATCH_REPROBE_TURNS:
                     _verifier_log.info(
-                        "completion-verifier skipped — latched after %d hard "
-                        "failures with no successful verdict between them "
+                        "completion-verifier skipped — latched (%s) "
                         "(skip %d/%d before re-probe); "
                         "continuation=%d/%d tool_rounds=%d",
-                        self._verifier_hard_failures, self._verifier_latch_skips,
+                        "deterministic denial: billing or model access"
+                        if self._verifier_latch_reason == "denied"
+                        else f"{self._verifier_hard_failures} hard failures "
+                        "with no successful verdict between them",
+                        self._verifier_latch_skips,
                         _VERIFIER_LATCH_REPROBE_TURNS, continuation,
                         self._max_continuations, tool_round,
                     )
@@ -4542,6 +4572,17 @@ class ChatSession:
                     )
                     if not retrying:
                         break
+                except _DENIED_VERDICT_ERRORS as exc:
+                    # Deterministic denial — see _DENIED_VERDICT_ERRORS (and the
+                    # ordering note there: this clause must stay ahead of the
+                    # transient tuple). Retrying, diagnosing, or telling the
+                    # user buys nothing: handled below by latching silently.
+                    verdict_failure = "denied"
+                    _verifier_log.info(
+                        "completion-verifier verdict=DENIED budget=%d error=%s",
+                        budget, _safe_error_detail(exc),
+                    )
+                    break
                 except _TRANSIENT_VERDICT_ERRORS as exc:
                     # Explicitly NOT a latch candidate. The latch is for a model
                     # that cannot produce a verdict at all (kimi-K3 rejecting the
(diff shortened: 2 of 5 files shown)


alecantu7 (commented):
## Self-review — head `06f9bc75`, base `origin/staging` (verified fresh against the API)

The two traps the ticket named for this layer are both real and both pinned — I mutated rather than read. `tests/test_verifier_failsafe.py` → **21 passed**.

| Mutation | Result |
| -- | -- |
| Move the denied clause **after** `_TRANSIENT_VERDICT_ERRORS` | `test_denied_verdict_latches_silently_on_first_occurrence[model-404]` **fails** |
| Drop the `verdict_failure = "denied"` sentinel (bare `break`) | **4 tests fail**, including `test_denied_reprobe_stays_latched_and_silent` |

The first is worth noting precisely: only the **`ModelUnavailableError`** case catches the reordering, and that's correct rather than a gap — `TokenLimitExceeded` is a plain `Exception`, so moving the clause doesn't change its routing. The parametrised 404 case is the one carrying the ordering contract, which is exactly the trap the comment warns about.

Also verified by reading the path: the `denied` branch is handled in the `else` **before** the hard-failure accounting, sets both `_verifier_latched` and `_verifier_latch_reason`, logs its own wording, and `break`s the outer `while` — so no diagnosis stream and no SYSTEM injection. The skip log no longer prints "latched after 0 hard failures" for a denied latch.

---

### One thing to fix — the comment's numbers don't match the ticket's

`session.py`:

> ENG-1632 — **208** aux-call wallet-402s across **33** users in 14 days

ENG-1632's description says **296 wallet-402s across 39 distinct users**. Both are probably right over different populations (aux-only vs all roles), but neither the comment nor the ticket says which, and this ticket family has already had three numbers retracted this week. One qualifying word — "aux-role" or "of the 296" — settles it.

### Not verified, flagging rather than implying otherwise

- I did **not** open `tests/test_turn_cost_terminals.py` to check the `ended_by` taxonomy on the new silent path, which the ticket lists as a trap.
- I did **not** read the new test bodies to confirm they assert on **history content** rather than `ended_by`. The mutations prove they're load-bearing; they don't prove they assert the right thing.

### Delivery

A and B must ride the same train — anton reaches desktop only via the cowork-server version bump (ENG-1094). Nothing in either PR enforces that.

alecantu7:
## Review response — `f7f65312` (also cherry-picked to the hotfix twin #358 as `e34d0b4b`)

**Hole 2 (latch scope overstated) — accepted, documented in code.** The latch is ChatSession state and Cowork rebuilds the session per message, so in production it spans one message's continuations: the steady-state cost of a persistent denial is **one silent verdict call per message**, top-up recovery is simply the next message, and the `_VERIFIER_LATCH_REPROBE_TURNS` re-probe window is CLI/long-session behaviour. A scope note now sits on the denied branch so the next reader doesn't infer a cross-message guarantee. No code-behavior change: Layer A's contract is "the user is never told the turn failed"; the cross-message fix is cowork-server's wallet-aware resolution (mindsdb/cowork-server#322), and the ticket's Done-when + residual estimates were corrected to per-message accounting (consolidation, which has no latch, included).

Verifier suite re-run on both branches: 21/21.

alecantu7 (commented):
## Re-review — head `f7f65312`

Documentation only, and it closes a real gap rather than restating the commit. "Latches for the session" did read as cross-message; under Cowork the ChatSession is rebuilt per message (`harness.py:389`, no cache), so the latch spans one message's continuations, the steady-state cost of a persistent denial is one silent verdict call per message, and the re-probe window is CLI/long-session behaviour. Naming that also makes the division of labour explicit: this branch guarantees the user is never told the turn failed; cowork-server#322 is what stops the denial recurring.

No behaviour change, so the previous pass's mutation results still stand — the ordering trap and the sentinel trap are both pinned.

### Still open from the previous pass

Two of the ticket's own named traps for this layer remain unverified, and this commit didn't touch them:

- `tests/test_turn_cost_terminals.py` — the `ended_by` taxonomy on the new silent path.
- Whether the new tests assert on **history content** rather than `ended_by`, which is what the ticket asks for (and what makes the assertion meaningful, since anton's tool loop swallows exceptions and `ended_by` assertions can pass vacuously).

The mutations prove those tests are load-bearing; they don't prove they assert the right thing. Worth closing before merge — it's a short check.

alecantu7:
## Closing the re-review's open items — commit `c0337241` (hotfix twin: same change)

**1. The `ended_by` taxonomy on the silent path — now verified AND pinned.** Read `tests/test_turn_cost_terminals.py`: the existing `test_verifier_failure_reports_handback_verifier_failure` uses a `RuntimeError` (a genuine hard failure), so it still routes to the diagnosis and is unaffected by the denied clause. A denied turn books **`ended_by="completed"`** — and that's now a deliberate, tested contract rather than an accident: the turn's work succeeded and the user saw a normal reply, so `group by ended_by` error-rate queries must not count a priced-out check as a broken turn. No signal is lost — the denied probe remains countable from the gateway-side ERROR observation in Langfuse (that's also what the ticket's post-release re-measure keys on). New test `test_denied_verdict_reports_completed_not_verifier_failure`.

One honest note on verification hygiene: my first mutation check of that new test was **vacuous** — I stashed `session.py`, but the denied clause was already committed, so I'd only stashed a comment edit and the test "surviving" proved nothing. Re-ran properly against pre-fix `session.py` from `origin/staging`: the test **fails** there (books `handback_verifier_failure`) and passes with the fix. That's the real load-bearing proof.

**2. The new tests assert on history content, confirmed on record.** `test_denied_verdict_latches_silently_on_first_occurrence` asserts three independent user-facing signals and zero `ended_by` values: (a) `session._history` contains no `"task-completion check failed"` in any message content — the injection that persists into every later turn's payload; (b) no `StreamTaskProgress` event carries `"Something went wrong"`; (c) the diagnosis plan call never fires (`state["n"] == 2` per turn). The re-probe test asserts (a) and (c) again across `_VERIFIER_LATCH_REPROBE_TURNS + 2` turns. There is no `ended_by` assertion anywhere in `test_verifier_failsafe.py`, so the vacuous-`ended_by` trap the ticket warns about can't apply.

**3. Numbers reconciled.** The `_DENIED_VERDICT_ERRORS` comment now reads "of that ticket's 14-day baseline of 296 wallet-402s across 39 users, 208 were aux-surface calls like this one, across 33 of those users" — same population split the ticket's evidence addendum records, one sentence, no ambiguity.

Suites: `test_turn_cost_terminals.py` + `test_verifier_failsafe.py` → 37 passed on both this branch and the hotfix twin.

alecantu7 (commented):
## Re-review — head `c0337241`

Both open traps are closed, and I verified each rather than taking the commit message for it.

**Trap 1 — the `ended_by` taxonomy** is now a pinned contract. Mutation-verified independently: letting the denied path fall through to the hard handler fails `test_denied_verdict_reports_completed_not_verifier_failure`, and only that test (1 failed / 15 passed).

**Trap 2 — asserting on history content** turns out to have been satisfied all along; I had flagged it as *unverified*, not missing, and it holds:

```python
assert not any(
    "task-completion check failed" in str(m.get("content", ""))
    for m in session._history if isinstance(m, dict)
), "the internal-error injection reached history on a denied verdict"
```

That is the right assertion rather than the convenient one — `ended_by` assertions can pass vacuously in this loop, so reading history directly is what makes it mean something.

**The evidence figures now reconcile** — "of that ticket's 14-day baseline of 296 wallet-402s across 39 users, 208 were aux-surface calls like this one, across 33 of those users" — instead of two numbers competing with no stated population.

---

### One finding — the taxonomy is right, but the turn keeps no record that verification was skipped · low · confirmed

No argument with `completed`: a priced-out check is not a broken turn, and counting it as one would corrupt error-rate queries. The consequence is downstream of that decision.

`turn_completed` carries no field recording that the verdict never ran — checked the payload, there is no verifier/verdict/latch property. So a denied-verification turn is **byte-identical in the analytics event** to a turn that was verified and passed. Honest-stop rate and Phase 0 criterion 3 are computed by grouping that event on `ended_by`, and after anton#345 those rows land in PostHog — so unverified turns silently join the "completed" population, concentrated in exactly the wallet-locked cohort you would most want to isolate.

**Refutation attempted, and it partly lands.** The comment's claim that *"the denied probe stays countable from the gateway-side ERROR trace in Langfuse, so no signal is lost"* is true for **counting denials** — that is how this ticket produced 296/208 in the first place. What is not recoverable without a per-conversation Langfuse hop is the **join**: which `turn_completed` rows were unverified.

This is the same shape as `root_cause_classify_errors`, added to anton#348 today for precisely this reason — a metric must not let a skipped instrument read as a clean result. One level over, same defect.

**Smallest fix:** one boolean on the payload (`verification_skipped` / `verifier_latched`) — `_posthog_body` is a pass-through, so it needs nothing else. Or, if that is out of scope here, one line in the comment saying the honest-stop denominator now includes unverified turns and how to exclude them.

alecantu7:
## Review response — `014c3f3a` (hotfix twin: `9ae784b2`)

**The verification_skipped finding — accepted and implemented as the payload boolean, not the comment-only fallback.** The refutation was right that denials stay *countable* from the gateway ERROR traces; what was missing was the *join* on `turn_completed`, and a documented caveat wouldn't have survived contact with a dashboard. So:

- **`TurnCost.verification_skipped`**, stamped at all **four** no-verdict sites — denied latch, latched skip, second-hard-failure latch, and the failed re-probe (the finding named the denied path; the other three produce the identical byte-identical-to-verified `completed` row and would have kept the skew alive, just smaller). Stamped at the sites rather than read from `self._verifier_latched` at emit, per the dataclass's own stamped-at-open rule — a late finalizer would report a later turn's latch state.
- Emitted on `turn_completed` as `"true"`/`"false"` and on the `turn_cost` log line (the sink that survives collector allowlist drops, per the ENG-1355 precedent). Property documented in `analytics.py`'s event register. Handback terminals carry no flag — their `ended_by` already distinguishes them.
- Hotfix-twin note: `main` predates ENG-1495, so there the property rides the collector relay (kwargs are the schema) rather than the direct PostHog post, and the register-doc hunk was dropped in the cherry-pick resolution — the flag itself is identical on both branches.
- Tests: the denied-terminal test now asserts `"true"`, and a new verified-pass control asserts `"false"`. Both fail without the emission (checked by stashing the session.py changes — a real mutation this time, since these edits were uncommitted).

Suites: `test_turn_cost_terminals.py` + `test_verifier_failsafe.py` + `test_turn_cost.py` → 70 passed on both branches.

For the measurement side: the harness-measure honest-stop recipe should add `verification_skipped = 'false'` (or exclude `'true'`) to the `turn_completed` denominator once this ships — noting it here so the query change is traceable to this thread.

alecantu7 (commented):
## Re-review — head `014c3f3a`

The flag is implemented better than I specified. Stamping at the four no-verdict sites rather than reading live latch state at emit is the right call — reading at emit would let a late finalizer report a *later* turn's state, which is the trap this file has been bitten by before. Emitting on the `turn_cost` log line as well as the event matters too: that's the sink that survives a collector drop. `38 passed`.

Mutations confirm the core is pinned:

| Mutation | Result |
| -- | -- |
| drop the emission from the `turn_completed` payload | **2 failed** — the denied case and the verified-pass control |
| drop the **denied-latch** stamp only | **1 failed** |

**A candidate of mine died on inspection**, and the docstring had already answered it: I expected round-capped turns to slip through unstamped, since they skip verification entirely. They book `ended_by="round_cap"` (`session.py:2966`, `:3888`), so they are already excludable — which is exactly what *"Handback terminals don't need it: their `ended_by` already distinguishes them"* means. Coverage is complete: a first hard failure hands back, and everything else that books `completed` without a verdict is stamped.

---

### One finding — three of the four stamp sites are unpinned · low · confirmed

```
mutation: drop the LATCHED-SKIP stamp
  -> 38 passed          (survived)
```

Only the denied-latch stamp has a test behind it. Remove the latched-skip one and the suite stays green — and that is the site covering the **steady state**: the denied latch fires once, then every subsequent turn in the session takes the latched-skip path. Same for the failed-re-probe and second-hard-failure stamps.

It matters most exactly where the scope note in `f7f65312` puts it. Under Cowork the session rebuilds per message, so the denied path re-fires every message and the tested site is the one that runs. The latched-skip path is the **CLI / long-session** behaviour — so the untested site is the one that has to hold across many turns, which is where a silent regression would do the most damage to the denominator this flag exists to protect.

**Refutation attempted:** the four sites are three lines each and visually identical, which makes them feel self-evidently correct. That is the same argument that failed on anton#348's legacy guard and anton#339's `PROGRESS_MARKER` writer — both "obviously the same as the one beside it", both unpinned until a mutation found them. This is the third instance of that shape today: several writers of one piece of state, one test.

**Smallest fix:** one test driving two turns against a denied verifier and asserting `verification_skipped == "true"` on the **second** turn's event. It covers the latched-skip site and goes red under the mutation above.

alecantu7:
## Review response — `24d1dfae` (hotfix twin: `7cc867ca`)

**Three-of-four-sites-unpinned — accepted, and the pattern call was fair.** Two multi-turn tests now cover the remaining sites:

- `test_latched_skip_turns_also_stamp_verification_skipped` — the reviewer's smallest fix, verbatim: a denied verifier over two turns in one session, asserting `"true"` on the **second** turn's event. That's the steady-state site.
- `test_hard_latch_and_failed_reprobe_turns_also_stamp_verification_skipped` — a hard-failing verifier over `_VERIFIER_LATCH_REPROBE_TURNS + 2` turns: turn 1 hands back (`handback_verifier_failure`, `verification_skipped="false"` — its terminal already distinguishes it), turn 2 pins the hard-latch site, and the window's tail covers every latched skip plus the failed re-probe.

**Per-site mutation matrix, each stamp dropped individually:**

| Stamp dropped | Result |
| -- | -- |
| latched-skip | both new tests fail |
| failed re-probe | the long test fails (exactly it — verified with the failure name, after my first mutation string produced a noisy count) |
| second-hard-failure latch | the long test fails |
| denied latch | already pinned (previous pass) |

All four writers of the flag now have a test that goes red if any one of them is silently removed. Suites: 19/19 terminals, 40/40 terminals+failsafe on the hotfix twin, full targeted set 72/72 on this branch.

With this, every finding across the five review passes on this PR is closed with a load-bearing test or an explicit disposition. Ready for human review.

alecantu7:
## 📍 Ready for human review — final state at `24d1dfae` (supersedes the earlier 📍 at `014c3f3a`)

**What this PR is:** a typed `denied` path in the completion verifier — wallet 402s / allowance 429s (`TokenLimitExceeded`) and 404'd/403'd models (`ModelUnavailableError`) latch silently on first occurrence instead of telling the user *"The task-completion check failed to run (internal error)"* about a turn whose work succeeded. Plus `verification_skipped` on `turn_completed` (stamped at all four no-verdict sites) so unverified turns are excludable from honest-stop denominators.

**Review trail, in order — five passes, all closed:**
1. Self-review (`06f9bc75`) — ordering + sentinel traps mutation-pinned.
2. Latch scope (`f7f65312`) — per-message under Cowork, documented at the branch.
3. Taxonomy + numbers (`c0337241`) — denied books `completed`, pinned; figures reconciled.
4. `verification_skipped` (`014c3f3a`) — the flag, stamped at four sites, emitted on event + log.
5. Stamp-site coverage (`24d1dfae`) — all four stamp writers individually mutation-pinned (previously only the denied site was).

**The two contracts to eyeball:** the except-clause ordering (`_DENIED_VERDICT_ERRORS` must precede `_TRANSIENT_VERDICT_ERRORS` — `ModelUnavailableError` is an `OSError` subclass; comment on the constant) and the `"denied"` sentinel handled before the hard-failure accounting (a bare `break` would fall into the diagnosis stream this PR exists to suppress).

**Tests:** 72 across `test_verifier_failsafe.py` + `test_turn_cost_terminals.py` + `test_turn_cost.py`; 8 new, every one mutation-verified against code with its guard removed. Full suite on this base: 2170 passed. CI green at this head.

**Merge order:** merges to `staging` first (before hotfix twin #358, which is content-identical modulo the pre-ENG-1495 analytics plumbing) so the later main→staging back-merge resolves trivially. Independent of cowork-server#322 and cowork#663.

pnewsam (approved):
**Approve — Layer A (silent denial latch), staging original.** Same load-bearing safety claims I verified on the #358 twin, checked against `origin/staging`:

- `TokenLimitExceeded` is billing-only and code-exact — produced only on a 402/429 carrying a `wallet_empty` / `included_allowance_exhausted` gate code (`openai.py:141-152`); context overflow is a *separate* `ContextOverflowError` (`provider.py:391`). So the first-occurrence silent latch cannot swallow a genuine context-limit verifier failure.
- The except-clause ordering is genuinely required: `ModelUnavailableError(ConnectionError)` (`provider.py:605`) subclasses `OSError`, so `_DENIED_VERDICT_ERRORS` must precede the transient tuple — the `model-404` parametrized case proves it latches instead of being swallowed as transient.
- This is the staging original; the only delta vs. the #358 hotfix twin is the `anton/analytics.py` doc-comment for the `verification_skipped` field, which correctly lands here. Session / turn_cost / test hunks are content-identical to #358, so the staging→main back-merge stays clean.
- The `verification_skipped` stamp lives on `TurnCost` at each skip/deny/latch site (never read from live latch state at emit), and the steady-state stamp sites are pinned by tests. Good.

Low-sev, not blocking (same note as #358): `openai.py:107` maps any 429-with-`detail` to `TokenLimitExceeded`; worth confirming gateway velocity/rate-limit 429s carry no `detail`, else they'd silent-latch here (self-heals via the re-probe).

_Reviewed with Claude Code; the load-bearing assumptions were checked against source at `monorepo-3`._

PR #358, opened 2026-08-14
ENG-1632 hotfix: silent verifier latch on deterministic denial (main)

alecantu7:
## Hotfix twin of #357

Same fix as #357 (which targets `staging` for the normal train), cherry-picked onto `main` for an out-of-band release: wallet/allowance denials and 404'd models on the verifier's verdict call latch silently on FIRST occurrence — no more per-turn "internal error" handback for turns whose work succeeded. Full rationale, evidence, and test plan on #357 and [ENG-1632](https://linear.app/mindsdb/issue/ENG-1632).

**Verified on this base:** cherry-pick was clean; `tests/test_verifier_failsafe.py` 21/21 (the 4 new tests are mutation-verified on the staging twin).

**Release sequencing (ENG-1094):** anton reaches desktop only via cowork-server — after this merges, the cowork-server hotfix (mindsdb/cowork-server#323) needs its `anton-agent` lock refreshed in the same release so the pypi channel carries both halves. Merging here alone ships nothing to desktops.

**If both this and #357 merge:** the weekly staging→main release will see the same change on both sides — the cherry-pick is content-identical, so the release merge stays clean.

Merge is a human call — this PR only stages the option.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -3953,6 +3953,16 @@ class ChatSession:
                     # runs; a re-probe denied again lands back here and stays
                     # latched, so adding credits self-heals within
                     # _VERIFIER_LATCH_REPROBE_TURNS turns.
+                    #
+                    # Scope note: the latch is ChatSession state, and Cowork
+                    # rebuilds the session PER MESSAGE — there it only spans
+                    # one message's continuations, the steady-state cost of a
+                    # persistent denial is one silent verdict call per
+                    # message, and top-up recovery is simply the next message.
+                    # The re-probe window above is long-session (CLI)
+                    # behaviour. The cross-message fix is server-side model
+                    # resolution (cowork-server, same ticket); this branch is
+                    # the guarantee the user is never told the turn failed.
                     self._verifier_latched = True
                     self._verifier_latch_reason = "denied"
                     _verifier_log.info(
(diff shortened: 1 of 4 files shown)


pnewsam (approved):
**Approve — Layer A (silent denial latch).** Verified the load-bearing safety claims against `origin` source, not just the diff:

- `TokenLimitExceeded` is billing-only and code-exact: `_raise_for_status_error` produces it only on a 402/429 carrying a `wallet_empty` / `included_allowance_exhausted` gate code (`openai.py:141-152`); context overflow is a *separate* `ContextOverflowError` (`provider.py:391`). So the first-occurrence silent latch cannot swallow a genuine context-limit verifier failure.
- The except-clause ordering is genuinely required, not defensive: `ModelUnavailableError(ConnectionError)` (`provider.py:605`) really subclasses `OSError`, so `_DENIED_VERDICT_ERRORS` must precede the transient tuple — the `model-404` parametrized case is the proof it latches instead of being swallowed as transient.
- Diffed against #357: identical text on `session.py` / `turn_cost.py` / both test files; #358 omits only the `analytics.py` doc-comment hunk. Content-identical hunks → the staging→main back-merge stays clean.

Two flags, neither blocking:
- **Delivery:** this reaches desktop only via the cowork-server bump — coordinate with cowork-server#323's `anton-agent` repin so both halves ship in the same train.
- **Low-sev:** `openai.py:107` maps any 429-with-`detail` to `TokenLimitExceeded`; if the gateway ever emits velocity/rate-limit 429s carrying a `detail`, they'd silent-latch here (pre-existing behavior; self-heals via the re-probe). Worth confirming rate-limit 429s carry no `detail`.

_Reviewed with Claude Code; the load-bearing assumptions were checked against source at `monorepo-3`._

PR #363, opened 2026-08-18
ENG-1693: only OUR gateway may select a billing verdict

alecantu7:
Closes [ENG-1693](https://linear.app/mindsdb/issue/ENG-1693). The half that [cowork-server#332](https://github.com/mindsdb/cowork-server/pull/332) (ENG-1686) could not reach.

## Why the previous two fixes didn't close this

ENG-1537 gated cowork-server's body-`code` carrier. ENG-1686 gated its header twin. **Both are in cowork-server, and anton reads the same two carriers one layer earlier — ungated.**

anton converts a wallet denial into a typed `TokenLimitExceeded`, and cowork-server matches that at the **top** of its ladder (`is_token_limit_error`), above all origin logic. So cowork-server's gate is correct and never consulted. Executed against ENG-1686's head:

```
raw third-party header             -> None                                      # its gate works
anton-shaped (TokenLimitExceeded)  -> ('token_limit', "You're out of credits…")
host resolved for the hostile case : openrouter.ai
gate says known-third-party?       : True                                       # never asked
```

The trigger is only `status ∈ {402, 429}` plus the header **or** the body code — both fully third-party controlled on a BYOK `OPENAI_COMPATIBLE` endpoint. A hostile or misconfigured provider could put our out-of-credits card, with its top-up CTA, in front of a user whose MindsHub wallet is fine.

## After

Verified through the real mapper with real SDK errors, not hand-built fixtures:

| Origin | Carrier | Result |
|---|---|---|
| `openrouter.ai` | header | `ConnectionError` → generic copy |
| `openrouter.ai` | body `code` | `ConnectionError` |
| `mindshub.ai.evil.com` | header | `ConnectionError` |
| `api.mindshub.ai` | header | **`TokenLimitExceeded`** → credits card |
| `api.mindshub.ai` | body `code` | **`TokenLimitExceeded`** |
| `api.mdb.ai` (legacy) | header | **`TokenLimitExceeded`** |

## Three decisions worth reviewing

**The host match is not the substring idiom.** This package already tests `"mindshub.ai" in base_url` in three places, but `mindshub.ai.evil.com` satisfies it. That's fine for picking an API dialect and unfit for a trust decision, so this matches the exact domain or a dot-delimited subdomain. Existing uses are untouched — they aren't trust decisions.

**Three-valued, mirroring ENG-1686.** Only a *provable* third party is refused; unknown origin stays trusted. That's what keeps the two **mid-stream** wallet paths working unchanged — they carry no response, so no host is resolvable. Treating unknown as hostile would have silently broken them.

**`classify_transient`'s wallet check is deliberately left ungated**, and now says so in the code. It suppresses *retries*, where being conservative is right regardless of origin — gating it would make a hostile wallet code **retryable**, which is worse than the card it would deny.

## Seven existing tests broke — I fixed the fixture, not the assertions

`_sdk_error` used a `gateway.test` placeholder. The host was incidental before and is load-bearing now, so every test named `test_gateway_*` was silently exercising the third-party path. They now use a real MindsHub host, which is what their names always claimed. **No assertion was changed** — worth checking that in review, since "tests broke so I edited them" is the shape of a bad fix.

## Verification

Mutation-verified against all three ways to get this wrong:

| Mutation | Result |
|---|---|
| no gate (pre-fix state) | **8 red** |
| the substring form (lookalike bypass) | **5 red** |
| unknown origin treated as hostile | **1 red** |

New tests parametrise origin × carrier, so neither lane can reopen alone, and cover the legacy `mdb.ai` host and the ENG-1169 direction (a genuine denial still cards).

Full suite: **2169 passed, 28 skipped, 0 failed.**

## Security pass

Narrows an existing capability; adds none. No new user input reaches a sink — the change is a hostname comparison, and all user-facing copy remains module constants. ENG-1169's Add-credits card is unaffected for real gateway denials, asserted in both mappers.

## Accepted tradeoff: a relay in front of MindsHub loses the credits card

Stated plainly because I reached it by accident, not by choice. The gate keys on the host being `mindshub.ai`/`mdb.ai` or a subdomain, so a user pointed at a corporate relay or LLM proxy that forwards our denials resolves a non-MindsHub host. Executed:

```
relay, wallet 402  -> ConnectionError        (credits card lost)
direct, wallet 402 -> TokenLimitExceeded     (card works)
relay, velocity429 -> TransientProviderError (wait preserved)
```

That is ENG-1169's symptom returning for that deployment shape. **And I treated the same shape two opposite ways in this PR**: `_velocity` is deliberately left ungated with a comment citing exactly "a relay or corporate proxy in front of MindsHub", while the wallet carrier is gated and breaks it.

The asymmetry is defensible — a spoofed billing card asks the user for money, a spoofed wait does not, so the carriers deserve different trust — but it should be a decision, not a side effect. Accepted as the cost of closing the spoof. If relayed MindsHub becomes a supported shape, the fix is a configurable trusted-host list rather than widening the match.

An earlier version of this section said "ENG-1169's Add-credits card is unaffected for real gateway denials". True as written, but it reads as fully unaffected. It isn't, for this shape.

## When this actually reaches users

**Merging to `staging` does not close the vector for desktop users.** cowork-server vendors anton as `anton-agent = { git = …, branch = "main" }`, and anton's staging is currently **119 commits ahead of main** — so this ships on the weekly staging→main train, not on merge. "Ships before or with cowork-server#332" below is about ordering between the two PRs, not about when the hole shuts.

**Ships before or with cowork-server#332** — the two are independent, but this is the one that closes the case the ticket describes.

---

# Self-review round — four more lanes were open

An adversarial pass on this PR found the gate real on two carriers and inert or absent on four others. Same shape as the two rounds before it: I fixed the layer I was looking at.

**The mid-stream lane was fully open, and my comment claimed the opposite.** `response_origin_host` read only `exc.response`; a mid-stream failure raises a bare `openai.APIError` with no `.response` but a `.request` carrying the real URL. Executed pre-fix: `evil.example.com -> host=None foreign=False -> TokenLimitExceeded`. My comment said "nothing a remote can pick" — serving 200-with-an-SSE-error instead of 402 is entirely the remote's choice, and it's the *cheaper* attack. The repo's own e2e fixture proved it: those "midstream wallet" tests ran against `http://mock/v1` and passed.

**The 429 `detail` branch bypassed the gate entirely** — one JSON key, no header, and the only carrier whose text is partly attacker-authored:

```
Server returned 429 — PWNED: visit https://evil.example to fix
Visit https://console.mindshub.ai to upgrade or top up your tokens.
```

**Also fixed:** the 403 plan-copy branch (same class — names a MindsHub plan, links our upgrade page); an **ENG-1537 regression I introduced** (the gate suppressed the header read wholesale, and `_velocity` reads it too, so a relay in front of MindsHub lost its `Retry-After` wait); the anthropic twin's missing `isinstance(str)` guard; and my factually wrong justification for leaving `classify_transient` ungated.

**Coverage was the review's central finding** — three of five gate sites had none. Six mutations now verified red: the `.request` fallback, both `detail` gates, the 403 gate, the anthropic wallet gate, and re-gating velocity.

Full suite: **2180 passed, 28 skipped, 0 failed.**

### DIFF
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -600,14 +600,22 @@ def response_origin_host(exc: BaseException) -> str | None:
 
     ``httpx.Response.url`` is a property that RAISES when no request is
     attached, so this cannot be a bare ``getattr`` chain.
+
+    Falls back to the exception's OWN ``request``, which is what makes this
+    work on the mid-stream lane. A mid-stream failure surfaces as a bare
+    ``openai.APIError`` — constructed as ``APIError(message, request, body=…)``,
+    so it has **no** ``.response`` but does carry ``.request`` with the real
+    URL. Reading only ``.response`` made the gate silently inert exactly where
+    an attacker has the freest hand: answering 200 and smuggling the wallet
+    code into an SSE frame is the remote's choice, not a quirk of our plumbing.
     """
     resp = getattr(exc, "response", None)
-    if resp is None:
-        return None
     try:
-        url = getattr(resp, "url", None)
-        if url is None:
+        url = getattr(resp, "url", None) if resp is not None else None
+        if url is None and resp is not None:
             url = getattr(getattr(resp, "request", None), "url", None)
+        if url is None:
+            url = getattr(getattr(exc, "request", None), "url", None)
         if url is None:
             return None
         host = getattr(url, "host", None)
@@ -761,11 +769,15 @@ def classify_transient(
         if etype == "insufficient_quota":
             return None
         if wallet_denial_code(b):
-            # Deliberately NOT origin-gated (ENG-1693). This guard suppresses
-            # RETRIES, and being conservative is right no matter who sent it:
-            # retrying a third party's quota denial is wasteful and cannot
-            # succeed either. Gating here would make a hostile wallet code
-            # RETRYABLE, which is worse than the card it would deny.
+            # Deliberately NOT origin-gated (ENG-1693), for a plainer reason
+            # than an earlier version of this comment claimed. It said gating
+            # would make a hostile wallet code "retryable"; that is false —
+            # both this branch and the fallthrough reach the same count-based
+            # retry, so retryability is unchanged either way. The real reasons
+            # are that this call site cannot see the origin (it receives only
+            # status + body, never the exception), and that suppressing a
+            # retry is conservative regardless of who sent the code: retrying
+            # a third party's quota denial cannot succeed either.
             return None
         # session_backoff=True, unlike every other request-time status here
         # (ENG-1537). The flag means "should the SESSION spend its budget on
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -574,6 +574,75 @@ _TRANSIENT_ERROR_TYPES = frozenset(
 _WALLET_DENIAL_CODES = frozenset({"wallet_empty", "included_allowance_exhausted"})
 
 
+# Hosts that ARE the MindsHub gateway. Only a response from one of these may
+# select a billing verdict — see :func:`origin_is_known_third_party` (ENG-1693).
+_MINDSHUB_HOSTS = ("mindshub.ai", "mdb.ai")
+
+
+def is_mindshub_host(host: str | None) -> bool:
+    """Whether ``host`` is the MindsHub gateway or one of its subdomains.
+
+    Deliberately NOT the ``"mindshub.ai" in base_url`` substring test used
+    elsewhere in this package for flavour detection and trace-header opt-in.
+    That form is fine for choosing an API dialect and unfit for a trust
+    decision: ``mindshub.ai.evil.com`` satisfies it. This matches the exact
+    domain or a dot-delimited subdomain of it, so an attacker cannot buy a
+    name that merely contains ours.
+    """
+    if not host:
+        return False
+    h = str(host).strip().lower().rstrip(".")
+    return any(h == d or h.endswith("." + d) for d in _MINDSHUB_HOSTS)
+
+
+def response_origin_host(exc: BaseException) -> str | None:
+    """Hostname the failing request was sent to, or ``None`` if unknowable.
+
+    ``httpx.Response.url`` is a property that RAISES when no request is
+    attached, so this cannot be a bare ``getattr`` chain.
+    """
+    resp = getattr(exc, "response", None)
+    if resp is None:
+        return None
+    try:
+        url = getattr(resp, "url", None)
+        if url is None:
+            url = getattr(getattr(resp, "request", None), "url", None)
+        if url is None:
+            return None
+        host = getattr(url, "host", None)
+        if not host:
+            from urllib.parse import urlparse
+
+            host = urlparse(str(url)).hostname
+    except Exception:
+        return None
+    return str(host).lower() if host else None
+
+
+def origin_is_known_third_party(exc: BaseException) -> bool:
+    """Whether this failure provably came from somewhere that is NOT our gateway.
+
+    Three-valued on purpose, mirroring cowork-server's gate (ENG-1686): it
+    answers "do we KNOW it was someone else", so an unknown origin stays
+    trusted rather than being treated as hostile. A real SDK error always
+    carries its request, so every genuine HTTP response resolves a host;
+    unknown origin means a synthetic or mid-stream error, which no remote
+    server can choose.
+
+    Used to stop a BYOK endpoint selecting a MindsHub billing verdict by
+    echoing our private ``X-MindsHub-Reason`` header or wallet ``code``
+    (ENG-1693). NOT used in :func:`classify_transient`'s wallet check — see
+    the comment there.
+    """
+    return origin_is_known_third_party_host(response_origin_host(exc))
+
+
+def origin_is_known_third_party_host(host: str | None) -> bool:
+    """Host-level form of :func:`origin_is_known_third_party`."""
+    return host is not None and not is_mindshub_host(host)
+
+
 def wallet_denial_code(body: Any) -> str | None:
     """The M3 gate's out-of-credits code carried in an error body, if any.
 
@@ -692,6 +761,11 @@ def classify_transient(
         if etype == "insufficient_quota":
             return None
         if wallet_denial_code(b):
+            # Deliberately NOT origin-gated (ENG-1693). This guard suppresses
+            # RETRIES, and being conservative is right no matter who sent it:
+            # retrying a third party's quota denial is wasteful and cannot
+            # succeed either. Gating here would make a hostile wallet code
+            # RETRYABLE, which is worse than the card it would deny.
             return None
         # session_backoff=True, unlike every other request-time status here
         # (ENG-1537). The flag means "should the SESSION spend its budget on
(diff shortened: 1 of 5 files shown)


anton/core/llm/provider.py
```diff
@@ -574,6 +574,83 @@ def __init__(
 _WALLET_DENIAL_CODES = frozenset({"wallet_empty", "included_allowance_exhausted"})
 
 
+# Hosts that ARE the MindsHub gateway. Only a response from one of these may
+# select a billing verdict — see :func:`origin_is_known_third_party` (ENG-1693).
+_MINDSHUB_HOSTS = ("mindshub.ai", "mdb.ai")
```
pnewsam, line 579: **Minor (verification question):** `_MINDSHUB_HOSTS = ("mindshub.ai", "mdb.ai")` is the sole allowlist, so any real gateway denial served from a production host outside those two apexes (or their subdomains) silently loses the credits card — a targeted ENG-1169 regression that fails closed to generic copy. The tests only exercise `api.mindshub.ai`, `mindshub.ai`, `api.mdb.ai`. Please confirm the production MindsHub gateway is always served from one of these two apex domains, including any regional or vanity host, since a missing apex produces the generic message rather than the card.

anton/core/llm/provider.py
```diff
@@ -574,6 +574,83 @@ def __init__(
...
+    work on the mid-stream lane. A mid-stream failure surfaces as a bare
+    ``openai.APIError`` — constructed as ``APIError(message, request, body=…)``,
+    so it has **no** ``.response`` but does carry ``.request`` with the real
+    URL. Reading only ``.response`` made the gate silently inert exactly where
+    an attacker has the freest hand: answering 200 and smuggling the wallet
+    code into an SSE frame is the remote's choice, not a quirk of our plumbing.
+    """
+    resp = getattr(exc, "response", None)
+    try:
+        url = getattr(resp, "url", None) if resp is not None else None
```
pnewsam, line 614: **Nit (low confidence, robustness):** `getattr(resp, "url", None)` triggers `httpx.Response.url`, which *raises* `RuntimeError` when no request is attached rather than returning `None`. That exception is caught by the outer `except Exception: return None`, so the intended `exc.request` fallback below is short-circuited whenever a response object is present but request-less. In the real SDK paths this can't trigger — request-time responses always carry a request, and mid-stream errors have `.response is None` — so the gate is correct today. It's only a defense-in-depth ordering note: a response-present-but-url-raising case resolves to `None` (trusted) even if `exc.request` held a foreign host. The anthropic tests set `exc.response.request` manually precisely because that construction otherwise lacks it, which is what surfaced the theoretical gap. No change required unless you want the fallback reachable in that case.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff across five files plus the complete `anthropic._raise_for_status_error` source at head. The change narrows an existing capability — only a response whose origin host is provably MindsHub (`mindshub.ai`/`mdb.ai` or a dot-delimited subdomain) may mint a billing verdict (credits card, plan-upgrade CTA, `console.mindshub.ai` copy). The host match correctly rejects the substring-lookalike (`mindshub.ai.evil.com`), the gate is three-valued so unknown/synthetic origins stay trusted, and every carrier (header, body `code`, `detail`, 403 plan-copy, mid-stream SSE) is gated in both mappers with matching tests. No confirmed merge-blocking defect. The two inline notes are verification questions, not proven defects.

### Validation

- Read the complete PR diff (5 files, +453/−22), including all new tests in `test_status_error_mapper.py` and `test_transient_retry_e2e.py`.
- Read the full `anton/core/llm/anthropic.py` mapper at head `e2f1de1c` — confirmed it has **no** ungated 403 `model_access_denied` branch (that branch is openai-only), and `gate_reason`/`_body_code` are both defined and read.
- Traced each gated branch's fall-through for a foreign origin: 402/429 wallet, 429 `detail`, 403 plan-copy, and both mid-stream sites all bypass the card and land on `classify_transient`/generic copy. `classify_transient`'s ungated wallet check returns `None` (never `TokenLimitExceeded`), so leaving it ungated cannot reopen the card.
- Did **not** run the suite — the `anton` repo is not checked out locally. Relying on the PR's stated `2180 passed, 0 failed` and mutation results.

The self-review round is unusually rigorous — the mid-stream `.request` fallback, the `detail`/403 gates, the ENG-1537 velocity un-gating, and the anthropic `isinstance(str)` guard are all real gaps that are now closed and covered by red-verified mutations. The fixture-host correction (not assertion edits) is the right call, and the assertions are unchanged.

anton/core/llm/provider.py
alecantu7: Confirmed, and now pinned in code — follow-up in #367 (this PR merged before the nits were addressed, my mistake).

**The allowlist is complete for every host that can serve a gateway billing denial.** Checked against the terraform host inventory (source of truth for zones and certs) plus every base URL configured across anton, cowork, cowork-server, mdb-ai and mindshub_inference:

- prod `api.mindshub.ai`; staging/dev `api.staging.mindshub.ai`, `api.dev.mindshub.ai`; ephemeral per-PR envs `api-pr<N>.dev.mindshub.ai`
- legacy vanity gateway `llm.mdb.ai`, `llm.staging.mdb.ai`
- white-label surfaces `writer.mdb.ai`, `terabase.dev.mdb.ai`, `view.mindshub.ai`

Terraform declares no other apex zone, and `*.mindshub.ai` / `*.mdb.ai` cover anything added beneath them later. I also checked the Kinaxis white-label branch specifically for a customer-owned gateway domain — there isn't one.

Rather than leave that answer here where it goes stale, #367 turns those nine hosts into a parametrised test, so narrowing the allowlist fails there instead of silently degrading a real out-of-credits denial.

One thing your question surfaced that is worth recording: **`4nton.ai` is a real production zone that is not on the allowlist.** It serves agent provisioning, per-instance hosts (`sp_<hash>`, `cw-<id>`) and artifact publishing, never inference — so excluding it is correct today. #367 asserts that exclusion and puts a note on `_MINDSHUB_HOSTS` saying it must be added if inference is ever routed onto it, so the cost of failing closed isn't discovered in production.

anton/core/llm/provider.py
alecantu7: You were right, and it was **understated** — fixed in #367.

Two things came out of chasing it:

1. The function's own docstring already claimed `.url` "cannot be a bare `getattr` chain", and the code was a bare `getattr` chain. So this wasn't only defense-in-depth: the comment asserted a guarantee the code didn't provide.
2. **`httpx.Response.request` has the identical trap.** Guarding only `.url` — which is what your nit named, and what I implemented first — still fails the new test, because the second fallback was `getattr(getattr(resp, "request", None), "url", None)` and that inner `getattr` raises the same `RuntimeError`. Verified against httpx 0.28.1 rather than assumed:

```
getattr(resp,'url',None)     RAISED RuntimeError
getattr(resp,'request',None) RAISED RuntimeError
```

Both reads are now contained individually so each fallback stays reachable. The test asserts its own premise (`pytest.raises(RuntimeError)` on the bare response), so it can't quietly stop testing anything if httpx changes that behaviour — and it reds under both the shipped chain and the `.url`-only fix, so it discriminates the complete fix from the plausible-looking one.

Agreed on your reachability call: not triggerable through the real SDK paths today. Worth fixing anyway because the failure direction is "silently trust a hostile origin".

PR #364, opened 2026-08-18
ENG-1633: mark every clip in the verify transcript, widen the judged reply (hotfix)

alecantu7:
> **Hotfix — targets `main`, so merging ships to production immediately.** Not stacked on anything; `_render_verify_transcript` is byte-identical on `main` and `origin/staging`, so this applies cleanly to both. **The back-merge is automated** — `.github/workflows/sync-main-to-staging.yml` fires on push to `main` and names this exact case; it succeeded on the last three pushes including the ENG-1632 hotfix. Confirm it ran rather than merging by hand.

Fixes ENG-1633.

## What was wrong

`_render_verify_transcript` clipped conversational text with a bare `text[:text_cap]`. Any reply over 2,000 characters reached the completion verifier ending mid-word with **nothing indicating it had been cut** — and, asked whether a complete answer was delivered, the verifier reported what it was shown.

It was not judging badly. It was reasoning correctly from a false premise: it believed it was looking at the whole reply, and that reply appeared to stop mid-sentence. Both models quote the cut back verbatim — *"cuts off mid-sentence ('That is')"* — where the 2,000-char slice ends exactly on `That is`.

Tool results have gone through `_clip_keep_cause` since ENG-836. **The conversational path in the same function never got the same treatment.**

## Impact, measured

| | |
| -- | -- |
| final replies over 2,000 chars (i.e. clipped, unmarked) | **31.1%** |
| production continuations carrying a truncation-worded reason | **74 of 169 — 44%** |
| user-billed tokens per false continuation | **~137,000** (full-context re-run × 2.6 downstream calls; the allowance charges cache reads at full weight) |

n = 1,401 anton final replies / 15,786 observations, gateway traffic, 51h.

## What changed

1. **Both conversational sites route through `_clip_keep_cause`** — head + `[... N chars elided ...]` + tail-biased remainder. Every cap in the function now goes through one helper, so nothing is clipped silently. This finishes a migration that stopped halfway rather than introducing a pattern (convention 7).
2. **The message under judgment gets its own budget: `final_cap = 12000`.** Prior turns stay at `text_cap = 2000`.

### Why a flat cap was not enough — the finding that changed the design

The obvious fix (route everything through `_clip_keep_cause` at one cap) **regresses**, in the dangerous direction. Measured live on a fixture where the reply opens *"All three done"* and buries *"(3) Email. I could NOT send this"* in the middle:

| arm | COMPLETE verdicts / 6 runs |
| -- | -- |
| shipped today | 0 — catches it |
| flat-cap fix | **3 — false COMPLETE on haiku, 3/3** |
| no-clip control | 0 — catches it |

Head-clipping deletes the *tail*; a flat both-ends clip deletes the *middle*. Any fixed window has a hole — so the window for the one message actually being judged has to be wide enough that it rarely opens. Widening only that entry also bounds the cost to one message instead of the ten the transcript carries.

`test_buried_failure_admission_survives_in_the_judged_reply` is the deterministic guard for this; it reds under both the shipped code and the flat-cap variant.

### Tradeoff, settled

`_clip_keep_cause` is tail-biased because a tool failure names its cause at the end. That is right for a conversational reply too, for the same underlying reason rather than by luck: in both cases the tail is *the part that answers the question being asked of this text* — the cause for a failure, the conclusion for a reply. A completion verifier asks "did the assistant finish?", and the evidence for finishing is the last paragraph, which is exactly what the head-slice discarded.

**Side-effect worth having:** today a genuinely truncated reply (ENG-1042) and a clipped complete one are byte-identical in shape — both end mid-word at 2,000, both unmarked. After this, the real one still ends mid-word *inside the tail* while a complete one ends on a period, so the verifier can separate them itself. Verified: the ENG-1042 shape is still graded INCOMPLETE 3/3.

## Cost

`final_cap = 12000` sized from the measured distribution (median 1,095 / p90 6,679 / p95 9,504 / p99 17,405):

- **+276 tokens/turn mean, +2,900 worst case** — one message, cannot compound.
- Sends **97.2%** of replies whole, up from 68.9%.
- Break-even false-continuation rate is 2.1% even at the maximum added cost; the measured rate is **5.4%**, so the change is net-positive by 2.6× in its own worst case. Measured over the same window: **+376k tokens added, 10.1M removed.**

8,000 was the first pick; 12,000 costs +48 mean tokens more and cuts residual clipping 2.8× (7.9% → 2.8%).

## Residual limit

A reply over 12,000 chars (2.8%) whose decisive negative evidence sits in the elided middle can still read COMPLETE. Today's equivalent exposure is 31.1% of replies losing their entire tail. Note also that **41% of the 12k+ population is itself generated by this bug's retry loop** — the nudge says *"pick up where you left off and finish the remaining work"* — so the residual should shrink after this lands rather than hold.

## Deliberately not changed

- **No global `text_cap` raise.** Measured: cap 6000 == cap 2000 == uncapped on every clean fixture.
- **No rubric change.** A prompt sentence explaining the elision marker scored identically to the marker alone; the rubric is a known binding lever (ENG-836 moved 12/12 on wording), so a change with no measured gain is pure regression risk.
- **`_summarize_history` (`:1908-1920`) has the identical bare-slice pattern**, but feeds the *compactor*, not the verifier. Different surface, different blast radius, no evidence yet on frequency. Left for its own ticket.

## Also fixed (small scope extension)

The same line clipped **USER** messages. Measured: a 3,466-char request rendered at 2,006 chars, silently dropping its trailing `IMPORTANT:` requirement — so the verifier judged completion against a truncated ask. Only bites prior turns (the current request is also passed unclipped in the header), but it is the same line and the same defect.

## Test plan

- **6 new deterministic unit cases** in `tests/test_verify_transcript.py`, all **mutation-verified**: reverting to the bare head-slice reds all 6; the flat-cap variant reds 3, including the buried-admission guard.
- **1 live verdict-quality case** in `tests/test_verifier_verdict_live.py` (ENG-1211's harness), pinned in `test_verifier_eval_gate.py`. Mutation-verified against the shipped code: **INCOMPLETE 3/3 on each model**, reasons quoting `That is`. Passes on this branch, both models.
- Full unit suite: **2,161 passed**.
- Fixture is prose, not a table — table and search-result shapes were tried first and discarded because their *unclipped controls* also returned INCOMPLETE (invented-description / unverified-source grounds), so they could not isolate the clip.

## Security check

Performed. No new surface: this changes how much of an already-in-context transcript is shown to an already-existing internal call — no new input, no new endpoint, no deserialization, no path handling, nothing logged. The elision marker is built from an integer character count (`len(text) - head - tail`) and never from user text, so it cannot be spoofed into the transcript by construction. No credentials or secrets touched.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -520,6 +520,7 @@ def _render_verify_transcript(
     max_tool: int = 12,
     tool_cap: int = 400,
     text_cap: int = 2000,
+    final_cap: int = 12000,
 ) -> str:
     """Render a compact, text-only view of the recent conversation for the
     completion verifier.
@@ -535,9 +536,36 @@ def _render_verify_transcript(
     injections are dropped before budgeting so they don't consume slots. Keeps
     the call cheap and free of tool_use/tool_result pairing constraints
     (ENG-716).
+
+    EVERY cap here goes through ``_clip_keep_cause``, so no text reaches the
+    verifier cut without an explicit ``[... N chars elided ...]`` marker. The
+    conversational path used to use a bare ``text[:text_cap]``, which handed the
+    verifier a reply ending mid-word with nothing saying it had been cut — and,
+    asked whether a complete answer was delivered, the verifier correctly
+    reported what it was shown. It was not judging badly; it was reasoning from
+    a false premise. 31% of anton's replies exceed 2,000 chars, and 44% of all
+    production continuations carried a truncation-worded reason (ENG-1633).
+
+    Two budgets, because the verifier is judging exactly ONE message:
+    ``final_cap`` for the last assistant turn (the reply under judgment) and
+    ``text_cap`` for everything else.
+
+    A flat cap is NOT sufficient, and this is the part that is easy to get
+    wrong. Measured live: with one cap for every entry, a reply that opens "all
+    three done" and buries "(3) I could NOT send the email" in the elided middle
+    is graded COMPLETE 3/3 — while both the shipped code and an unclipped
+    control catch it. Head-clipping deletes the tail; a flat both-ends clip
+    deletes the middle. Any fixed window has a hole, so the window for the one
+    message actually being judged has to be wide enough that the hole rarely
+    opens. Raising the cap for that entry alone also bounds the added cost to a
+    single message rather than ``max_convo`` of them: +276 tokens/turn on
+    average at 12,000, which sends 97.2% of real replies whole.
     """
-    convo: list[tuple[int, str]] = []   # (orig_index, line) — user/assistant text
-    tools: list[tuple[int, str]] = []   # (orig_index, line) — tool activity
+    # (orig_index, speaker, unclipped text) — conversational entries are held
+    # raw until the loop ends, because which one is "last" (and therefore gets
+    # ``final_cap``) is not known until then.
+    convo: list[tuple[int, str, str]] = []
+    tools: list[tuple[int, str]] = []   # (orig_index, line) — already rendered
 
     for i, msg in enumerate(history):
         role = msg.get("role")
@@ -547,7 +575,7 @@ def _render_verify_transcript(
             text = content.strip()
             if not text or (role == "user" and text.startswith("SYSTEM:")):
                 continue
-            convo.append((i, f"{speaker}: {text[:text_cap]}"))
+            convo.append((i, speaker, text))
         elif isinstance(content, list):
             # Assistant text emitted alongside a tool call is preamble/narration
             # ("Processing step 4"), not a conversational turn — route it to the
@@ -564,17 +592,40 @@ def _render_verify_transcript(
                     text = (block.get("text") or "").strip()
                     if not text:
                         continue
-                    bucket = tools if preamble else convo
-                    bucket.append((i, f"{speaker}: {text[:text_cap]}"))
+                    if preamble:
+                        # A tool_use rides with this text, so it is never the
+                        # reply under judgment — clip it now, at the small cap.
+                        tools.append(
+                            (i, f"{speaker}: {_clip_keep_cause(text, text_cap)}")
+                        )
+                    else:
+                        convo.append((i, speaker, text))
                 elif btype in ("image", "image_url"):
-                    convo.append((i, f"{speaker}: [image]"))
+                    convo.append((i, speaker, "[image]"))
                 elif btype == "tool_use":
                     tools.append((i, f"ASSISTANT called tool: {block.get('name')}"))
                 elif btype == "tool_result":
                     rendered = _render_tool_result_content(block.get("content"), tool_cap)
                     tools.append((i, f"TOOL RESULT: {rendered}"))
 
-    kept = convo[-max_convo:] + tools[-max_tool:]
+    # The reply under judgment is the most recent ASSISTANT entry in the
+    # conversational budget — the only one that gets ``final_cap``. Computed
+    # over the whole list rather than the kept tail because it is at or near the
+    # end by construction, so the ``max_convo`` slice below never drops it.
+    judged = max(
+        (n for n, (_, speaker, _) in enumerate(convo) if speaker == "ASSISTANT"),
+        default=-1,
+    )
+    convo_lines = [
+        (
+            i,
+            f"{speaker}: "
+            f"{_clip_keep_cause(text, final_cap if n == judged else text_cap)}",
+        )
+        for n, (i, speaker, text) in enumerate(convo)
+    ]
+
+    kept = convo_lines[-max_convo:] + tools[-max_tool:]
     kept.sort(key=lambda entry: entry[0])
     return "\n".join(line for _, line in kept) or "(no conversation)"
 
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -513,6 +513,12 @@ def _render_tool_result_content(content, cap: int) -> str:
     return _clip_keep_cause(str(content), cap)
 
 
+# Stands in for a multimodal block in the verify transcript. Named because
+# the judged-entry selection below has to be able to tell it apart from a
+# real reply.
+_IMAGE_PLACEHOLDER = "[image]"
+
+
 def _render_verify_transcript(
     history: list[dict],
     *,
@@ -601,31 +607,39 @@ def _render_verify_transcript(
                     else:
                         convo.append((i, speaker, text))
                 elif btype in ("image", "image_url"):
-                    convo.append((i, speaker, "[image]"))
+                    convo.append((i, speaker, _IMAGE_PLACEHOLDER))
                 elif btype == "tool_use":
                     tools.append((i, f"ASSISTANT called tool: {block.get('name')}"))
                 elif btype == "tool_result":
                     rendered = _render_tool_result_content(block.get("content"), tool_cap)
                     tools.append((i, f"TOOL RESULT: {rendered}"))
 
-    # The reply under judgment is the most recent ASSISTANT entry in the
-    # conversational budget — the only one that gets ``final_cap``. Computed
-    # over the whole list rather than the kept tail because it is at or near the
-    # end by construction, so the ``max_convo`` slice below never drops it.
+    # Budget only what survives, and pick the judged entry from that same
+    # window — chosen over the full list, ``final_cap`` could land on an entry
+    # the slice then discards. Slicing first also stops us clipping entries that
+    # are about to be thrown away.
+    kept_convo = convo[-max_convo:]
+    # The reply under judgment is the most recent ASSISTANT *text* entry. One
+    # message contributes one entry per content block, so an image placeholder
+    # trailing the reply would otherwise be picked as "the last assistant
+    # entry", silently dropping the reply itself back to ``text_cap`` — the flat
+    # cap this function exists to avoid (caught in self-review on #364).
     judged = max(
-        (n for n, (_, speaker, _) in enumerate(convo) if speaker == "ASSISTANT"),
+        (
+            n
+            for n, (_, speaker, text) in enumerate(kept_convo)
+            if speaker == "ASSISTANT" and text != _IMAGE_PLACEHOLDER
+        ),
         default=-1,
     )
-    convo_lines = [
+    kept = [
         (
             i,
             f"{speaker}: "
             f"{_clip_keep_cause(text, final_cap if n == judged else text_cap)}",
         )
-        for n, (i, speaker, text) in enumerate(convo)
-    ]
-
-    kept = convo_lines[-max_convo:] + tools[-max_tool:]
+        for n, (i, speaker, text) in enumerate(kept_convo)
+    ] + tools[-max_tool:]
     kept.sort(key=lambda entry: entry[0])
     return "\n".join(line for _, line in kept) or "(no conversation)"
 
--- a/tests/test_verifier_eval_gate.py
+++ b/tests/test_verifier_eval_gate.py
@@ -316,6 +316,10 @@ def test_single_valued_cases_still_demand_that_exact_verdict():
         "genuine_question": "WAITING",
         "environment_wall": "STUCK",
         "stopped_partway": "INCOMPLETE",
+        # ENG-1633: INCOMPLETE here IS the incident — a complete reply
+        # failed as "truncated" because the TRANSCRIPT was clipped, not
+        # the answer.
+        "long_complete_reply": "COMPLETE",
     }
     for name, verdict in single.items():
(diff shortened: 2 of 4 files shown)


alecantu7 (commented):
Self-review before requesting human review (convention 6). Two findings, both mine, both sharing one fix — plus the checks that failed to kill them.

**Not findings, but worth recording as checked:** no consumer of `_render_verify_transcript` / `_clip_keep_cause` / `_build_verify_request` exists outside `anton/` (grepped all of MindsDB_Repos), and nothing anywhere parses the rendered transcript shape — so there is no cross-repo contract here. No process boundary is crossed, so an old cowork-server with a new anton is unaffected. `verdict-eval` genuinely executed rather than skipping (`collected=13 executed=13 skipped=0 out_of_money=0`), and all five pre-existing eval cases still pass, which is the evidence the change did not disturb existing verdict behaviour.

Follow-up commit addressing both is on its way.

anton/core/session.py
```diff
@@ -564,17 +592,40 @@ def _render_verify_transcript(
...
                     rendered = _render_tool_result_content(block.get("content"), tool_cap)
                     tools.append((i, f"TOOL RESULT: {rendered}"))
 
-    kept = convo[-max_convo:] + tools[-max_tool:]
+    # The reply under judgment is the most recent ASSISTANT entry in the
+    # conversational budget — the only one that gets ``final_cap``. Computed
+    # over the whole list rather than the kept tail because it is at or near the
+    # end by construction, so the ``max_convo`` slice below never drops it.
+    judged = max(
+        (n for n, (_, speaker, _) in enumerate(convo) if speaker == "ASSISTANT"),
```
alecantu7: **Finding 1 — low, confirmed by execution, introduced by this PR.** A trailing non-text block steals `final_cap` from the reply being judged.

`judged` is the last ASSISTANT entry in `convo`, but one assistant message can contribute **several** entries — one per content block. A trailing `image` block becomes "the judged entry", and the actual reply text falls back to `text_cap`.

Executed, with a 5,263-char reply (over `text_cap`, under `final_cap`):

```
[1] string reply appended last (production shape)   judged reply whole: True
[2] assistant content = [text, image]               judged reply whole: False   <- steals it
[3] assistant content = [text, text]                first block whole:  False
```

When it fires, the reply degrades to the flat-cap behaviour — the exact variant measured producing a **false COMPLETE** on the buried-admission fixture.

**Refutation attempted, and it largely succeeded, which is why this is low rather than high.** Every assistant-append site was traced: `:3078` and `:4029` are the only list-shaped builders and both emit at most *one* text block followed by `tool_use` blocks, so anton never constructs `[text, image]` or `[text, text]`. The production verify path appends a plain string (`reply = llm_response.content or ""`, `:4517-4518`) immediately before the verdict call — which is why case [1] passes. Reachable only via host-supplied history (`cloud_turn/session.py:294` takes `request.history` from the request). And even when it fires the result still beats `main`, since the marker and tail survive — a lost improvement, not a regression.

**Fix** — choose `judged` inside the kept window and skip placeholders. This also lets the slice happen *before* rendering, so we stop clipping entries that are then discarded:

```python
kept_convo = convo[-max_convo:]
judged = max(
    (n for n, (_, sp, text) in enumerate(kept_convo)
     if sp == "ASSISTANT" and text != _IMAGE_PLACEHOLDER),
    default=-1,
)
```

**Test that fails before and passes after:** assistant content `[{"type": "text", "text": <3k reply>}, {"type": "image"}]` -> assert the reply appears whole.

anton/core/session.py
```diff
@@ -564,17 +592,40 @@ def _render_verify_transcript(
...
                     tools.append((i, f"ASSISTANT called tool: {block.get('name')}"))
                 elif btype == "tool_result":
                     rendered = _render_tool_result_content(block.get("content"), tool_cap)
                     tools.append((i, f"TOOL RESULT: {rendered}"))
 
-    kept = convo[-max_convo:] + tools[-max_tool:]
+    # The reply under judgment is the most recent ASSISTANT entry in the
+    # conversational budget — the only one that gets ``final_cap``. Computed
+    # over the whole list rather than the kept tail because it is at or near the
+    # end by construction, so the ``max_convo`` slice below never drops it.
```
alecantu7: **Finding 2 — low, confirmed by execution. The comment is introduced by this PR; the behaviour it describes is pre-existing.**

This says the `max_convo` slice "never drops" the judged entry. It can. Twelve trailing user image blocks push it out of the `[-max_convo:]` window.

**Refutation attempted -> the behaviour is genuinely pre-existing.** Rather than reasoning about it, I loaded `origin/main`'s renderer and ran the same history through it:

```
MAIN renderer, 12 trailing user image blocks:
  reply present at all: False   -> main drops it too
```

So there is no regression here and nothing blocking. But it is a load-bearing claim in code I just wrote, and it is false. Finding 1's fix makes it true rather than requiring the claim to be hedged, so the two should land together.

alecantu7:
Both self-review findings addressed in f7695e41.

The judged entry is now selected from `convo[-max_convo:]` rather than the full list, skipping the image placeholder (now a named constant, so the two sites can't drift). That closes Finding 1 and makes Finding 2's comment true rather than hedged — the slice can no longer discard the entry that was given `final_cap`. Slicing before rendering also stops clipping entries that are immediately thrown away, so the fix is smaller than what it replaced.

Mutation-verified: dropping the `!= _IMAGE_PLACEHOLDER` guard reds `test_a_trailing_image_block_does_not_steal_the_judged_budget` and nothing else.

**Accepted, not fixed:** an assistant message split across several text blocks gives `final_cap` only to the last one. No code path produces that shape — both list-shaped assistant appends emit at most one text block plus `tool_use` blocks — and the last block is the reply's tail, which is the part the verifier is judging.

Full suite 2,162 passed. Ready for human review; leaving the merge (and the follow-up `main` → `staging` merge) to a human per the guardrail.

alecantu7:
All three review findings verified independently and addressed in a07766aa. Thank you — #1 in particular caught a real hole in my own verification.

**1 — unpinned slice-alignment half. Confirmed, and it was the finding that mattered.** I reproduced the mutation rather than taking it on trust: `enumerate(kept_convo)` → `enumerate(convo)` survived all 18 tests, and it is behaviour-changing — on an 18-entry history with a 5.2k-char reply the reply went from whole to clipped with a marker. Your diagnosis of the mechanism is exactly right: `judged` indexes the sliced window, so computing it over the full list compares it against a different enumeration and `final_cap` lands on nothing.

The gap was structural: every case in the file built ≤10 conversational entries, at or under `max_convo`, so none of them *could* see it — and long threads are precisely where the oversized replies and the retry loops live. `test_the_judged_reply_survives_a_thread_longer_than_the_convo_budget` now builds 18 entries and reds under that mutation and no other. Both halves are pinned by one test each, so the ticket's mutation-verified clause now holds for all seven rather than six.

**2 — the back-merge instruction. Confirmed, and I was wrong twice.** `sync-main-to-staging.yml` triggers on `push: branches: [main]`, its header names this exact case, and it succeeded on the last three pushes including the ENG-1632 hotfix (`32d30652`). PR body corrected to "confirm it ran rather than merging by hand". The source was a stale note of mine asserting the manual step; I have corrected that note too, since the failure mode was repeating an operational instruction from memory without re-checking the repo.

**3 — name outran the assertion. Agreed.** Renamed to `test_buried_failure_admission_survives_under_final_cap` and the comment now says why the fixture sits under the cap deliberately: above it the middle really is elided and the admission really is lost. I did not add a companion asserting the residual — a test asserting a known limitation persists would red on an improvement rather than a regression, and the boundary is already pinned by `test_oversize_judged_reply_keeps_both_ends_and_marks_the_elision`. Say the word if you'd rather have it explicit.

Also noting your refutations for the record — the cost-duplication hypothesis was the right thing to check first, and `request_header` echoing only the user's request is what kills it.

alecantu7:
Self-review complete and all findings addressed — ready for human review.

**Three rounds of findings, all verified by execution before acting on them:**
- My own self-review: a trailing non-text block stealing `final_cap` from the judged reply, and a comment asserting a false invariant (f7695e41).
- External review: the slice-alignment half of that fix was unpinned — the mutation survived all 18 tests while measurably changing behaviour on threads longer than `max_convo`; a wrong back-merge instruction in the body; and a test name that outran its assertion (a07766aa).

Both halves of the judged-entry fix are now pinned by one test each, so the ticket's mutation-verified clause holds for all seven new tests rather than six.

**State:** 3 commits, 11/11 checks green, `verdict-eval` executed for real (`collected=13 executed=13 skipped=0`) with all five pre-existing cases still passing.

**For whoever merges:** this targets `main`, so merging ships to production immediately. The `main` → `staging` back-merge is automated by `sync-main-to-staging.yml` — confirm it ran rather than doing it by hand. And publishing anton is not delivery: desktop users get this only when cowork-server next bumps its `anton-agent` pin.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Re-reviewed `_render_verify_transcript` against its single production call site and independently reproduced the three mutation claims the PR rests on. The judged-reply selection is correct for the production path, the common short-reply path is unchanged, and every load-bearing claim held. No actionable findings.

### What I verified independently

**The core invariant is real, not assumed.** The verifier has one caller — `_build_verify_request` → `_render_verify_transcript(self._history)`, defaults only. At that point the assistant's final reply was just appended to `history` (`session.py:4620`), so it is always the last conversational entry. The judged reply therefore can never be evicted from the `convo[-10:]` window in production. The window-eviction edge the commits note needs trailing content *after* the reply, which only a host-supplied `initial_history` can produce, and `origin/main` already behaves identically there — so it is not a regression.

**Reproduced every mutation claim** (mutating the real source, running the suite, restoring):
- Faithful pre-fix bug (`text[:text_cap]`, no `final_cap`) → reds all 8 new deterministic tests.
- Flat-cap variant (`_clip_keep_cause` at one cap) → reds `test_buried_failure_admission_survives_under_final_cap`, the test that justifies the two-budget design.
- Slice-alignment mutation (`enumerate(convo)` vs `enumerate(kept_convo)`) → reds exactly one test, only visible on the 18-entry thread case.

**No collateral change to already-whole replies.** For any reply ≤ `text_cap`, `_clip_keep_cause` returns it whole — identical to the old `text[:2000]`. Only replies in (2000, 12000] change (now whole) and those > 12000 (now marked both-ends instead of silently head-clipped).

### Validation
- `test_verify_transcript.py` + `test_verifier_eval_gate.py` → 35 passed on head `a07766aa`.
- Mutation matrix run against live source (above) — the "mutation-verified" clause is accurate.
- CI green on the head SHA including `run-tests` and `verdict-eval` (the live verdict-quality gate that would catch a verdict regression).
- `_render_verify_transcript` is byte-identical on `origin/main` and `origin/staging`, so the automated back-merge premise holds.

### Merge-readiness (hotfix)
Directionally risk-reducing: the old failure mode (false INCOMPLETE → retry → ~137k billed tokens) is strictly worse than any plausible new failure, the added cost is bounded to one message and cannot compound, and rollback is a one-commit revert of a self-contained pure function. The accepted residual (a >12k-char reply with decisive negative evidence in the elided middle) is genuine, strictly better than today's 31% tail-loss, pinned by the oversize test, and documented. It does not block.

Reviewed with Claude Code.

alecantu7:
Thanks — and one correction to my own comment above: pnewsam's count is the right one. There are **8** new deterministic tests, not seven; my "all seven rather than six" was off by one (6 in the first commit, +1 for the image-placeholder case, +1 for the long-thread case). The substance is unchanged — every one of the 8 reds under the faithful pre-fix mutation, and both halves of the judged-entry fix are pinned by exactly one test each.

Approval is anchored to the current head (`a07766aa`), so it is live rather than carried over from an earlier push.

anton/core/session.py
alecantu7: Fixed in f7695e41. `judged` is now selected from `convo[-max_convo:]` with the image placeholder excluded (`_IMAGE_PLACEHOLDER`, a named constant so the produce and compare sites cannot drift). Pinned by `test_a_trailing_image_block_does_not_steal_the_judged_budget`, which reds when the guard is dropped and nothing else does. Independently reproduced by pnewsam.

anton/core/session.py
alecantu7: Fixed in f7695e41 — the claim is now true rather than hedged. Selecting `judged` from the kept window means the slice can no longer discard the entry that was given `final_cap`, so the comment describes what the code actually guarantees. The separate slice-alignment risk this exposed was then pinned in a07766aa by `test_the_judged_reply_survives_a_thread_longer_than_the_convo_budget` (18 entries), after review caught that half as unpinned.

PR #366, opened 2026-08-18
ENG-1100: fix build-html-dashboard/build-fullstack-backend SKILL.md guidance on uploaded assets

StpMax:
## Summary
- Resolve the contradiction in `build-html-dashboard/SKILL.md` between the "MUST be a single .html file" line and the existing sibling-file exception for large JSON, and add an "UPLOADED IMAGES AND BINARY ASSETS" section that tells the agent where uploaded/pasted files live and how to bring them into the artifact folder (inline as `data:` URI under 100KB, copy as sibling file at/above 100KB).
- Clarify in `build-fullstack-backend/SKILL.md` that any uploaded/pasted asset brought into a fullstack artifact must live under `<artifact_path>/static/`, not the artifact root, and make the html-dashboard snippet's copy destination context-aware so it doesn't hardcode the artifact root (which would break publish for fullstack builds).
- Text-only changes to two `SKILL.md` prompt files — no application code.

Part of [ENG-1100](https://linear.app/mindsdb/issue/ENG-1100/pasted-clipboard-images-all-named-imagepng-artifact-skill-doesnt). Companion PR in `cowork` handles the clipboard-paste filename side of the ticket.

## Test plan
- [x] `tests/test_builtin_skills.py` — 18/18 passing
- [x] Full suite (`pytest -q -m "not slow"`) — 2205 passed, 28 skipped, 0 failures
- [x] Python snippet in the new SKILL.md section validated with `ast.parse`
- [x] No Cyrillic text in either changed file (grep-verified)
- [x] Single 100KB threshold used consistently across both files, no stray second number


ea-rus (approved):
Worth checking:

Contradictions an agent resolves arbitrarily

- build-fullstack-backend/SKILL.md:140 — "Include all CSS and JS inlined (no external file references)" survives two lines above the line this PR edits to mandate sibling assets under static/. That file also contains no threshold at all, so the PR body's "single 100KB threshold used consistently across both files" is not accurate — its effective threshold is 0.
- build-html-dashboard/SKILL.md:52 — "If the user refers to a file whose path you cannot see, ask them for it" directly contradicts the harness system prompt, which says "Do not ask them to type or paste a filesystem path; you are not permitted to read files that are neither in the project nor attached" (harness.py:278-285). The recalled skill lands later in context and reads as the more specific rule. Suggest: "…ask them to attach it to the conversation — do not ask for a filesystem path."

StpMax:
Both findings are valid, fixed in 91ea8fdb:

- `build-fullstack-backend/SKILL.md`: scoped the "additional frontend assets... MUST live under static/" rule to assets that aren't inlined into `index.html`, so it no longer contradicts "Include all CSS and JS inlined (no external file references)" two lines above.
- `build-html-dashboard/SKILL.md`: replaced "ask them for it" with "ask them to attach it to the conversation — never ask for a filesystem path", matching the harness's hardened rule against requesting filesystem paths (harness.py:303-310, post-ENG-1357).

Thanks for catching these — same class of latent contradiction this PR was already fixing elsewhere in these files.

PR #367, opened 2026-08-18
ENG-1693: contain both raising httpx properties so the billing-origin fallback works

alecantu7:
Follow-up to **#363**, which merged approved with two review nits from @pnewsam outstanding. Targets `staging` (#363's base), not a hotfix.

## Nit 2 — the ordering issue was real, and understated

`response_origin_host`'s own docstring says:

> `httpx.Response.url` is a property that RAISES when no request is attached, so this cannot be a bare `getattr` chain.

The code was a bare `getattr` chain. `getattr(resp, "url", None)` does **not** rescue a `RuntimeError` — its default only covers `AttributeError`. The exception reached the outer `except Exception: return None`, which means *"origin unknown"* — a state this gate deliberately **trusts** — so a request-less response shadowed a foreign host carried on `exc.request`, and a third party could select the credits card again.

**The review named `.url`. `httpx.Response.request` has the identical trap.** Guarding only `.url` still fails the new test, because the second fallback was `getattr(getattr(resp, "request", None), "url", None)` and that inner `getattr` triggers the same `RuntimeError`. Verified rather than assumed, against httpx 0.28.1:

```
getattr(resp,'url',None)     RAISED RuntimeError
getattr(resp,'request',None) RAISED RuntimeError
```

Both reads are now contained individually so each fallback stays reachable, and the docstring's existing claim becomes true.

Agreed with the review that this is **not reachable through the real SDK paths today** — request-time responses always carry a request, mid-stream errors have `.response is None`. This is defense-in-depth on a trust boundary where the failure direction is "silently trust a hostile origin".

## Nit 1 — the allowlist question, answered and pinned in code

The question was whether the production gateway is *always* served from `mindshub.ai` or `mdb.ai`, including regional and vanity hosts, since a missing apex fails closed to generic copy and reopens ENG-1169's user-visible bug.

**Answer: confirmed complete.** Checked against the terraform host inventory — the source of truth for zones and certs — plus every base URL configured across `anton`, `cowork`, `cowork-server`, `mdb-ai` and `mindshub_inference`:

| Surface | Host | Covered |
| -- | -- | -- |
| prod inference | `api.mindshub.ai` | ✅ |
| staging / dev | `api.staging.mindshub.ai`, `api.dev.mindshub.ai` | ✅ |
| ephemeral per-PR envs | `api-pr123.dev.mindshub.ai` | ✅ |
| legacy vanity gateway | `llm.mdb.ai`, `llm.staging.mdb.ai` | ✅ |
| white-label surfaces | `writer.mdb.ai`, `terabase.dev.mdb.ai`, `view.mindshub.ai` | ✅ |

Terraform declares **no other apex zone**, and the wildcards `*.mindshub.ai` / `*.mdb.ai` cover anything added later beneath them. I also checked the Kinaxis white-label branch for a customer-owned gateway domain — there isn't one.

Rather than leave that answer in a PR comment where it goes stale, those nine hosts are now a parametrised test. Narrowing the allowlist fails there instead of silently degrading a real out-of-credits denial.

**`4nton.ai` is the one real production zone deliberately excluded.** It serves agent provisioning, per-instance hosts (`sp_<hash>.4nton.ai`, `cw-<id>.4nton.ai`) and artifact publishing — never LLM inference. Its exclusion is now asserted, and the comment on `_MINDSHUB_HOSTS` records that it must be added if inference is ever routed onto it, so the cost of failing closed isn't discovered in production.

## Test plan

Three new cases in `tests/test_status_error_mapper.py`, all mutation-verified:

| mutation | reds |
| -- | -- |
| revert to the merged bare `getattr` chain | `test_a_request_less_response_does_not_shadow_a_foreign_host` |
| guard only `.url` (the incomplete fix) | same test — it discriminates the complete fix from a plausible one |
| narrow the allowlist to `mindshub.ai` only | `test_every_real_gateway_host_is_trusted[writer.mdb.ai]`, `[terabase.dev.mdb.ai]` |

The ordering test asserts its own premise (`pytest.raises(RuntimeError)` on the bare response) so it can't silently stop testing anything if httpx changes that behaviour.

`tests/test_status_error_mapper.py` 80 passed · touched-area suite 487 passed.

## Security check

Performed, and this change *is* the security surface. It widens the reachability of the origin fallback, so a hostile host that was previously resolved as "unknown" (trusted) is now resolved and correctly distrusted.

**Being precise about the direction, because it is the one that costs a user something.** The change can only move a verdict from trusted to distrusted, never the reverse — and trusted→distrusted is the *card-losing* direction, the one ENG-1169 exists to prevent. Executed on the only path this PR changes:

```
request-less response + RELAY    OLD  card=shown  ->  NEW  card=LOST
request-less response + GATEWAY  OLD  card=shown  ->  NEW  card=shown
```

That is safe for a specific reason rather than by construction: a relay in front of MindsHub **already** loses the card in the normal `response+request` path, as ENG-1693's "Accepted tradeoff" records. So this closes an accidental exception to a decision already taken instead of introducing a new loss. Self-review caught the original wording, which presented this as unconditionally harmless. No new input is parsed, nothing is logged, and the host comparison remains exact-domain-or-dot-delimited-subdomain, so a lookalike like `mindshub.ai.evil.com` is still rejected (existing test). The allowlist is unchanged in value; only its documentation and test coverage grew.

### DIFF
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -576,6 +576,26 @@ _WALLET_DENIAL_CODES = frozenset({"wallet_empty", "included_allowance_exhausted"
 
 # Hosts that ARE the MindsHub gateway. Only a response from one of these may
 # select a billing verdict — see :func:`origin_is_known_third_party` (ENG-1693).
+#
+# VERIFIED COMPLETE for every host that can serve a gateway billing denial
+# (review question on #363, checked 2026-08-18 against the terraform host
+# inventory, which is the source of truth for zones and certs). Two apexes cover
+# it because every environment and vanity host is a subdomain of one of them:
+# prod `api.mindshub.ai`, `api.staging.mindshub.ai`, `api.dev.mindshub.ai`,
+# per-PR envs `api-pr<N>.dev.mindshub.ai`, and the white-label surfaces
+# `llm.mdb.ai`, `llm.staging.mdb.ai`, `writer.mdb.ai`, `terabase.dev.mdb.ai`,
+# `view.mindshub.ai`. Terraform declares no other apex zone, and the wildcards
+# `*.mindshub.ai` / `*.mdb.ai` cover anything added later under them. The
+# positive cases are pinned in tests/test_status_error_mapper.py so this
+# paragraph cannot quietly go stale.
+#
+# `4nton.ai` is a real production zone and is DELIBERATELY absent: it serves
+# agent provisioning and per-instance hosts (`sp_<hash>.4nton.ai`,
+# `cw-<id>.4nton.ai`) plus artifact publishing, never LLM inference. If an
+# inference endpoint is ever routed onto it, it MUST be added here — otherwise a
+# genuine out-of-credits denial from that host silently degrades to generic copy
+# and reopens ENG-1169's user-visible bug. Failing closed is the right default;
+# this note exists so the cost of that default is not discovered in production.
 _MINDSHUB_HOSTS = ("mindshub.ai", "mdb.ai")
 
 
@@ -611,9 +631,28 @@ def response_origin_host(exc: BaseException) -> str | None:
     """
     resp = getattr(exc, "response", None)
     try:
-        url = getattr(resp, "url", None) if resp is not None else None
-        if url is None and resp is not None:
-            url = getattr(getattr(resp, "request", None), "url", None)
+        url = None
+        if resp is not None:
+            # BOTH `httpx.Response.url` and `httpx.Response.request` are
+            # properties that RAISE RuntimeError when no request is attached, and
+            # `getattr(resp, name, None)` does NOT rescue that — getattr's
+            # default only covers AttributeError. Either one reaching the outer
+            # handler returned None, i.e. "origin unknown", a state this gate
+            # deliberately TRUSTS — so a request-less response shadowed a
+            # foreign host sitting on `exc.request`. Each read is contained
+            # individually so the next fallback stays reachable.
+            #
+            # The review nit on #363 named `.url`; `.request` has the identical
+            # trap, and guarding only `.url` still failed the test below.
+            try:
+                url = resp.url
+            except Exception:
+                url = None
+            if url is None:
+                try:
+                    url = resp.request.url
+                except Exception:
+                    url = None
         if url is None:
             url = getattr(getattr(exc, "request", None), "url", None)
         if url is None:
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -577,10 +577,12 @@ _WALLET_DENIAL_CODES = frozenset({"wallet_empty", "included_allowance_exhausted"
 # Hosts that ARE the MindsHub gateway. Only a response from one of these may
 # select a billing verdict — see :func:`origin_is_known_third_party` (ENG-1693).
 #
-# VERIFIED COMPLETE for every host that can serve a gateway billing denial
-# (review question on #363, checked 2026-08-18 against the terraform host
-# inventory, which is the source of truth for zones and certs). Two apexes cover
-# it because every environment and vanity host is a subdomain of one of them:
+# VERIFIED COMPLETE for every host MindsHub ITSELF serves (review question on
+# #363, checked 2026-08-18 against the terraform host inventory, which is the
+# source of truth for zones and certs). Deliberately not the broader claim
+# "every host that can serve a gateway billing denial" — see the relay note
+# below, where that is false by design. Two apexes cover the served set because
+# every environment and vanity host is a subdomain of one of them:
 # prod `api.mindshub.ai`, `api.staging.mindshub.ai`, `api.dev.mindshub.ai`,
 # per-PR envs `api-pr<N>.dev.mindshub.ai`, and the white-label surfaces
 # `llm.mdb.ai`, `llm.staging.mdb.ai`, `writer.mdb.ai`, `terabase.dev.mdb.ai`,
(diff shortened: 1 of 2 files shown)


alecantu7 (commented):
Self-review before requesting human review (convention 6). Three findings, all **low**, and all about *claims* rather than behaviour — one code comment, one test docstring, one line of this PR description. No logic changes: the 11 non-comment logic lines and all 75 test assertion lines are unaffected.

Worth recording, because it nearly became a fourth finding: I could not find any test covering the **anthropic** mapper's origin gate, which is exactly the gap ENG-1693's `Done when` #4 flags. Rather than report it, I mutation-tested — deleting `origin_is_known_third_party` from `anthropic.py` reds `test_the_anthropic_twin_gates_the_same_carriers[402-body0]` and `[429-body1]`. Coverage exists, `Done when` #4 is met, and my greps were simply wrong. The mutation is the only reason that didn't get filed as noise.

Also refuted, with executed evidence: the concern that this PR's wider fallback regresses ENG-1169's credits card. It does change one path (`request-less response + relay`: card shown → card LOST), but the relay **already** loses the card in the normal `response+request` path by the accepted tradeoff recorded on the ticket — so the change makes an accidental exception consistent with a decision already taken, rather than introducing a loss. That is a stronger argument than the one currently in the PR body, which is finding 2.

Fix commit follows.

anton/core/llm/provider.py
```diff
@@ -576,6 +576,26 @@ def __init__(
 
 # Hosts that ARE the MindsHub gateway. Only a response from one of these may
 # select a billing verdict — see :func:`origin_is_known_third_party` (ENG-1693).
+#
+# VERIFIED COMPLETE for every host that can serve a gateway billing denial
```
alecantu7: **Finding 1 — low, confirmed, introduced by this PR.** This comment overclaims what was actually verified.

> VERIFIED COMPLETE for every host that can serve a gateway billing denial

ENG-1693's own **"Accepted tradeoff"** section contradicts that:

> a user pointed at a corporate relay or LLM proxy that forwards our denials resolves a non-MindsHub host … `relay, wallet 402 -> ConnectionError (credits card LOST)`

A relay **does** serve genuine gateway billing denials, from a host deliberately not on this list. Executed, identically on `origin/staging` and this head:

```
relay, response+request   OLD  host=llm-proxy.corp.example  distrusted=True  card=LOST
relay, response+request   NEW  host=llm-proxy.corp.example  distrusted=True  card=LOST
```

**Refutation attempted:** I checked whether "gateway billing denial" could be read narrowly enough to exclude a relayed one — it can't, the ticket calls it exactly that and records the card being lost. The sentence's own scope list (prod / staging / dev / per-PR / white-label) is accurate; the defect is only the universal quantifier.

**Impact:** a future reader concludes the relay shape is covered and skips the remedy the ticket names — *"a configurable trusted-host list, not a looser host match"*. This is the same class of defect as the docstring error this PR exists to fix: a comment asserting a guarantee the code does not provide.

**Fix:** scope the claim to *every host MindsHub itself serves*, and cross-reference the accepted tradeoff so the relay case reads as a known, deliberate exclusion rather than an oversight.

tests/test_status_error_mapper.py
```diff
@@ -904,3 +904,78 @@ def test_a_relayed_gateway_rate_limit_still_earns_the_session_wait():
...
+    assert is_mindshub_host(host) is True, (
+        f"{host} is a real MindsHub host but the allowlist rejects it — a genuine "
+        f"out-of-credits denial from it would lose the credits card (ENG-1169)"
+    )
+
+
+def test_the_agent_instance_zone_is_deliberately_untrusted():
+    """`4nton.ai` is a real production zone that serves agent provisioning and
+    per-instance hosts, never inference — so it must NOT be able to select a
+    billing verdict. If inference is ever routed onto it, this test is the
```
alecantu7: **Finding 3 — low, confirmed, introduced by this PR.** The docstring calls this test a tripwire and then admits it isn't one.

> If inference is ever routed onto it, this test is the tripwire: it will still pass

A test that still passes is not a tripwire. This is a current-state assertion — `4nton.ai` is untrusted today and should be — which is worth having. The actual tripwire is the note on `_MINDSHUB_HOSTS` telling a future author to add the host.

Harmless in effect, but it's a claim about our own safety net that doesn't hold, and this PR is specifically about not leaving those lying around. **Fix:** drop the tripwire framing and point at the comment that does the job.

anton/core/llm/provider.py
alecantu7: Fixed in 470b0155. The claim is now scoped to *every host MindsHub itself serves*, and the relay exclusion is recorded as the deliberate decision it is — including why the asymmetry with the ungated `_velocity` carrier is correct (a spoofed billing card asks the user for money, a spoofed wait does not), and that the remedy if relayed MindsHub becomes supported is a configurable trusted-host list rather than a looser match here.

tests/test_status_error_mapper.py
alecantu7: Fixed in 470b0155 — the tripwire framing is gone. It now says plainly that this asserts today's behaviour and that if inference is ever routed onto `4nton.ai` the test still passes while the card degrades, with the warning living in the comment on `_MINDSHUB_HOSTS`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff against current `staging` (which already contains merged #363). The one behavioral change — containing `resp.url` and `resp.request.url` in individual `try/except` blocks — is correct, verified against httpx 0.28.1, and moves verdicts only in the safe direction (unknown → distrusted). No actionable findings.

### Validation
- Confirmed the premise on httpx 0.28.1: both `httpx.Response(402).url` and `.request` raise `RuntimeError`, and `getattr(resp, "url", None)` does **not** rescue it — so the old bare-getattr chain fell through to the outer `except Exception: return None` ("origin unknown" = trusted). The bug is real.
- Traced the request-less-response path: `resp.url` raises → None, `resp.request.url` raises → None, then falls back to `exc.request.url` → `evil.example.com` → correctly distrusted. Non-raising paths are behavior-preserving.
- Ran the suite on a clean PR-head worktree: `tests/test_status_error_mapper.py` **80 passed**; broader provider/mapper/host subset **231 passed**.
- Checked both new host tests against `is_mindshub_host`: `api-pr123.dev.mindshub.ai` etc. match via dot-delimited subdomain; `4nton.ai` / `sp_abc123.4nton.ai` correctly rejected.

### Notes (non-blocking)
- Documentation-heavy — the `_MINDSHUB_HOSTS` comment grew ~30 lines. It's accurate and this is a trust boundary, so the verbosity is defensible; flagging only so it's a conscious choice.
- Genuinely defense-in-depth: as the PR states, this path isn't reachable through real SDK flows today. It closes an accidental exception rather than a live exploit — correctly framed.

Both review nits are answered, and the fixes are pinned by mutation-verified tests so they can't silently regress.

PR #371, opened 2026-08-19
[ENG-1635] scratchpad auto install any module

tino097:
https://linear.app/mindsdb/issue/ENG-1635/security-the-scratchpad-auto-installs-any-module-name-the-model

- Removed the exception-triggered pip-install-and-retry in scratchpad_boot.py — a missing import now surfaces as a normal error with a hint, no install attempt
- Only install path left is the pre-existing explicit packages array
- Added telemetry (scratchpad_package_installed, package name only) for installs that do happen
- Clarified in tool description/system prompt that get_llm/agentic_loop/web_search/sample/progress are already globals, no import needed
- Updated docs/configure/analytics.md to disclose the new event
- Fixed the hint ordering so it doesn't break root-cause classification (traceback's last line)
- Reworded SCRATCHPAD_INSTALL_NUDGE to stop claiming an install was attempted
- Deduped the install-success/failure check into shared helpers
- Marked the now-dead ENG-1275 install-span code as dead in comments

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -174,7 +174,9 @@ _pending_lock = threading.Lock()
 #
 # An event NOT listed here keeps the collector path, so moving one is an
 # explicit decision rather than something that happens by default.
-_POSTHOG_EVENTS = frozenset({"turn_completed", "rule_retrieval"})
+_POSTHOG_EVENTS = frozenset(
+    {"turn_completed", "rule_retrieval", "scratchpad_package_installed"}
+)
 
 # `$lib` names the sender, matching the convention the other emitters follow
 # (`cowork-desktop`, `mindshub-site-beacon`), so a per-emitter breakdown in
--- a/anton/core/artifacts/backend_launcher.py
+++ b/anton/core/artifacts/backend_launcher.py
@@ -115,8 +115,8 @@ async def launch_artifact_backend(
                 install_log.write(banner.encode("utf-8"))
                 install_log.write(install_result.encode("utf-8"))
                 install_log.write(b"\n")
-            if install_result.startswith("Install failed") or install_result.startswith(
-                "Install timed out"
+            if install_result.startswith(
+                ("Install failed", "Install timed out", "Install refused")
             ):
                 return (
                     "Error: dependency install failed for `requirements.txt`.\n"
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -1248,6 +1248,16 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
         needed = [p for p in packages if p.lower() not in self._installed_packages]
         if not needed:
             return "All packages already installed."
+        # Belt for every caller (the tool paths reject earlier, with a typed
+        # reason): nothing flag-, URL- or path-shaped may reach pip's argv
+        # (ENG-1635 — a single "--index-url=…" entry redirects resolution for
+        # the whole call). Lazy import: this module must stay importable
+        # without the tools layer.
+        from anton.core.utils.scratchpad import reject_invalid_packages
+
+        refused = reject_invalid_packages(needed)
+        if refused:
+            return refused
         self._ensure_venv()
 
         uv = self._find_uv()
--- a/anton/core/backends/local.py
+++ b/anton/core/backends/local.py
@@ -656,10 +656,9 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
         if uv:
             env["ANTON_UV_PATH"] = uv
 
-        # The in-cell auto-installer (scratchpad_boot) runs pip/uv under this
-        # budget. Pass the resolved setting so the worker's timer and the
-        # parent's kill windows in _read_result run off one number — they
-        # used to be independent constants that drifted apart (ENG-1275).
+        # scratchpad_boot no longer auto-installs on ModuleNotFoundError, so
+        # this var (and _read_result's install-span handling below) is dead;
+        # left wired rather than reworking the kill-window logic in one pass.
         env["ANTON_CELL_INSTALL_TIMEOUT"] = str(CoreSettings().cell_install_timeout)
 
         # Namespace snapshot path (ENG-1124). The boot script reads
@@ -1050,12 +1049,9 @@ class LocalScratchpadRuntime(ScratchpadRuntime):
         last_notice = 0.0
         last_output = 0.0
 
-        # In-cell auto-install span (ENG-1275). While the worker's installer
-        # runs, both kill windows defer to the install budget — the same
-        # cell_install_timeout the worker was handed at spawn — plus a grace
-        # margin, so the worker's own named install error wins the race
-        # against a parent kill. The cell's budget resumes, install time
-        # excluded, when the end marker arrives.
+        # Dead since scratchpad_boot dropped its in-cell auto-installer: the
+        # markers this deferral watches for are never emitted anymore. Left
+        # in place rather than unpicking it from the kill-window logic below.
         install_budget = float(s.cell_install_timeout)
         installing: str | None = None
         install_started = 0.0
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -9,6 +9,7 @@ import dill
 
 from anton.core.backends.wire import (
     CELL_DELIM,
+    MISSING_MODULE_HINT,
     RESULT_START,
     RESULT_END,
     heal_surrogate_source,
@@ -1187,12 +1188,7 @@ while True:
             # Hint goes before the traceback: callers key off its last line.
             hint = ""
             if _mnf.name:
-                hint = (
-                    f"'{_mnf.name}' was not auto-installed. If your code "
-                    "deliberately needs it, list it in the exec call's "
-                    "'packages' array (or use the scratchpad's install "
-                    "action) and retry.\n"
-                )
+                hint = MISSING_MODULE_HINT.format(name=_mnf.name)
             error = hint + traceback.format_exc()
         except Exception:
             error = traceback.format_exc()
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -822,8 +822,6 @@ import threading
 
 from anton.core.backends.wire import (
     HEARTBEAT_MARKER,
-    INSTALL_END_MARKER,
-    INSTALL_START_MARKER,
     PROGRESS_MARKER,
     STDOUT_CHUNK_MARKER,
 )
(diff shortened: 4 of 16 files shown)


alecantu7 (commented):
Reviewed at `09a8c7d3` against a fresh `origin/staging` (`322a3b40`), in a scratch worktree. The core fix is real and I **mutation-verified** it: restoring `scratchpad_boot.py` from `staging` makes `test_module_not_found_does_not_auto_install` fail with *"an undeclared import must never trigger an install"*. That's a well-designed test — the sentinel-touching stub installer is exactly the right shape. 131 tests pass across the four touched files.

One finding I'd like addressed before merge; the rest are smaller or pre-existing.

---

### 1. The new hint and nudge coach the model to re-declare the hallucinated name — medium, introduced

The hint in `scratchpad_boot.py`, and `SCRATCHPAD_INSTALL_NUDGE` in `prompts.py:458`:

> `'{name}' was not auto-installed. If your code deliberately needs it, list it in the exec call's 'packages' array (or use the scratchpad's install action) and retry.`

> `Declare the package explicitly, either in the exec call's 'packages' array or via the scratchpad's install action (double-check the PyPI name), then rerun the same cell.`

For the exact case ENG-1635 documents — `from functions import web_search`, seen across four users — the sequence becomes: `ModuleNotFoundError` → hint naming `functions` → model lists `functions` in `packages` → `pip install functions` runs. Same package, same venv, same inherited `DS_*` credentials, one model turn later. The ticket's security note called this out: *"the fix must not simply move the install behind a prompt the agent can auto-approve on the user's behalf."* There's no user confirmation and no allowlist on the surviving path, so the agent is the only approver.

**Partly mitigated, and worth crediting:** adding *"get_llm, agentic_loop, web_search, sample, and progress are already available as globals — do not import them"* to both `tool_defs.py` and `prompts.py` should suppress the specific observed hallucination, which is most of the real traffic. And the install is now a visible tool call with telemetry rather than an invisible in-cell one — a genuine improvement. The residual is the general class: any *other* hallucinated importable name still routes to a declared install.

**Smallest fix:** have the hint name the module without instructing re-declaration — say the import failed and that a package must be a real PyPI distribution the task actually needs. "double-check the PyPI name" is advice to an LLM, not a control.

### 2. No validation on `packages` — flag-shaped entries are parsed as uv/pip flags — high, pre-existing, now the sole install path

`local.py:1255`:

```python
cmd = [uv, "pip", "install", "--python", self._venv_python, *needed]
```

`packages` is `{"type": "array", "items": {"type": "string"}}` with no validation anywhere between the tool call and argv. Confirmed by execution:

```
$ uv pip install --python .venv/bin/python --dry-run "--index-url=http://127.0.0.1:9/simple" requests
error: Failed to fetch: `http://127.0.0.1:9/simple/requests/`

$ uv pip install --python .venv/bin/python --dry-run requests
 + requests==2.34.2      # control resolves normally
```

A single model-authored string controls the index every package resolves from — the same dependency-confusion outcome the ticket is about, through the path this PR leaves open. The agent reads untrusted content (web pages, files), so the string isn't purely model-internal.

Pre-existing and not this PR's regression, but ENG-1635 *is* the security review for this vector, and this PR makes it the only install path — so I think it belongs on this ticket rather than a follow-up. **Fix:** reject any entry that isn't a valid PEP-508 name/specifier before it reaches argv.

### 3. A successful install whose pip output mentions a retry is classified as failed — medium, pre-existing, exposure increased

`scratchpad.py:20`: `_INSTALL_FAILURE_MARKERS = ("Install failed", "timed out")`. Executed:

```
input:  WARNING: Retrying ... ReadTimeoutError("... Read timed out.") ...
        Successfully installed numpy-2.1.0

classified as FAILED: True
would emit telemetry: False
```

A network hiccup during an otherwise successful install returns `ToolOutcome(ok=False, reason="package_install_failed")`, so the cell never runs despite the package being present, and the install event is suppressed. The substring check is a faithful extraction of the pre-existing line, so this isn't introduced — but with auto-install gone every install now flows through it. Mitigating: uv is the usual binary and its retry wording differs; the pip fallback is the exposed case.

### 4. Two new tests restate production strings instead of importing them — low, introduced

`test_root_cause.py` hand-rolls the extraction rather than calling it:

```python
reason = error.splitlines()[-1][:160]
```

Production is `tool_handlers.py:541` — identical today, but the test would still pass if that line changed to `[0]`, which is precisely the regression it exists to prevent.

`test_undeclared_import_routes_to_install_nudge` has the same shape: it carries its own copy of the hint text. The nudge routes on the substring `"auto-install"` surviving inside the negation *"was not auto-installed"* (`session.py:1302`) — genuinely fragile. Reword the production hint to "was not installed automatically" and routing silently dies while the test stays green on its stale copy.

**Fix:** export the hint as a constant, import it in both the boot script and the test; call the production extractor in the root-cause test.

### 5. Fossil tests alongside the retained dead code — low, introduced

Two surviving tests assert nudge routing for strings the code can no longer emit:

```python
"Auto-install of 'torch' was killed after 120s (cell_install_timeout) ..."
"Auto-install failed:\nERROR: No matching distribution found for torhc"
```

They pass, but pin a mechanism that's gone. Marking `ANTON_CELL_INSTALL_TIMEOUT` (`local.py:662`) and the parent's install-span handling (`local.py:1049-1211`) as dead in comments is honest and I'd rather have that than a silent removal — but the next reader still sees live tests asserting auto-install behaviour. Worth deleting these two with the rest.

---

### Checked and cleared

- **Telemetry covers both install paths** — I initially suspected the hook was only on the `install` action; `prepare_scratchpad_exec` emits it too. Event carries `package` only, never cell contents, per the ticket's security note. `analytics_enabled` opt-out is respected, and the docs disclosure in `analytics.md` is thorough.
- **The dead marker branch is not spoofable** by cell output — it extends the watchdog kill window, so I checked: cell stdout ships JSON-encoded inside a single `STDOUT_CHUNK_MARKER` line, so a cell printing the marker can't produce a matching raw line.
- **`cell_install_timeout` is not dead config** — still live for the parent-side `install_packages`.
- **`verifier-eval` not running is correct**, not a gate gap — it filters on `session.py`/`structured.py`/`client.py`, none touched.
- **The deleted watchdog contracts were legitimately obsolete** — all four triggered installs via `pad.execute("import ...")`, i.e. the path being removed.
- **Root-cause classification is unaffected by the hint** — `acc.py:502` reads kill *reasons*, not cell error text, and the two first-line consumers read `stdout`, not `error`.

### On the ticket's `Done when`

- ✅ No unattended install on `ModuleNotFoundError` — mutation-verified.
- ✅ A test asserts a nonexistent import doesn't pip-install that name.
- ⚠️ "visible to the user" is half met — telemetry yes, but I found no UI surface (item 3). Plausibly cowork's side rather than anton's; worth saying so on the ticket so QA doesn't file it as a miss.
- ⬜ ENG-392's status reconfirmation is still open — it sets the blast radius.

Not approving yet on the strength of finding 1, but this is close and the fix there is a wording change.

alecantu7:
Pushed `f7d1c796` addressing my review findings directly, so this doesn't stall another round-trip — @tino097 please sanity-check the follow-up since it's on your branch.

**Finding 1 (the blocker)** — the hint and `SCRATCHPAD_INSTALL_NUDGE` no longer instruct re-declaring the failed name. The hint now lives as `MISSING_MODULE_HINT` in `wire.py` (the shared parent↔worker constants module) and says to question the import itself; the nudge names the real globals and says to install only a nameable, genuinely-needed PyPI distribution. Guarded by `test_hint_and_nudge_do_not_coach_redeclaring_the_name`.

**Finding 2** — `packages` entries are now gated as plain PEP 508 requirements (`packaging.Requirement`, direct-URL requirements rejected) before anything reaches pip argv: at both model-facing call sites (exec's `packages`, the `install` action) with a typed `package_install_rejected` reason (mapped `TIER_SELF` in root_cause, so a refusal can never read as a wall), plus a belt inside `LocalScratchpadRuntime.install_packages` covering every other caller (`requirements.txt` launches included). `--index-url=…`, `-e .`, `name @ https://…`, and path-shaped entries are all refused; names, pins, and extras pass.

**Finding 3** — install failure classification is anchored to the message head (`Install failed` / `Install timed out` / `Install refused`) instead of `"timed out"` anywhere, so a successful install whose pip output mentions a retried network read no longer fails the cell or suppresses the telemetry event.

**Finding 4** — nudge routing now keys on `ModuleNotFoundError` and the install-message prefixes rather than the hint's wording (`"auto-install"` kept only as a legacy matcher for old workers), so a hint reword can't silently kill the route. `cell_failure_reason` is exported and both tests call the production extractor/constant instead of restated copies.

**Finding 5** — the two fossil auto-install routing tests are deleted; the kill-during-install one is kept but reframed as an explicit legacy-worker guard.

**Verification**: full suite green locally (2259 passed, 30 skipped), and each new test was mutation-verified — old substring markers back → retry-warning test fails; validator call removed → refusal test fails; routing reverted → both routing tests fail; coaching hint restored → wording test fails; extractor flipped to first-line → root-cause hint test fails.

**Security pass over this follow-up**: the new surface is the validator itself — it treats every entry as untrusted, rejects on parse failure (fail-closed), and the refusal message echoes entries only `repr()`-quoted inside a string result, never into argv. Telemetry still carries package names only. Also reconfirmed on the ticket: ENG-392 is Canceled with both PRs closed unmerged, so the full-env inheritance (and thus the stated blast radius) is current.

alecantu7 (approved):
Re-reviewed at `f7d1c796`. Finding 1 (the blocker) is resolved — the hint/nudge no longer coach re-declaring the failed name — and findings 2–5 are addressed with mutation-verified tests; full suite green. The follow-up commit is mine, so treat this approval as covering the original change plus my own trail: @tino097, glance at `f7d1c796` before merging since it's on your branch. Merge is yours/Alejandro's call once CI is green.

pnewsam (commented):
## Code Review

**Verdict: COMMENT (no blocking issues — LGTM)**

Reviewed the full diff: the removal of exception-triggered auto-install in `scratchpad_boot.py`, the PEP 508 reject gate, the belt in `install_packages`, telemetry, nudge/root-cause routing, docs, and tests. The security objective holds and the round-2 findings are genuinely addressed. Nothing blocks merge — one low-probability nit below.

### Validation
- **Reject gate, empirically:** ran `invalid_package_spec`'s logic against 14 cases — every flag/URL/path/direct-ref/newline-smuggle form is refused; plain names, pins, extras, ranges, and env markers pass.
- **argv, not shell:** `install_packages` uses `create_subprocess_exec(*cmd)` with each entry as a discrete token, so refusing flag-shaped entries actually prevents `--index-url` redirection. No `shell=True` on the path.
- **`packaging>=21.0`** is declared (`anton/pyproject.toml`), so the lazy `Requirement` import can't fail at runtime.
- **No format-string injection:** `MISSING_MODULE_HINT.format(name=_mnf.name)` substitutes the untrusted value as an argument, not into the template.
- **All `install_packages` callers checked:** exec path, install action, `backend_launcher` (already strips `#` comments and `-` flag lines before the gate), `datasource/verify.py`, `chat.py` — inputs are identifier-shaped or trusted config; the newly-blocked direct-URL requirements.txt lines are the intended ENG-1635 behavior.
- **Classification/routing:** `cell_failure_reason` reads the traceback's last line so the prepended hint doesn't shadow root-cause; nudge routing keys on the exception name / message prefixes, not hint wording; `package_install_rejected` maps to TIER_SELF / non-tripping.

### Findings

**Nit — `anton/commands/datasource/verify.py:54`** (not a changed line; unblocking):
The failure check is `if "failed" in install_result.lower()`, but the belt can now return `"Install refused: …"`, which contains neither "failed" nor "timed out". This PR makes that string newly reachable from `install_packages`. If a curated `engine_def.pip` ever held a flag/URL, verify would miss the refusal and fall through to a confusing import error instead of the clean message. Near-unreachable with trusted/curated input — worth a one-line prefix check only if you're already touching that file.

Bottom line: clean, well-tested security fix that removes the unattended-install vector and gates the one remaining install path. Safe to merge.

alecantu7:
@pnewsam thanks for the empirical pass — agreed on the `datasource/verify.py:54` nit and agreed it shouldn't reopen this PR: the path's inputs are curated `engine_def.pip` names, and a new commit here would restart CI and the standing approvals right before the release train. Tracked instead on **ENG-1854** (the poisoned-venv cleanup follow-up, which touches the same install-path code): its Done-when now includes switching that check to the shared `install_call_failed()` helper so `Install refused` is classified correctly there too.

PR #372, opened 2026-08-19
fix(cli): name the endpoint that actually serves requests (ENG-1634)

SailingSF:
## Description

The setup summary decided the provider label from `minds_url`:

```python
if settings.minds_url and ("mindshub.ai" in settings.minds_url or "mdb.ai" in settings.minds_url):
    provider_label = "MindsHub"
```

`minds_url` is set for anyone signed in, so this labelled the provider "MindsHub" regardless of where requests actually go. A user pointed at a local model was told they were running on MindsHub — the one claim a local model is installed to be able to trust.

`is_minds_host` now judges from the base URL, the value that decides routing. It matches host **and** port, so a self-hosted gateway and a local model server sharing a machine stay distinct, and the `mindshub.ai` / `mdb.ai` suffixes are dot-anchored so a lookalike host misses.

This is the self-report half of ENG-1634; the routing half is in cowork (mindsdb/cowork#688) and cowork-server (mindsdb/cowork-server#356).

Related to ENG-1634

## Type of change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

## Testing

New `tests/test_provider_label_host.py` covers each branch: a LAN endpoint while signed in, the public hosts, a self-hosted gateway matched via `minds_url`, same host with a different port, a lookalike host, and an unset base.

Full suite: 2341 passed, 17 skipped, 0 failures.

### DIFF
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -72,6 +72,11 @@ def is_minds_host(base: str, minds_url: str = "") -> bool:
     Judged from the endpoint, never from whether a MindsHub key or URL happens to
     be configured: signing in sets those, so a user running a local model has
     them too. *minds_url* covers self-hosted gateways on a private host.
+
+    A schemeless base ("api.mindshub.ai/v1") names no host to ``urlparse`` and
+    so reads as not-MindsHub. Deliberate: this only picks a display label, and
+    declining to claim MindsHub is the safe direction. Cowork's equivalent
+    assumes a scheme, because there the answer decides routing.
     """
     from urllib.parse import urlparse
 
@@ -92,6 +97,30 @@ def is_minds_host(base: str, minds_url: str = "") -> bool:
     return host in ("mindshub.ai", "mdb.ai") or host.endswith(
         (".mindshub.ai", ".mdb.ai")
     )
+
+
+def openai_compatible_label(settings) -> str:
+    """Display label for an ``openai-compatible`` provider.
+
+    Names the endpoint that will actually serve requests, read from
+    ``openai_base_url``. Not from ``minds_url``: signing in to MindsHub sets
+    that, so a user pointed at a local model has it too and would be told their
+    prompts go to MindsHub when they do not.
+
+    Shared with the ``/setup`` summary so the two surfaces cannot disagree
+    about where requests go.
+    """
+    base = settings.openai_base_url or ""
+    if is_minds_host(base, settings.minds_url or ""):
+        return "MindsHub"
+    host = url_hostname(base)
+    if host == "generativelanguage.googleapis.com" or host.endswith(
+        ".generativelanguage.googleapis.com"
+    ):
+        return "Google Gemini"
+    if base:
+        return f"OpenAI-compatible ({base})"
+    return "OpenAI-compatible"
 
 
 def _reexec() -> None:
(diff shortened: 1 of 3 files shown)


alecantu7 (commented):
Unsolicited review (came to this from #688). Reviewed at `3054c495`, base fresh at `322a3b40`. CI green; `tests/test_provider_label_host.py` → 6 passed.

**The logic is right.** Judging from the base URL rather than `minds_url` is the correct call, host-AND-port is the right comparison (a self-hosted gateway and a local model server really do share machines), and the dot-anchored suffixes are right. Two of the three mutations I tried are properly caught:

| Mutation | Result |
|---|---|
| Drop the port comparison (host only) | **RED** — `test_same_host_different_port_is_not_the_gateway` |
| Un-anchor the suffixes (`endswith("mindshub.ai")`) | **RED** — `test_lookalike_host_is_not_minds` |
| **Revert the call site to the original `minds_url` substring check** | **GREEN — 6 passed** |

### 1. The fix itself isn't pinned

That third mutation is the bug this PR exists to fix. I put the original line back verbatim:

```python
if settings.minds_url and ("mindshub.ai" in settings.minds_url or "mdb.ai" in settings.minds_url):
```

and the whole suite still passes. The new tests exercise `is_minds_host` in isolation and never reach `_build_frame`, so the one-line change that actually fixes the mislabel can be reverted — by a refactor, a bad merge, anyone — and CI won't notice.

Worth a test that drives the label itself: settings with `planning_provider="openai-compatible"`, `openai_base_url="http://192.168.0.39:1234/v1"`, `minds_url="https://api.mindshub.ai"` → assert the rendered label is **not** "MindsHub". That fails before the call-site change and passes after, which is the property the helper tests don't cover.

### 2. The same check survives one file over

`anton/commands/setup.py:42-48` has it verbatim:

```python
def _provider_label(provider: str) -> str:
    if provider == "openai-compatible":
        base = settings.openai_base_url or ""
        if settings.minds_url and (
            "mindshub.ai" in settings.minds_url or "mdb.ai" in settings.minds_url
        ):
            return "MindsHub"
```

Its result feeds `provider_display = _provider_label(settings.planning_provider)` at line 64. So a user on the `/setup` path is still told "MindsHub" while pointed at a local endpoint — the same false claim, same mechanism, untouched by this PR. Since `is_minds_host` is now exported from `anton.cli`, this should be a two-line change to call it.

Given the ticket frames this as *"the one claim a local model is installed to be able to trust"*, I'd rather both surfaces landed together than leave a second one saying the wrong thing.

### 3. Minor: divergence from the cowork-side predicate

`cowork`'s `isMindsBaseUrl` (in #688's new `src/shared/minds-endpoint.ts`) normalises a schemeless base by prepending `http://`; `urlparse` here does not, so a hand-typed `api.mindshub.ai/v1` yields `hostname=None` and returns `False`:

```
api.mindshub.ai/v1          -> hostname=None
http://192.168.0.39:1234/v1 -> hostname='192.168.0.39' port=1234
```

The two predicates therefore disagree on schemeless input. It errs toward *not* claiming MindsHub, which is the safe direction for this bug, and it's a display label — so I'd leave it as-is rather than add parsing. Flagging only so the divergence is a known one rather than a surprise later.

alecantu7 (approved):
Approving — both findings from my last pass are fixed, and I re-verified rather than assumed. Head `5a994fb5`, base fresh at `322a3b40`.

**The duplicate is gone the right way.** `openai_compatible_label(settings)` is now one shared function and `setup.py`'s 17-line copy delegates to it, so the two surfaces can't drift apart. That's a better fix than patching the second copy.

**The call site is pinned now.** The mutation that slipped through last time no longer does:

| Mutation | Before | Now |
|---|---|---|
| Revert the label to the `minds_url` substring check | **GREEN** ← the gap | **RED** (3 tests) |
| Drop the port comparison (host only) | RED | RED |
| Un-anchor the suffixes | RED | RED |

Tests 6 → 11; full suite 2332 passed, 31 skipped.

Two notes, neither blocking:

- **Minor:** replacing `setup.py`'s `return openai_compatible_label(settings)` with a hardcoded `"MindsHub"` still leaves the suite green, so that surface isn't independently pinned. With the duplication removed this is belt-and-braces rather than real exposure — worth a line only if you touch the file again.
- **Known divergence, deliberate:** cowork's `isMindsBaseUrl` normalises a schemeless base by prepending `http://`; `urlparse` here doesn't, so a hand-typed `api.mindshub.ai/v1` yields `hostname=None` and returns `False`. It errs toward *not* claiming MindsHub — the safe direction for this bug — and this is a display label, so I'd leave it. Flagging so it's a known difference rather than a surprise.

Outstanding for the ticket rather than this PR: the end-to-end repro (clean profile, LAN endpoint, send a message, confirm where the request actually went) still hasn't been run by anyone. That's the check that closes ENG-1634's data-residency claim; the label fix and the routing fixes are all verified at unit level only.

PR #373, opened 2026-08-19
ENG-1459: declare which surface a turn came from, on every path

alecantu7:
> [!NOTE]
> **Now covers the pod (web) path too**, added after the review traced where web turns actually run.
>
> Web turns don't execute in cowork-server — they run in a `minds-anton-scratchpad` pod via `anton.cloud_turn`, and that pod can't derive its own attribution (no cowork-server installed, and only the deployment knows the surface). Measured on prod: **0 of 68** `harness=cloud` traces carried `surface`, `cowork_server_version` or `install_channel`.
>
> So `TurnRequestV1` gains a `trace` block, the pod reads `surface` from it onto the session config, and the build fields ride the turn's `trace_metadata`.
>
> Siblings, **any merge order**: [cowork-server#357](https://github.com/mindsdb/cowork-server/pull/357) sends it, [scratchpad-controller#24](https://github.com/mindsdb/scratchpad-controller/pull/24) relays it.
>
> **Worth a reviewer's attention:** three mutations initially *survived* on the pod side — the pod ignoring `surface`, the entrypoint dropping `trace_metadata`, and the entrypoint leaking `surface` into the metadata as well. The contract tests covered the wire and nothing covered the two consumers. Fixed with behavioural tests driving the real `build_cloud_chat_session` and `stream_turn`. Same "tested the ends, not the middle" gap this PR already hit once on the session seam.

Part 1 of 2 for ENG-1459. **Merge this first** — cowork-server#<pr> supplies the value for desktop/web.

Adds `surface` to `TraceContext` and `ChatSessionConfig`, emitted on Langfuse traces as a `surface:<value>` tag plus a plain `surface` metadata key. Values: `desktop` | `web` | `cli`.

## Why

The SaaS web build is live (first `cowork-web` analytics events 2026-08-18) and **nothing on a trace told it apart from the desktop app** — both run this codebase and both report `harness="anton"`. So "are web users behaving differently?" had no answer.

`surface` says **where the user was**; `harness` says **which agent ran**. Orthogonal: a web turn and a desktop turn can both run anton. ENG-1694 finishes the job by moving the host meaning out of `harness` entirely, and depends on this.

## Three choices worth reviewing

**1. The tag is prefixed (`surface:cli`, not `cli`).** `cli` is currently a legal value of *both* fields while ENG-1694 is in flight, so a bare tag would make `surface=cli` and `harness=cli` indistinguishable in a filter — reintroducing the exact ambiguity this pair of fields exists to remove. `origin:` (ENG-1289) set the same convention one change earlier. Pinned by `test_surface_is_prefixed_so_it_cannot_be_confused_with_the_harness_tag`.

**2. `cloud` is deliberately not a surface value.** `anton/cloud_turn` is a *cloud-safe execution mode* — headless, pod-side — and its own docstring names cowork as the caller (*"cowork sends the tenant's slots per turn"*), so one web turn could legitimately be both. Folding it in would recreate two-vocabularies-in-one-field. It is also 42 traces / staff-only today. Pinned by `test_cloud_is_deliberately_not_a_surface`.

**3. An unrecognised value is dropped with a warning, never forwarded.** A junk row in every breakdown is bad; a **wrong** surface is worse, because it silently moves a turn into the population it is being compared against. Absent stays honestly unknown — the ENG-1495 lesson, where `""` meant both "CLI" and "the host did not say".

## For anyone querying this

**Tags are filterable, not groupable.** `dimensions: [{field: "tags"}]` keys on the whole tag array as a tuple and returns ~one row per trace (9,678 rows for 9,679 traces, measured on prod). Measure a surface with filtered counts, one query per surface:

```
filters: [{column: "tags", operator: "any of", value: ["surface:web"], type: "arrayOptions"}]
```

The metadata copy is unprefixed so a human reading one trace sees the plain value.

## Tests

17 new tests (`tests/test_surface.py`, plus 5 in `tests/test_trace_headers.py`). **Mutation-verified — 8 mutations, all caught.**

Two of them initially **SURVIVED** and are worth flagging, because it is the "tests that lie" shape: deleting `surface=self._surface` from the `TraceContext`, and dropping validation at the session seam. I had covered both *ends* — the validator and the header builder — and not the middle, so the feature could have been fully disconnected with every test green. Fixed by `TestItReachesTheWireFromTheConfig`, which drives a real turn and reads what the provider would actually send.

Full suite: **2,341 passed, 31 skipped, 0 failed.**

> Mutation runs need `PYTHONDONTWRITEBYTECODE=1` — a mutation that swaps two same-length identifiers is byte-identical in size, so Python reuses a stale `.pyc` (cache keys on mtime+size) and reports a false pass while `inspect.getsource` shows correct code.

## Security check

Performed. `surface` is a fixed low-cardinality enum, validated against a closed set before use — no user content, hostnames, credentials, endpoints or auth surface. Host-supplied, so treated as untrusted: normalised, checked, and dropped if unrecognised. Tag values are additionally run through the existing `_sanitize_langfuse_tag` (commas and control chars would corrupt the header). Validation cannot raise — telemetry must never fail a turn, covered by `test_validation_never_raises`.

## No gateway change needed

Verified against the gateway branch: client-supplied tags are merged through untouched, so `surface:web` arrives without any `mindshub_inference` change. (ENG-1289's tag strip is scoped to the `origin:` prefix only.)

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -20,6 +20,7 @@ from anton.clipboard import (
     replace_at_image_paths,
     save_clipboard_image,
 )
+from anton.core.llm.tracing import SURFACE_CLI
 from anton.core.session import ChatSession, ChatSessionConfig, _is_provider_auth_error
 from anton.core.llm.prompt_builder import SystemPromptContext
 from anton.core.llm.provider import (
@@ -1396,6 +1397,10 @@ async def _chat_loop(
         # ENG-1495: see the note on the same field in chat_session.py — an unset
         # `harness` is indistinguishable from a host that forgot to set one.
         harness="cli",
+        # WHERE this ran, as opposed to which agent ran (ENG-1459).
+        # ENG-1694 will move the host meaning out of `harness` entirely,
+        # at which point this becomes the only place "cli" is stated.
+        surface=SURFACE_CLI,
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
         output_dir=settings.artifacts_dir,
--- a/anton/chat_session.py
+++ b/anton/chat_session.py
@@ -88,6 +88,7 @@ def rebuild_session(
     """Rebuild LLMClient + ChatSession after settings change."""
     from anton.core.llm.client import LLMClient
     from anton.chat import ChatSession
+    from anton.core.llm.tracing import SURFACE_CLI
     from anton.core.session import ChatSessionConfig
     from anton.tools import DEFAULT_SESSION_TOOLS
 
@@ -121,6 +122,10 @@ def rebuild_session(
         # didn't identify itself", so the two could never be told apart and the
         # ambiguity got worse with every new host.
         harness="cli",
+        # WHERE this ran, as opposed to which agent ran (ENG-1459).
+        # ENG-1694 will move the host meaning out of `harness` entirely,
+        # at which point this becomes the only place "cli" is stated.
+        surface=SURFACE_CLI,
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
         output_dir=settings.artifacts_dir,
--- a/anton/cloud_turn/__main__.py
+++ b/anton/cloud_turn/__main__.py
@@ -204,7 +204,17 @@ async def stream_turn(raw_line: str, emit, session_builder=None) -> None:
         tool_args_len: dict[str, int] = {}
         seen_tool_progress: set[str] = set()
         last_progress_wire = 0.0
-        async for event in session.turn_stream(turn_content):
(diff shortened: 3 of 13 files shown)


pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: gpt-5.6-terra**

Reviewed the trace context, header emission, CLI, and cloud-turn changes. No actionable findings.

### Validation

- Inspected all 13 changed files at `360274aef6c0d615ce83021a5d76ab68d6ef38ec`.
- Ran 124 focused tests locally.
- GitHub CI, image build, and evaluation workflows passed.

PR #375, opened 2026-08-20
ENG-1291: budget the history summariser off the model's window

ea-rus:
**Changes:**
1. Input budget now derives from the summariser model's context window (`context_window`, extracted from `provider.py`), minus the carried-forward summary and system prompt. The clip became a backstop, not a routine step.
2. Clipping reuses `_clip_keep_cause` (head + marker + tail) and logs what it dropped. Tool-call inputs keep a head slice — a call's target is in its first fields.
3. Output: one call at 8192 tokens; a truncated record is logged and kept, not retried.
4. Prompt length instruction: `~2000 tokens` → `~1500 words` (the model can't see its own tokenisation).

**Not implemented:** step 3 of the ticket (`generate_with_truncation_retry`). It fires on a forced-schema call that never arrived; this call has no schema, so the ladder would never retry. Full reasoning in the ticket comment.

Fixes https://linear.app/mindsdb/issue/ENG-1291/history-summariser-wastes-the-context-window-and-pre-truncates-from


StpMax (approved):
## Findings

### 1. `anton/core/session.py:2144` — on the reactive path, `return False` discards more history than the placeholder it replaced `[CONFIRMED]`

The branch is reachable: `raise_on_empty_response` (`provider.py:864`) is `if content or tool_calls or stop_reason: return`, and Anthropic sets `stop_reason` on every non-streaming `complete()` (`anthropic.py:286`), so an empty response with `stop_reason="max_tokens"` passes through unraised. The PR's comment on this is accurate, and it ships a test for it.

The four proactive call sites (`:3199`, `:3370`, `:4126`, `:4723`) handle `False` harmlessly: `_compaction_failed_this_turn = True`, history intact, turn continues.

But `plan_with_recovery` (`:2448`) and `plan_stream_with_recovery` (`:2832`) use the return value only to set a flag / emit `StreamContextCompacted`, then **unconditionally** proceed to the retry → a second `ContextOverflowError` → `hard_truncate_history()` (`:2457` / `:2846`), which keeps **at most 4 tail messages** (fewer after orphaned `tool_result` stripping) and clears `_last_compacted_count`. `_compaction_failed_this_turn` does not gate those calls.

- **Before the PR:** history became `[placeholder, separator, *recent_turns]` — the old ~60% removed, the newest ~40% retained, and the retry usually succeeded.
- **After the PR:** history is untouched → the retry is guaranteed to overflow again → ~4 messages survive.

Note that `hard_truncate_history` inserts its own fact-free placeholder (`"[Earlier conversation was truncated due to persistent context overflow.]"`). So the ENG-1274 rationale ("substituting a placeholder would replace real turns with a fact-free line") does not actually avoid a placeholder — it defers one while discarding far more turns.

Mitigating: hard truncation is a pre-existing, deliberate tier-3 last resort reached by other `return False` paths too, and it emits an honest event instead of falsely announcing a successful summarisation. Net: a real regression in retained context, moderate severity, not an outright correctness bug.

### 2. `anton/core/session.py:2053` — the second lossy layer the ticket names is untouched, and in the measured regime it is what discards almost everything `[PLAUSIBLE]`

The per-block caps `2000` / `1000` / `500` / `500` were left hardcoded; only the outer clip became budget-derived.

Simulating the actual serializer at the trigger point (~500k chars of raw history = 0.7 × 200k tokens × 4):

| Average raw block | `old_text` | Clip fires? |
|---|---|---|
| 200 chars | 282,484 | yes |
| 500 chars | 285,912 | yes |
| 750 chars | 246,476 | yes |
| 1,500 chars | 166,500 | no |
| 3,000 chars | 101,102 | no |
| 8,000 chars | 44,210 | no |
| 30,000 chars | 13,188 | no |

The crossover sits at an average raw block of ~1,000–1,200 chars (~250–400 folded messages, ~600–950 entries). Because capping is contractive, `old_text ≈ 0.6 × raw × (500/R)`: for realistic agent tool output (R ≈ 3k–30k chars) the serialized form is only 2.5–20% of the folded material, i.e. **the caps discard 80–97.5% while the budget discards 0%**. The PR's own production figure (max 17,078 chars = 8.5% of the budget) confirms this and implies R ≈ 20–30k.

Not REFUTED, because the caps apply **per block**, not per message (`compacted_count` counts messages), so a tool-chatty session with 400+ blocks of 500 chars does reach the budget.

Actual gain delivered: **8,000 → 17,078 chars (~2.1×)**, plus — and this is the part that matters — the clip is now tail-biased instead of dropping the newest folded turns. The title "budget off the model's window" describes a mechanism that does not bind in typical traffic; the ticket's second lossy layer remains.

### 3. `anton/core/session.py:2137` — the empty-record guard omits `.strip()`, against this same file's convention `[PLAUSIBLE]`

`if not summary_response.content:` versus the file's convention: `:4011` `silent = not (llm_response.content or "").strip()`, `:4069`, `:2779`, and `thalamus.py:333`.

`raise_on_empty_response` does not strip (`if content or tool_calls or stop_reason: return`) and returns early on any truthy `stop_reason`, while the providers assign text raw (`anthropic.py:272`, `openai.py:1024`) — so `content="\n\n"` gets through. Past the guard: `summary = summary_response.content`, `_last_compacted_count = compacted_count`, then an unconditional `self._history = [summary_msg] + recent_turns`. There is no length check, no minimum-size assertion, and no later `.strip()`; the `new_old_len < 0.10 * total` gate runs **before** the call, on input, so it cannot catch a blank output. `looks_truncated` only logs.

This is exactly the unhealable blank record (ENG-1274) the branch exists to prevent. The PR's test uses `content=""` and passes either way. One-word fix: `if not (summary_response.content or "").strip():`.

The window is narrow (it needs a provider returning whitespace-only text at a max_tokens cut), hence PLAUSIBLE rather than CONFIRMED.

### 4. `anton/core/session.py:2112` — a defensive accessor for `max_tokens` already exists in this file and the PR bypasses it `[reuse]`

`_turn_max_tokens` (`:3950`):

```python
budget = getattr(self._llm, "max_tokens", None)
return budget if isinstance(budget, int) and budget > 0 else 8192
```

documented as a fallback for "mocks, exotic hosts". The new code reads `self._llm.max_tokens` directly, so `min(8192, ...)` raises `TypeError` on a non-number (measured: `min(8192, AsyncMock().max_tokens)` and `min(8192, None)` both raise).

Unreachable in the current anton suite: `_summarize_history` returns early on `if len(self._history) < 6`, and the four bare-`AsyncMock` doubles (`test_cloud_turn_session.py:227`, `test_web_tools.py:529/607`, `test_tools.py:27`, `test_datasource.py:173`) never reach compaction. No production path either: `max_tokens: int = 8192` in `LLMClient.__init__`, pydantic settings typed `int`, and `cowork-server/cowork/services/providers.py:1111` omits the parameter entirely.

However **`cowork-server/tests/test_history_compaction_plateau.py:50`** `_mock_llm()` sets `coding_model` but neither `router_model` nor `max_tokens` — there the `TypeError` does fire and is swallowed by `except Exception` as "summarization failed". Cross-repo test fragility plus a masked failure mode.

### 5. `anton/core/session.py:634` — `int(context_window(model) * 0.25) * 4` is the identity function on `context_window(model)` `[simplification]`

Verified:

```
200000  -> 200000
128000  -> 128000
400000  -> 400000
1000000 -> 1000000
```

Since `0.25 * 4 == 1` and every window in `_CONTEXT_WINDOWS` is divisible by 4, the `int()` never even rounds. Two named constants and a cast that provably change no value — and `_CHARS_PER_TOKEN` carries a warning about cross-coupling to hippocampus/cortex, so a future reader will believe they are calibrating a budget when they are scaling a number that had no derivation.

Equivalent: `return max(context_window(model) - reserved, 0)` with one comment.

Related (not filed separately): `reserved = len(prior_summary) + len(_SUMMARY_SYSTEM_PROMPT)` subtracts a fixed ~975–1300 chars from 200,000 — a 0.5% correction inside an estimate whose own error bar is a factor of several. And `budget` is bound twice in the function with different units (input chars at `:2079`, output tokens at `:2112`), so two warnings print the same word "budget" with values ~25× apart and no unit.

### 6. `anton/core/session.py:618-635` — the new helper's docstring states a false premise `[correctness of documentation]`

"Compaction fires at 70% pressure on the *coding* model" — in fact the main loop calls `plan`/`plan_stream` with `model=self._planning_model` (`client.py:111/132`), so pressure is measured against the **planning** model. Inert at defaults (planning `claude-sonnet-4-6` and coding `claude-haiku-4-5-20251001`, both 200k), but the entire `ponytail` argument about router/coding asymmetry is built on the wrong model.

### 7. `tests/test_history_compaction.py:104` — the PR's central claim has no test `[coverage]`

`test_newest_folded_turn_reaches_the_summarizer`: 40 messages × ~1207 chars (all under the 2000 cap), 24 folded → `old_text` ≈ 29,160 against a budget of 199,025. A 6.8× margin, so the clip never fires — the test asserts "no clipping happened", not "the tail survives clipping". The assertions `"turn-0 " in sent` / `"turn-23 " in sent` would still pass with `old_text[:budget]`, i.e. under a revert to the head-cut.

No test asserts the `"history summarization: input clipped"` warning, none drives `len(old_text) > budget`, and none covers `budget == 0`. The two remaining budget tests are pure arithmetic over `_summarizer_input_budget` and never touch the call site.

Fix: either a direct unit assertion on `_clip_keep_cause(old_text, small_cap)`, or monkeypatch `_summarizer_input_budget` to a small value in this test.

### 8. `anton/core/session.py:586` — `_CHARS_PER_TOKEN = 4` is a third copy of one calibration, and the PR knows it `[reuse]`

The copies: `anton/core/memory/hippocampus.py:91` (`char_budget = token_budget * 4`, inside `_filter_by_token_budget` — the converter behind `recall_lessons`/`get_lessons`/`recall_scratchpad_wisdom`) and `anton/core/memory/cortex.py:229` (`_RULES_BUDGET_CHARS = 6000`, commented "~6000 chars ≈ ~1500 tokens"). No shared helper exists — grepping `CHARS_PER_TOKEN|estimate_tokens|approx_tokens|count_tokens` finds only the new constant.

The PR already touches `provider.py`, where a `token_budget_to_chars()` would sit naturally next to the new `context_window()`; instead it adds copy #3. Recalibrating for a code-heavy or CJK corpus (where 4 chars/token is off by ~2×) now means editing three unrelated files, nothing links them, and there is no test asserting they agree — so the drift is silent.

---

## Checked and dismissed

| Candidate | Why REFUTED |
|---|---|
| `summary_response.usage.output_tokens` dereferenced unguarded although `looks_truncated` tolerates `usage=None` | `LLMResponse.usage` is `field(default_factory=Usage)` (`provider.py:72`) and every in-repo constructor populates it (`openai.py:1051/1314/1610/1672`, `anthropic.py:292/465`). The same unguarded pattern already exists next door at `:4046`. Reachable only via an out-of-repo duck-typed client |
| `_clip_keep_cause(text, 0)` returns `marker + the whole text` | The mechanism is real and measured (5000 → **5029** chars, because `text[-0:] == text[0:]`), and `cap=0` is the only misbehaving value. But it is unreachable: the floor of `window_chars` is `int(128000*0.25)*4` = 128,000, while `reserved` is bounded at ~33k (the record is replaced, not accumulated). Every other call site passes the literals `2000`/`1000`/`500`/`400`/`12000`. The new head branch handles `cap=0` correctly — the asymmetry favours the new code |
| The multi-line marker breaks the `[role]:` transcript framing | Pre-PR entries were already routinely multi-line: `content[:2000]` preserves every `\n`, and this is raw user prompts, assistant prose, and stringified tool output. A "one `[role]:` per physical line" invariant never existed, and the old outer cap was line-unsafe too. Residue: a minor inconsistency — the single-line marker was added for one of four call sites |
| A truncated record is kept instead of retried; `## Remaining` is the last section | **This refutes what the first-pass assessment claimed.** Pre-PR a truncated record was kept too — silently, and at a 4× smaller budget (`max_tokens=2048` + `content or "(summary unavailable)"`). The section order is byte-identical and not touched by the diff. The two branches discriminate on **content presence**, not `stop_reason`: zero information → keep the turns, partial → keep the partial. Headroom is ~4× (the prompt asks for ~1500 words ≈ 2000 tokens against 8192; production input maxes at ~4.3k tokens). The PR strictly improves here |
| `router_model` read outside the `try` → AttributeError bypasses recovery | Structurally accurate (read at `:2079-2082`, the `try:` opens at `:2106`, and `plan_with_recovery` does not catch AttributeError), but there is no trigger: `LLMClient.__init__` always sets `self._router_model = router_model or coding_model` (`client.py:67`), and `AsyncMock().router_model` returns a child mock rather than raising. No `spec=`/`autospec` mock or hand-rolled `llm_client` fake exists in the repo |
| `Mock.startswith` is truthy → 200k instead of the conservative default | Reproduced (`bool(Mock().startswith(...)) is True`, first table row = 200,000), but it predates the PR: on `staging`, `compute_context_pressure(Mock(), 1000) == 0.005`. The diff only extracts the loop into a function. Test-only — model ids are always `str` in production |
| The summary can exceed what it replaces → history grows, repeated billing | The `new_old_len < 0.10 * (new_old_len + recent_len)` gate does in practice force the folded part to be larger: at trigger, the minimum passing fold is ~50,000 chars against a maximum record of 32,768 chars (W=200k). Only on W=128k does the minimum (~29,840) fall marginally below — worst case ~**+1%**, not +45%. The loop is damped by `_compacted_this_turn` and by in-place replacement of `prior_summary` (the slot saturates at ~32.7k rather than accumulating) |
| The reactive path is unguarded → up to 25 summariser calls per turn | Pre-existing, outside the diff (both lines are byte-identical on `staging`), and deliberate: the docstring at `:1899-1902` and `tests/test_chat.py:203-228` fix "overflow plus one proactive retry is the ceiling". Self-limiting at 2–3 (a successful compaction removes ~60% of messages; otherwise `hard_truncate_history` or propagation). Amplification is ~2.1× on input, not 25×. The proposed fix would be a **regression**: a round-2 overflow would go straight to the destructive `hard_truncate_history()` |
| Conventions (CLAUDE.md / AGENT.md / CONTRIBUTING.md) | No violations. No Russian text; comments are English. anton has no `CLAUDE.md`, `AGENT.md`, or `AGENTS.md` at any depth. `CONTRIBUTING.md` states workflow rules only ("New behavior should come with tests" — satisfied). Logging is content-safe: only `len()`, budgets, `output_tokens`, `stop_reason` — no user-derived text. `pyproject.toml` sets no line-length and CI runs no Python linter. `ponytail` appears nowhere else in the repo, but no rule governs it either — nothing to flag |

---

## Verdict

**Blocking before merge:**

1. **#1** — on the reactive path the PR retains less history than before (`return False` → `hard_truncate_history(keep=4)` instead of keeping the newest ~40%). Either preserve the placeholder behaviour on the reactive path, or gate `hard_truncate_history` differently.
2. **#3** — add `.strip()`. One word; restores the file's convention and closes the ENG-1274 path.
3. **#7** — a test in which the clip actually fires; otherwise the PR's central behaviour ships without regression protection.

**Follow-up:**

4. **#2** — record that the per-block caps remain the effective ceiling (discarding 80–97.5% versus 0% for the new budget), so "budget off the window" is not read as a solved problem. Deeper fix: derive the per-message caps from the budget too (`budget // max(len(old_turns), 1)` with a floor).
5. **#5**, **#6** — the identity arithmetic and the false premise in the docstring; cosmetic, but both mislead.
6. **#4**, **#8** — reuse `_turn_max_tokens`; extract `token_budget_to_chars()` next to `context_window()` instead of a third copy of the calibration.

**Deliberately out of scope, worth tracking:** the `[:N]` family is not fixed beyond this file. Head-cuts on model-facing text whose meaning lives in the tail remain at `session.py:3871` and `:3850` (`assistant_text_parts[:2000]` into episodic — literally ENG-1633, named in the ticket), `anton/core/memory/cerebellum.py:172/178/182` (`stdout[:500]`, `stderr[:400]`, `error[:400]` into an LLM prompt — ENG-836), and `anton/commands/skills.py:136/187`. `_clip_keep_cause` is already importable and is the house solution, so each is a one-line substitution.

PR #377, opened 2026-08-20
ENG-1694: harness names which agent ran, not where

alecantu7:
> [!NOTE]
> **The gate is SATISFIED as of 2026-08-26. This is now mergeable on the merits.**
>
> All three conditions verified against production, not inferred:
>
> 1. ✅ **ENG-1459 merged** — anton `695dcee7`, cowork-server `06c9a4b9`, scratchpad-controller `d21a0e1a` (21 Aug).
> 2. ✅ **Pod image pin moved past it.** scratchpad-controller [#32](https://github.com/mindsdb/scratchpad-controller/pull/32) merged 26 Aug 08:53 UTC (`b6b55ce9`), released as `v2.26.8.26.1`. It splits the pins rather than letting prod inherit staging: `values.yaml` → `17dc5b57`, `values-prod.yaml` → `2a3ad8ea`. **Both contain ENG-1459's anton half** (verified by content — `695dcee7` is not an *ancestor* of `anton/main`, because the release squashes, so an ancestry check misleads here).
> 3. ✅ **Production web turns carry `surface=web`** — **49 traces in 72h** on prod Langfuse. A real one:
>
> ```
> ['cloud', 'origin:harness', 'prod', 'surface:web',
>  'organization_id:…', 'user_id:7d8022d3-…', '<user email>']
> ```
>
> So `harness="cloud"` can be retired without losing the web/desktop split — `surface:web` has replaced it on live traffic. Companion tags also confirmed live: `surface:desktop` 3,155, `surface:cli` 145.
>
> **Note for anyone querying across the changeover:** historical traces keep `harness=cloud`, so a window spanning 26 Aug needs `harness=cloud OR surface=web`, not one or the other.
>
> **Also re-verified after the gate cleared:** merges cleanly into `staging` across the 15 commits added since this branch was rebased, and the **full suite passes at 2,653** on the merged result — including ENG-1486's `tool_completed`, which landed in the same emit path.

Closes ENG-1694 (pending the gate above).

## What

`harness` becomes a pure agent identity. `surface` (ENG-1459) owns "where".

```
chat.py / chat_session.py    harness="cli"    →  HARNESS_ANTON + surface=cli
cloud_turn/session.py        harness="cloud"  →  HARNESS_ANTON
```

Adds `HARNESS_ANTON` / `HARNESS_HERMES` / `VALID_HARNESSES` beside the surface vocabulary, so both live in one place and disjointness is assertable — `test_no_value_means_both_an_agent_and_a_place` fails if any value ever appears in both.

## Why `cloud` was wrong, not just overloaded

The pod image is `minds-anton-scratchpad` — *"anton + boot"*. **The agent running there IS anton.** And the web path can only ever run anton: `turnqueue/producer.py` hardcodes `op="anton_turn"`, the controller's only turn op.

Nothing is lost by removing it: every web turn runs in a pod and only web turns do, so *"ran in a pod"* is fully determined by `surface=web`. It earns its own axis only if that 1:1 breaks.

## The trace-name fix is free

The gateway composes the display name from `harness`, so CLI turns arrived as `cli:turn-1` while every other anton turn was `anton:turn-N` — splitting name-based grouping, which is the most natural thing to group on in a dashboard. Verified: a CLI context now yields `anton:turn-1`, **no gateway change**.

## Two existing tests updated deliberately, not flipped

Both pinned the old contract:

- `test_cloud_turn_session.py` asserted `cfg.harness == "cloud"` → now `anton` plus `surface is None` (that fixture sends no trace block).
- `test_session_tools.py::test_both_session_builders_identify_the_host_as_cli` was **ENG-1495's guard**. Its property is unchanged — a host that doesn't identify itself must stay distinguishable from the CLI — but the field carrying it moved. It now requires **both** `surface=SURFACE_CLI` and `harness=HARNESS_ANTON`, and rejects `harness="cli"`.

> Process note, since it bit twice: I found both by running the **full** suite (one first surfaced by a mutation run), not my own new file. New tests passing says nothing about existing tests that pinned the old contract.

## ENG-1540's inherited criterion — discharged

Audited every in-repo query keying on `harness`. **None filters on harness absence.** One keys on the trace-*name* prefix and this change moves traffic into its population:

`harness-measure/scripts/measure.py:460` — `startswith('anton:turn')`

CLI (`cli:turn-N`) and pod (`cloud:turn-N`) turns are excluded today and will now **join**. Direction is right — they *are* harness-identified anton turns — but it is a **discontinuity**: the scoped denominator grows by ≈5 CLI traces plus every pod turn (68 prod / 506 staging). No fix needed there; recorded so the jump is explained rather than investigated.

## Tests

7 new in `tests/test_harness_is_the_agent.py`. **Mutation-verified — 5 mutations, all caught:** reverting either call site, readmitting `cloud` *or* `cli` into `VALID_HARNESSES` (the latter recreates the exact agent/place overlap), and dropping the CLI's surface.

## Security check

Performed. Read-only observability. Both fields are closed low-cardinality enums set by first-party code — no user content, hostnames, credentials, endpoints or auth surface. Values narrow rather than widen (`cli`/`cloud` leave the harness vocabulary), so nothing new reaches a header; tag sanitisation is unchanged.

## Self-review findings (2026-08-20)

Three, all recorded rather than code-fixed — the change itself held up.

**1. A vintage boundary this PR creates, and my body didn't mention.** Historical traces keep their old values: a `harness=cloud` trace stays `cloud` forever. So *after* this ships, "which turns were web?" is `harness=cloud` **before** the boundary UNION `surface=web` **after** it. A single-predicate query spanning the change undercounts one side silently. Same class as ENG-1289's tagging-era boundary. Scale: 68 prod / 506 staging `cloud` traces over recent windows.

**2. ENG-1540's inherited criterion is only PARTIALLY dischargeable by me, and the ticket overstated it.** Verified programmatically: no in-repo query filters on harness absence, and **0 of 93 saved PostHog insights** (projects 424726 + 355390) reference `harness` at all. **Not verifiable from here:** saved *Langfuse* views — the public API doesn't expose them (`/api/public/{dashboards,saved-views,views,widgets}` all 404). **A human must check the Langfuse UI** for views keyed on `harness = cli` / `= cloud` / absence. Ticket corrected from "DISCHARGED" to "partially".

**3. A measurement discontinuity in our own tooling.** `harness-measure/scripts/measure.py:460` scopes on `startswith('anton:turn')`. CLI (`cli:turn-N`) and pod (`cloud:turn-N`) turns are excluded today and will **join** that population — correct in direction, but the scoped denominator jumps with no code change behind it.

### Refuted during the review

- **Other consumers of the harness value.** The gateway's only use is the trace name (`context.py:380-384`) — no allowlist, no branching, so `anton` passes cleanly and produces the intended name. Grepped anton, cowork-server, anton_services and mdb-ai: nothing branches on `harness == "cli"` or `"cloud"`.
- **PostHog breakage.** `turn_completed` carries `harness`, and the value distribution shifts (cli 9 + cloud 108 fold into anton) — but no saved insight or skill keys on it, so nothing breaks.
- **`Message.harness`** (cowork-server DB column, values like `cowork-direct`) — untouched; different field, confirmed out of scope.

### Known weakness in the tests

The call-site assertions are **source inspection** (`inspect.getsource`), matching the existing ENG-1495 guard and for the reason its docstring gives — exercising `rebuild_session` means standing up an LLMClient, Cortex and knowledge refresh. They catch removal and the wrong value, but would be fooled by the string appearing in a comment. The trace-name test is genuinely behavioural (builds real headers via the real provider).

### DIFF
--- a/anton/chat.py
+++ b/anton/chat.py
@@ -20,7 +20,7 @@ from anton.clipboard import (
     replace_at_image_paths,
     save_clipboard_image,
 )
-from anton.core.llm.tracing import SURFACE_CLI
+from anton.core.llm.tracing import HARNESS_ANTON, SURFACE_CLI
 from anton.core.session import ChatSession, ChatSessionConfig, _is_provider_auth_error
 from anton.core.llm.prompt_builder import SystemPromptContext
 from anton.core.llm.provider import (
@@ -1396,10 +1396,11 @@ async def _chat_loop(
         session_id=current_session_id,
         # ENG-1495: see the note on the same field in chat_session.py — an unset
         # `harness` is indistinguishable from a host that forgot to set one.
-        harness="cli",
-        # WHERE this ran, as opposed to which agent ran (ENG-1459).
-        # ENG-1694 will move the host meaning out of `harness` entirely,
-        # at which point this becomes the only place "cli" is stated.
+        # WHICH AGENT: the CLI runs anton, so that is what it reports
+        # (ENG-1694). It said "cli" until then, which described the host and
+        # left "did this run anton or hermes?" unanswerable.
+        harness=HARNESS_ANTON,
+        # WHERE it ran — the other axis, and now the only place "cli" appears.
         surface=SURFACE_CLI,
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
--- a/anton/chat_session.py
+++ b/anton/chat_session.py
@@ -88,7 +88,7 @@ def rebuild_session(
     """Rebuild LLMClient + ChatSession after settings change."""
     from anton.core.llm.client import LLMClient
     from anton.chat import ChatSession
-    from anton.core.llm.tracing import SURFACE_CLI
+    from anton.core.llm.tracing import HARNESS_ANTON, SURFACE_CLI
     from anton.core.session import ChatSessionConfig
     from anton.tools import DEFAULT_SESSION_TOOLS
 
@@ -121,10 +121,11 @@ def rebuild_session(
         # telemetry as "" — which meant BOTH "this is the CLI" and "the host
         # didn't identify itself", so the two could never be told apart and the
         # ambiguity got worse with every new host.
-        harness="cli",
-        # WHERE this ran, as opposed to which agent ran (ENG-1459).
-        # ENG-1694 will move the host meaning out of `harness` entirely,
-        # at which point this becomes the only place "cli" is stated.
+        # WHICH AGENT: the CLI runs anton, so that is what it reports
+        # (ENG-1694). It said "cli" until then, which described the host and
+        # left "did this run anton or hermes?" unanswerable.
+        harness=HARNESS_ANTON,
+        # WHERE it ran — the other axis, and now the only place "cli" appears.
         surface=SURFACE_CLI,
         proactive_dashboards=settings.proactive_dashboards,
         act_first=settings.act_first,
--- a/anton/cloud_turn/session.py
+++ b/anton/cloud_turn/session.py
@@ -26,6 +26,7 @@ from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
 
 from anton.cloud_turn.contract import TurnRequestV1
+from anton.core.llm.tracing import HARNESS_ANTON
 from anton.core.tools.skill_format import SKILL_FILE
 
 if TYPE_CHECKING:
@@ -587,7 +588,12 @@ def build_cloud_chat_session(request: TurnRequestV1) -> "ChatSession":
         settings=settings,
         workspace=workspace,
         session_id=request.conversation_id,
-        harness="cloud",
+        # WHICH AGENT: this pod image is "anton + boot" — the agent running here
+        # IS anton, so "cloud" was factually wrong, not merely overloaded
+        # (ENG-1694). Where it ran comes from `surface` below, which cowork
+        # sends; every web turn executes in a pod and only web turns do, so
+        # "ran in a pod" needs no field of its own today.
+        harness=HARNESS_ANTON,
         # WHERE the user was, which this pod cannot know on its own — only the
         # deployment does, so cowork sends it (ENG-1459). Absent when the pod is
         # driven directly rather than by cowork; an unset surface reads as
--- a/anton/core/llm/tracing.py
+++ b/anton/core/llm/tracing.py
@@ -23,6 +23,26 @@ from contextvars import ContextVar, Token
 from dataclasses import dataclass, replace
 
 
+# WHICH AGENT ran the turn (ENG-1694). The canonical definition, alongside the
+# surface vocabulary below — the two are orthogonal and must stay so.
+#
+#   HARNESS_ANTON   the anton agent. Every first-party caller today: the CLI,
+#                   cowork-server (desktop and web), and the cloud pod.
+#   HARNESS_HERMES  cowork-server's other harness. Emits no langfuse traces yet
+#                   (`hermes_harness/harness.py` accepts trace_tags/metadata and
+#                   ignores them), so this value has NO volume — a query showing
+#                   100% anton is not evidence hermes is unused.
+#
+# Until ENG-1694 this field also held "cli" and "cloud", which name *places*.
+# One field with two vocabularies could answer neither question: a "cli" trace
+# could not say which agent ran, and an "anton" trace could not say where. If a
+# value answers "where", it belongs in `surface`; a third question gets a third
+# field rather than a third meaning here.
+HARNESS_ANTON = "anton"
+HARNESS_HERMES = "hermes"
+VALID_HARNESSES = frozenset({HARNESS_ANTON, HARNESS_HERMES})
+
+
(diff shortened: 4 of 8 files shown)


pnewsam (commented):
## Code Review

**Verdict: COMMENT** (code-approve, hold-merge)

Reviewed the full `+202/-36` diff across all 8 files plus the gateway consumer (`mindshub_inference`). The code is correct and the change is well-founded — no actionable defects. This is COMMENT rather than APPROVE for one reason only: your own `DO NOT MERGE` gate is real and, as far as I can verify, still unsatisfied. Nothing in the code needs to change; the hold is operational.

### Validation

- **Ran the affected tests** (`test_harness_is_the_agent.py` + `test_cloud_turn_session.py` + the updated `test_session_tools` case) → **63 passed** locally.
- **Grepped the whole monorepo** (anton, cowork-server, mindshub_inference/services, cowork) for `harness == "cli"/"cloud"` branching → the only match is the test this PR updates. The safety claim holds: nothing branches on the deleted values.
- **Read the gateway** (`mindshub_inference/.../context.py`): `harness` is only *composed* into `trace_name = f"{harness}:turn-{turn_id}"` and *presence-checked* in `_origin_tag`. No allowlist, no value branch — `"anton"` flows cleanly and still yields `origin:harness`.
- **Confirmed the cloud path** already reads `surface` from `request.trace` (`cloud_turn/session.py:595`) and that `_build_trace_headers` (`openai.py:947,952`) emits `harness` in metadata and the prefixed `surface:` tag — so the behavioural test's assumptions are accurate.
- **CI**: all checks green; `mergeable: CLEAN`. Nothing technical blocks the merge — which is exactly why the gate carries the weight.

### Findings

**Major (blocks merge; operational, not a code defect):** Merging now would make production web turns unmeasurable. `harness="cloud"` is currently the only field separating a web turn from a desktop one (the pod hardcodes it; web is the only surface routed to pods), and this diff deletes it. The replacement signal — `surface=web` sent by cowork — must be confirmed flowing end-to-end in production *first*. I verified the code side (the pod reads `request.trace.surface`; nothing consumes the old values), but the two blocking conditions live in another repo and in prod telemetry, which I cannot reach from here: the deployed `SCRATCHPAD_CONTROLLER__SCRATCHPAD_IMAGE` pin (scratchpad-controller `values.yaml`) and an observed prod web turn carrying `surface=web`. Do not merge until the gate's steps 2 and 3 are satisfied — which your PR description already documents and self-blocks on.

**No other actionable findings.** The three call-site changes, the vocabulary constants, the comment rewrites, and both deliberately-updated tests are all correct. The vintage-boundary concern (historical `harness=cloud` traces keep their old value, so a query spanning the change must `UNION` the two predicates) is already recorded in your self-review and needs no code change here.

_Verdict calibration: the code itself would earn APPROVE on quality; I'm withholding it purely because the unsatisfied gate makes merging **now** unsafe, and the `CLEAN` mergeable state means nothing but that gate stands between this and an accidental merge._

alecantu7:
@pnewsam — the two conditions you couldn't reach from there are **both satisfied**, and were already satisfied about 11 hours before your review. Evidence below; PR body updated so the `DO NOT MERGE` banner no longer contradicts reality.

## Gate step 2 — the pod image pin

scratchpad-controller [#32](https://github.com/mindsdb/scratchpad-controller/pull/32) merged **26 Aug 08:53 UTC** (`b6b55ce9`), released as `v2.26.8.26.1` a minute later.

Max split the pins rather than letting prod inherit staging:

| | pin | contains ENG-1459's anton |
|---|---|---|
| `values.yaml` (staging) | `17dc5b57` | ✅ |
| `values-prod.yaml` | `2a3ad8ea` | ✅ |

**Verified by content, not ancestry** — worth flagging because the obvious check is wrong here. `695dcee7` is *not* an ancestor of `anton/main` (the release squashes), so `merge-base --is-ancestor` says "no" while the code is plainly present. I made exactly that mistake first.

## Gate step 3 — a production web turn carrying `surface=web`

**49 traces in 72h** on prod Langfuse. A real one:

```
['cloud', 'origin:harness', 'prod', 'surface:web',
 'organization_id:…', 'user_id:7d8022d3-…', '<user email>']
```

`surface:web` and `harness=cloud` on the same trace, with a genuine `user_id`. Companions also live: `surface:desktop` 3,155, `surface:cli` 145.

So the replacement signal is flowing on real traffic, and retiring `harness="cloud"` no longer loses the web/desktop split. That was the whole basis of the hold.

## Re-verified after the gate cleared

- Merges cleanly into current `staging` across the **15 commits** added since this branch was rebased.
- **Full suite: 2,653 passed** on the merged result — run deliberately because ENG-1486's `tool_completed` landed in the same emit path after your review, and a clean textual merge is not evidence the two agree.

## One thing I took from your review

The vintage-boundary point is now in the PR body rather than only in the self-review, since it is the thing most likely to bite a reader later: **historical traces keep `harness=cloud`**, so any window spanning 26 Aug needs `harness=cloud OR surface=web` — not one or the other.

Your calibration was right, and the hold did its job: it survived four days precisely because the gate was real. Happy for you to re-check the two external facts yourself before converting — they are both in prod and I would rather you saw them than took my word.

PR #379, opened 2026-08-20
ENG-1692: don't report script traffic as turns

alecantu7:
Closes ENG-1692.

A turn with **no session id AND zero LLM calls** did nothing and belongs to nobody — it is an eval, replay or test driver. Two lines guard the analytics sink; the log line is untouched.

## The problem, measured

14 days to 2026-08-20, `turn_completed` in project 424726 — 11,734 events:

| `conversation_id` | `llm_calls` | events | installs | tokens | what it is |
|---|---|---|---|---|---|
| **missing** | **zero** | **8,209 (70%)** | **15** | **0** | ← script traffic |
| present | one or more | 3,119 (27%) | 348 | 757,570,982 | the real signal |
| present | **zero** | 338 (3%) | 88 | 0 | real turns that **failed** before reaching a model |
| **missing** | one or more | 68 (0.6%) | 22 | 4,854,880 | the **connector probe** — real spend |

70% of volume, zero tokens, from 15 machines. It dragged the per-turn median to **zero** — impossible for a completed turn — and ENG-1286's spend ceiling would have been set from it.

## Why the conjunction, not either half

- **`llm_calls == 0` alone** deletes 338 real turns that failed pre-model (`error` 157, `retry_exhausted` 132, `cancelled` 49, across 88 installs). Those are real users hitting real errors — exactly the population a reliability question needs.
- **Missing session id alone** deletes cowork-server's connector probe, which legitimately runs a turn without one and spent 4.8M real tokens.

Four tests pin all four quadrants so this can't later be "simplified" to one condition.

## Why behavioural and not the version test the ticket proposed

The ticket proposed gating on `.dev` in the version. Re-measured, it doesn't work:

```
.dev version gate  →  catches 2,907 of 11,734,  MISSES 5,302 contaminated
this gate          →  catches 8,209 of 11,734,  misses 0
```

The largest single contaminant is **`2.26.8.18.1rc2`** — 5,253 events, no `.dev` in it, and explicitly protected by the ticket's own *"do not catch rc"* rule. Nor does "exclude rc" work: `2.26.8.15.2rc2` is 19 real turns and 3.3M real tokens.

**A version says what the software is, not who is running it.** The same build is installed by a real tester *and* driven by a script — `2.26.8.18.1rc2` holds 5,253 script turns **and** 16 real ones. No version rule can split one build; a per-turn behavioural test doesn't need to.

Two things this removes from the original plan:

- **No `ANTON_ANALYTICS_FORCE` override.** A developer running a *real* turn has a session id and LLM calls, so it still emits. The ticket's worry — *"anyone verifying analytics changes has to defeat the gate by hand, which is how gates get commented out"* — doesn't arise.
- **No dependence on version derivation.** The ticket flagged that release builds being non-`.dev` is *"a property of how the version is derived, not a guarantee"*, with real users silenced invisibly as the failure mode. This gate can't have that.

## Why both signals are trustworthy

- **`conversation_id`** is `str(self._session_id or "")` (`session.py:2538`). Emptiness is a structural marker of "not driven by a product host".

  **Corrected after review** (thanks @pnewsam) — my original enumeration was incomplete. There are **three** non-test builders that omit a session id, not two:

  | Builder | Runs a turn? | Effect |
  |---|---|---|
  | `cli.py`'s datasource-connect | no | never emits |
  | cowork-server's connector probe | yes | **kept** — `llm_calls > 0` |
  | `anton chat` with episodic memory **off** | yes | **dropped if it fails pre-model** |

  The third is real: `chat.py` sets `current_session_id = None` whenever `settings.episodic_memory` is False, which is user-settable (default True). Such a turn failing before its first model call has neither half of the conjunction.

  **The collateral is measured, and currently zero.** Over 14 days every one of the 8,569 script-shaped events carries an **empty** `harness`, while all 26 `harness="cli"` turns are kept — no CLI turn falls in the dropped bucket today.

  That also shows `harness` *would* discriminate the two, contrary to my assumption that eval drivers report `cli`. Deliberately unused: it would make a fix worth 70% of the volume depend on a field ENG-1695 flags as unreliable (33 turns already carry an empty `harness` from a host that should set one), so script traffic acquiring a `harness` value would silently disable the guard. Trading a measured zero for that brittleness is the wrong direction. Reasoning recorded at the guard itself, not just here.
- **`llm_calls`** is incremented by the LLM client itself (`TurnCost.add` is installed as `LLMClient.usage_listener`, `turn_cost.py:156`). It can't be wrong independently of the call happening. The codebase already relies on this exact signal for the same purpose (`turn_cost.py:137`).

## 19 existing fixtures now pass a `session_id`

They drove `turn_stream` with a mock LLM that never reports usage *and* no session id — a shape no product path produces, since cowork-server and the CLI both always set one. **The fixtures were unrealistic and the guard exposed it**; supplying the id makes them match production rather than working around the gate. A header note in each file records why.

## Tests

5 new in `tests/test_turn_cost.py`. **Mutation-verified — 6 mutations, all caught:** removing the guard, `or` instead of `and`, each half alone, inverting it, and moving it above the log line.

All 478 tests touching `send_event` / `turn_completed` pass.

## Expect a ~70% drop in event volume

That is the fix, not a regression. **ENG-1546's floor becomes a prerequisite rather than a companion** — any alert threshold calibrated on today's volume is calibrated on 70% contamination.

## Also fixed, outside this PR

The `turn-cost` skill was applying the same wrong version logic and reporting on fake data: it picked *"the busiest build without `dev`"* — `2.26.8.18.1rc2` — and pinned every query to it, reporting **"5,268 completed, 100.0%"** off 5,253 script events. Replaced with this same behavioural filter across all builds; the real distribution has nine outcomes, 27% non-completion, and `spend_ceiling` burning 25.5% of tokens from 4.8% of turns. That fix is local tooling, not in this diff.

## Security check

Performed. Read-only observability change that only ever *suppresses* an outbound event — no new data collected, no new field, no endpoint or auth surface touched. Both values read are already sent in the same event today.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2460,6 +2460,40 @@ class ChatSession:
             " ".join(f"{k}={v}" for k, v in _root_cause_fields.items()),
         )
 
+        # Script traffic never reaches the analytics sink (ENG-1692).
+        #
+        # A turn with NO session id AND zero LLM calls did nothing and belongs to
+        # nobody: it is an eval, replay or test driver. Measured over 14 days to
+        # 2026-08-20, that shape was 8,209 of 11,734 `turn_completed` events —
+        # **70% of all volume, carrying 0 tokens** — from 15 machines. It made
+        # the per-turn median come out at ZERO, which is impossible for a
+        # completed turn, and would have set ENG-1286's spend ceiling from a
+        # fabricated distribution.
+        #
+        # BOTH conditions are required, and each half alone deletes real data:
+        #   * `llm_calls == 0` alone would drop 338 real turns that failed
+        #     before reaching a model (ended_by error / retry_exhausted /
+        #     cancelled, 88 installs) — real users hitting real errors.
+        #   * a missing session id alone would drop cowork-server's connector
+        #     probe, which legitimately runs a turn without one and spent 4.8M
+        #     tokens over the same window.
+        #
+        # Deliberately BEHAVIOURAL, not a version test. The original plan gated
+        # on `.dev` in the version; re-measured, that catches only 2,907 of the
+        # 8,209 and misses the single largest contaminant (`2.26.8.18.1rc2`,
+        # 5,253 events, no `.dev` in it). A version says what the software IS,
+        # not who is running it — the same build is installed by a real tester
+        # AND driven by a script, so `2.26.8.18.1rc2` holds 5,253 fake turns and
+        # 16 real ones. No version rule can split that; this one does not try.
+        #
+        # No env override is needed to test the sink by hand: a developer running
+        # a REAL turn has a session id and LLM calls, so it still emits.
+        #
+        # The log line above is deliberately NOT gated — whoever is running the
+        # driver keeps full local diagnosability.
+        if not self._session_id and tc.llm_calls == 0:
+            return
+
         # Analytics sink — same settings-resolution pattern as the
         # ds_connect_* events (anton/tools.py): the session's settings when
         # the host provided AntonSettings, else a fresh resolve so
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2478,6 +2478,24 @@ class ChatSession:
         #     probe, which legitimately runs a turn without one and spent 4.8M
         #     tokens over the same window.
         #
+        # Two hosts run turns without a session id, not one (#379 review):
+        # cowork-server's connector probe, and `anton chat` when episodic memory
+        # is off — `chat.py`'s `current_session_id` is None whenever
+        # `settings.episodic_memory` is False (user-settable, default True). The
+        # probe survives on `llm_calls > 0`; a CLI turn that fails BEFORE its
+        # first model call has neither half and is dropped with the scripts.
+        #
+        # That collateral is accepted, and measured rather than assumed: over 14
+        # days, all 8,569 script-shaped events carry an EMPTY `harness`, while
+        # every `harness="cli"` turn (26) is kept. So the observed loss is zero.
+        #
+        # `harness` is therefore a discriminator that could narrow this guard if
+        # the loss ever becomes non-zero — deliberately not used yet, because it
+        # would make a fix worth 70% of the volume depend on a field ENG-1695
+        # flags as unreliable (33 turns already carry an empty harness from a
+        # host that should set one). Trading a measured zero for that brittleness
+        # is the wrong direction; revisit only with a real CLI case in hand.
+        #
         # Deliberately BEHAVIOURAL, not a version test. The original plan gated
         # on `.dev` in the version; re-measured, that catches only 2,907 of the
         # 8,209 and misses the single largest contaminant (`2.26.8.18.1rc2`,
--- a/tests/test_root_cause_wiring.py
+++ b/tests/test_root_cause_wiring.py
@@ -75,7 +75,7 @@ def _session(workspace, outcomes, n_tool_calls: int):
 
     llm.plan_stream = plan_stream
     llm.plan = plan
-    s = ChatSession(ChatSessionConfig(llm_client=llm, workspace=workspace))
+    s = ChatSession(ChatSessionConfig(llm_client=llm, workspace=workspace, session_id="conv-rc"))
     s._max_turn_tokens = 0  # ceiling off — this ticket must be independent of it
     pending = list(outcomes)
     s.tool_registry.dispatch_tool = AsyncMock(
@@ -369,7 +369,7 @@ async def test_the_scratchpad_exec_path_records_its_traceback(workspace):
         return gen()
 
     llm.plan_stream = plan_stream
-    session = ChatSession(ChatSessionConfig(llm_client=llm, workspace=workspace))
+    session = ChatSession(ChatSessionConfig(llm_client=llm, workspace=workspace, session_id="conv-rc"))
     session._max_turn_tokens = 0
 
     pad = MagicMock()
--- a/tests/test_turn_cost.py
+++ b/tests/test_turn_cost.py
@@ -659,3 +659,31 @@ class TestScriptTrafficIsNotReported:
         with patch("anton.analytics.send_event") as send:
             s._emit_turn_cost()
         assert send.called
+
+    def test_a_dropped_turn_still_closes_its_books(self):
+        """Position matters: the guard must sit BELOW the books-closing block.
+
+        Dropping the event must not drop the cleanup. If the guard is ever moved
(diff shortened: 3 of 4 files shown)


anton/core/session.py
```diff
@@ -2460,6 +2460,40 @@ def _emit_turn_cost(
...
+        # not who is running it — the same build is installed by a real tester
+        # AND driven by a script, so `2.26.8.18.1rc2` holds 5,253 fake turns and
+        # 16 real ones. No version rule can split that; this one does not try.
+        #
+        # No env override is needed to test the sink by hand: a developer running
+        # a REAL turn has a session id and LLM calls, so it still emits.
+        #
+        # The log line above is deliberately NOT gated — whoever is running the
+        # driver keeps full local diagnosability.
+        if not self._session_id and tc.llm_calls == 0:
```
pnewsam, line 2512: **Minor:** The PR states the only turn-running builder that omits a session id is cowork-server's connector probe (kept because `llm_calls > 0`). That enumeration misses the local `anton chat` CLI: `current_session_id = episodic._session_id if episodic.enabled else None` (`chat.py:1371`), where `episodic.enabled` is `settings.episodic_memory` — user-settable, default `True` (`settings.py:131`). With episodic memory disabled, every CLI turn runs with `session_id=None`, and any such turn that fails *before its first LLM call* (auth error, `retry_exhausted`, cancel) now matches this guard and is dropped — the same pre-model-failure population the PR argues it preserves, just for a no-id host. Failed connector probes (0 tokens → 0 `llm_calls`) fall in the same bucket.

Consequence is bounded: it's an opt-out local CLI (default on), the events carry no join key (`conversation_id=""`) and no spend, and they're genuinely indistinguishable from script traffic under the chosen definition — so this is acceptable collateral, not a defect in the code. The ask is only to the *narrative*: correct the "only the connector probe omits the id" claim so a future maintainer knows CLI-with-episodic-off pre-model failures are also silenced. Confidence the code behaves as written: high; that the dropped data matters to anyone: low.

pnewsam (commented):
## Code Review

**Verdict: COMMENT**
**Model: claude-opus-4-8**

Reviewed the full diff at head `fffe879` (guard in `anton/core/session.py`, plus 21 test-fixture `session_id` additions and 7 new tests). The change is sound: I verified the guard's position, the `llm_calls` signal, and independently enumerated the production session builders. One non-blocking finding, inline — the claim that the connector probe is the *only* turn-running path that omits a session id is incomplete, which slightly widens what gets dropped. Not a merge blocker.

### Validation

- Confirmed guard placement in `_emit_turn_cost` (`session.py:2494`): it sits **after** the log line, after the `tc.emitted` latch, and after the owner cleanup that clears `self._turn_cost` / `usage_listener`. The position claims and the two position-pinning tests hold.
- Confirmed `tc.llm_calls` is incremented only inside `TurnCost.add`, installed as the LLM client's `usage_listener` (`turn_cost.py:156`) — so `== 0` reliably means "no usage reported."
- Confirmed `session_id` defaults to `None` and the guard reads session-constant `self._session_id`, so a late finalizer emitting another turn's books can't be misjudged by it.
- Enumerated every non-test `ChatSessionConfig(` site: `cli.py` connect/edit omit the id but only mutate `session._history` — they never run a turn, so never emit (matches the PR). `runtime.py` and `cloud_turn/session.py` take the id from the caller/request.
- Did **not** execute the suite — reviewed from a read-only checkout of head, not the anton working tree. Relying on the reported 478 passing + 6-mutation verification for the test results.

No blocking or major issues. The both-conditions-required design and the test coverage (including the position and double-emit latches) are unusually thorough.

anton/core/session.py
alecantu7: Verified and fixed in b7784787 — thank you, the enumeration was wrong.

Both references confirmed on `origin/staging`: `chat.py` does `current_session_id = episodic._session_id if episodic.enabled else None`, and `episodic_memory: bool = True` is user-settable. So there are **two** no-session-id turn hosts, not one, and a CLI turn failing before its first model call has neither half of the conjunction.

**Measured the collateral rather than reasoning about it — it is currently zero.** 14 days, `turn_completed`, split by whether the event matches the guard:

| `harness` | dropped (script-shaped) | kept |
|---|---:|---:|
| *(empty)* | **8,569** | 71 |
| `anton` | 0 | 3,441 |
| `cloud` | 0 | 193 |
| `cli` | **0** | 26 |

Every script-shaped event carries an **empty** `harness`; every `harness="cli"` turn is kept. Not one CLI turn currently falls in the dropped bucket.

**That also surfaces something better than the narrative fix: `harness` would discriminate the two cleanly.** I had assumed it couldn't, on the theory that eval drivers run through the CLI and would report `cli` — the data says they report empty instead. So `and harness == ""` would remove the collateral entirely.

**Deliberately not doing that**, and the reasoning is now in the code comment: it would make a fix worth 70% of the volume depend on a field ENG-1695 flags as unreliable — 33 turns already carry an empty `harness` from a host that should set one. If script traffic ever acquires a `harness` value, the guard silently stops working and the contamination returns. Trading a **measured zero** for that brittleness is the wrong direction. Revisit only with a real CLI case in hand.

Put the correction at the guard rather than only in the PR body, since that is where the next maintainer looks — including the discriminator and why it is unused, so this isn't re-derived. PR body updated too.

Your read on impact matches mine: opt-out local CLI, no join key, no spend, indistinguishable from script traffic under the chosen definition. Acceptable collateral, now documented as such rather than implied.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

Re-review at head `b778478`. My prior review left one Minor, non-blocking finding: the narrative claimed the connector probe was the *only* turn-running host that omits a session id, missing `anton chat` with episodic memory off. Commit `b778478` addresses it exactly — a comment-only change (18 lines added, 0 removed) at the guard itself. No actionable findings.

### Validation

- Confirmed the guard code is unchanged from my prior review: `if not self._session_id and tc.llm_calls == 0: return`, still below the log line, the `tc.emitted` latch, and the owner cleanup. This commit adds no behavior.
- Verified the new comment's two factual claims on head: `anton/chat.py:1371` sets `current_session_id = ... if episodic.enabled else None`, and `anton/config/settings.py:131` is `episodic_memory: bool = True` (user-settable, default on).
- The comment records the second no-session-id host, the measured collateral (currently zero across 14 days), and `harness` as a discriminator deliberately left unused per ENG-1695 — placed at the guard, where the next maintainer looks.
- All checks green (run-tests, verdict-eval, CodeQL, Snyk). Relied on CI for the suite; did not run it locally.

PR #380, opened 2026-08-20
Guarantee terminal event

ZoranPandovski:
Some cloud turns were ending as a bare "An unexpected error occurred" with no detail. The cause: a teardown crash (an httpcore GeneratorExit → RuntimeError during stream close, or a CancelledError) unwinds past except Exception, so stream_turn exits without emitting turn_completed or turn_failed. The controller then sees a stream that closed with no terminal event and reports the generic error — and the turn isn't cleanly retry-able.

This adds a finally guard that emits a turn_failed if no terminal event went out, so a turn can never end silently. It turns those silent deaths into a normal, retry-able failure.


tino097 (approved):
LGTM

PR #381, opened 2026-08-20
ENG-1689: stamp endpoint_class and error_type on turn_completed

alecantu7:
Closes gaps 2 and 3 of ENG-1689. `turn_completed` could report *how much* a turn cost but not *where its tokens went* or *why it failed*. Two properties, both anton-only, both purely additive.

## `endpoint_class` — `local` | `mindshub` | `third-party`

**Every existing proxy for this is wrong.** `llm_provider` names the *emitting host's own vocabulary*, not the destination, so our gateway arrives under **two** values — `minds-cloud` from the desktop DB overlay (written by `setattr`, bypassing the validator) and `openai-compatible` from anything constructed through `AntonSettings` (where `_map_minds_cloud_to_openai_compatible` rewrites it). Mechanism documented on ENG-1695.

Measured 14 days, real turns only (ENG-1692's script filter), model set from the **live** prod `/v1/models` catalog rather than a hand-list:

| | turns |
|---|---:|
| `llm_provider = minds-cloud` | 1,739 |
| `llm_provider = openai-compatible`, catalog-only model | **382 — also our gateway** |
| `openai-compatible`, non-catalog | 1,363 |

**Counting the gateway as `minds-cloud` alone runs 18% low — 382 turns, 146 people, 25.2M tokens.** `planning_model` is no better: it needs a live allowlist to interpret (73 distinct models in two days) and can never be authoritative, since nothing stops a user naming a local model `sonnet`.

### The values partition the observed host, not who pays

That distinction is what makes them mutually exclusive. "Is this loopback" and "is this our domain" are facts about a string we hold; "whose money paid" depends on a server-side gate verdict anton cannot see. Keyed on intent, a loopback proxy forwarding to our gateway would land in two buckets at once.

Two consequences, both deliberate and both documented in the module:

- **A local proxy forwarding to our gateway reports `local`** — that is genuinely all anton observes. `local` means "the endpoint was on this machine or LAN", *not* "definitely not our gateway".
- **Private/LAN addresses count as `local`,** not only loopback. Ollama and LM Studio are routinely reached across the network, and a loopback-only rule would file those as `third-party` and understate the exact population this measures.

### The provider gate is the load-bearing part

`AntonSettings.model_post_init` derives `openai_base_url` from `minds_url` — but **only when a provider is `openai-compatible`**. So for a vendor-fixed provider that field can be set, stale, and pointed at us while the turn never goes near us. Consulting it there would invent gateway traffic.

| `planning_provider` | classified from |
|---|---|
| `minds-cloud` | fixed → `mindshub` |
| `anthropic`, `gemini` | fixed → `third-party` (no base URL exists for them in `AntonSettings`; `gemini` only ever arrives via cowork-server's overlay) |
| `openai`, `openai-compatible` | the base URL |
| anything else | `""` — absent beats a guess |

**Note a correction to the ticket:** its table put plain `openai` under "third-party, vendor endpoint". That is wrong — `from_settings` constructs it with `base_url=settings.openai_base_url`, so a user *can* point it at us, and fixing it at `third-party` would misfile them. Ticket updated to match; mutation-tested (dropping `openai` from the URL-decided set fails a test).

### Not built on the existing helper

`cli.py`'s `_looks_like_mdb_ai` matches `{minds_url}` and `{minds_url}/api/v1` but **not `{minds_url}/v1`** — precisely what `model_post_init` builds for `https://api.mindshub.ai`, our production gateway. Its docstring claims parity with `LLMClient.from_settings`, which was fixed while it was not. Reusing it would have reproduced a known bug in a new field (ENG-1695).

Domain matching here is **exact-or-subdomain**, not the substring form `LLMClient`'s vision gate uses — `mindshub.ai.attacker.example` satisfies `"mindshub.ai" in host` and would inflate gateway share. Pinned by test.

Unifying the three existing "is this ours?" idioms onto this helper is **left to ENG-1695 as a separate PR**, deliberately: two of them *gate real behaviour* (native web-search passthrough, Anthropic-vs-OpenAI image format), so fixing them changes what some installs do and deserves its own review rather than riding along with a telemetry change. Sequential, not stacked.

## `error_type` — the exception class at the two terminals that discarded it

Real turns, 14 days: `error` **199 (5.6%)** and `retry_exhausted` **240 (6.7%)**, both with a **median of zero LLM calls and zero rounds** — they die *before the first model call*. A failure to start, not a failure mid-work, and previously indistinguishable.

`error` is a catch-all: a parse failure, a tool crash, a bad config and an anton bug all landed there with the exception object in scope at the assignment and thrown away. `retry_exhausted` is worse — it already formats the exception into a SYSTEM message the model *and the user* read, then never attached it to telemetry.

**Class name only**, via a new `_safe_error_type` sibling to the existing `_safe_error_detail`. The existing helper appends status codes and pydantic field paths, which suit a log line a human reads and are wrong for a property whose purpose is a *groupable* distribution — a compound value spreads one failure mode across many rows. Both are content-free; they differ only in cardinality, and the docstring says so.

Not stamped on `cancelled`, so ordinary user Stops cannot inflate an error breakdown. Not stamped on `handback_verifier_failure` (221 turns, 6.2%) either — that is a verdict, not a raise, and diagnosing it is a separate problem.

## Also fixed here

The emit-site comment claiming `llm_provider` *"separates gateway traffic from BYOK in queries"*. It cannot, and that belief is what produced the 18% undercount. Corrected in place rather than deferred, since leaving a false claim on the line directly above the field that *does* answer it is worse than the one-line overlap with ENG-1695.

## What is NOT in this PR

- **`key_source` is dropped from the ticket entirely** — a layer error, not a deferral. ENG-1611's `key_source` is a column on `usage_events`, *"resolved from the gate verdict at write time and never trusted from the caller"*. The populations are disjoint (a local-Ollama turn produces no ledger row at all, so `local` is unreachable there), "BYOK" means different things in the two tickets, and anton cannot compute it regardless — both key kinds start with `mdb_`. Nothing was ever blocked on Lucas; that was my misreading, now corrected on the ticket.
- Gaps 1 and 4 (`aid`, `anton_version` on a `cowork-desktop` event) — cowork repo, and they need ENG-713's over-merge constraint honoured (property, never alias).

## Security pass

- The **raw base URL is never emitted** — a closed set of three labels, asserted by a test that also checks no digit, dot, colon or slash survives into the value. This is the whole reason it is a class rather than the URL.
- **Exception messages are never emitted**, only class names — messages carry file paths and payload fragments.
- The domain check is **not substring-spoofable** (test covers four lookalike hosts).
- Both new helpers are **exception-proof**: `classify_endpoint` runs on the analytics path, which must never disturb the turn; `_safe_error_type` runs inside an `except` handler, where an escape would turn a handled failure into a dead turn. A hostile settings object whose property raises is covered by test.
- No new input is parsed from an untrusted source: the base URL comes from the user's own settings.

## Verification

- **42 new tests** in `tests/test_endpoint_class.py` — host partition (loopback, IPv6, bare `host:port`, private ranges, link-local, `.local` mDNS, all three gateway shapes, case-insensitivity), the provider gate, four spoofing attempts, the privacy constraint, and the three real `model_post_init` derivations exercised through **actual `AntonSettings`** rather than hand-written URLs.
- **5 new tests** in `tests/test_turn_cost_terminals.py`, driving the real `turn_stream` rather than poking the classifier — matching that file's stated philosophy, since these bugs live in the wiring.
- **An AST seam guard with a positive control.** Deleting a comparable emit-site argument once passed the entire suite because no test drove that path, so the seam is pinned structurally — and the guard's own predicate is verified against a source that omits the keyword, so a bug in the predicate cannot make it vacuous.
- **Full suite: 2372 passed, 31 skipped** on Python 3.12 (pinned via `UV_PYTHON` to match CI — a local 3.14 resolve produces failures that are purely environmental).
- **Mutation-verified with `PYTHONDONTWRITEBYTECODE=1`** (a same-length edit otherwise reuses cached bytecode and reports a false pass). Nine mutations, **all nine caught**: dropping either emit-site argument; removing the provider gate; loopback-only local; substring domain match; dropping `openai` from the URL-decided set; removing the `//` prefix in `_host_of`; and removing each of the two `error_type` stamp sites.

## Merge notes

No ordering constraint against the other open Harness PRs — this adds two new property names and touches no field they read. ENG-1695's helper unification should land *after* this, from `staging`, not stacked on it.

Unblocks ENG-1690, which needed six live prod gateway probes and a Langfuse cross-check to establish only that a failure was anton-side.

### DIFF
--- a/anton/core/llm/endpoints.py
+++ b/anton/core/llm/endpoints.py
@@ -1,0 +1,175 @@
+"""Where a turn's tokens actually went — MindsHub, a third party, or on-box.
+
+ENG-1689. ``turn_completed`` could report *how much* a turn cost but not
+*where it went*, and every available proxy for that is wrong:
+
+- ``llm_provider`` names the **client shape**, not the destination. Our own
+  gateway arrives under two different values — ``minds-cloud`` from the desktop
+  DB overlay and ``openai-compatible`` from anything constructed through
+  ``AntonSettings`` (the ``_map_minds_cloud_to_openai_compatible`` validator
+  rewrites it, and cowork-server's ``setattr`` overlay bypasses the validator;
+  ENG-1695). Measured 2026-08-20: counting the gateway as ``minds-cloud`` alone
+  undercounts it by 18% — 382 turns, 146 people, 25.2M tokens over 14 days.
+- ``planning_model`` needs a live ``/v1/models`` allowlist to interpret (73
+  distinct models in two days) and can never be authoritative anyway, since
+  nothing stops a user naming a local model ``sonnet``.
+
+Only anton knows the base URL it resolved, so only anton can answer this.
+
+**The values are a partition over the observed host, not over who pays.** That
+distinction is what makes them mutually exclusive: "is this loopback" and "is
+this our domain" are facts about a string we hold, whereas "whose money paid"
+depends on a server-side gate verdict anton cannot see (that lives on
+``usage_events``; see ENG-1611). A rule keyed on intent would put a loopback
+proxy forwarding to our gateway in two buckets at once.
+
+Consequences of reading the host and nothing else, both deliberate:
+
+- A local proxy forwarding to our gateway reports ``local``, because that is
+  genuinely all anton observes. ``local`` therefore means "the endpoint was on
+  this machine or LAN", NOT "definitely not our gateway".
+- A private/LAN address counts as ``local``, not only loopback — Ollama and
+  LM Studio are routinely reached across the network, and a loopback-only rule
+  would file those as ``third-party`` and understate on-box use, which is the
+  population this exists to measure.
+
+The raw base URL is **never** emitted: those carry internal corporate
+hostnames. Only the three labels below leave the process.
+"""
+
+from __future__ import annotations
+
+import ipaddress
+from typing import TYPE_CHECKING
+from urllib.parse import urlsplit
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from anton.config.settings import AntonSettings
+
+#: The endpoint was on this machine or its local network.
+ENDPOINT_LOCAL = "local"
+#: The endpoint was the MindsHub gateway (any of its host/path shapes).
+ENDPOINT_MINDSHUB = "mindshub"
+#: The endpoint was somebody else's — OpenRouter, Databricks, a vendor API.
+ENDPOINT_THIRD_PARTY = "third-party"
+#: Unrecognised provider. Absent beats wrong: a value here would be a guess.
+ENDPOINT_UNKNOWN = ""
+
+VALID_ENDPOINT_CLASSES = frozenset(
+    {ENDPOINT_LOCAL, ENDPOINT_MINDSHUB, ENDPOINT_THIRD_PARTY}
+)
+
+# Registrable domains we serve the gateway from. `mdb.ai` is the legacy host
+# and still live. Matched as an exact host or a dot-delimited subdomain —
+# never as a substring, which `mindshub.ai.attacker.example` would satisfy.
+_MINDSHUB_DOMAINS = ("mindshub.ai", "mdb.ai")
+
+# Hostname suffixes that mean "this machine / this network" without being IPs.
+# `.local` is mDNS (Bonjour), how an Ollama box on the LAN is usually named.
+_LOCAL_SUFFIXES = (".localhost", ".local", ".internal", ".lan", ".home.arpa")
+
+# Providers whose destination is fixed by the provider itself, so the base URL
+# is irrelevant (and, for `anthropic`, not even passed — see
+# `LLMClient.from_settings`). `gemini` is absent from AntonSettings entirely
+# and only ever arrives via cowork-server's overlay (ENG-1695).
+_VENDOR_FIXED = {
+    "anthropic": ENDPOINT_THIRD_PARTY,
+    "gemini": ENDPOINT_THIRD_PARTY,
+    "minds-cloud": ENDPOINT_MINDSHUB,
+}
+
+# Providers that DO read `openai_base_url`, so the URL decides. Both are in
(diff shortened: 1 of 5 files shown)


pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

Reviewed ENG-1689 Gaps 2 and 3 — `endpoint_class` (a provider-gated host partition) and `error_type` (the exception class at the `error` and `retry_exhausted` terminals). Checked the two failure modes most likely to be silently wrong; both are handled. No merge-blocking issues; one non-blocking test-coverage question below.

### Validation
- Confirmed no `NoneType` pollution: at the `retry_exhausted` stamp (`session.py:3783`), `_agent_exc` is the live `except Exception as _agent_exc` clause variable, so it is always a real exception; the `error` branch guards on `exc is not None`.
- Confirmed the stale-`error_type` clear: the `cancelled` branch zeroes a value a prior `retry_exhausted` stamped, so an ordinary Stop after failed retries cannot enter an error breakdown — matches `test_stop_after_exhausted_retries_does_not_report_an_error_type`.
- Confirmed `_turn_settings` is captured before the ambient-`AntonSettings()` fallback, so `classify_endpoint` never reports the process environment as the turn's endpoint; `CoreSettings`/`None` → `""`.
- Spot-checked the classifier: exact-or-subdomain match rejects `mindshub.ai.attacker.example`; vendor-fixed `anthropic` with a stale MindsHub base URL → `third-party`; all three `model_post_init` URL shapes → `mindshub`.

### Findings

**Minor:** the ticket's "Done when" list asks for a test that a turn switching endpoint type mid-conversation is tagged correctly per turn. `endpoint_class` is computed per turn from `self._settings` at each `_emit_turn_cost`, so the mechanism is correct, but I do not see a test asserting that two turns in one conversation get different labels. Worth adding that case or confirming the checklist item is considered satisfied by the per-turn construction. Same read-at-emit behavior as the existing `llm_provider`, so there is no regression either way.

PR #383, opened 2026-08-20
fix(workspace): retry initialize once on stale NFS handle (ESTALE)

hamishfagg:
Live pods crash the turn with `OSError: [Errno 116] Stale file handle` on `<workspace>/.anton/anton.md` when another EFS client (cowork-server's per-turn `stage_project_instructions` unlink/re-copy, or a conversation-dir wipe) removes an inode the pod's NFS client still has cached. The failed stat invalidates the cached dentry, so a single retry performs a fresh lookup and succeeds.

`initialize()` now runs its body via `_initialize_once()` and retries exactly once on ESTALE; other OSErrors propagate unchanged. Unit tests cover both paths.

Base is `main` (hotfix line); needs the usual sync back into staging.

### DIFF
--- a/anton/workspace.py
+++ b/anton/workspace.py
@@ -8,6 +8,7 @@ Handles:
 
 from __future__ import annotations
 
+import errno
 import os
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


SailingSF (approved):
lgtm

PR #388, opened 2026-08-21
Run eval in PRs (CI)

ea-rus:
**An MVP version:**

trigger:
- in anton repo you put 'run-evals' label on pr
  - it runs eval with: anton=current_branch, cowork-server=staging
- to run once again: remove/add label

result:
- added comment into pr with result of evals
- just numbers, without comparison with some base line

test:
- https://github.com/mindsdb/cowork_evals/blob/main/datasets/dialog_context.jsonl
  - 2 min and 0.5$ per run
- one model: MindsHub Air
- 3 repeats

Depends: https://github.com/mindsdb/cowork_evals/pull/17

Fixes: https://linear.app/mindsdb/issue/ENG-1828/run-evals-in-prs

### DIFF
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -98,7 +98,12 @@ jobs:
           set -euo pipefail
           : "${MINDS_API_KEY:?MINDSHUB_API_KEY secret is not set}"
           : "${REPO_TOKEN:?no read token — set the GH_PULL_REQUEST_READ_TOKEN secret or pass repo_token}"
-          : "${LANGFUSE_PUBLIC_KEY:?no Langfuse key — token and cost columns would be empty}"
+          # Not required for now — Langfuse secrets aren't provisioned in CI yet.
+          # eval_service degrades to local-only scoring when they're absent, so
+          # this just means empty token/cost columns, not a failure.
+          if [ -z "${LANGFUSE_PUBLIC_KEY:-}" ]; then
+            echo "::warning::no Langfuse key — token and cost columns will be empty"
+          fi
 
       - name: Check out anton
         if: steps.fork.outputs.skip != '1'
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -1,0 +1,214 @@
+name: Anton evals
+
+# Behavioural baseline for the Anton harness: runs the `dialog_context` dataset
+# through a cowork-server built from this PR's anton sha, then posts accuracy,
+# tokens and latency as a PR comment.
+#
+# This reports numbers; it does not gate. Agent runs vary between passes, and a
+# red check on noise trains people to click "re-run" without reading. The job
+# fails only when it could not measure anything at all.
+#
+# Required repository secrets:
+#   MINDSHUB_API_KEY   prod gateway key — also seeded into the built cowork-server.
+#   LANGFUSE_HOST / LANGFUSE_PUBLIC_KEY / LANGFUSE_SECRET_KEY
+#                      the project the PROD gateway writes to. A dev-project key
+#                      here yields empty token/cost columns, not an error.
+#   GH_PULL_REQUEST_READ_TOKEN
+#                      read access to mindsdb/cowork-server + mindsdb/cowork_evals.
+#                      The automatic GITHUB_TOKEN is scoped to this repository
+#                      alone — same-org membership does not extend it.
+#
+# Credentials come from secrets ONLY. This repository is public, and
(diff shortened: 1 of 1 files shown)


.github/workflows/evals.yml
```diff
@@ -0,0 +1,239 @@
...
+          {
+            echo "$marker"
+            cat report.md
+            echo ""
+            echo "<sub>anton \`${SHA:0:8}\` · cowork-server \`$COWORK_REF\` · $REPEATS replicates</sub>"
+          } > comment.md
+          # One comment per PR, edited in place: a fresh comment per run buries the
+          # conversation under near-identical tables.
+          id=$(gh api "repos/$GITHUB_REPOSITORY/issues/$PR/comments" \
+                 --jq "[.[] | select(.body | startswith(\"$marker\")) | .id] | first // empty")
```
tino097, line 241: This `gh api` call has no `--paginate`, so it only sees the first page (30) of PR comments. Once a PR passes 30 total comments — common on a long-lived PR in this repo — the marker lookup returns empty every time, `id` stays unset, and the workflow always falls into the POST branch. That's a fresh duplicate comment on every run instead of the edit-in-place behavior the comment above says this is for.

Suggest adding `--paginate` (and `jq -s 'add'` or equivalent to flatten pages) to the search.

.github/workflows/evals.yml
```diff
@@ -0,0 +1,239 @@
...
+      # exist.
+      - name: Fail loudly on a missing credential
+        if: steps.fork.outputs.skip != '1'
+        env:
+          MINDS_API_KEY: ${{ secrets.MINDSHUB_API_KEY }}
+        run: |
+          set -euo pipefail
+          : "${MINDS_API_KEY:?MINDSHUB_API_KEY secret is not set}"
+          : "${REPO_TOKEN:?no read token — set the GH_EVALS_READ_TOKEN secret or pass repo_token}"
+          : "${LANGFUSE_PUBLIC_KEY:?no Langfuse key — token and cost columns would be empty}"
```
tino097: This guard only asserts `LANGFUSE_PUBLIC_KEY`, but `LANGFUSE_HOST` and `LANGFUSE_SECRET_KEY` are also required (both are wired into `env` above). If either of those is unset while the public key is present, this step passes silently and the failure only surfaces later as a confusing eval/report error, or as silently empty token/cost columns — the exact failure mode this step is meant to catch early.

Suggest asserting all three.

.github/workflows/evals.yml
```diff
@@ -0,0 +1,239 @@
...
+    # plus a `uv sync` per harness build. Anything past this is hung rather than
+    # slow, and the GitHub default of 360 would hold a runner for six hours to
+    # establish that.
+    timeout-minutes: 120
+    env:
+      COWORK_REF: ${{ inputs.cowork_ref || 'staging' }}
+      BASELINE_REF: ${{ inputs.baseline_ref || 'staging' }}
+      EVALS_REF: ${{ inputs.evals_ref || 'main' }}
+      REPEATS: ${{ inputs.repeats || '3' }}
+      ANTON_SHA: ${{ github.event.pull_request.head.sha || inputs.anton_ref || github.sha }}
```
tino097, line 72: `BASELINE_REF` is resolved to a commit sha via `git rev-parse` before use elsewhere in the workflow, but `ANTON_SHA` is taken verbatim from `inputs.anton_ref` for manual `workflow_dispatch` runs — it's never resolved. A dispatch with `anton_ref: staging` and the default `baseline_ref: staging` ends up comparing `staging` (branch name) against a resolved sha for what's actually the same commit.

cowork_evals' `_drop_duplicate_variants` compares these ref fields as raw strings, so the two variants won't collapse as duplicates in that case, silently losing the intended 2x runtime savings on manual dispatch runs. Narrow (manual-dispatch-only) but worth resolving `ANTON_SHA` the same way `BASELINE_REF` is resolved, for consistency.

tino097 (changes requested):
Reviewed the workflow diff. One bug I'd treat as blocking, two narrower edge cases left as suggestions.

.github/workflows/evals.yml
ea-rus: changed staging to sha

.github/workflows/evals.yml
ea-rus: added checks for this keys too

.github/workflows/evals.yml
ea-rus: added --paginate

tino097 (approved):

ZoranPandovski (approved):

PR #389, opened 2026-08-21
[ENG-1795] Artifact publishing TLS certificate verification failure

tino097:
https://linear.app/mindsdb/issue/ENG-1795/artifact-publishing-not-possible-in-cowork-desktop-tls-certificate

- Artifact publishing ("Share to the Web") failed on Windows with "Connection failed during TLS certificate verification", most likely behind a TLS-inspecting corporate proxy/AV whose root CA is trusted by the OS but not picked up by CPython's own Windows cert-store enumeration.
- minds_request() now builds its SSL context via truststore instead of relying on urllib's default context.
- This makes it use OS-native certificate verification (SChannel on Windows), same as the browser.
- Applies to every call that shares this transport: publish, list_published, unpublish, list_versions, activate_version, and the Minds product API calls.
- The existing ANTON_MINDS_SSL_VERIFY=false opt-out is unchanged.
- Same fix pattern as #705. Related to ENG-1819.

Root cause not confirmed against the reporter's actual machine — no raw traceback was available, only the generic UI error message.

### DIFF
--- a/anton/minds_client.py
+++ b/anton/minds_client.py
@@ -16,6 +16,8 @@ import urllib.request
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+import truststore
+
 from anton.core.llm.openai import build_chat_completion_kwargs
 
 if TYPE_CHECKING:
@@ -82,9 +84,13 @@ def minds_request(
     req.add_header("Accept-Encoding", "identity")
     req.add_header("Connection", "keep-alive")
 
-    ctx = None
+    # truststore delegates verification to the OS-native trust APIs
+    # (SChannel on Windows, Security framework on macOS) instead of
+    # CPython's own cert-store enumeration, which some Windows machines
+    # behind a TLS-inspecting corporate proxy/AV don't pick up even though
+    # the injected root CA is trusted by the OS (and thus the browser).
+    ctx = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
     if not verify:
-        ctx = ssl.create_default_context()
         ctx.check_hostname = False
         ctx.verify_mode = ssl.CERT_NONE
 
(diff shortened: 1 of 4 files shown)


pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

I reviewed the complete diff (4 files, +63/-2): the `minds_request` SSL-context change in `anton/minds_client.py`, the new test, and the `truststore>=0.10.4` dependency addition. The change swaps urllib's implicit default SSL context for a `truststore.SSLContext`, delegating verification to OS-native trust stores — a correct, low-risk fix for the reported Windows corporate-proxy TLS failure. No actionable findings.

### Validation

- Read the full `minds_request` body at head SHA `447104e`. The `verify=False` opt-out sets `check_hostname = False` before `verify_mode = ssl.CERT_NONE` — correct ordering, avoids the `ValueError` from disabling verify while hostname-checking is on.
- Confirmed `ssl` is still imported at module top; `ssl.PROTOCOL_TLS_CLIENT` and `ssl.CERT_NONE` remain resolvable after the change.
- Checked `requires-python = ">=3.11"` against truststore's 3.10+ floor — compatible on all supported interpreters.
- CI: all checks pass, including `run-tests` (4m43s), Snyk security/license, and CodeQL.

### Notes (non-blocking)

- **Nit:** `tests/test_minds_client_ssl_context.py` mocks `urllib.request.urlopen`, so both tests only assert how the context is *constructed* (type + `verify_mode`/`check_hostname`), not that truststore actually skips OS verification when `CERT_NONE` is set. That's a reasonable boundary for a unit test — a real handshake can't be unit-tested — but the tests don't prove the `verify=False` opt-out still bypasses verification end-to-end. truststore does honor `CERT_NONE` (it guards `_verify_peercerts` on `verify_mode != CERT_NONE`), so the behavior is correct; just flagging that the test name slightly overstates what's verified.
- The PR body is honest that the root cause is unconfirmed against the reporter's machine. The fix is a superset improvement regardless — truststore also closes the known macOS python.org keychain gap — so shipping it as a probable fix is sound even without a confirmed traceback.

PR #390, opened 2026-08-21
ENG-1486: emit a per-tool-call tool_completed analytics event

alecantu7:
Closes [ENG-1486](https://linear.app/mindsdb/issue/ENG-1486/anton-computes-a-per-tool-success-verdict-and-throws-it-away-nothing).

Anton computes a definitive per-tool success verdict for the UI's `tool_done` marker and threw it away — nothing could answer *which tools fail, how often, or how slowly* (the gateway is structurally the wrong vantage point: 67% of its `tool:*` spans are structured-output schemas, and anton's own handling hides failures from it).

## What

- **`session.py`** — emit `tool_completed` via `send_event` at the point every **executed** tool call flows through: the streaming dispatch loop's common tail — covering all three branches (inline scratchpad exec, interactive, general) and both failure shapes (a raise AND `ToolOutcome.ok=False` — the PR #304 distinction) — plus the non-streaming `turn()` loop's own dispatch site (zero production callers today; instrumented so the per-call contract holds on every path — self-review finding 2). `name` is bounded to 200 chars — it is model output (self-review finding 3). Discarded/damaged calls `continue` before the timer and never emit; structured-output extractions (`generate_object_code`) never enter tool dispatch at all.
  - Payload: `name`, `ok` (`"true"`/`"false"`/`"unknown"` — `None` = unmigrated handler or cancelled exec, reported honestly rather than coerced), `duration_ms` (human wait subtracted), `error_type` (exception **class name** when the failure was a raise, `""` otherwise — never the message), plus two join keys: `conversation_id` and `turn_index`, mirroring `turn_completed`'s values exactly (same names, same derivation) so a tool row joins to its parent turn row in PostHog and, via `conversation_id` → Langfuse `sessionId`, to the gateway trace. All string values, per `send_event`'s wire contract.
  - New `_tool_error_type` variable kept apart from `tool_reason`, which on the `ToolOutcome` path is prose (traceback line / handler message) and can carry paths and user input.
  - `answer_wait_s` now resets with `_tool_t0` for **every** dispatched call, not only on the general branch. This is a **prerequisite of the new cross-branch emit**, not a fix of a pre-existing bug: before this PR the general branch was both the only resetter and the only subtractor, so the old maths was self-consistent (self-review finding 1 corrected an earlier claim here).
- **`analytics.py`** — `tool_completed` registered in `_POSTHOG_EVENTS` (the ENG-1495 direct sink) with its properties documented in the register. A new event *name* through the collector would reach nothing (ENG-1355).

## Ticket decisions (as recommended in the ticket)

1. **Per-call event** (not per-turn counters) — volume checked against current ingest, see below.
2. **`error_type` included**, exception class name only.
3. **PostHog, not Langfuse** — client-side product telemetry keyed on install. The `conversation_id` join key is the bridge for trace forensics: PostHog → Langfuse `sessionId` → the turn's gateway trace.

## Relationship to ENG-1689 / #381 (no duplication)

#381's `error_type` is **turn-level** — the exception that *ended the turn*. Tool exceptions are caught by the dispatch loop and converted into a `Tool 'x' failed:` result the model reacts to; they never escape to end the turn, so they are invisible to #381 **by construction**, and a turn-ending exception never books here. Same name and semantics (class name only) on two disjoint populations. `git merge-tree` confirms the branches merge cleanly. If #381 lands first, a one-line rebase switches `type(exc).__name__` to its `_safe_error_type` helper.

## Volume check (decision 1's pre-merge condition)

Measured 2026-08-21, project 424726, last 7 days: 4,075 real turns (`llm_calls > 0`) totalling 15,506 rounds (avg 3.8/turn). At ~1–2 tool calls per round this is roughly **15–30k `tool_completed` events/week**, against a current total project ingest of ~53k events/week — a +30–60% relative increase but small in absolute terms.

## Verified end-to-end (not only unit-tested)

Ran a real turn on this branch (real LLM via the gateway, real analytics, no mocks) prompting a scratchpad exec of `raise ValueError('eng1486_e2e_probe')`. The event is visible in PostHog project 424726:

```
2026-08-21T13:46:12-07:00  distinct_id=a64c660a636b44be
{'name': 'scratchpad', 'ok': 'false', 'duration_ms': '270', 'error_type': '',
 'conversation_id': 'eng1486-e2e-probe', 'turn_index': '1'}
```

This exercised the **inline scratchpad-exec branch** (cell-error verdict, no raise → `error_type=""`), which the unit harness can't reach — so both the raise path (unit) and the `ToolOutcome`/cell path (unit + live) are confirmed, join keys included.

## Tests (all mutation-verified — each mutation flips a named test to FAIL)

`tests/test_tool_completed.py`, driving the real `turn_stream`:
- raise → `ok="false"`, `error_type="RuntimeError"`; `str(exc)` (a path-bearing message) appears in **no** property — killed by `error_type = str(exc)` mutation
- `ToolOutcome.ok=False` → `ok="false"`, `error_type=""` (the prose `reason` is not smuggled in)
- `ok=None` → `"unknown"`, never coerced — killed by coerce-to-true mutation
- payload is exactly `{name, ok, duration_ms, error_type, conversation_id, turn_index}`, all strings, result content absent — killed by extra-property mutation
- join keys equal the same run's `turn_completed` values — killed by blanking `conversation_id` and by deriving `turn_index` from a different counter on the tool side only
- duration subtracts real wall-clock human wait — killed by dropping the subtraction
- a consumer stalling 0.4s right after `tool_done` does not inflate the booked duration, which equals the `eta_seconds` the UI displayed — killed by re-reading the clock at the emit (review nit from @pnewsam; measured 401ms billed to a ~0ms tool before the fix)
- one event per call; turn survives `send_event` raising
- routing: `tool_completed` takes the PostHog path, never the collector — killed by unregistering it

Two more from self-review, both mutation-verified: the non-streaming `turn()` path emits (killed by deleting that emit) and the model-generated `name` is bounded to 200 chars (killed by dropping the truncation).

Full suite: **2336 passed, 31 skipped**.

## Security check

Performed. The payload carries tool name, verdict, duration, exception class name and the two join ids only — no tool arguments, no result content, no `str(exc)` (asserted in a test with a path-bearing message), no file paths, no credentials. `conversation_id` and `turn_index` already ride `turn_completed` through this same sink — no new privacy surface. `analytics_enabled` opt-out and the CI drop apply unchanged (the new call site goes through `send_event`, which owns both). `distinct_id` is the existing per-install `aid`. `$process_person_profile=False` as with the other direct-sink events.

## Open question the first week of data answers

Is the observed near-zero tool-failure rate real, or does anton absorb failures the gateway never sees? This event is what distinguishes the two.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -4763,6 +4763,26 @@ class ChatSession:
                     # would make the booked duration disagree with the
                     # displayed one (#390 review). Branches that never stopped
                     # the clock fall back to reading it now.
+                    #
+                    # The fallback's `- answer_wait_s` is inert TODAY and kept
+                    # for the day it isn't: `elicit()` is the only writer of
+                    # that counter, and its only callers (`handle_select_path`,
+                    # `handle_ask_user`) both route to the branch above, which
+                    # never reaches this fallback. Verified dynamically too —
+                    # asserting the counter is 0.0 here passes the whole suite,
+                    # across the 29 tests that do reach this line.
+                    #
+                    # KNOWN LIMIT, not covered by that subtraction: the
+                    # interactive branch's tools (`connect_new_datasource`,
+                    # `publish_or_preview` publish) collect human input via
+                    # `prompt_or_cancel`, which does NOT feed `answer_wait_s`.
+                    # Their duration therefore includes however long the human
+                    # spent typing — measured at 401ms for a 0.4s stand-in, so
+                    # minutes of real credential entry. Read p95 for those two
+                    # names as "wall-clock of an interactive flow", never as
+                    # tool performance. Fixing it needs a session-scoped
+                    # accumulator behind `prompt_or_cancel`, which is a change
+                    # to the CLI prompt path rather than to this event.
                     self._emit_tool_completed(
                         name=tc.name,
                         ok=tool_ok,
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2768,7 +2768,11 @@ class ChatSession:
             send_event(
                 settings,
                 "tool_completed",
-                name=name,
+                # `tc.name` is model-generated: a hallucinated tool name is
+                # real signal and SHOULD emit, but unbounded model output has
+                # no business being a property value verbatim. Same idiom as
+                # the dispatch loop's args_summary[:120].
+                name=name[:200],
                 ok="unknown" if ok is None else str(ok).lower(),
                 duration_ms=str(int(duration_ms)),
                 error_type=error_type,
@@ -3427,6 +3431,10 @@ class ChatSession:
                     tool_results.append(damaged)
                     continue
 
+                import time as _time  # same local-import idiom as turn_stream
+
+                _tool_t0 = _time.monotonic()
+                _tool_error_type = ""
                 try:
                     outcome = await self.tool_registry.dispatch_tool(
                         self, tc.name, tc.input, tool_call_id=tc.id,
@@ -3439,6 +3447,19 @@ class ChatSession:
                         ok=False,
                         reason=type(exc).__name__,
                     )
+                    _tool_error_type = type(exc).__name__
+                # Same per-tool-call analytics as the streaming loop's tail
+                # (ENG-1486) — without this, a host on the non-streaming API
+                # silently undercounts. Raw elapsed, no human-wait deduction:
+                # elicitation is structurally unavailable on this path (no
+                # emitter to render a question), so there is no wait to
+                # subtract.
+                self._emit_tool_completed(
+                    name=tc.name,
+                    ok=outcome.ok,
+                    duration_ms=(_time.monotonic() - _tool_t0) * 1000.0,
+                    error_type=_tool_error_type,
+                )
                 result = outcome.content
 
                 if isinstance(result, list):
@@ -4447,12 +4468,14 @@ class ChatSession:
                         continue
 
                     _tool_t0 = _time.monotonic()
-                    # Per dispatched call, not per general-branch call: every
-                    # branch below can elicit() (which accumulates the wait),
-                    # and the duration maths at the bottom of this block runs
-                    # for all of them. Resetting only on the general branch
-                    # left a stale wait from an earlier call in the round to
-                    # be subtracted from the wrong tool's runtime.
+                    # Reset alongside the timer, before ANY branch dispatches:
+                    # the tool_completed emit at the bottom of this block is
+                    # the first reader that runs for every branch, and every
+                    # branch can elicit() (which accumulates the wait). The
+                    # reset used to live on the general branch only — correct
+                    # while that branch was also the only subtractor, but the
+                    # cross-branch emit would have deducted another call's
+                    # leftover wait from this one's runtime.
                     self.answer_wait_s = 0.0
 
                     # The handler's own failure verdict, when it gave one —
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2732,9 +2732,16 @@ class ChatSession:
         is reported as ``"unknown"`` rather than coerced either way.
 
         Payload is deliberately name + verdict + duration + exception CLASS
-        and nothing else. Arguments, result content and ``str(exc)`` routinely
-        carry file paths, user data and credentials-adjacent strings — none of
-        them may ever appear here.
+        plus the two join keys, and nothing else. Arguments, result content
+        and ``str(exc)`` routinely carry file paths, user data and
+        credentials-adjacent strings — none of them may ever appear here.
+
+        ``conversation_id`` / ``turn_index`` mirror ``turn_completed``'s
+        values exactly (same names, same derivation), so a tool failure spotted
+        in PostHog joins to its parent turn row there and, via
+        ``conversation_id`` → Langfuse ``sessionId``, to the gateway trace of
+        the turn it happened in. Both already ride ``turn_completed`` through
+        this same sink — no new privacy surface.
         """
         try:
             # Same settings resolution as `_emit_turn_cost` above: the
@@ -2750,6 +2757,14 @@ class ChatSession:
             # send_event takes STRING values only — every extra is a wire
             # parameter (tests/test_ask_user.py:496 exists because a first
             # draft assumed a (name, props) shape).
+            # The same derivation `_emit_turn_cost` uses at close-of-books:
+            # the live TurnCost carries the authoritative index (an explicit
+            # turn_id on the cloud path, the local counter otherwise), so the
+            # values here and on the turn's own row can never disagree.
+            _tc = getattr(self, "_turn_cost", None)
+            turn_index = (
+                getattr(_tc, "turn_index", 0) or (self._turn_count + 1)
+            )
             send_event(
                 settings,
                 "tool_completed",
@@ -2757,6 +2772,8 @@ class ChatSession:
                 ok="unknown" if ok is None else str(ok).lower(),
                 duration_ms=str(int(duration_ms)),
                 error_type=error_type,
+                conversation_id=str(self._session_id or ""),
+                turn_index=str(turn_index),
             )
         except Exception:
             # Analytics must never affect the tool call that just ran.
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2710,6 +2710,58 @@ class ChatSession:
             # Reporting must never affect the turn that just ran.
             pass
 
+    def _emit_tool_completed(
+        self,
+        name: str,
+        ok: bool | None,
+        duration_ms: float,
+        error_type: str,
+    ) -> None:
+        """Per-tool-call analytics event (ENG-1486).
+
+        The verdict the dispatch loop already computes for the UI's
+        ``tool_done`` marker, sent where it can be aggregated: without this,
+        "which tools fail, how often, how slowly" is unanswerable — the
+        gateway's vantage point is structurally wrong (67% of its tool spans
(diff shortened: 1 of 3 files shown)


alecantu7:
**Self-review findings (adversarial pass @ 7b0730ba)** — three issues, fixes follow as commits on this branch.

**1. The "stale wait" claim is wrong (comment + PR body, medium as a docs defect).** The reset-move comment and the PR body describe a pre-existing bug — "a stale wait from an earlier call in the round could be subtracted from the wrong tool's runtime." Reading the pre-change code refutes this: the only subtraction site was the general branch, and the old reset sat in that same branch before its own dispatch, so the old maths was self-consistent. The reset move is a prerequisite of the new tail emit (which is the first cross-branch reader), not a fix of an old defect. Fixing the comment, the PR body, and the `test_wait_reset_is_per_call_not_per_general_branch` docstring, which overclaims the same thing — that test only guards the negative-clamp; both harness-reachable paths reset before dispatch either way.

**2. The non-streaming `turn()` path dispatches tools without emitting (low).** `session.py:3315` is a second `dispatch_tool` site inside `turn()`'s own loop, uncovered by the tail emit — so "one event per executed tool call" is false there. `.turn(` has zero production callers today (grepped anton, cowork-server, scratchpad-controller, cowork — one cowork-server test file only), which is why this is low rather than high. Fix: emit from that loop too (verdict and exception already in hand; raw elapsed, since elicitation — the human-wait case — is structurally unavailable on the non-streaming path), plus a test driving `turn()` with a failing tool.

**3. `name` is an unbounded model-generated string (low).** `tc.name` comes from the model's tool call; a hallucinated or degenerate name flows into the PostHog property verbatim, unbounded. An unknown-tool failure SHOULD emit (that's real signal), but the value needs a bound. Fix: truncate at the emit (`name[:200]`), matching the `args_summary[:120]` idiom in the same loop.

Also verified while hunting (no action): conftest kills analytics suite-wide so local pytest runs can't leak events, and the fresh-`AntonSettings` fallback in the emitter honors it; the blank-`analytics_url` kill-switch sits upstream of the PostHog routing so opted-out installs stay dark on the new event; control-flow scan found no `continue`/`break` between the timer and the emit.

alecantu7:
All three self-review findings addressed in 1bdcdb7a: (1) reset-move comment, test docstring and PR body reworded — the earlier reset is a prerequisite of the cross-branch emit, not a fix of a pre-existing bug; (2) the non-streaming `turn()` dispatch loop now emits too, with a seam-guard test (mutation-verified: deleting the emit fails it); (3) `name` bounded to 200 chars at the emit, test asserts the bound (mutation-verified). Full suite 2335 passed.

anton/core/session.py
```diff
@@ -4511,6 +4616,27 @@ async def _stream_and_handle_tools(
...
+                    # ToolOutcome.ok=False failure shapes — and only executed
+                    # calls: discarded/damaged calls `continue` before
+                    # `_tool_t0`, and structured-output extractions
+                    # (`generate_object_code`) never enter tool dispatch at
+                    # all. Human wait is subtracted so an ask_user answered
+                    # after four minutes is not booked as a slow tool.
+                    self._emit_tool_completed(
+                        name=tc.name,
+                        ok=tool_ok,
+                        duration_ms=max(
```
pnewsam, line 4792: **Nit (non-blocking):** `duration_ms` is recomputed here rather than reusing the `_tool_elapsed` already computed a few lines up (session.py:4586). Between those two `monotonic()` reads the code yields `tool_done` (and a `StreamToolResult` for a scratchpad `dump`), so the emitted duration includes however long the consumer takes to pull those events. The bias is always upward and typically sub-millisecond, but it lands on the exact "how slowly" metric this event exists to answer. Capturing one duration before the `tool_done` yield and passing it to the emit would make the booked duration match the displayed one and exclude consumer-pull latency. Telemetry is fire-and-forget and the magnitude is small, so this is not merge-blocking.

pnewsam (approved):
**Model: claude-opus-4-8**

Scope: `_emit_tool_completed`, its two call sites (streaming tail + non-streaming `turn()`), the `answer_wait_s` reset move, the `_POSTHOG_EVENTS` registration, and the new test file. The design holds up — the emit sits at the one point every executed call converges, both failure shapes (`raise` and `ToolOutcome.ok=False`) map to `ok="false"`, `error_type` is the exception class name only, and the payload is exactly the six string keys with no arguments or result content. No blocking findings; one non-blocking nit inline.

**Validation**
- All three streaming branches set `tool_ok` and `_tool_error_type` before the shared emit, so it never reads an unbound name.
- Join keys can't disagree: the tool emit and `turn_completed` both read the same `self._turn_cost.turn_index`, set before `_stream_and_handle_tools` runs.
- The `answer_wait_s` reset move is safe — the only two readers both subtract it and both run after the per-call reset; no turn-level reader exists.
- Routing confirmed: `tool_completed` takes the direct `/capture/` PostHog path with `$process_person_profile=False`, not the collector that drops unknown names.
- CI is fully green (12 checks).

Unusually well-tested (mutation-verified seam tests plus a live end-to-end run) and the privacy surface is genuinely unchanged from `turn_completed`.

anton/core/session.py
alecantu7: Good catch — fixed in c5701ae4, and it was worth more than a nit: the bias is unbounded, not sub-millisecond, because the consumer sets the pace. A repro test where the consumer stalls 0.4s right after `tool_done` booked **401ms for a tool that ran in ~0ms** on the old code.

Not a direct reuse of `_tool_elapsed`, though — it's assigned only inside the general branch, so referencing it at the shared emit would raise `UnboundLocalError` on the scratchpad-exec and interactive paths. Instead it's hoisted to `None` beside `_tool_t0`; the emit prefers it when a branch stopped the clock and falls back to reading the clock otherwise. So the general branch now books exactly the value `tool_done` displays (the test asserts that equality, not just a bound), and the other two branches are unchanged.

Mutation-verified: reverting the emit to re-read the clock fails the new test. Full suite 2336 passed.

alecantu7:
**Re: the fallback path's human-wait subtraction — checked the caveat first, and it resolves to not-needed. But the worry was right via a different mechanism, so 829d31d7 does something about that instead.**

**1. The requested test would assert on an unreachable state.** `answer_wait_s` has exactly one writer — `elicit()` at `interaction/elicit.py:218`. `elicit()` has exactly two callers: `handle_select_path` and `handle_ask_user` (`tool_handlers.py:911,960,1081`). Neither is an interactive-branch tool: that branch is `connect_new_datasource` and `publish_or_preview` publish only, and `select_path` is *explicitly excluded* from it by the branch's own comment. Both eliciting tools therefore route to the general branch, which assigns `_tool_elapsed` and never reaches the fallback.

Verified dynamically as well as statically — I replaced the fallback expression with one that asserts `answer_wait_s == 0.0` and ran the full suite: **2336 passed**, across the **29 tests that do reach that line** (confirmed separately by booby-trapping the fallback to raise: 29 failures). So the term is provably inert, not merely untested. Your three-way mutation table is reproduced and explained by this: dropping it changes nothing because it always subtracts zero.

Kept rather than deleted — it is correct-by-construction the day an eliciting tool is routed there — but the comment now says it is currently inert instead of implying it does work.

**2. The interactive branch does have real human wait, and it is NOT tracked at all.** Both of its tools collect input via `prompt_or_cancel` (`commands/datasource/connect.py:202,420,424,477,565` and `tools.py:525`), which never touches `answer_wait_s`. So no subtraction can help them. Measured with a probe — a `connect_new_datasource` whose dispatch stalls 0.4s standing in for typing:

```
{'name': 'connect_new_datasource', 'ok': 'true', 'duration_ms': '401', ...}
answer_wait_s at end: 0.0
```

401ms booked for 0.4s of "typing"; real credential entry is minutes. This is newly visible because nothing measured that branch's duration before — it yields no `tool_done`. Fixing it properly needs a session-scoped accumulator behind the CLI prompt path, which is a change to `prompt_or_cancel`, not to this event — happy to file that as a follow-up if you want it.

For now it is documented at the emit and **pinned by a test**, so `p95(connect_new_datasource)` is read as interactive wall-clock rather than tool speed. That test also closes a hole nobody had noticed: **nothing asserted the interactive branch reaches the shared emit at all.** Mutation-verified — restricting the emit to the general branch fails it.

Full suite: 2337 passed, 31 skipped.

alecantu7:
**Rebased onto fresh `staging` (`bf70d3a7`) — and that also explains why CI went quiet on the previous head.**

`829d31d7` had no CI runs at all beyond the CLA check. Cause: `staging` moved underneath us (five commits, including **#381 / ENG-1689** and #371) and the branch went `CONFLICTING`. GitHub builds `pull_request` runs against the `refs/pull/N/merge` ref, which cannot be created for a conflicted branch — so the checks were never queued, rather than queued-and-failing. Worth knowing as a general trap: **a silent CI is not a green CI**, and the PR page shows it as an absence rather than a warning.

Resolved in the rebase:
- `_POSTHOG_EVENTS` — union, keeping staging's new `scratchpad_package_installed` alongside `tool_completed` (the only textual conflict).
- **Adopted `_safe_error_type`** at both `tool_completed` sites, as this PR said it would once #381 landed. One definition of "class name only, never raises" across both events instead of a local `type(exc).__name__` (convention 7). Behaviour is identical apart from the helper's defensive guard; the `str(exc)`-leak mutation test still kills it.

Verified after the rebase, against the merged ENG-1689 code rather than the old base: **2436 passed, 31 skipped** locally, and all 8 CI checks green on `bf70d3a7`. Approval predates the rebase — content is unchanged apart from the two items above, but re-glance if you'd rather.

PR #391, opened 2026-08-21
ENG-1689: record that the install-id format is depended on cross-repo

alecantu7:
Docstring only. Follow-up to review feedback on [cowork#707](https://github.com/mindsdb/cowork/pull/707), doing the *"or vice versa"* half so the coupling is discoverable from the **producer** as well as the consumer.

## Why

`get_installation_id()` is now read across three repos, and **the format matters, not just the value**:

- `cowork-server` serves it on `/health` ([#374](https://github.com/mindsdb/cowork-server/pull/374))
- `cowork`'s renderer validates it as **lowercase hex** before stamping it as the join key that connects anton's per-turn cost events to an identified user ([#707](https://github.com/mindsdb/cowork/pull/707))

Nothing in this repo hinted that anyone depends on the encoding. The review put it well — a format change here would break a consumer that has no shared source of truth for it, and the symptom would be **every join returning zero rows**, which reads as *"no data"* rather than *"the key was discarded"*.

## What the note records

**Widening is safe; leaving hex is not.** The consumers deliberately do *not* pin the width (`{8,64}`, not `16`), precisely so this side stays free to change `[:16]`. What they cannot absorb is a move off lowercase hex.

**`"unknown"` is filtered downstream, and why.** Every unfingerprintable machine returns that same string, so a join on it would merge them all into one identity — ENG-713's over-merge outcome without an alias.

**`"unknown"` is now in the `Returns:` section.** It previously promised only *"a 16-character hex string"* while the `except` branch returns `"unknown"`. That gap is not academic: it is exactly why the first draft of cowork-server#374 passed the sentinel through as a join key. The docstring under-described the contract, and the code downstream believed it.

## Verification

Docstring only — no executable change. Full suite **2361 passed / 31 skipped** on this head.

## Merge notes

Independent of everything. Safe in any order relative to anton#381, cowork-server#374 and cowork#707; it documents a contract those PRs rely on rather than changing it.

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -238,9 +238,25 @@ def get_installation_id() -> str:
     moving off lowercase hex is not, and would silently drop the key rather
     than fail loudly.
 
-    The ``"unknown"`` return below is filtered by those consumers too: every
-    unfingerprintable machine reports the same string, so a join on it would
-    merge them into one identity.
+    Two limits worth knowing before changing either, since both are the change
+    someone would plausibly make:
+
+    - **The width freedom stops at 64.** The consumer bound is ``{8,64}``, so
+      anything up to a full SHA-256 hexdigest is fine, including dropping the
+      truncation entirely. A longer digest left untruncated is not — SHA-512 is
+      128 hex characters and would be rejected.
+    - **A replacement sentinel must stay NON-HEX.** cowork-server filters
+      ``"unknown"`` by name, but cowork's renderer rejects it only because it
+      fails the hex test — and the renderer is the side that mints the join
+      key. So a hex-shaped sentinel (``"0000000000000000"`` being the obvious
+      accident) would pass validation and stamp every unfingerprintable machine
+      with the *same* key, merging them into one identity. That is the exact
+      over-merge this value is guarded against, arriving through the door the
+      guard is assumed to cover.
+
+    The ``"unknown"`` return below exists for that reason: every
+    unfingerprintable machine reports the same string, so it must never be
+    usable as a join key.
 
     Returns:
         A 16-character hex string (64 bits of entropy), or ``"unknown"`` when
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -230,8 +230,21 @@ def get_installation_id() -> str:
     networking), a random UUID is persisted to ``~/.anton/.installation_id``
     as a one-time fallback. Computed once per process and cached.
 
+    **The format is depended on outside this repo.** cowork-server serves this
+    value on ``/health`` and cowork's renderer validates it as **lowercase
+    hex** before stamping it as the join key that connects anton's per-turn
+    cost events to an identified user (ENG-1689). Those consumers deliberately
+    do *not* pin the width, so widening the ``[:16]`` truncation is safe —
+    moving off lowercase hex is not, and would silently drop the key rather
+    than fail loudly.
+
+    The ``"unknown"`` return below is filtered by those consumers too: every
+    unfingerprintable machine reports the same string, so a join on it would
+    merge them into one identity.
+
     Returns:
-        A 16-character hex string (64 bits of entropy).
+        A 16-character hex string (64 bits of entropy), or ``"unknown"`` when
+        the machine cannot be fingerprinted at all.
     """
     global _cached_aid
     if _cached_aid is not None:


alecantu7 (commented):
Self-review at `d67985f1`, base fresh at `06675efb`.

## Verified

**Docstring only** — AST comparison of `anton/analytics.py` with docstrings stripped is identical to `origin/staging`. Suite **2361 passed / 31 skipped** on CI's 3.13, matching the body.

**Every cross-repo claim holds**, checked against the consumer PRs rather than the prose describing them:

| Claim | Evidence |
| -- | -- |
| cowork-server serves it on `/health` | `cowork/api/v1/endpoints/health.py` in cowork-server#374 |
| cowork's renderer validates lowercase hex | `const AID_SHAPE = /^[0-9a-f]{8,64}$/;` in cowork#707 |
| consumers deliberately do not pin the width | the comment says so in as many words — *"**Hex, but deliberately NOT a fixed width.** … Pinning 16 here would…"* |
| `"unknown"` is filtered by those consumers | cowork-server#374 defines `_AID_SENTINEL = "unknown"` with "the sentinel must not escape"; cowork#707 has `it('rejects the "unknown" sentinel, which would MERGE distinct machines')` |
| the old docstring under-described the contract | `origin/staging` promised only *"A 16-character hex string (64 bits of entropy)."* while the `except` branch sets `_cached_aid = "unknown"` |

So the note is accurate. Two boundaries it does not state, though, and both are exactly the change someone would plausibly make.

## 1. The width freedom has a ceiling

The docstring says *"widening the `[:16]` truncation is safe"*. The consumer regex is `{8,64}`, so that is true up to **64 characters** — a full SHA-256 hexdigest, which is the natural end of that road. Removing the truncation entirely is fine; moving to a longer digest without truncating (SHA-512 is 128 hex chars) would be rejected. One clause — "safe up to a full SHA-256 digest, 64 chars" — closes it, and this docstring exists precisely to state that kind of limit.

## 2. `"unknown"` is filtered by name in one consumer and only incidentally in the other

cowork-server#374 checks the sentinel explicitly. cowork#707 does **not** — it relies on `"unknown"` failing the hex test, and its own comment says the check *"also catches a future sentinel without needing to know its name."* That is only true for a non-hex sentinel:

```
unknown            passes AID_SHAPE = False
0000000000000000   passes AID_SHAPE = True
deadbeefdeadbeef   passes AID_SHAPE = True
ffffffffffffffff   passes AID_SHAPE = True
ABCDEF0123456789   passes AID_SHAPE = False   (uppercase)
a1b2c3             passes AID_SHAPE = False   (too short)
```

So changing the sentinel to anything hex-shaped — `"0000000000000000"` is the obvious accident — would pass the renderer's validation and stamp every unfingerprintable machine with the same join key. That is the ENG-713 over-merge the docstring warns about, arriving through the door it says is guarded. cowork-server would still catch it, but the renderer is the side that mints the join key, so the protection that matters is the incidental one.

The docstring should say the sentinel must stay **non-hex**, not merely that it is filtered. That is a stronger and more useful constraint than the one written, and it is the same shape as the lowercase-hex rule already recorded: a property of the string, not a promise about a consumer.

Both are one-line additions to a docstring whose entire purpose is to be the place this is written down. Worth doing here rather than as a follow-up, since the next person to read it is the person who needs them.

## Not raised

The three producing paths cowork#707 enumerates (`sha256(...).hexdigest()[:16]`, `uuid4().hex[:16]`, and the persisted-file read) collapse to the two this docstring describes, since the file read returns what the UUID path wrote. No gap.

alecantu7:
Both addressed in `7e0d353c`. I verified each against the consumer's actual code rather than the prose, and both hold.

## 1. The width ceiling — added

Confirmed empirically against `AID_SHAPE = /^[0-9a-f]{8,64}$/`:

```
64 chars  -> accepted
65 chars  -> rejected
```

So "widening is safe" is true up to a full SHA-256 hexdigest, and dropping the truncation entirely is fine — SHA-512 untruncated (128 hex) is not. Now stated, since leaving the ceiling implicit in the one docstring whose job is to record this kind of limit was the gap.

## 2. The sentinel must stay non-hex — added, and this was the real find

The note claimed `"unknown"` "is filtered by those consumers". Half true, exactly as you say: cowork-server filters it **by name**, cowork's renderer only rejects it because it **fails the hex test** — and the renderer is the side that mints the join key. Verified against that regex:

```
unknown            -> rejected
0000000000000000   -> ACCEPTED
deadbeefdeadbeef   -> ACCEPTED
```

So renaming the sentinel to anything hex-shaped stamps every unfingerprintable machine with the *same* key and merges them into one identity — the precise over-merge this value is guarded against, arriving through the door the guard is assumed to cover.

It is now written as a property of the string ("a replacement sentinel must stay non-hex") rather than a promise about a consumer, matching the lowercase-hex rule already recorded. That framing is the part I took from your comment, not just the content.

**Deliberately not fixing this in the renderer.** Adding a name blocklist there would undo the reason the shape check exists — *"catches a future sentinel without needing to know its name"* — and would be defeated again the moment the name changed. A hex-shaped sentinel passes any shape check by construction, so the producer is the only place the constraint can actually hold. That is this docstring.

## Also

Took the AST check from your review and made it part of the evidence: `anton/analytics.py` with docstrings stripped is **identical** to `origin/staging`, so this is provably docstring-only. Full suite 2361 passed / 31 skipped.

Agreed on the not-raised point — the persisted-file read returns what the UUID path wrote, so the three producing paths do collapse to the two described. No gap there.

PR #392, opened 2026-08-21
ENG-1852: require user confirmation for a model-supplied single select_path candidate

alecantu7:
## What

`select_path`'s PICK mode auto-resolved whenever exactly one candidate survived filtering — including when that one candidate came from the model's own `candidates` list. In a fresh cowork project exactly one directory exists (`skills`), so "I'd like to give you access to a folder on my computer" reliably ended with the agent's own `skills` subdirectory reported as connected, and the turn marked COMPLETE. Across 348 users in the 2026-08-14 → 08-21 window, not one connected a folder they chose.

Closes [ENG-1852](https://linear.app/mindsdb/issue/ENG-1852).

## How

- **A lone `pattern` match keeps auto-resolving** — a glob search is a genuine discovery, so `auto_resolved: true` is unchanged there.
- **A lone explicit candidate is never accepted silently.** It is the model's guess echoed back, so the user confirms it — enforced in the handler, not the prompt:
  - **Choice-capable hosts (every cowork session today):** a yes/no confirmation card ("Only one candidate was found — use this folder?"). Yes → `resolved` (no `auto_resolved` flag). No / free-typed answer → `cancelled`, with the user's typed words passed through verbatim and a message telling the model a project path is never an answer to a request for one outside it — say so plainly and point at attachment.
  - **Path-picker hosts (CLI):** falls through to the existing picker with the single option — picking it is the confirmation.
  - **Neither (e.g. non-streaming `turn()`, question budget spent):** new `needs_confirmation` status carrying the candidate and the same guidance; never `resolved`.
- Both tool definitions (`SELECT_PATH_TOOL`, `SELECT_PATH_TOOL_PICK_ONLY`) document the new behaviour and status, and the pick-only variant's description + injected prompt now forbid offering an in-project path as a substitute for an out-of-project request.

## Notes / tradeoffs

- The confirmation card draws on the shared 3-questions-per-turn budget. On exhaustion the result is an `error` instructing the model not to use the candidate unconfirmed — deliberately not `resolved`.
- The chip half of ENG-1852 (the first-run "Point it at a folder" promise) is a separate cowork PR — independent, no merge-order constraint.
- The verifier half ("a turn whose only outcome is auto_resolved isn't the task") mostly dissolves with this change, since `auto_resolved` can no longer fire on the first-run path; anything further belongs to ENG-1204.

## Tests

- 10 new tests in `tests/test_select_path.py` covering: no-elicitor → `needs_confirmation`; card confirmed → `resolved` without `auto_resolved`; declined / dismissed → `cancelled` with attach guidance; free-typed decline carries the user's words; no-emitter → `needs_confirmation`; budget exhausted → `error`; path-picker fall-through (1 option) + cancel; pattern single match still auto-resolves even on a choice host; both definitions document `needs_confirmation`.
- **Mutation-verified:** with the handler guard reverted to unconditional auto-resolve, all 9 behavioural tests fail; with the fix, the full suite is green (2423 passed, 31 skipped).

## Security pass

Reviewed the changed surface: candidate paths remain confined to the project root by the untouched `_is_within` guard; the only new data flow is the user's typed card answer being quoted back into the tool result (model-visible text, same trust level as any user message — no shell/log/exec sink); no credentials, no new endpoints, no deserialization.

### DIFF
--- a/anton/core/tools/tool_handlers.py
+++ b/anton/core/tools/tool_handlers.py
@@ -829,6 +829,94 @@ def _status(status: str, message: str = "", **extra) -> str:
     return json.dumps({"status": status, **({"message": message} if message else {}), **extra})
 
 
+def _needs_confirmation(candidate: "Path", label: str) -> str:
+    """The model-supplied candidate cannot be silently accepted and no
+    confirmation card can render on this host: hand the decision back."""
+    return _status(
+        "needs_confirmation",
+        f"You supplied '{label}' as the only candidate — that is your guess, not "
+        "the user's choice, so it was not accepted. Ask the user to confirm it "
+        "before using it, and do not present it as chosen or connected until "
+        "they do. If they asked for a file or folder outside the project, no "
+        "path inside the project is an answer: tell them plainly that this host "
+        "cannot reach paths outside the project, and ask them to attach the "
+        "relevant files to the conversation.",
+        candidates=[str(candidate)],
+    )
+
+
+async def _confirm_single_candidate(
+    session: "ChatSession", candidate: "Path", root: "Path", timeout_s: "int | None"
+) -> str:
+    """Confirm a model-supplied single candidate with the user (ENG-1852).
+
+    A lone entry in ``candidates`` is the model's own guess echoed back, not a
+    discovery: in a fresh project exactly one directory exists (``skills``), so
+    silently resolving it reported folders nobody chose as connected — to the
+    first-run prompt "give you access to a folder on my computer", no less.
+    Confirmation is enforced here rather than in the prompt, because the model
+    already narrates such guesses as decisions ("I'll use it as the folder").
+
+    Renders a yes/no choice card (supported on every cowork host, which has no
+    path picker). Where no card can render either, returns
+    ``needs_confirmation`` so the model must ask in plain text.
+    """
+    from anton.core.interaction.elicit import AskOption, AskRequest, elicit
+
+    try:
+        label = str(candidate.relative_to(root))
+    except ValueError:
+        label = str(candidate)
+    noun = "folder" if candidate.is_dir() else "file"
+    request = AskRequest(
+        prompt=f"Only one candidate was found — use this {noun}?",
+        kind="choice",
+        timeout_s=timeout_s,
+        options=(
+            AskOption(value="yes", label=f"Yes — use {label}", kind=noun, style="primary"),
+            AskOption(value="no", label="No — that's not it"),
+        ),
+    )
+    try:
+        answer = await elicit(session, f"path:{uuid.uuid4().hex}", request)
+    except Exception as exc:  # noqa: BLE001 — the card is host code
+        _log.warning("select_path confirmation elicitor failed: %s", exc, exc_info=True)
+        return _status("error", f"Selection failed: {exc}")
+    if answer.status == "unavailable":
+        return _needs_confirmation(candidate, label)
+    if answer.status == "cancelled":
+        return _status(
+            "cancelled",
+            "The user dismissed the confirmation without choosing. Ask how they "
+            "would like to proceed.",
+        )
+    if answer.status == "limit":
+        return _status(
+            "error",
+            "Too many questions this turn; do not use the candidate without the "
+            "user's confirmation — ask in plain text.",
+        )
+    if answer.status != "answered":
+        return _status("error", f"The confirmation did not return an answer ({answer.status}).")
+    typed = (answer.text or "").strip()
+    if "yes" in answer.values:
+        # Deliberately NOT auto_resolved: the user confirmed this path.
+        return _status(
+            "resolved",
+            f'The user added: "{typed}"' if typed else "",
+            path=str(candidate),
+        )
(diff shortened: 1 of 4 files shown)


alecantu7 (commented):
Self-review (adversarial pass) — three findings, none blocking:

1. **The pattern loophole stays open, deliberately.** A model that wants to skip confirmation can pass `pattern: "skills"` instead of `candidates: ["skills"]` and still get `auto_resolved`. I did not close it because (a) the observed failure mode in all 9 traces is an explicit `candidates` list, (b) a lone pattern match is the tool's legitimate main case (`**/report.csv` finding the one report), and (c) the pick-only definition's prose now explicitly forbids offering in-project paths as substitutes for out-of-project requests. If a post-fix weekly review shows the model migrating to patterns to dodge the card, the follow-up is to require confirmation for single-segment literal patterns (no glob metacharacters) too.

2. **The card renders generic on cowork.** cowork-server's stream formatter serializes options as `{value, label, detail}` only — the `style="primary"` and `kind="folder"` hints on the Yes option are dropped on the wire. Verified this is cosmetic: the card still renders as a standard two-button question with free-text input, and the answer broker keys purely on `question_id`, so no cowork-server change is needed for this PR to work end to end.

3. **A user who types "yes" as free text (instead of clicking) is treated as a decline** carrying their words. Deliberate: parsing typed affirmations ("yes", "si", "yes but a different one") is a heuristic that fails worse than a re-ask costs, and the model sees the verbatim text either way.

Also verified: full suite green (2423 passed), and all 9 behavioural tests fail when the guard is reverted to unconditional auto-resolve (mutation check).

anton/core/tools/tool_handlers.py
```diff
@@ -829,6 +829,94 @@ def _status(status: str, message: str = "", **extra) -> str:
...
+    if answer.status == "limit":
+        return _status(
+            "error",
+            "Too many questions this turn; do not use the candidate without the "
+            "user's confirmation — ask in plain text.",
+        )
+    if answer.status != "answered":
+        return _status("error", f"The confirmation did not return an answer ({answer.status}).")
+    typed = (answer.text or "").strip()
+    if "yes" in answer.values:
```
pnewsam, line 902: **Nit:** a user who *types* the literal word "yes" here (rather than clicking the Yes button) is read as a decline. `"yes" in answer.values` inspects only structured option values; a free-typed answer arrives as `text` with `values=()`, so it takes the `cancelled`/"declined" branch — the model is then told "The user declined 'skills'" while also receiving `The user said: "yes"`.

This is the deliberate "free text = decline carrying intent" design (and the valuable case — "I meant /some/other/path" — works correctly), so the literal-"yes" collision is a narrow subset, only reachable if the cowork card renders a text input for this 2-option prompt (`allow_custom` defaults to `True`). Low severity — worth a follow-up thought, not a blocker.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

Reviewed the handler control flow, the `elicit`/`AskAnswer` contract it depends on, candidate collection and boundary guards, the tool-def copy, and the tests. The core property holds: no model-supplied single candidate reaches `resolved` without an explicit user action, while a genuine lone-pattern discovery still auto-resolves. One non-blocking nit inline.

### Validation
- Traced every path to `resolved` for `len(candidates)==1`: picker host → 1-option picker (pick = confirm); card host → yes/no card (Yes = confirm); no card → `needs_confirmation`; decline/dismiss → `cancelled`; budget/timeout/other → `error`. Auto-resolve fires only on `not has_candidates` (pattern) — and `has_candidates` uses the same `isinstance(list) and bool(...)` predicate as `_collect_selection_candidates`'s explicit-vs-pattern branch, so it provably means "from a glob."
- `AskAnswer.values` defaults to `()`, so `"yes" in answer.values` can't throw on a free-typed answer; a `timeout` status falls into the non-`answered` → `error` branch, never `resolved`.
- `consider()` still gates on `_is_within(root, …)` **and** `.exists()` **and** kind, so any candidate reaching the confirmation card is real, in-project, and correctly typed — the traversal guard is untouched as claimed.
- `git grep`: no Python code branches on the select_path status set, so adding `needs_confirmation`/`cancelled` needs no downstream handler change.
- Ran `tests/test_select_path.py` at head `8fa2fa10` in an isolated worktree: 36 passed. (I did not independently re-run the mutation check the PR describes.)

anton/core/tools/tool_handlers.py
alecantu7: Addressed in 2dad71b3 — you're right that the branch asserted a decline the user never made. Rather than parsing typed affirmations (still a worse heuristic than one re-ask: "yes but a different one", other languages), the text-only branch now relays the reply verbatim with neutral wording: the candidate is *not confirmed*, act on the reply, and re-run select_path for an explicit confirmation if it reads as agreement. An explicit **No** click keeps the "declined" wording. Regression test updated to assert "declined" is absent for typed replies (`test_single_explicit_candidate_typed_reply_is_relayed_not_asserted_as_decline`); suite green at the new head.

PR #394, opened 2026-08-25
Let Workspace's artifacts root be overridden by the caller

torrmal:
## Summary

Part of the fix for [ENG-1933](https://linear.app/mindsdb/issue/ENG-1933) — artifact cards leaking across concurrent sibling conversations in the same project on desktop.

`Workspace` always derived `artifacts_dir` from a fresh `AntonSettings()`, ignoring any conversation-scoped root the embedding host process might want to use instead. This is the hook `cowork-server` needs so each desktop conversation can get its own artifacts folder (matching how org mode already isolates a pod's artifacts per conversation), instead of every conversation in a project sharing one folder.

- `Workspace(base, artifacts_dir=...)` — new optional param, defaults to the existing behavior when omitted.
- `anton.md`/`.env`/memory are unaffected — they still resolve under `base`.
- `tool_handlers.py`'s `_artifact_store()` already reads `workspace.artifacts_dir`, so no tool-handler changes are needed.

## Test plan

- [x] `uv run pytest tests/test_workspace.py` — 38 passed (3 new: override sets `artifacts_dir`, override is created by `initialize()` without also creating the default path, override doesn't affect `anton.md`).

### DIFF
--- a/anton/workspace.py
+++ b/anton/workspace.py
@@ -27,15 +27,21 @@ Created: {date}
 class Workspace:
     """Manages the .anton/ workspace directory and its files."""
 
-    def __init__(self, base: Path) -> None:
+    def __init__(self, base: Path, artifacts_dir: Path | None = None) -> None:
         self._base = base
         self._anton_dir = base / ".anton"
         self._anton_md = self._anton_dir / "anton.md"
         self._env_file = self._anton_dir / ".env"
         self._anton_md_last_read: datetime | None = None
 
-        settings = AntonSettings()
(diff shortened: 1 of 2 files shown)


torrmal:
Superseded — see mindsdb/cowork-server#384's closing comment. The `Workspace` artifacts-root override existed to let cowork-server point each conversation at its own artifacts folder; the replacement approach keeps one shared project-wide folder and stamps attribution at tool-call time instead, so this override is no longer needed. A different anton change (stamping `record_turn()` in the artifact tool handlers) will land in its place.

PR #395, opened 2026-08-25
[ENG-1893] Agent stops mid session

tino097:
https://linear.app/mindsdb/issue/ENG-1893/agent-stops-mid-task-near-the-per-turn-token-budget-instead-of

* agent no longer stops to ask permission just because it hit a round or token cap
* grants a small one-time extension when the completion verifier says the remaining work is close to done
* round cap and mid loop spend ceiling get the same one time extension, unconditionally, since no verdict exists yet at that point
* extension is one-shot per turn, the hard ceiling is unchanged, genuine stuck or uncertain cases still ask
* updated the round cap and spend ceiling tests whose counts were calibrated to the old always ask behavior

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -14,7 +14,7 @@ import sys
 from typing import TYPE_CHECKING, List, Literal
 import os
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from anton.core.backends.base import Cell, ScratchpadRuntimeFactory
 from anton.core.backends.local import local_scratchpad_runtime_factory
@@ -363,6 +363,13 @@ class _VerifierVerdict(BaseModel):
             "than assuming it's almost done."
         ),
     )
+
+    @field_validator("close_to_done", mode="before")
+    @classmethod
+    def _null_falls_back_to_default(cls, v: object) -> object:
+        # An explicit `null` here otherwise fails validation for the WHOLE
+        # verdict, not just this field — fall back to the field default.
+        return False if v is None else v
 
 
 # Output budgets for the verdict call: first attempt, then the retry used when
@@ -878,14 +885,15 @@ def _role_calls(tc: TurnCost, role: str) -> str:
 
 
 def _record_grace(turn_cost: TurnCost | None, tag: str) -> None:
-    """Append `tag` ("round" or "ceiling") to `turn_cost.grace_granted`,
-    comma-joined and deduplicated."""
+    """Add `tag` ("round" or "ceiling") to `turn_cost.grace_granted`,
+    comma-joined, deduplicated, and sorted — alphabetical rather than
+    firing order, so the two-grace value is one fixed string either way."""
     if turn_cost is None:
         return
     tags = turn_cost.grace_granted.split(",") if turn_cost.grace_granted else []
     if tag not in tags:
         tags.append(tag)
-        turn_cost.grace_granted = ",".join(tags)
+        turn_cost.grace_granted = ",".join(sorted(tags))
 
 
 def _build_verify_request(
@@ -3132,9 +3140,12 @@ class ChatSession:
 
         `_spend_ceiling_grace_tokens` adds on top of that bound, not inside it:
         the reserve term above cancels out of the gate-plus-reserve landing
-        point, so a granted grace overshoots the ceiling by (approximately)
-        its own size, once per turn. See `_grant_spend_ceiling_grace` for why
-        that term is kept well under the reserve's own size.
+        point, so a granted grace overshoots the ceiling by roughly one more
+        call, once per turn. Known gap: the grace is capped at
+        `max_turn_tokens // 4`, so a single call costing more than that share
+        overshoots past this bound — see `test_total_is_bounded_by_the_ceiling`.
+        See `_grant_spend_ceiling_grace` for why the grace stays well under
+        the reserve's own size otherwise.
         """
         peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
         reserve = min(
@@ -3658,18 +3669,22 @@ class ChatSession:
         resilience_nudged: set[str] = set()
         # Mirrors turn_stream's one-time grants, the same way the round-cap
         # and spend-ceiling checks below already mirror it. This path has no
-        # continuation loop, so each grant is simply spent once for the
-        # whole call rather than once per continuation.
-        _round_cap_grace_active = False
-
+        # continuation loop — unlike `_stream_and_handle_tools`, which needs
+        # its own per-continuation flag — so `_round_cap_grace_used` alone
+        # already tracks whether THIS call earned the wider cap.
         while response.tool_calls:
             tool_round += 1
-            if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
+            # `> 0` excludes a configured 0 (or misconfigured negative) cap:
+            # that means "no tool rounds", not "one round of free grace".
+            if (
+                tool_round > self._max_tool_rounds
+                and not self._round_cap_grace_used
+                and self._max_tool_rounds > 0
+            ):
                 self._round_cap_grace_used = True
-                _round_cap_grace_active = True
                 _record_grace(self._turn_cost, "round")
             effective_round_cap = self._max_tool_rounds + (
-                _ROUND_CAP_GRACE_ROUNDS if _round_cap_grace_active else 0
+                _ROUND_CAP_GRACE_ROUNDS if self._round_cap_grace_used else 0
             )
             if self._turn_cost is not None and tool_round <= effective_round_cap:
                 self._turn_cost.rounds += 1
@@ -4046,6 +4061,11 @@ class ChatSession:
         assistant_text_parts: list[str] = []
         _max_auto_retries = 2
         _retry_count = 0
+        # One-time cap grace, reset once per turn_stream() call rather than in
+        # `_stream_and_handle_tools`, which retries re-enter and would re-arm.
+        self._spend_ceiling_grace_used = False
+        self._spend_ceiling_grace_tokens = 0
+        self._round_cap_grace_used = False
         # ENG-673: a mid-stream provider failure that had NO prior retry (an
         # overload smuggled into an HTTP-200 stream) gets budget-bounded
         # backoff-and-retry — separate from the instant, count-bounded recovery
@@ -4583,10 +4603,6 @@ class ChatSession:
         system = await self._build_system_prompt(user_message)
         self._compacted_this_turn = False
         self._compaction_failed_this_turn = False
-        # One-time cap grace (ENG-1893), reset per turn like the flags above.
-        self._spend_ceiling_grace_used = False
-        self._spend_ceiling_grace_tokens = 0
-        self._round_cap_grace_used = False
 
         response: StreamComplete | None = None
 
@@ -4674,8 +4690,14 @@ class ChatSession:
                 # mid-loop spend-ceiling grace — no verdict exists yet at
                 # this point, so the grant is a single small window rather
                 # than a self-report. Decided before the accounting below so
-                # the granted round is counted like any other.
-                if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
+                # the granted round is counted like any other. `> 0` excludes
+                # a configured 0 (or negative) cap — "no tool rounds", not
+                # "one round of free grace".
+                if (
+                    tool_round > self._max_tool_rounds
+                    and not self._round_cap_grace_used
+                    and self._max_tool_rounds > 0
+                ):
                     self._round_cap_grace_used = True
                     _round_cap_grace_active = True
                     _record_grace(self._turn_cost, "round")
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -449,6 +449,13 @@ _VERIFIER_LATCH_REPROBE_TURNS = 10
 # call has reported usage.
 _SPEND_CEILING_RESERVE = 200_000
 
+# Floor for the one-time spend-ceiling grace — deliberately lower
+# than `_SPEND_CEILING_RESERVE`'s own floor. That reserve is sized to cover
+# TWO calls guaranteed to land after the gate trips; the grace is sized for
+# roughly ONE more, so granting it does not cancel the reserve outright and
+# put the turn back up near the ceiling. See `_grant_spend_ceiling_grace`.
+_SPEND_CEILING_GRACE_FLOOR = 100_000
+
 # One-time round-cap extension (ENG-1893), granted the first time a turn
 # breaches max_tool_rounds. Sized to the "2-3 more tool calls" this exists
 # for — add a closing step, save, verify — not a way to talk past the cap
@@ -847,6 +854,17 @@ def _role_tokens(tc: TurnCost, role: str) -> str:
 def _role_calls(tc: TurnCost, role: str) -> str:
     slice_ = tc.by_role.get(role)
     return str(slice_.calls if slice_ else 0)
+
+
+def _record_grace(turn_cost: TurnCost | None, tag: str) -> None:
+    """Append `tag` ("round" or "ceiling") to `turn_cost.grace_granted`,
+    comma-joined and deduplicated."""
+    if turn_cost is None:
+        return
+    tags = turn_cost.grace_granted.split(",") if turn_cost.grace_granted else []
+    if tag not in tags:
+        tags.append(tag)
+        turn_cost.grace_granted = ",".join(tags)
 
 
 def _build_verify_request(
@@ -2676,12 +2694,14 @@ class ChatSession:
             _root_cause_fields = {}
         logger.info(
             "turn_cost session=%s turn=%d ended_by=%s verification_skipped=%s "
+            "grace_granted=%s grace_tokens=%d "
             "tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
             "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d "
             "by_role=%s %s",
             self._session_id, turn_index, tc.ended_by,
-            str(tc.verification_skipped).lower(), tc.total_tokens,
+            str(tc.verification_skipped).lower(),
+            tc.grace_granted or "-", tc.grace_tokens, tc.total_tokens,
             tc.input_tokens, tc.output_tokens, tc.cache_read_tokens,
             tc.cache_creation_tokens, tc.llm_calls, tc.rounds,
             tc.continuations, tc.peak_context_tokens, tc.duration_ms,
@@ -2795,6 +2815,11 @@ class ChatSession:
                 # ENG-1632) must be excludable from the honest-stop
                 # denominator without a per-conversation Langfuse hop.
                 verification_skipped=str(tc.verification_skipped).lower(),
+                # One-time grace(s) granted this turn — see
+                # `TurnCost.grace_granted`. Empty string, not "-", to stay
+                # consistent with how other unset string properties go out.
+                grace_granted=tc.grace_granted,
+                grace_tokens=str(tc.grace_tokens),
                 tokens_total=str(tc.total_tokens),
                 input_tokens=str(tc.input_tokens),
                 output_tokens=str(tc.output_tokens),
@@ -3012,6 +3037,12 @@ class ChatSession:
         against `total_tokens`, which also counts output, so a turn can land
         over the ceiling by roughly two output budgets. See
         `test_total_is_bounded_by_the_ceiling` for the asserted form.
+
+        `_spend_ceiling_grace_tokens` adds on top of that bound, not inside it:
+        the reserve term above cancels out of the gate-plus-reserve landing
+        point, so a granted grace overshoots the ceiling by (approximately)
+        its own size, once per turn. See `_grant_spend_ceiling_grace` for why
+        that term is kept well under the reserve's own size.
         """
         peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
         reserve = min(
@@ -3023,20 +3054,30 @@ class ChatSession:
         )
 
     def _grant_spend_ceiling_grace(self) -> None:
-        """Raise the gate once this turn (ENG-1893), sized like one more call.
+        """Raise the gate once this turn, sized like ONE more call.
 
-        Reuses the reserve's own formula rather than a flat constant, for the
-        same reason the reserve does: a turn with a big context needs a bigger
-        exception to actually cover 1-3 more calls. Idempotent — a caller that
-        checks `_spend_ceiling_grace_used` first never calls this twice, but a
-        stray second call would just recompute the same amount, not stack it.
+        Deliberately smaller than the reserve `_spend_ceiling_gate` nets this
+        against: that reserve is sized for the TWO calls guaranteed to land
+        after the gate trips (the crossing call, the hand-back). Reusing that
+        same size for the grace cancelled the reserve outright and let a
+        granted turn land back up near the ceiling — measured up to 1.5x the
+        configured ceiling. Scaled by `peak_context_tokens` for the same
+        reason the reserve is (a big-context turn's next call costs more),
+        but at 1x peak and a lower floor, not 2x and the
+        reserve's own floor, so the grant stays closer to "a bit more room"
+        than "the reserve, again". Idempotent — a caller that checks
+        `_spend_ceiling_grace_used` first never calls this twice, but a stray
+        second call would just recompute the same amount, not stack it.
         """
         peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
         self._spend_ceiling_grace_tokens = min(
-            max(_SPEND_CEILING_RESERVE, 2 * peak),
-            max(self._max_turn_tokens // 2, 1),
+            max(_SPEND_CEILING_GRACE_FLOOR, peak),
+            max(self._max_turn_tokens // 4, 1),
         )
         self._spend_ceiling_grace_used = True
+        if self._turn_cost is not None:
+            self._turn_cost.grace_tokens = self._spend_ceiling_grace_tokens
+        _record_grace(self._turn_cost, "ceiling")
 
     def _spend_ceiling_notice(self) -> str:
         """The SYSTEM injection for a ceiling stop.
@@ -3496,6 +3537,12 @@ class ChatSession:
         system = await self._build_system_prompt(user_msg_str)
         self._compacted_this_turn = False
         self._compaction_failed_this_turn = False
+        # Same reset turn_stream does — without it a grace granted in an
+        # earlier turn_stream call leaks into every later turn() call on
+        # this session, permanently widening its round cap / spend ceiling.
+        self._spend_ceiling_grace_used = False
+        self._spend_ceiling_grace_tokens = 0
+        self._round_cap_grace_used = False
 
         response = await self.plan_with_recovery(system=system, tools=tools)
 
@@ -3517,12 +3564,24 @@ class ChatSession:
         tool_round = 0
         error_streak: dict[str, int] = {}
         resilience_nudged: set[str] = set()
+        # Mirrors turn_stream's one-time grants, the same way the round-cap
+        # and spend-ceiling checks below already mirror it. This path has no
+        # continuation loop, so each grant is simply spent once for the
+        # whole call rather than once per continuation.
+        _round_cap_grace_active = False
 
         while response.tool_calls:
             tool_round += 1
-            if self._turn_cost is not None and tool_round <= self._max_tool_rounds:
+            if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
+                self._round_cap_grace_used = True
+                _round_cap_grace_active = True
+                _record_grace(self._turn_cost, "round")
+            effective_round_cap = self._max_tool_rounds + (
+                _ROUND_CAP_GRACE_ROUNDS if _round_cap_grace_active else 0
+            )
+            if self._turn_cost is not None and tool_round <= effective_round_cap:
                 self._turn_cost.rounds += 1
-            if tool_round > self._max_tool_rounds:
+            if tool_round > effective_round_cap:
                 # Mirror turn_stream's cap mark (#309 review) — this path is
                 # public API even without in-repo callers, and the books are
                 # wired here too.
@@ -3535,7 +3594,7 @@ class ChatSession:
                     {
                         "role": "user",
                         "content": (
-                            f"SYSTEM: You have used {self._max_tool_rounds} tool-call rounds on this turn. "
+                            f"SYSTEM: You have used {effective_round_cap} tool-call rounds on this turn. "
                             "Pause here. Summarize what you have accomplished so far and what remains. "
                             "If you believe you are on a good track and can finish the task with more steps, "
                             "tell the user and ask if they'd like you to continue. "
@@ -3558,22 +3617,28 @@ class ChatSession:
             # all, so this path is exactly where a tiny ceiling would break the
             # loop on round 1 having run nothing (#344 review).
             if self._spend_ceiling_stops_the_tool_loop():
-                if self._turn_cost is not None:
-                    self._turn_cost.ended_by = "spend_ceiling"
-                logger.info(
-                    "spend ceiling reached (turn): tokens=%d ceiling=%d "
-                    "tool_round=%d — pausing to ask the user",
-                    self._turn_cost.total_tokens if self._turn_cost else 0,
-                    self._max_turn_tokens, tool_round,
-                )
-                self._append_history(
-                    {"role": "assistant", "content": response.content or ""}
-                )
-                self._append_history(
-                    {"role": "user", "content": self._spend_ceiling_notice()}
-                )
-                response = await self.plan_with_recovery(system=system)
-                break
+                # One-time, unconditional extension, mirroring turn_stream's
+                # mid-loop grant — no verdict exists on this verifier-less
+                # path either, so the grant is unconditional here too.
+                if not self._spend_ceiling_grace_used:
+                    self._grant_spend_ceiling_grace()
+                if self._spend_ceiling_stops_the_tool_loop():
+                    if self._turn_cost is not None:
+                        self._turn_cost.ended_by = "spend_ceiling"
+                    logger.info(
+                        "spend ceiling reached (turn): tokens=%d ceiling=%d "
+                        "tool_round=%d — pausing to ask the user",
+                        self._turn_cost.total_tokens if self._turn_cost else 0,
+                        self._max_turn_tokens, tool_round,
+                    )
+                    self._append_history(
+                        {"role": "assistant", "content": response.content or ""}
+                    )
+                    self._append_history(
+                        {"role": "user", "content": self._spend_ceiling_notice()}
+                    )
+                    response = await self.plan_with_recovery(system=system)
+                    break
 
             # Build assistant message with content blocks
             assistant_content: list[dict] = []
@@ -4499,6 +4564,13 @@ class ChatSession:
             tool_round = 0
             error_streak: dict[str, int] = {}
             resilience_nudged: set[str] = set()
+            # Per-continuation, unlike `_round_cap_grace_used`: only the
+            # continuation that actually earns the grant (the first ever
+            # breach this turn) gets the wider cap. Without this,
+            # `_round_cap_grace_used` staying True for the rest of the turn
+            # let EVERY later continuation start pre-extended to M+3, turning
+            # a one-shot +3 into +3 per continuation.
+            _round_cap_grace_active = False
 
             while llm_response.tool_calls:
                 tool_round += 1
@@ -4509,8 +4581,10 @@ class ChatSession:
                 # the granted round is counted like any other.
                 if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
                     self._round_cap_grace_used = True
+                    _round_cap_grace_active = True
+                    _record_grace(self._turn_cost, "round")
                 effective_round_cap = self._max_tool_rounds + (
-                    _ROUND_CAP_GRACE_ROUNDS if self._round_cap_grace_used else 0
+                    _ROUND_CAP_GRACE_ROUNDS if _round_cap_grace_active else 0
                 )
                 # Accumulate, don't assign: `tool_round` is loop-local and
                 # resets to 0 on every verifier-forced continuation, so
@@ -4527,7 +4601,7 @@ class ChatSession:
                         self._turn_cost.ended_by = "round_cap"
                     self._acc_observe(
                         "cap_exhausted",
-                        {"cap": self._max_tool_rounds},
+                        {"cap": effective_round_cap, "configured_cap": self._max_tool_rounds},
                         severity=9,
                         round_idx=tool_round,
                     )
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -351,6 +351,17 @@ class _VerifierVerdict(BaseModel):
         )
     )
     reason: str = Field(description="One brief sentence explaining the verdict.")
+    close_to_done: bool = Field(
+        default=False,
+        description=(
+            "Only meaningful when status is INCOMPLETE: true if the "
+            "remaining work is small — roughly 1-3 more tool calls — with "
+            "nothing blocking or uncertain. False for anything that needs "
+            "substantially more work, or where you aren't sure. Defaults to "
+            "false, so an unsure model errs toward asking the user rather "
+            "than assuming it's almost done."
+        ),
+    )
 
 
 # Output budgets for the verdict call: first attempt, then the retry used when
@@ -437,6 +448,12 @@ _VERIFIER_LATCH_REPROBE_TURNS = 10
 # will cost, and ENG-1288 already tracks it. This floor only applies before any
 # call has reported usage.
 _SPEND_CEILING_RESERVE = 200_000
+
+# One-time round-cap extension (ENG-1893), granted the first time a turn
+# breaches max_tool_rounds. Sized to the "2-3 more tool calls" this exists
+# for — add a closing step, save, verify — not a way to talk past the cap
+# repeatedly: see the grant sites for the one-shot guard.
+_ROUND_CAP_GRACE_ROUNDS = 3
 
 # Appended to the verifier system prompt. Shortens the preamble on narrating
 # models but is not sufficient alone (0/3 at 256 with it), so it pairs with the
@@ -1039,6 +1056,13 @@ class ChatSession:
         # older settings object doesn't break — absent means "no ceiling",
         # matching the pre-ENG-1286 behaviour rather than silently applying one.
         self._max_turn_tokens = getattr(s, "max_turn_tokens", 0)
+        # One-time exception to the always-ask ceiling (ENG-1893): granted at
+        # most once per turn, and only when the verifier reports the
+        # remaining work as small. Reset per turn in turn_stream, not here —
+        # this is __init__ so a long CLI session gets it too.
+        self._spend_ceiling_grace_used = False
+        self._spend_ceiling_grace_tokens = 0
+        self._round_cap_grace_used = False
         # Tally of classified tool failures (ENG-1492). MEASUREMENT ONLY —
         # nothing reads this to change behaviour. The control that will
         # (ENG-1531) is deliberately unbuilt until this has reported real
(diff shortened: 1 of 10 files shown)


alecantu7 (commented):
## Adversarial review — ENG-1893

Reviewed at head `7714dc8f` against fresh `origin/staging`. Deep pass: findings were generated per failure-dimension and then each was handed to an independent agent prompted to **refute** it — three were killed that way and are listed at the bottom so nobody re-litigates them. Everything below survived, and the two headline items I re-verified myself with executed repros.

**Verdict: request changes on two items (#1, #2).** The idea is right and I measured it working — in the near-done band, `staging` asks the user on **3/3** consecutive turns and this branch asks on **0/3**. What needs another pass is the *sizing* and the *description*: two of the four claims in the PR body are false, and the one that's most wrong understates the round-cap effect by ~6x.

---

### 1. 🔴 The round-cap grace is not one-shot per turn — it permanently raises the cap for the rest of the turn

`anton/core/session.py:4510`. The flag is reset at `:4425`, at the top of `_stream_and_handle_tools` — **outside** the continuation loop. `tool_round` resets at `:4499`, **inside** it.

```python
if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
    self._round_cap_grace_used = True
effective_round_cap = self._max_tool_rounds + (
    _ROUND_CAP_GRACE_ROUNDS if self._round_cap_grace_used else 0
)
```

So once any continuation breaches the cap, every subsequent continuation starts already extended to `M+3`. More consequentially: on `staging` the first breach sets `_max_rounds_hit` and **ends the turn** (the break past verification at `:5169`); with grace it no longer does, so the turn survives into continuations it would previously never have reached.

Measured — same scripted model, `M=5`, `max_continuations=3`, PR vs. the same code with `_ROUND_CAP_GRACE_ROUNDS` patched to 0:

```
tool calls per continuation = 6    BASE:  5 rounds / round_cap  ->  PR: 24 rounds / handback_budget   (+19, 4.8x)
tool calls per continuation = 8    BASE:  5 rounds / round_cap  ->  PR: 32 rounds / handback_budget   (+27, 6.4x)
tool calls per continuation = 12   BASE:  5 rounds / round_cap  ->  PR:  8 rounds / round_cap         (+3,  1.6x)
```

Only the last row matches the advertised `+3`. This contradicts the PR body ("extension is one-shot per turn") and the constant's own comment at `:449-453` — *"not a way to talk past the cap repeatedly: see the grant sites for the one-shot guard."*

None of the three tests in `tests/test_round_cap_grace.py` drive a verifier continuation, so this path is uncovered.

**Fix:** move the reset inside the `while True:` continuation loop at `:4490` if "+3 once" is the intent; or keep it turn-wide and correct both the PR body and the `:449-453` comment to the real bound (`max_continuations x grace`, plus the survivability change). Either way it should be a deliberate choice rather than a consequence of where the reset landed.

**Test:** drive `_max_continuations` continuations with `close_to_done=False` INCOMPLETE verdicts and `per_continuation` in `(M, M+3]`; assert total `rounds <= M + _ROUND_CAP_GRACE_ROUNDS`. Fails today (32 vs 8).

---

### 2. 🔴 "the hard ceiling is unchanged" is not true — the grace exactly cancels the reserve, on every ceiling-crossing turn

`anton/core/session.py:3020-3039`. `_grant_spend_ceiling_grace` uses the **identical formula** as the reserve it's subtracted against:

```python
return max(self._max_turn_tokens + self._spend_ceiling_grace_tokens - reserve, 1)
```

The reserve exists to pre-pay the two calls still in flight when the gate trips (your own docstring at `:2991-2998`). Cancelling it moves the landing point up by exactly the reserve. Measured driving the real `turn_stream` through this file's own fixtures:

| ceiling | peak ctx | base lands | this branch lands |
|---|---|---|---|
| 1,000,000 | 75,000 | 900,000 (90%) | 1,100,000 (**110%**) |
| 1,250,000 | 75,000 | 1,200,000 (96%) | 1,400,000 (**112%**) |
| 1,250,000 | 190,001 | 1,266,670 (101%) | 1,520,004 (**122%**) |
| 1,250,000 | 337,500 | 1,350,000 (108%) | 1,800,000 (**144%**) |

1,250,000 is the real default in both `anton/core/settings.py:52` and cowork-server's `default_max_turn_tokens`. Analytic bound: `M + min(max(200k, 2*peak), M//2)` — up to **1.5x M**. All four rows report `ended_by=spend_ceiling`, so the ceiling was the binding constraint, not the round cap.

Two aggravators:
- The mid-loop grant at `:4573-4574` is **unconditional** — no `close_to_done`, no verdict. So this applies to every long turn, not to the close-to-done ones the PR body describes.
- The `_spend_ceiling_gate` docstring at `:3013-3015` still says the overshoot is "roughly two output budgets". Left unedited, so the next reader inherits the old bound.

I tried three ways to kill this and couldn't: the overshoot isn't just the pre-existing in-flight calls (those are priced *below* M by the reserve, and the measured delta equals the reserve exactly); the round cap doesn't bind first; and it can't run away unbounded (`peak_context_tokens` is monotonic so the gate is clamped at M — which is what keeps it at 1.5x, but doesn't rescue the claim).

**Fix:** state the real bound (PR body + the `:3013-3015` docstring + the ticket), **or** size the grace to one call's context rather than the two-call reserve so the turn still lands at/under M, **or** drop the unconditional mid-loop grant and keep only the verdict-gated one.

---

### 3. 🟠 The one test pinning the money guarantee was relaxed, and the relaxation is only correct at the fixture's small context

`tests/test_spend_ceiling.py:149-166` — `test_total_is_bounded_by_the_ceiling` widened from `<= CEILING + slack` to `<= CEILING + slack + _SPEND_CEILING_RESERVE`. The relaxation is honestly documented, but it's a **flat 200k** while the actual grace is `min(max(200k, 2*peak), M//2)`. At the 190k contexts this codebase cites in its own comments the grace is 380k and the new bound is already violated — it only holds because the fixture's `PER_CALL` is small.

**Fix:** parametrise at production-shaped numbers and assert against the real formula, not the flat constant.

---

### 4. 🟠 Nothing records that a grace was granted, so the fix's own effect and cost are unmeasurable

Both grant sites are silent, `TurnCost` is untouched, and `_grant_spend_ceiling_grace` logs nothing. A turn that used its grace and finished is byte-identical in `turn_completed` to one that never neared a limit — and it isn't derivable post-hoc either, since the event carries `rounds` but not `max_tool_rounds`, `tokens_total` but not `max_turn_tokens`, and both are per-org settings with wide ranges.

The observable consequence is that `ended_by=round_cap` and `ended_by=spend_ceiling` counts go **down** — which the harness ledger will read as an unambiguous win — while the same change raises spend on every ceiling-hitting turn. This is exactly the failure mode `TurnCost.verification_skipped` was added for, and `turn_cost.py:117-121` says so verbatim: *"WITHOUT this flag they are byte-identical in analytics to turns that were verified and passed."*

**Fix:** one field — `grace_granted: str = ""` (`""` / `"round"` / `"ceiling"` / `"round,ceiling"`) — stamped at both grant sites and emitted next to `ended_by`. Ideally `grace_tokens` alongside it, so the size of the concession is queryable.

---

### 5. 🟠 `close_to_done` is an un-evaluated model self-report authorising spend, from a view blind to the ticket's own task shape

`anton/core/session.py:354`. The verifier's entire input is `_render_verify_transcript` with `max_convo=10, max_tool=12, tool_cap=400` — last 10 turns, last 12 tool events, results clipped to **400 characters**, no filesystem access. The failure class that creates is documented in this same file at `:729-751` (ENG-1633). ENG-1893's repro is a **slide-deck build** — a file-producing task whose remaining work is precisely what that view cannot see.

`verifier-eval.yml` triggers on `session.py`, so this PR does run the eval — but `tests/test_verifier_verdict_live.py:749-769` asserts only `[v.status for v in verdicts]`. `close_to_done` appears nowhere in it, including on the single INCOMPLETE fixture where the field is the only thing that's meaningful. And `tests/test_verifier_verdict_schema.py` — which exists specifically because this repo learned that verdict-field *wording* is the behavioural lever — pins nothing for the new field.

If a model never sets the flag it silently defaults to `False`, the judgment-gated half degrades to a no-op, the suite stays green, and `ended_by="spend_ceiling"` looks identical either way.

**Fix:** score `close_to_done` in the live eval on `_PARTIAL` plus one deliberately-far-from-done fixture across both models in `_MODELS`; add a wording pin in `test_verifier_verdict_schema.py` in the style of the existing three; log the observed rate so production can answer "does any model actually set this".

---

### 6. 🟡 The grace never reaches the non-streaming `turn()` loop, breaking a review-enforced mirror

`anton/core/session.py:3525` / `:3560` still use the raw `self._max_tool_rounds` and the un-graced `_spend_ceiling_stops_the_tool_loop()`. The comments at `:3526-3528` and `:3549-3559` explicitly encode a mirror requirement from the **#309 and #344 reviews** — they now claim a mirror that no longer exists.

I enumerated callers (anton itself, the CLI, `cloud_turn`, and greps against `origin/staging` in cowork-server / cowork / scratchpad-controller): **no production callers outside tests**, so no user impact today. The cost is stale comments and convention drift.

**Fix:** mirror the two graces, or retire the mirror explicitly in those comments and say why it's streaming-only.

---

### 7. 🟡 `_spend_ceiling_grace_tokens` leaks out of a streamed turn into a later `turn()` on the same session

`turn()` resets `_compacted_this_turn` / `_compaction_failed_this_turn` at `:3496` but not the three grace fields. A grace granted in `turn_stream` leaves `_spend_ceiling_grace_tokens` set, permanently inflating `_spend_ceiling_gate()` for every later `turn()` on that session. Same reachability caveat as #6.

**Fix:** reset the three fields where `turn()` opens the books (`:3466`).

---

### 8. 🟡 `cap_exhausted` records the raw cap for a break that used the effective one

`anton/core/session.py:4530` — `{"cap": self._max_tool_rounds}` sits next to `round_idx=tool_round`, which *is* the effective number. Every other value at this terminal was moved to `effective_round_cap` (the break at `:4524`, the accounting at `:4522`, the SYSTEM text at `:4541`). Produces cortex records reading `cap=25, round_idx=29`.

It doesn't reach analytics (`RootCauseLedger.event_fields()` emits only tier counters), so this is forensic only — but the cortex record is what persisted Lessons get derived from.

**Fix:** `{"cap": effective_round_cap, "configured_cap": self._max_tool_rounds}`.

---

## Things I checked that came back clean

Worth stating explicitly, because they're the objections a reader raises first — all three were investigated with executed counter-evidence and **do not** hold:

- **`close_to_done` is not dead code.** It's shadowed in the long-tool-loop shape (where the verifier isn't invoked at all — `verdict_calls=0`), but at this file's realistic `per_call=100_000` it fires from `:5530` with 4, 5 and 6 tool rounds and flips `ended_by` from `spend_ceiling` to `completed`. Mutating the branch to `if False and ...` kills exactly its two tests and no others — it's a mutation-verified guard.
- **The PR does fix the reported symptom.** On md5-verified isolated copies of base and head, sweeping distance-to-done across 3 consecutive turns: at `n_tools = 7` and `8`, base asks on 3/3 turns and this branch asks on 0/3, with `ended_by` flipping `spend_ceiling` -> `completed`. An earlier "no change" result came from a harness scripting a model that never finishes — which is the genuine-stuck case this PR deliberately keeps asking about.
- **The e2e edits are honest recalibrations, not loosened assertions.** `test_loop_safety` and `test_circuit_breaker_evasion` still assert the backstop fires and still assert the message text, now derived from `_EFFECTIVE_CAP`. Not vacuous.

Security pass over the diff: no new logging (the two `logger.info` blocks are pre-existing, re-indented into the new nested `if`, format args unchanged and all `%d`-bound ints); the only new interpolation is `effective_round_cap`, an int derived from a bounded setting plus a module constant; no secrets, paths, tool arguments or conversation content added to any sink; no new endpoint, authz surface, or deserialization change beyond one `bool` with `default=False`. One bounded note, not a finding: `close_to_done` is a model-asserted boolean that unlocks paid allowance, and the verifier sees tool output that can be attacker-controlled — but it's capped, one-shot, and the mid-loop path grants more unconditionally anyway, so it isn't an escalation.

## Two open questions, not blocking

- **Grace flags also reset on the in-turn error auto-retry.** `_stream_and_handle_tools` is re-entered from `:4151` (`_max_auto_retries = 2`), re-running the `:4425` reset. Spend impact is bounded — grace is assigned, not accumulated, so the gate never exceeds M — and the round impact is subsumed because `tool_round` resets on re-entry anyway. It does mean "one-shot per turn" is inaccurate here too.
- **The SYSTEM notices are never stripped from the planning history** (pre-existing — `:765` is the only `SYSTEM:` filter and it lives in the verifier transcript builder, not the planning path). Measured 3 ceiling notices still in history after 3 turns. That may be the actual bug behind the ticket: *"rather than risk running long **again**"* only makes sense if the model can see a previous interruption, which would explain asks that fire before any gate trips. Not this PR's to fix, but worth a look before we conclude ENG-1893 is closed — ideally against a real Langfuse trace from the reporter, which the ticket doesn't have yet.

anton/core/session.py
```diff
@@ -4380,6 +4487,10 @@ async def _stream_and_handle_tools(
         system = await self._build_system_prompt(user_message)
         self._compacted_this_turn = False
         self._compaction_failed_this_turn = False
+        # One-time cap grace (ENG-1893), reset per turn like the flags above.
```
tino097: **Finding 1 (critical):** The one-time round-cap and spend-ceiling grace flags are reset at the top of `_stream_and_handle_tools` (this line), but `_turn_stream_inner`'s retry loop (`while not routed_direct:` at line 4049) re-invokes `_stream_and_handle_tools` on a caught retryable error (`continue` at lines 4152 and 4218), re-arming both one-time grants against the same persistent `self._turn_cost`.

Empirically reproduced: with max_tool_rounds=5 (intended effective cap 8), one injected TransientProviderError mid-turn caused the turn to run 16 tool rounds and `_record_grace(tc, "round")` to fire twice in one `turn_stream()` call. Any transient provider blip after a grace has already been used lets the turn get the +3-round and/or spend-ceiling grace again, contradicting the "one-time, once per turn" design intent stated throughout this diff's comments.

tino097: **Finding 2:** The `TransientProviderError` session_backoff retry path (~4086-4152) re-enters `_stream_and_handle_tools` without being bounded by `_max_auto_retries` at all — only by a wall-clock/velocity budget — so it can retrigger the reset-and-regrant bug (finding 1) far more than the 3x a count-bounded retry would suggest. During an extended provider incident or a bursty rate limit this can re-arm both graces many times within the time budget, untested by any test in test_round_cap_grace.py or test_spend_ceiling.py.

tests/test_round_cap_grace.py
```diff
@@ -0,0 +1,195 @@
...
+        responses=[_tool_call(i) for i in range(1, max_rounds + _ROUND_CAP_GRACE_ROUNDS + 3)],
+    )
+    with patch("anton.analytics.send_event") as send:
+        await session.turn("do the thing")
+    assert send.call_args.kwargs["ended_by"] == "round_cap"
+    assert session._round_cap_grace_used
+    assert int(send.call_args.kwargs["rounds"]) == max_rounds + _ROUND_CAP_GRACE_ROUNDS
+
+
+async def test_round_cap_grace_lets_a_near_finish_wrap_up_without_asking(workspace):
```
tino097, line 137: **Finding 3:** This test and `test_round_under_the_cap_is_untouched` (line 147) never mock `session._llm.generate_object_code`, so their tool-loop shapes reach the real completion-verifier codepath and get back an unconfigured `MagicMock` instead of a `_VerifierVerdict`.

Verified by running both tests with instrumentation: `MagicMock.status` never equals COMPLETE/WAITING/STUCK (identity equality), so the turn falls into the INCOMPLETE branch, forces a spurious extra continuation, and appends a SYSTEM history message embedding a literal `<coroutine object AsyncMockMixin._execute_mock_call ...>` as the verifier's reason. Both tests' assertions still pass by coincidence, so they provide no real coverage of what their docstrings claim.

anton/core/session.py
```diff
@@ -2988,13 +3037,47 @@ def _spend_ceiling_gate(self) -> int:
...
+        `_spend_ceiling_grace_used` first never calls this twice, but a stray
+        second call would just recompute the same amount, not stack it.
+        """
+        peak = self._turn_cost.peak_context_tokens if self._turn_cost else 0
+        self._spend_ceiling_grace_tokens = min(
+            max(_SPEND_CEILING_GRACE_FLOOR, peak),
+            max(self._max_turn_tokens // 4, 1),
+        )
+        self._spend_ceiling_grace_used = True
+        if self._turn_cost is not None:
```
tino097, line 3181: **Finding 5:** `_grant_spend_ceiling_grace` overwrites `self._turn_cost.grace_tokens` here (plain assignment, not accumulation), and `_record_grace`'s tag dedup means a repeat grant of the same tag is invisible in `grace_granted`.

Under the retry-loop double-grace bug (finding 1), a turn granted the ceiling grace twice still reports `grace_granted='ceiling'` and `grace_tokens` equal to only the LAST grant's size — undercounting the true cumulative overshoot in exactly the telemetry this diff added to make grace usage measurable.

anton/core/turn_cost.py
```diff
@@ -124,6 +124,15 @@ class TurnCost:
     # note below on stamped-at-open fields). Handback terminals don't need it:
     # their ended_by already distinguishes them.
     verification_skipped: bool = False
+    # One-time grace(s) granted this turn: "", "round", "ceiling", or
```
tino097, line 129: **Finding 6:** This comment documents `grace_granted` as one of "", "round", "ceiling", or "round,ceiling", but `_record_grace` appends whichever tag fires first, so a turn whose ceiling grace trips before its round-cap grace produces "ceiling,round" — a fifth value not documented here. A downstream analytics query pattern-matching on the two documented orderings (rather than substring/set membership) would misclassify such turns.

anton/core/session.py
```diff
@@ -4453,24 +4564,44 @@ async def _stream_and_handle_tools(
...
+            _round_cap_grace_active = False
 
             while llm_response.tool_calls:
                 tool_round += 1
+                # One-time extension (ENG-1893), unconditional like the
+                # mid-loop spend-ceiling grace — no verdict exists yet at
+                # this point, so the grant is a single small window rather
+                # than a self-report. Decided before the accounting below so
+                # the granted round is counted like any other.
+                if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
```
tino097: **Finding 7:** This condition fires on the very first tool round when `max_tool_rounds <= 0`; `anton/core/settings.py:30` has no lower-bound validation on `max_tool_rounds` (unlike cowork-server's `ge=5`).

An operator setting `ANTON_MAX_TOOL_ROUNDS=0` intending to disallow any tool round instead gets the opposite: round 1 immediately "breaches" the cap and silently grants `_ROUND_CAP_GRACE_ROUNDS=3` extra rounds, contrary to the configured intent. No test covers `max_tool_rounds<=1`.

anton/core/session.py
```diff
@@ -351,6 +351,17 @@ class _VerifierVerdict(BaseModel):
         )
     )
     reason: str = Field(description="One brief sentence explaining the verdict.")
+    close_to_done: bool = Field(
```
tino097, line 355: **Finding 8:** `close_to_done: bool` is non-Optional and validated via plain `model_validate` (structured.py:343), Pydantic v2 lax mode — an explicit JSON `null` for this key raises ValidationError rather than falling back to the field default.

A model emitting `"close_to_done": null` alongside an otherwise-valid verdict gets the whole thing discarded via the generic `except Exception: verdict_failure = "hard"` handler, pushing toward the verifier hard-failure latch after two occurrences. Lax bool coercion also accepts "1"/"yes"/"on", working against the field's own "an unsure model errs toward asking" intent.

anton/core/session.py
```diff
@@ -4453,24 +4564,44 @@ async def _stream_and_handle_tools(
...
                     self._turn_cost.rounds += 1
-                if tool_round > self._max_tool_rounds:
+                if tool_round > effective_round_cap:
                     _max_rounds_hit = True
                     if self._turn_cost is not None:
                         self._turn_cost.ended_by = "round_cap"
                     self._acc_observe(
                         "cap_exhausted",
-                        {"cap": self._max_tool_rounds},
+                        {"cap": effective_round_cap, "configured_cap": self._max_tool_rounds},
```
tino097, line 4722: **Finding 9:** This widens `cap_exhausted`'s detail payload to `{"cap": effective_round_cap, "configured_cap": ...}`, but `anton/core/memory/acc.py:144`'s vocabulary comment still documents `cap_exhausted` as carrying `detail: {}`, and the new `cap` value is now the grace-inflated cap, not the configured one. No functional bug today (the one detector reading this kind ignores `detail`), but the doc/emitter drift is exactly what that file's own header warns against.

tests/test_spend_ceiling.py
```diff
@@ -148,13 +155,26 @@ async def test_total_is_bounded_by_the_ceiling(workspace):
...
     """
-    session = _session(workspace, responses=[_tool_call(i) for i in range(1, 40)])
+    session = _session(workspace, responses=[_tool_call(i) for i in range(1, 40)],
+                       per_call=per_call)
     with patch("anton.analytics.send_event") as send:
         await _run(session)
-    slack = 2 * (PER_CALL // 4)  # `_usage` puts a quarter of each call in output
+    slack = 2 * (per_call // 4)  # `_usage` puts a quarter of each call in output
+    peak = per_call - (per_call // 4)  # `_usage`'s context share of one call
+    grace = min(max(_SPEND_CEILING_GRACE_FLOOR, peak), CEILING // 4)
```
tino097: **Finding 10:** This tolerance term re-derives the exact production `_grant_spend_ceiling_grace` formula instead of asserting an independent, tighter bound, so a regression in that shared formula would move the production value and the test's expected bound in lockstep and never be caught. The loosened bound is also materially wide: for per_call=300_000 it permits up to a 37.5% overshoot of the configured ceiling, versus roughly 5-10% before this PR.

anton/core/session.py
```diff
@@ -3475,12 +3564,24 @@ async def turn(self, user_input: str | list[dict]) -> str:
...
+        # Mirrors turn_stream's one-time grants, the same way the round-cap
+        # and spend-ceiling checks below already mirror it. This path has no
+        # continuation loop, so each grant is simply spent once for the
+        # whole call rather than once per continuation.
+        _round_cap_grace_active = False
 
         while response.tool_calls:
             tool_round += 1
-            if self._turn_cost is not None and tool_round <= self._max_tool_rounds:
+            if tool_round > self._max_tool_rounds and not self._round_cap_grace_used:
```
tino097: **Finding 11 (cleanup):** The round-cap grace-activation block and its `effective_round_cap` formula are duplicated near-verbatim here and in `_stream_and_handle_tools` (~4582-4588); the ceiling grant-then-recheck idiom is similarly duplicated across three sites. Drift is already visible: only the streaming path calls `_acc_observe("cap_exhausted", ...)`. A shared helper for "check cap -> maybe grant grace -> recheck" would remove the duplication and the drift risk.

anton/core/session.py
```diff
@@ -438,6 +449,19 @@ class _VerifierVerdict(BaseModel):
...
 _SPEND_CEILING_RESERVE = 200_000
 
+# Floor for the one-time spend-ceiling grace — deliberately lower
+# than `_SPEND_CEILING_RESERVE`'s own floor. That reserve is sized to cover
+# TWO calls guaranteed to land after the gate trips; the grace is sized for
+# roughly ONE more, so granting it does not cancel the reserve outright and
+# put the turn back up near the ceiling. See `_grant_spend_ceiling_grace`.
+_SPEND_CEILING_GRACE_FLOOR = 100_000
+
+# One-time round-cap extension (ENG-1893), granted the first time a turn
```
tino097, line 467: **Finding 12 (convention):** New comments/docstrings added by this diff repeatedly embed the ticket id ENG-1893 (this line, plus ~1077-1080, 3056-3071, 4577-4581, 4644-4650, 5597-5602, and test_round_cap_grace.py's module docstring), and several exceed 2 lines. This violates the repo CLAUDE.md: "No ticket numbers, issue IDs, or ticket names in code, comments, commit messages, PR descriptions, tests, or docs" and "At most 2 lines per comment."

anton/core/session.py
```diff
@@ -3475,12 +3564,24 @@ async def turn(self, user_input: str | list[dict]) -> str:
         tool_round = 0
         error_streak: dict[str, int] = {}
         resilience_nudged: set[str] = set()
+        # Mirrors turn_stream's one-time grants, the same way the round-cap
+        # and spend-ceiling checks below already mirror it. This path has no
+        # continuation loop, so each grant is simply spent once for the
+        # whole call rather than once per continuation.
+        _round_cap_grace_active = False
```
tino097: **Finding 13 (cleanup):** `_round_cap_grace_active` in `turn()` is dead complexity — unlike `_stream_and_handle_tools`, `turn()` has one flat loop with no continuation wrapper, so this local flag stays in exact lockstep with `self._round_cap_grace_used` for the whole call and could be deleted in favor of reading that field directly. Separately, `_spend_ceiling_grace_used` is fully derivable from `_spend_ceiling_grace_tokens != 0` once granted.

anton/core/session.py
```diff
@@ -2988,13 +3037,47 @@ def _spend_ceiling_gate(self) -> int:
...
         reserve = min(
             max(_SPEND_CEILING_RESERVE, 2 * peak),
             max(self._max_turn_tokens // 2, 1),
         )
-        return max(self._max_turn_tokens - reserve, 1)
+        return max(
+            self._max_turn_tokens + self._spend_ceiling_grace_tokens - reserve, 1
+        )
+
+    def _grant_spend_ceiling_grace(self) -> None:
```
tino097, line 3159: **Finding 14:** No test exercises the round-cap grace AND the spend-ceiling grace both firing within the same turn, so the true worst-case combined overshoot (round grace + token grace stacked) is unverified, even though the two are independent one-shot flags that can both trip in one turn.

anton/core/session.py
```diff
@@ -2771,6 +2815,11 @@ def _emit_turn_cost(
                 # ENG-1632) must be excludable from the honest-stop
                 # denominator without a per-conversation Langfuse hop.
                 verification_skipped=str(tc.verification_skipped).lower(),
+                # One-time grace(s) granted this turn — see
+                # `TurnCost.grace_granted`. Empty string, not "-", to stay
+                # consistent with how other unset string properties go out.
+                grace_granted=tc.grace_granted,
```
tino097, line 2901: **Finding 15:** The new `grace_granted`/`grace_tokens` properties are added to the `turn_completed` analytics event with no mentioned update to any downstream collector allowlist — the nearby comment (a few lines up) documents a prior incident (ENG-1355) where such an allowlist silently dropped `turn_completed` properties for weeks. Worth confirming the allowlist (if any) was updated for these two new properties.

tino097:
**Finding 4** (posted as a general comment, not inline, since `tests/test_verifier_failsafe.py` is not touched by this diff so GitHub won't anchor a line comment on it):

`tests/test_verifier_failsafe.py:484`, `test_max_tool_rounds_diagnosis_is_persisted_as_streamed`, sets `max_tool_rounds=1` and mocks `generate_object_code` to raise `AssertionError("no verdict call on the max-rounds path")`, asserting the OLD unconditional round-cap trip. With the new grace, round 2's breach is silently absorbed (effective cap becomes 4), so the round-cap block is never entered.

Confirmed by tracing and running the suite: the turn falls through to normal completion verification, `generate_object_code`'s `AssertionError` fires and is swallowed by the generic `except Exception` handler, and the turn ends via `ended_by='handback_verifier_failure'` instead of `'round_cap'`. The test still passes (both paths stream a similar diagnosis message) so nothing flags this, but any real session with a small configured round cap now pays for an extra verifier call and gets misclassified `ended_by` analytics.

alecantu7 (commented):
## Adjudication of the self-review findings — and a merge call

Worked through all 14 findings above plus the three still-live items from my earlier passes, at head `d1929ac4`. Every verdict below is executed, not read — and every finding I concluded was blocking got handed to a second pass told to default to *not* blocking, so the bar was applied against me as well as for me.

**Bottom line: nothing blocks. One six-line patch is worth applying first, then merge.**

Thanks for doing this properly — F1 and F3 are real, and F1 is one I had myself and under-rated. I raised the retry re-entry in my first review as an "unverified risk", reasoned that the round impact was "subsumed because `tool_round` resets on re-entry anyway", named the exact experiment that would settle it, and then rated it low without running it. That reasoning was wrong: `tool_round` resetting means the retry gets a *fresh* budget, and the re-granted grace extends that fresh budget again. It compounds rather than being subsumed.

---

## Three findings do not survive — worth dropping rather than deferring

### F15 — refuted

This was the one with teeth and it does not hold. `turn_completed` is on the **direct-PostHog** path (`_POSTHOG_EVENTS`), and being moved there was the fix for the very incident F15 invokes. There is no property allowlist anywhere on that path — the only filtering in `_posthog_body` drops `_`, `action` and `timestamp`. Adding a property to an event already on that path needs no allowlist work. `grace_granted`/`grace_tokens` will exist in project 424726 the moment this ships.

Residual, and it is only that: the register **comment** at `analytics.py:165-169` enumerates `turn_completed`'s properties and was not extended. That comment documents, it does not gate. Folds into the doc-drift bundle with F6/F9.

### F5 — refuted

The premise is that overshoot accumulates across grants and the field reports only the last slice. That is false by construction: the ceiling grace is an absolute **gate offset**, not a budget increment, so it cannot stack. Executed A/B — one grant vs two — gives byte-identical token totals. The value reported is the largest offset in force, which is exactly the overshoot bound the field exists to expose.

On the dedup: `grace_granted` is a tag **set**, and `_record_grace`'s own docstring says "comma-joined and deduplicated" (`session.py:860-861`). Reading it as a grant counter reads a semantics it never claimed.

The one true residual is narrow and belongs to F1, not here: under the unfixed reset the **round** grace does compound while the tag stays a single `"round"`. The F1 patch below closes that.

### F12 — refuted, and this is the one I would drop hardest

Three independent problems:

1. **The governing document does not exist.** anton has no `CLAUDE.md`, no `AGENTS.md`, no `.claude/` — not on disk, not in `origin/staging`.
2. The quoted text is real but comes from an **agent-instruction plan for a different ticket**, sitting on an unmerged branch in a separate worktree. It binds the agent doing *that* task in *those* files.
3. That plan itself concedes the point: *"Surrounding existing code in these files already does this in places — that's pre-existing convention."*

And the counts settle it: `session.py` carried **140 ticket references and 210 multi-line comment blocks before this PR**; 819 refs across 112 files repo-wide. A rule the file already breaks 140 times is not the house convention. This would generate real churn for no benefit.

---

## F1 — real, introduced, downgraded on measurement, and worth fixing anyway

Confirmed and reproduced independently: with `max_tool_rounds=5` and one injected `TransientProviderError`, rounds came out **15** against an intended effective cap of 8, with `_record_grace` firing **2x `['round','round']`** while `grace_granted` reported a single `'round'`.

But the alarming multiplier is a test-config artifact. Re-measured at the **shipping default** (`max_tool_rounds = 25`), blips placed identically in both arms:

| blips | pre-PR rounds | post-PR rounds |
| -- | -- | -- |
| 0 | 25 | 28 |
| 1 | 49 | 55 |
| 2 | 73 | 82 |
| 3 | 97 | 109 |

A flat **+3 per retry segment — +12%**, not +60%. The round *multiplication across retries is pre-existing*: 97 rounds on 3 blips ships on `staging` today. Only the `+3` is attributable to this PR.

**Zero extra cost, measured.** With the ceiling grace re-granted twice, `tokens_total` is byte-identical at 1,260,000 across 0–6 blips — `_spend_ceiling_grace_tokens` is *assigned*, never accumulated, and `_turn_cost.total_tokens` is turn-wide. At realistic per-call context (≥30k) the full production numbers are byte-identical between `GRACE=3` and `GRACE=0`: rounds 39, tokens 1,200,000, `ended_by=spend_ceiling`. The bug is invisible on any turn expensive enough to matter.

Also worth recording: the reset idiom is **pre-existing house style**. `_compacted_this_turn = False` sits at the identical spot inside `_stream_and_handle_tools`, re-armed by the identical retry loop, in production right now. This PR followed the idiom; it did not invent the class.

So: no extra spend, no corruption, bounded extra latency, same `round_cap` terminal, nothing new on screen. That is a follow-up on harm alone.

**What tips it to "fix it now" is that the fix is trivial and already verified.** Move the three resets out of `_stream_and_handle_tools` and into `_turn_stream_inner`, right after `_retry_count = 0`:

- `tests/test_round_cap_grace.py`, `test_spend_ceiling.py`, `test_turn_cost_terminals.py`, `test_transient_retry.py` → **114 passed**
- round grants 5→**1** and 8→**1**; rounds 35→12 and 56→12
- `ended_by` improves from `error` back to **`round_cap`**, the intended terminal
- the per-continuation `_round_cap_grace_active` semantics are preserved — it stays loop-local and still resets per continuation; `_round_cap_grace_used` correctly becomes turn-wide

Six mechanical lines, green, and it stops us shipping a hardening commit titled *"truly one-shot"* that one ordinary transient blip defeats.

Optional in the same patch: reset `self._turn_cost.grace_tokens` alongside `_spend_ceiling_grace_tokens`.

---

## F2 — real, pre-existing, and subsumed

The mechanical claim is correct: the session-backoff path `continue`s at `:4152` before `_retry_count += 1` at `:4182`, so it never consumes the count budget. But it is verbatim pre-existing ENG-673/ENG-1537 code, and it is deliberate — a rate limit has a known end time, so a wall-clock/slept budget is the right instrument and a count is the wrong one.

"Far more than 3x" overstates it. Measured maxima: **5** re-entries on the incident path (30s wall clock), **8** on the rate-limited path (90s slept), ~7 interleaved. Ceiling ~10 per `turn_stream()` call, not an open loop.

Decisive: with the F1 patch applied, the same 8-blip run produces **1** round grant instead of 8. F2 needs no fix of its own.

---

## The rest — real, none blocking

**F10** is the pick of them (medium): the bound test re-derives the production formula, so a regression in that formula moves the production value and the expected bound in lockstep. Worth an independent, tighter assertion.

**F3** (fixture never configures `generate_object_code`, so the turns hit the real verifier with an unconfigured `MagicMock` and embed a `<coroutine object …>` in history) — reproduced, real, and worth fixing; the assertions do still hold, so it is hygiene rather than a false green.

**F6, F9** and the F15 residual are documentation drift — one bundle. **F7** (`max_tool_rounds=0` grants 3 instead of 0) and **F8** (`close_to_done: null` discards the verdict) are real edges on values nothing sets today. **F11, F13** are cleanup.

## One item of mine that was never posted, still live

The spend-ceiling overshoot is quantised to **one call**, not to the grace. The PR's own asserted bound `total <= CEILING + slack + grace` holds to `per_call=500_000` and **breaks at 600_000** — total 1,800,000 against a bound of 1,550,000, 180% of a 1,000,000 ceiling — because `grace` is capped at `CEILING // 4` while one call costs more. The test parametrises only `[100_000, 300_000]`, so it stops two steps short of where its own claim fails.

Low: it needs a single call worth >25% of the whole turn budget. The fix is to state the true bound (`CEILING + slack + max(grace, one_call)`) and change the docstring from "overshoots by approximately its own size" to "by one more call".

---

## Recommendation

1. Apply the six-line F1 reset move — verified, 114 tests green.
2. **Merge.**
3. One follow-up ticket bundling F3, F10, F14, F6/F9/F15-residual, F7, F8, F11, F13, and the quantised bound.
4. Soften the PR body's headline claim: it is *"no longer stops when it is close to done"*, not *"no longer stops just because it hit a round or token cap"*. The graces are per-turn and finite, so a task needing more than one window still asks — and can still ask on consecutive turns, which is the symptom the ticket was filed on.

Everything I raised in the earlier rounds that `d1929ac4` addressed, I re-verified: round-cap re-arm per continuation (32→13 rounds, `ended_by` back to `round_cap`), ceiling sizing (110%→100%, 112%→104% at small per-call), `turn()` parity and the flag leak, `cap_exhausted`'s `configured_cap`, and the `close_to_done` wording pin. All good.

alecantu7 (commented):
## Re-review of `43a8a9df` — merge-ready

One real commit since the adjudication; the other 22 are the `staging` merge. Base is fresh (`behind_by=0`). It lands three fixes and **all three are mutation-verified** — I reintroduced each bug and watched the corresponding test fail:

| Fix | Mutation applied | Result |
| --- | --- | --- |
| Grace re-arm on retry | reset moved back into `_stream_and_handle_tools` | `test_round_cap_grace_survives_a_mid_loop_retry` → **14 rounds vs the expected 11** |
| Zero-cap guard | dropped `and self._max_tool_rounds > 0` | `test_a_zero_cap_denies_the_first_round_instead_of_granting_one` → **fails** |
| Null `close_to_done` | removed the `field_validator` | **ValidationError**, test fails |

The re-arm fix is right structurally, not merely test-green. `_round_cap_grace_used` is now turn-scoped — reset in `_turn_stream_inner` at `session.py:4065-4068`, *outside* the `while not routed_direct:` retry loop — while `_round_cap_grace_active` stays per-continuation. Granting is once per turn, applying is once per continuation, which is exactly the intended one-shot semantics. Each function has a single caller, so there is no path that reaches the loop with the reset skipped. This closes the finding I reproduced at 15 rounds against an intended cap of 8.

The narrow scope of the `close_to_done` validator is also correct rather than incidental: it is the only field on `_VerifierVerdict` with a default, so it is the only one that *can* fall back. A null `status` or `reason` has nothing to fall back to, and failing the whole verdict there is the right outcome.

**Full suite: 2,789 passed, 31 skipped, 0 failed.** All 13 checks green.

Two findings, neither blocking — both are comment/test-only and either could follow the merge.

---

### Finding A — the "known gap" docstring misdiagnoses the overshoot, and "tracked as a follow-up" is not true yet

`session.py:3145` and the `test_total_is_bounded_by_the_ceiling` docstring both attribute the large overshoot to the grace being capped at `max_turn_tokens // 4`. I measured it at the production default (`max_turn_tokens = 1_250_000`, `settings.py:52`) against a no-grace control:

| `per_call` | no grace | with grace | marginal |
| --- | --- | --- | --- |
| 200,000 | 1,200,000 | 1,400,000 | +200,000 |
| 400,000 | 1,200,000 | 1,600,000 | +400,000 |
| 500,000 | 1,500,000 | 1,500,000 | **0** |
| 600,000 | 1,800,000 | 1,800,000 | **0** |

At `per_call=600_000` the grace **is** granted (`grace_granted='ceiling'`, `grace_tokens=312500`) and contributes **nothing** — the totals are byte-identical with and without it. The 1,800,000 is entirely the pre-existing pre-call gate granularity: a gate that checks *before* a call can always be overshot by that call's full cost. The `// 4` cap is not the mechanism.

That matters because of where it sends the follow-up. Written from this docstring, the follow-up goes and tunes `// 4` — and changes nothing, since at the `per_call` values in question the cap is not what is binding. The real fix is a pre-call cost estimate, which is a different piece of work.

The accurate statement is also the simpler one: **the grace's marginal cost is exactly one more LLM call, whenever one more call fits.** That is a fair price for what the ticket is buying, and it is a bound a reader can reason about.

Separately, I searched Linear twice (`spend ceiling overshoot`, `max_turn_tokens grace overshoot one call`) and found no follow-up ticket. So "Tracked as a follow-up, not fixed here" is currently false — worth either filing it or dropping the clause, since an untracked claim of tracking is how this gets forgotten.

One note on method, because it cuts in the PR's favour: my first attempt at this measurement set `_spend_ceiling_grace_used = True` as the no-grace control and reported marginal=0 across the board. That was wrong — the new reset in `_turn_stream_inner` clears the flag at turn start, so the fix under review silently defeated my control. I had to no-op `_grant_spend_ceiling_grace` instead. The numbers above are from the corrected control.

### Finding B — `sorted()` in `_record_grace` is unpinned

Removing `sorted()` from `_record_grace` (`session.py:896`) leaves **all 8 round-cap tests passing**.

`test_both_graces_can_fire_in_the_same_turn` asserts `grace_granted == "ceiling,round"`, but in that fixture the ceiling grace fires first — so the unsorted value already *is* `"ceiling,round"`. The assertion passes for the wrong reason, and the sort it was added to protect is unguarded. If the firing order ever flips, `grace_granted` becomes `"round,ceiling"` and anything filtering on the literal string silently drops those turns — which is the failure mode the sort exists to prevent, and an analytics one, so it fails quietly.

Six lines fix it. I verified this fails with the sort removed and passes with it restored:

```python
def test_record_grace_is_order_independent():
    """The two-grace value must be one fixed string regardless of which grace
    fired first — consumers filter on the literal `"ceiling,round"`."""
    a = TurnCost(); _record_grace(a, "ceiling"); _record_grace(a, "round")
    b = TurnCost(); _record_grace(b, "round"); _record_grace(b, "ceiling")
    assert a.grace_granted == b.grace_granted == "ceiling,round"
```

---

### Context, not a finding

`turn()` still carries its own copy of the round-cap grace logic, and it has no production caller — anton's own CLI (`chat.py:1920`), cowork-server (`anton_harness/harness.py:411`, `connectors/probe.py:437`) and scratchpad-controller (`anton_turn.py:74`) all go through `turn_stream()`. Removing the dead `_round_cap_grace_active` there was the right call; the remaining duplication is a maintenance note, not a defect.

### What was reviewed

Full `anton` suite at `43a8a9df` on a clean checkout (2,789 passed / 31 skipped / 0 failed); four separate mutations executed against `session.py`, each restored and verified restored before the next; a purpose-built measurement harness for Finding A using the repo's own scripted-mock fixtures at the production `max_turn_tokens`; grace-flag lifecycle traced across `__init__` / `turn()` / `_turn_stream_inner` / `_stream_and_handle_tools`; caller graph checked across `anton`, `cowork-server` and `scratchpad-controller`; Linear searched for the claimed follow-up.

**Verdict: COMMENT — merge-ready.** Neither finding blocks. The merge is a human call, not mine.

alecantu7 (approved):
## Approving `43a8a9df`

All three fixes in `43a8a9df` are mutation-verified — I reintroduced each bug and watched the matching test fail:

| Fix | Mutation | Result |
| --- | --- | --- |
| Round-cap grace re-arming on retry | reset moved back into `_stream_and_handle_tools` | `test_round_cap_grace_survives_a_mid_loop_retry` → **14 rounds vs the expected 11** |
| Zero-cap guard | dropped `and self._max_tool_rounds > 0` | `test_a_zero_cap_denies_the_first_round…` → **fails** |
| Null `close_to_done` | removed the `field_validator` | **ValidationError**, test fails |

The re-arm fix is right structurally, not merely green: `_round_cap_grace_used` is now turn-scoped, reset in `_turn_stream_inner` **outside** the `while not routed_direct:` retry loop, while `_round_cap_grace_active` stays per-continuation. Granting once per turn, applying once per continuation — which is the intended one-shot semantics, and each function has a single caller so no path reaches the loop with the reset skipped. That closes the finding I reproduced at 15 rounds against an intended cap of 8.

Full suite at this head: **2,789 passed, 31 skipped, 0 failed.** All 13 checks green.

## The two open items are follow-ups, not conditions

Neither blocks, and I am taking both off this PR rather than asking for another round after fifteen self-review findings and three review passes.

**1. `sorted()` in `_record_grace` is unpinned.** Removing it leaves all 8 round-cap tests passing, because in the both-graces fixture the ceiling grace fires first and the unsorted value already reads `"ceiling,round"`. The assertion passes for the wrong reason. Six lines fix it, verified failing under mutation and passing after:

```python
def test_record_grace_is_order_independent():
    a = TurnCost(); _record_grace(a, "ceiling"); _record_grace(a, "round")
    b = TurnCost(); _record_grace(b, "round"); _record_grace(b, "ceiling")
    assert a.grace_granted == b.grace_granted == "ceiling,round"
```

**2. The "known gap" docstring misdiagnoses the overshoot.** `session.py:3145` and the `test_total_is_bounded_by_the_ceiling` docstring attribute it to the grace being capped at `max_turn_tokens // 4`. Measured at the production default (1,250,000) against a no-grace control:

| `per_call` | no grace | with grace | marginal |
| --- | --- | --- | --- |
| 200,000 | 1,200,000 | 1,400,000 | +200,000 |
| 400,000 | 1,200,000 | 1,600,000 | +400,000 |
| 600,000 | 1,800,000 | 1,800,000 | **0** |

At 600k the grace **is** granted (`grace_tokens=312500`) and contributes nothing — the totals are identical with and without it. The `// 4` cap is not the mechanism; the overshoot is pre-existing gate granularity, because a gate that checks before a call can always be overshot by that call's full cost. A follow-up written from this docstring would go tune `// 4` and change nothing. The accurate statement is simpler: **the grace's marginal cost is exactly one more LLM call, whenever one more call fits.**

I am filing that follow-up now, which also settles the "tracked as a follow-up" clause — it was not tracked anywhere; I searched Linear twice and found no ticket. The docstring correction and the `sorted()` test both belong in that ticket's PR, since they touch the same file.

## Notes for the merge

`behind_by=1` — one commit has landed on `staging` since. No conflict, but the suite has not run against the merged result, so a rebase is worth it if the merge is not immediate. And `staging-freeze` is active on this repo, so this approval clears the review gate and not the freeze.

Approving.

alecantu7:
Follow-up filed as **[ENG-2070](https://linear.app/mindsdb/issue/ENG-2070/a-turn-can-overshoot-its-token-ceiling-by-one-full-llm-call-the-gate)** — "A turn can overshoot its token ceiling by one full LLM call — the gate has no pre-call cost estimate".

It carries both open items from the approval above, so neither is lost on merge:

- The `session.py:3145` / `test_total_is_bounded_by_the_ceiling` docstring correction, with the measurement showing the `// 4` cap is **not** the mechanism (at 600k per call the grace is granted and contributes zero — the totals are identical with and without it).
- `test_record_grace_is_order_independent`, verified failing with `sorted()` removed and passing with it restored.

That also settles the "Tracked as a follow-up, not fixed here" clause in the docstring, which was not true at the time — I searched Linear twice before filing and found nothing tracking it. Pointing it at ENG-2070 makes it true.

Nothing here blocks this PR.

PR #397, opened 2026-08-25
ENG-1945: stamp surface on turn_completed and tool_completed

alecantu7:
Closes [ENG-1945](https://linear.app/mindsdb/issue/ENG-1945/turn-completed-cant-say-which-surface-a-turn-ran-on-surface-exists-on).

## What

`surface` (`desktop` / `web` / `cli`) has ridden Langfuse traces since ENG-1459 (#373) but was **deliberately left off anton's PostHog events** — ENG-1459's Out-of-scope reads *"PostHog's renderer events already stamp surface … do not redo it."* Those renderer events ride a different `distinct_id` (Keycloak sub) from anton's `aid`, so they cannot split `turn_completed` — and `turn_completed` is where per-turn cost and outcome live. Measured 2026-08-25: `surface` is `None` on **100%** of `turn_completed` rows, desktop included.

Two emit sites, one line each, beside the existing `harness=`:

- `_emit_turn_cost` → `turn_completed`
- `_emit_tool_completed` → `tool_completed` (same derivation, so a tool row splits by surface without joining to its parent turn)

`""` when the host did not say — never a guess; `_validated_surface` already dropped anything unrecognised at config time. `getattr(self, "_surface", None)` rather than the bare attribute so a host that builds a session unusually can never turn a telemetry field into a swallowed emit.

Also registers `endpoint_class` / `error_type` in `analytics.py`'s event register — #381 shipped them without adding the doc entries.

## Why this and not the ENG-1459 reasoning

ENG-1459 gave two reasons for leaving it off. Neither holds for this event: (1) the renderer's `surface` can't join to `aid`-keyed rows; (2) "web pods collapse users into one `aid`" is true of **every** field on a web `turn_completed`, not this one, and is irrelevant on desktop where all real traffic is. Recorded on the ticket so the decision isn't re-litigated from the old text. Web rows will carry `surface=web` under a shared `aid` until ENG-1459/ENG-1694 resolve pod identity — noted so nobody reads per-user web numbers off this event.

## Verification

- `test_turn_cost.py::test_surface_is_stamped_and_empty_when_the_host_did_not_say` — `"desktop"` when set, `""` when `None`.
- `test_tool_completed.py::test_payload_is_exactly_the_seven_keys_all_strings` (was six) + `test_surface_rides_the_tool_row_and_is_empty_when_the_host_did_not_say`.
- **Mutation-verified:** with both `surface=` kwargs deleted, exactly those three tests fail (3 failed / 53 passed); restored, all pass.
- Full suite (`uv sync --frozen --all-groups`, macOS): **2,501 passed, 31 skipped, 5 failed** — all five in `tests/test_cloud_turn_process.py`, and **identical on pure `origin/staging`** (stashed check: 5 failed / 1 passed). CI on `staging` at the same base `17dc5b5` is green, so those are local-environment failures, not this branch.

## Delivery / QA note

anton changes reach desktop users only after this lands on anton `main` and cowork-server picks it up — verify on the released build, not this PR. After it ships: `turn_completed` on the new `anton_version` shows `surface=desktop` on desktop rows; `.claude/skills/turn-cost` can then group by it (follow-up on the analytics side, not this repo). No wire change — `turn_completed` posts direct to PostHog (ENG-1495), so no collector allowlist to update.

## Security check

Observability-only. `surface` is a closed enum validated at config time (`_validated_surface`); the value is never user-supplied text, never a URL or path. No new endpoints, no new inputs. Same sink, same opt-out (`analytics_enabled`) and CI drop as the existing properties.

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -165,7 +165,9 @@ _pending_lock = threading.Lock()
 #                    cache_read_tokens, cache_creation_tokens, llm_calls,
 #                    rounds, continuations, peak_context_tokens, duration_ms,
 #                    {planning,coding,router}_{model,tokens,calls},
-#                    unknown_{tokens,calls}, llm_provider, harness,
+#                    unknown_{tokens,calls}, llm_provider, endpoint_class,
+#                    error_type, harness, surface (desktop / web / cli — WHERE
+#                    the user was, "" when the host did not say; ENG-1945),
 #                    anton_version, conversation_id, turn_index
 #
 #   rule_retrieval   outcome, when_rules, kept_rules, rules_chars,
@@ -180,6 +182,7 @@ _pending_lock = threading.Lock()
 #                    never the message, which carries paths and user input).
 #                    One event per executed tool call; tool arguments and
 #                    result content are deliberately absent (ENG-1486).
+#                    surface mirrors turn_completed's (ENG-1945).
 #                    conversation_id + turn_index mirror turn_completed's
 #                    values, so a tool row joins to its parent turn row and,
 #                    via Langfuse sessionId, to the gateway trace.
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2871,9 +2871,10 @@ class ChatSession:
         is reported as ``"unknown"`` rather than coerced either way.
 
         Payload is deliberately name + verdict + duration + exception CLASS
-        plus the two join keys, and nothing else. Arguments, result content
-        and ``str(exc)`` routinely carry file paths, user data and
-        credentials-adjacent strings — none of them may ever appear here.
+        + surface (a closed enum, ENG-1945) plus the two join keys, and
+        nothing else. Arguments, result content and ``str(exc)`` routinely
+        carry file paths, user data and credentials-adjacent strings — none
+        of them may ever appear here.
 
         ``conversation_id`` / ``turn_index`` mirror ``turn_completed``'s
         values exactly (same names, same derivation), so a tool failure spotted
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2828,6 +2828,15 @@ class ChatSession:
                 # URL itself — those carry internal corporate hostnames.
                 endpoint_class=classify_endpoint(_turn_settings),
                 harness=str(self._harness or ""),
+                # WHERE the user was (desktop / web / cli), as opposed to which
+                # agent ran (`harness`). Same value the Langfuse `surface:` tag
+                # carries (ENG-1459); it was deliberately left off this event
+                # then, on the grounds that PostHog's renderer events already
+                # had it — but those ride a different distinct_id and cannot
+                # split this row, and this row is where cost and outcome live
+                # (ENG-1945). "" when the host did not say; never a guess —
+                # `_validated_surface` already rejected anything unrecognised.
+                surface=str(getattr(self, "_surface", None) or ""),
                 anton_version=_anton_version,
                 # Join keys: the same session/turn identity the MindsHub
                 # trace headers carry, so an analytics row links back to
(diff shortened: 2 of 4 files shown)


alecantu7 (commented):
## Adversarial self-review (/pr-review procedure, standard mode)

Triage: no deep-mode signals — anton-internal, no wire contract (PostHog direct sink, no allowlist), no persisted state, no version skew (`_surface` is set in `ChatSession.__init__` at every tag since #373; `getattr` guards the `__new__`-built test doubles). Base `origin/staging` verified against GitHub at `17dc5b5`.

**One finding (low, confirmed) — fixed in the follow-up commit.** `_emit_tool_completed`'s docstring still said the payload was *"name + verdict + duration + exception CLASS plus the two join keys, and nothing else."* After this PR it also carries `surface`. On a payload whose whole design argument is privacy scope, a stale "nothing else" is exactly the sentence a reviewer trusts instead of reading the call. Corrected to list `surface (a closed enum)`.

**Attacks that failed (recorded so nobody re-runs them):**
- *Some test pins the exact key set of `turn_completed` and would now silently constrain it.* Grepped all `turn_completed` tests — none assert the full key set; `tool_completed`'s does, and it's updated (six → seven) with the new key named.
- *A host path never sets `_surface`, so the new property is `""` everywhere and the PR changes nothing.* `chat_session.py:128` / `chat.py:1403` → `SURFACE_CLI`; `cloud_turn/session.py:595` → from the request's `trace` block (empty today on pods until the scratchpad image is bumped — that is ENG-1459's open step, not this PR's); cowork-server → `surface_kwarg` (desktop). Every originating path declares it.
- *Conflicts with the open, gated anton#377 (ENG-1694).* #377 touches `session.py` but not either emit block — no overlapping hunks. Either merge order is safe.
- *`getattr(self, "_surface", None)` hides a real bug.* It only differs from the bare attribute for objects built without `__init__` (the `_bare_session` test double). Chosen so a telemetry field can never turn into a swallowed `except Exception: pass` around the whole emit — the same failure the harness field would have had. Not a behaviour change for real sessions.

**Verification:** 3 new/changed tests; deleting both `surface=` kwargs fails exactly those 3 (mutation-verified). Full suite 2,501 passed / 31 skipped / 5 failed — all five in `test_cloud_turn_process.py` and identical on pure `origin/staging` (stashed run), with CI green on that same base: local environment, not this branch.

**Unverified:** none worth naming. The web `aid` collapse is pre-existing and documented in the body.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed all four changed files at `5dfe1b6`. Observability-only; `surface=` rides the exact same wire path as the adjacent `harness=`, which already ships. No actionable findings.

### Validation

- **No allowlist drops the field:** `send_event(settings, action, **extra)` does `params.update(extra)` with no key filtering (analytics.py:480–505); `surface` reaches PostHog beside `harness`, and the emit is wrapped in a swallow-all try/except so it can't fail the turn.
- **`analytics.py` register is a comment, not enforcement** — confirmed by the passthrough — so the register/docstring edits are accuracy-only.
- **No exact-key test on `turn_completed` breaks** (there is none); `tool_completed`'s exact-keys test was correctly bumped six→seven with the matching set.
- **`_surface` is initialized in `__init__`** (session.py:1120) via `_validated_surface`, so the value is a closed enum or `""` — never user text; the `getattr(…, None)` guard covers unusually-built sessions.
- Did not run the suites; relied on the author's mutation evidence (deleting both kwargs fails exactly the three named tests) plus the static checks above.

### Findings

No actionable findings.

**Data caveat (already documented):** web pods collapse users into one `aid`, so `surface=web` rows aren't per-user-attributable until pod identity resolves — true of every field on a web `turn_completed`, so a read-time caveat, not a defect here.

PR #398, opened 2026-08-25
ENG-1849: fix connect from chat

ea-rus:
Changes
- connect_new_datasource's non-interactive path (known_variables) no longer needs a terminal — it returned "no console available" on every call from Cowork.
- Interactive mode now fails with a clear message instead of hanging/crashing when there's no console, and the tool stops advertising that mode to hosts that can't service it.
- Connection test now times out after 30s instead of hanging forever when the target server doesn't respond.

Fixes https://linear.app/mindsdb/issue/ENG-1849/connect-new-datasource-is-dead-in-cowork-a-console-guard-blocks-the

### DIFF
--- a/anton/commands/datasource/verify.py
+++ b/anton/commands/datasource/verify.py
@@ -18,6 +18,10 @@ if TYPE_CHECKING:
     from rich.console import Console
     from anton.core.backends.manager import ScratchpadManager
 
+# Bounds only the connect attempt itself (pad.execute of the test_snippet).
+# A ModuleNotFoundError retry's pad.install_packages() call is not covered —
+# it has its own, separate cell_install_timeout (default 120s) — so the
+# worst-case wall clock for a full test is higher than this number suggests.
 CONNECTION_TEST_TIMEOUT_SECONDS = 30
 
 
@@ -29,11 +33,17 @@ async def run_connection_test(
     credentials: dict[str, str],
     retry_fields: "list[DatasourceField]",
     retry_edit_callback: "Callable[[], Awaitable[bool]] | None" = None,
+    *,
+    interactive: bool = True,
 ) -> bool:
     """Inject flat DS_* vars, run engine_def.test_snippet, restore env.
 
     Returns True on success, False if the user declines retry after failure.
     Mutates credentials in-place when the user re-enters secrets on retry.
+    `interactive=False` (mode (a) callers with no terminal to prompt in)
+    skips the "retry?" prompt entirely and fails closed on the first error —
+    prompt_or_cancel drives a real terminal regardless of the `console`
+    passed in, so it must never be reached without one.
     """
     while True:
         console.print()
@@ -61,12 +71,23 @@ async def run_connection_test(
 
             cell = None
             for attempt in range(3):
-                try:
-                    cell = await asyncio.wait_for(
-                        pad.execute(engine_def.test_snippet),
-                        timeout=CONNECTION_TEST_TIMEOUT_SECONDS,
-                    )
-                except asyncio.TimeoutError:
+                task = asyncio.ensure_future(pad.execute(engine_def.test_snippet))
+                done, _pending = await asyncio.wait(
+                    {task}, timeout=CONNECTION_TEST_TIMEOUT_SECONDS
+                )
+                if task not in done:
+                    # The local backend catches its own internal
+                    # CancelledError and returns a Cell instead of
+                    # re-raising, so `await task` here would silently hand
+                    # back that Cell (with a generic kill-tree message)
+                    # rather than surfacing our timeout — cancel and
+                    # discard its result/exception either way, and always
+                    # report our own message.
+                    task.cancel()
+                    try:
+                        await task
+                    except (asyncio.CancelledError, Exception):
+                        pass
                     cell = Cell(
                         code=engine_def.test_snippet,
                         stdout="",
@@ -78,6 +99,7 @@ async def run_connection_test(
                         ),
                     )
                     break
+                cell = task.result()
                 if cell.error and "ModuleNotFoundError" in cell.error:
                     match = re.search(r"No module named '([^']+)'", cell.error)
                     if match:
@@ -98,6 +120,8 @@ async def run_connection_test(
             console.print()
             console.print(f"        Error: {last_line}")
             console.print()
+            if not interactive:
+                return False
             retry = await prompt_or_cancel(
                 "(anton) Would you like to re-enter your credentials?",
                 choices=["y", "n"], default="n",
--- a/anton/commands/datasource/verify.py
+++ b/anton/commands/datasource/verify.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
 import re
 from typing import TYPE_CHECKING, Awaitable, Callable
 
 from anton.commands.datasource.helpers import prompt_field_value
+from anton.core.backends.base import Cell
 from anton.core.datasources.data_vault import DataVault, LocalDataVault
 from anton.core.datasources.datasource_registry import DatasourceEngine, DatasourceField, DatasourceRegistry
 from anton.utils.datasources import parse_connection_slug, register_secret_vars, restore_namespaced_env
@@ -15,6 +17,8 @@ from anton.utils.prompt import prompt_or_cancel
 if TYPE_CHECKING:
     from rich.console import Console
     from anton.core.backends.manager import ScratchpadManager
+
+CONNECTION_TEST_TIMEOUT_SECONDS = 30
 
 
 async def run_connection_test(
@@ -57,7 +61,23 @@ async def run_connection_test(
 
             cell = None
             for attempt in range(3):
-                cell = await pad.execute(engine_def.test_snippet)
+                try:
+                    cell = await asyncio.wait_for(
+                        pad.execute(engine_def.test_snippet),
+                        timeout=CONNECTION_TEST_TIMEOUT_SECONDS,
+                    )
+                except asyncio.TimeoutError:
+                    cell = Cell(
+                        code=engine_def.test_snippet,
+                        stdout="",
+                        stderr="",
+                        error=(
+                            f"Connection test timed out after "
+                            f"{CONNECTION_TEST_TIMEOUT_SECONDS}s — the server "
+                            f"did not respond."
+                        ),
+                    )
+                    break
                 if cell.error and "ModuleNotFoundError" in cell.error:
                     match = re.search(r"No module named '([^']+)'", cell.error)
                     if match:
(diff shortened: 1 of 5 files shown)


tino097 (commented):
A few notes from a pass over this diff (see inline comments). One more that doesn't map to a changed line so I'm leaving it here: `anton/core/backends/remote.py` (not touched by this PR) — `RemoteScratchpadRuntime.execute_streaming` has no cancellation handling, so if that backend is ever in play, `verify.py`'s new `wait_for(timeout=30)` would abandon the await locally on timeout but leave the remote-side execution running unbounded. Not exercised by any caller in this repo today, just flagging as a latent gap in shared infra.

anton/tools.py
```diff
@@ -65,9 +66,12 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
...
 
-    console = session._console
-    if console is None:
-        return "Cannot connect datasource — no console available."
+    # Mode (a) below needs no terminal — it writes straight to the vault. A
+    # quiet console swallows its status prints instead of crashing on
+    # `None.print()`. The real console (if any) is required only for mode
+    # (b)'s prompts, gated further down.
+    from rich.console import Console
+    console = session._console or Console(quiet=True)
```
tino097, line 74: Mode (a) still flows through to `run_connection_test` here even when there's no real console — and that function's failure-retry prompt (`prompt_or_cancel` in `verify.py`) drives a real terminal session regardless of the `Console` object passed in. In a console-less host this can hang or crash on a failed test, which looks like the exact bug class this PR is fixing elsewhere.

anton/commands/datasource/verify.py
```diff
@@ -57,7 +61,23 @@ async def run_connection_test(
 
             cell = None
             for attempt in range(3):
-                cell = await pad.execute(engine_def.test_snippet)
+                try:
+                    cell = await asyncio.wait_for(
```
tino097: Against the default `LocalScratchpadRuntime`, `execute_streaming` catches `CancelledError` internally and returns a `Cell` instead of re-raising — so `wait_for` returns that cell rather than raising `TimeoutError`, and this `except` branch (and the friendlier "timed out after 30s" message) never actually fires. The kill still happens, just with the generic kill-tree error text instead.

anton/tools.py
```diff
@@ -396,6 +418,32 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
     handler = handle_connect_datasource,
 )
 
+# Console-less hosts (Cowork desktop: session._console is None) can't service
+# mode (b) — there's no terminal to prompt in, and handle_connect_datasource
+# refuses it with an actionable error rather than crashing (ENG-1849). Don't
+# advertise a mode the model would then have to be told, mid-turn, doesn't
+# work here — swapped in by ChatSession.__init__ (anton/core/session.py)
+# once session._console is known.
+CONNECT_DATASOURCE_TOOL_NO_CONSOLE = dataclasses.replace(
```
tino097, line 449: This only overrides `description` — `input_schema` is shared by reference with `CONNECT_DATASOURCE_TOOL` and still only requires `engine`, which contradicts the new description's "do NOT call with just an engine" guidance. Might be worth giving this an explicit (copied) `input_schema` too.

anton/core/session.py
```diff
@@ -1100,6 +1100,19 @@ def __init__(self, config: ChatSessionConfig) -> None:
         self._workspace = config.workspace
         self._data_vault = config.data_vault
         self._console = config.console
+        if self._console is None:
```
tino097, line 1103: This snapshots `config.console` once at construction rather than re-checking live `session._console` like the other capability-gated tools do — if a console gets attached later, the advertised tool description can go stale relative to reality.

anton/commands/datasource/verify.py
```diff
@@ -16,6 +18,8 @@
     from rich.console import Console
     from anton.core.backends.manager import ScratchpadManager
 
+CONNECTION_TEST_TIMEOUT_SECONDS = 30
```
tino097, line 25: This stacks on top of the scratchpad's own configurable timeout (~120s default) rather than passing `estimated_seconds` into it — worth double-checking this doesn't cut off legitimately slow-but-valid connections (cold warehouse spin-up, OAuth handshakes, tunneled DBs).

anton/tools.py
```diff
@@ -237,6 +241,18 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
                     f"in scratchpad code — never embed raw values."
                 )
 
+    # Below this point is mode (b), the interactive field-by-field prompt
+    # flow. It needs a real terminal (Rich prompts, Live spinners); a quiet
+    # console can't service it, so bail out with an actionable message
+    # instead of hanging or crashing on a prompt read.
+    if session._console is None:
```
tino097, line 253: When a scrub-marker placeholder gets dropped earlier in this function, the warning about it is printed to a `Console(quiet=True)` substitute when there's no real console, so it's discarded — the model just gets this generic message with no indication which field was rejected.

anton/commands/datasource/verify.py
```diff
@@ -57,7 +61,23 @@ async def run_connection_test(
 
             cell = None
```
tino097, line 72: Worth noting the real worst case here is more than 30s: a `ModuleNotFoundError` on attempts 1-2 triggers an uncapped `pad.install_packages(...)` each time before attempt 3 runs the (still up to 30s) connect.

anton/core/session.py
```diff
@@ -1100,6 +1100,19 @@ def __init__(self, config: ChatSessionConfig) -> None:
...
+            # connect_new_datasource's interactive mode needs a terminal to
+            # prompt in; without one, don't advertise it to the model (ENG-1849).
+            from anton.tools import (
+                CONNECT_DATASOURCE_TOOL,
+                CONNECT_DATASOURCE_TOOL_NO_CONSOLE,
+            )
+            self._extra_tools = [
```
tino097, line 1110: This matches by `tool.name` rather than identity, so a host-supplied tool that happens to share the name `connect_new_datasource` would be silently replaced. Not hit by any current caller, just a latent trap.

anton/tools.py
```diff
@@ -396,6 +418,32 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
...
+# Console-less hosts (Cowork desktop: session._console is None) can't service
+# mode (b) — there's no terminal to prompt in, and handle_connect_datasource
+# refuses it with an actionable error rather than crashing (ENG-1849). Don't
+# advertise a mode the model would then have to be told, mid-turn, doesn't
+# work here — swapped in by ChatSession.__init__ (anton/core/session.py)
+# once session._console is known.
+CONNECT_DATASOURCE_TOOL_NO_CONSOLE = dataclasses.replace(
+    CONNECT_DATASOURCE_TOOL,
+    description = (
+        "Connect a data source to Anton's Local Vault. Call this tool "
```
tino097: This paragraph is a near-duplicate of `CONNECT_DATASOURCE_TOOL`'s mode-(a) opening text above (line 372) — only the tail got extracted into `_CONNECT_DATASOURCE_DESCRIPTION_TAIL`. Might be worth pulling this one into a shared constant too so the two don't drift.

anton/tools.py
tino097: Fixed — added an `interactive` param to `run_connection_test` (default True, so the existing /connect and /edit paths are unaffected); mode (a) now passes `interactive=False`, which skips the retry-on-failure prompt entirely and fails closed on the first error instead of ever reaching `prompt_or_cancel`.

anton/commands/datasource/verify.py
tino097: Confirmed and fixed. Added a test with a `pad.execute` mock that mirrors `LocalScratchpadRuntime.execute_streaming`'s actual behavior (catches its own `CancelledError`, returns a Cell) — the existing test's plain `asyncio.sleep` mock propagates cancellation normally, so it never exercised this gap. Reworked the loop to run `pad.execute` as a Task, wait with a timeout via `asyncio.wait`, and on timeout cancel the task and always build our own Cell/message — regardless of what the task's own cancellation handling does with the exception.

anton/tools.py
tino097: Fixed — gave `CONNECT_DATASOURCE_TOOL_NO_CONSOLE` its own `input_schema` with `required: ["engine", "known_variables"]`, no longer sharing `CONNECT_DATASOURCE_TOOL`'s schema object by reference.

anton/core/session.py
tino097: Checked — `self._console` is never reassigned after `ChatSession.__init__` anywhere in this repo (only unrelated classes like `chat_ui.py`/`cli.py` reassign their own `_console`), and `_build_tools()` only registers `_extra_tools` once, gated behind `if not self.tool_registry:` — the same one-shot pattern `_build_core_tools()` already uses for its own live-state-derived text (e.g. the installed-package list in the scratchpad tool description). So there's no path today where this snapshot goes stale relative to a later live check. Leaving it as-is rather than adding machinery for a scenario nothing currently triggers — flag it again if a caller starts mutating `_console` post-construction.

anton/commands/datasource/verify.py
tino097: Added a comment above the constant clarifying scope: it bounds only the `pad.execute` connect attempt, not `install_packages` (which has its own separate `cell_install_timeout`, default 120s). Left the 30s value itself as-is since that's this PR's stated product call for the connect step — didn't see enough here to second-guess the number for legitimately slow connections without that context.

anton/tools.py
tino097: Fixed — the no-console bail-out message now names which fields were rejected as scrub-marker placeholders when that's why known_variables came in short, instead of only the generic 'no terminal to prompt in' text.

anton/commands/datasource/verify.py
tino097: Correct, and related to the comment above on the 120s stacking — `install_packages` already has its own bounded timeout (`cell_install_timeout`, default 120s) so this isn't literally unbounded, but the total for one `run_connection_test()` call can run well past 30s across two auto-installs plus three connect attempts. Documented via the comment above `CONNECTION_TEST_TIMEOUT_SECONDS` rather than changing the retry/install behavior, since that timeout is pre-existing and not introduced by this PR.

anton/core/session.py
tino097: Fixed — changed the match to `tool is CONNECT_DATASOURCE_TOOL` (identity) instead of `tool.name ==`.

anton/tools.py
tino097: Fixed — extracted the shared body into `_CONNECT_DATASOURCE_MODE_A_BODY`, used by both `CONNECT_DATASOURCE_TOOL` and `CONNECT_DATASOURCE_TOOL_NO_CONSOLE`.

PR #400, opened 2026-08-26
feat(anton): Turn-Key Data Vault for cloud OAuth connectors

martyna-mindsdb:
Linear ticket: https://linear.app/mindsdb/issue/ENG-208/mindshub-oauth-proxy-and-data-vault

## What was changed

- **Turn-Key Token Handoff** — new `TurnKeyDataVault` class (satisfying the same `DataVault` protocol `LocalDataVault` does) fetches a live connector token from `auth` using the turn's turn key, for cloud turns only. Wired into `build_cloud_chat_session()` in place of the previous `data_vault=None`. Desktop/local-mode turns are unaffected.

## Notes

- New ANTON_CLOUD_AUTH_BASE_URL env var defaults to auth.mindshub.ai, correct for production, but isn't set anywhere yet for other environments — needs to be explicitly set to the real auth hostname of whichever environment this gets tested in (e.g. staging's auth.staging.mindshub.ai, or a PR-environment's own per-instance auth URL).

TODO: set ANTON_CLOUD_AUTH_BASE_URL to the target environment's real auth hostname before testing there.

## Related PRs

- mindsdb/auth#327
- mindsdb/cowork-server#391
- mindsdb/cowork#730
- mindsdb/Mind-Castle#22
- mindsdb/scratchpad-controller#33

### DIFF
--- a/anton/cloud_turn/session.py
+++ b/anton/cloud_turn/session.py
@@ -2,14 +2,17 @@
 
 Assembles a :class:`~anton.core.session.ChatSession` scoped to the tenant's
 mounted workspace with the desktop-only, tenant-leaky behaviours OFF. It does
-NOT call the desktop ``build_chat_session`` (which loads workspace ``.env``,
-uses ``~/.anton`` personal memory, and injects vault creds into ``os.environ``).
+NOT call the desktop ``build_chat_session`` (which loads workspace ``.env``
+and uses ``~/.anton`` personal memory).
 
 Safety posture (all internal — nothing here is on the wire):
 
 * Trusted pod-side workspace mount, never taken from the request.
 * No dotenv loading (``AntonSettings(_env_file=None)``, shared into Workspace).
-* Connectors / data-vault / disk history OFF.
+* Data vault is a live :class:`~anton.core.datasources.data_vault.TurnKeyDataVault`
+  only when the turn carries an ``oauth`` block (Google Drive/Gmail via
+  auth's turn-key endpoint) — ``None`` (connectors OFF) otherwise. Disk
+  history stays OFF regardless.
 * Memory never persists pod-side: cowork sends the tenant's slots per turn, and
   writes are reported back for cowork to apply (see :func:`_build_cortex`).
 * Only reviewed, headless-safe tools are exposed (scratchpad + artifacts).
@@ -582,6 +585,20 @@ def build_cloud_chat_session(request: TurnRequestV1) -> "ChatSession":
 
     cortex = _build_cortex(request.memory, llm_client)
 
+    # Connectors ON only when this turn carries an oauth block (Google
+    # Drive/Gmail today). restore_namespaced_env() is the exact same
+    # function desktop's harness.py already calls for LocalDataVault: it
+    # clears+reinjects DS_* env vars and registers them for credential
+    # scrubbing (so a token can't leak into LLM-visible output) before the
+    # sandbox subprocess inherits this process's env via runtime_factory.
+    data_vault = None
+    if request.oauth:
+        from anton.core.datasources.data_vault import TurnKeyDataVault
+        from anton.utils.datasources import restore_namespaced_env
+
+        data_vault = TurnKeyDataVault(request.oauth)
+        restore_namespaced_env(data_vault)
+
     config = ChatSessionConfig(
         llm_client=llm_client,
         settings=settings,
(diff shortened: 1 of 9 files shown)


anton/core/datasources/data_vault.py
```diff
@@ -378,3 +382,150 @@ def next_connection_number(self, engine: str) -> int:
...
+#: cloud_turn/session.py). Lets a PR/staging pod point at a non-prod auth.
+ANTON_CLOUD_AUTH_BASE_URL_ENV = "ANTON_CLOUD_AUTH_BASE_URL"
+_DEFAULT_AUTH_BASE_URL = "https://auth.mindshub.ai"
+
+#: Non-secret fields the turn-key endpoint returns alongside access_token,
+#: same field names LocalDataVault already persists for these engines
+#: (cowork-server/cowork/services/connectors/oauth/google.py) — kept
+#: identical so scratchpad code sees the same DS_* var names on cloud and
+#: desktop.
+_TURNKEY_RESPONSE_FIELDS = ("access_token", "account_email", "token_type", "scope", "expires_at")
```
tino097, line 398: This allowlist drops `auth_type` and `_picked_files`, which `build_datasource_context()` needs to tell the agent about a connected Drive account or picked files. The picker feature is invisible to the agent in cloud turns even when credentials are injected correctly.

anton/core/datasources/data_vault.py
```diff
@@ -378,3 +382,150 @@ def next_connection_number(self, engine: str) -> int:
...
+        self._cache[cache_key] = fields
+        return fields
+
+    def _fetch_uncached(self, engine: str, name: str) -> dict[str, str] | None:
+        if not self._turn_key:
+            logger.warning("TurnKeyDataVault: no turn key available for %s/%s", engine, name)
+            return None
+        from anton.minds_client import minds_request  # local: avoid a module-load-time dep from core.datasources
+
+        url = f"{self._base_url}/v1/oauth/{engine}/token"
```
tino097, line 507: URL is built from `engine` only, no `name`. With two named connections for the same engine (e.g. two Drive accounts), both resolve to the same token, and the per-turn disabled-connection filter can be bypassed since `_fetch` never checks against `self._connections`.

anton/cloud_turn/session.py
```diff
@@ -582,6 +585,20 @@ def build_cloud_chat_session(request: TurnRequestV1) -> "ChatSession":
...
+    # clears+reinjects DS_* env vars and registers them for credential
+    # scrubbing (so a token can't leak into LLM-visible output) before the
+    # sandbox subprocess inherits this process's env via runtime_factory.
+    data_vault = None
+    if request.oauth:
+        from anton.core.datasources.data_vault import TurnKeyDataVault
+        from anton.utils.datasources import restore_namespaced_env
+
+        data_vault = TurnKeyDataVault(request.oauth)
+        restore_namespaced_env(data_vault)
```
tino097, line 600: `restore_namespaced_env()` now does blocking urllib HTTP per connection. `stream_turn()` in `__main__.py` starts the heartbeat task and then calls into this synchronously — the loop can't yield for the heartbeat until all these serial fetches finish (up to 15s each), so a turn with a couple of connectors can get killed by the controller as stalled.

anton/cloud_turn/session.py
```diff
@@ -582,6 +585,20 @@ def build_cloud_chat_session(request: TurnRequestV1) -> "ChatSession":
...
+    # Connectors ON only when this turn carries an oauth block (Google
+    # Drive/Gmail today). restore_namespaced_env() is the exact same
+    # function desktop's harness.py already calls for LocalDataVault: it
+    # clears+reinjects DS_* env vars and registers them for credential
+    # scrubbing (so a token can't leak into LLM-visible output) before the
+    # sandbox subprocess inherits this process's env via runtime_factory.
+    data_vault = None
+    if request.oauth:
+        from anton.core.datasources.data_vault import TurnKeyDataVault
+        from anton.utils.datasources import restore_namespaced_env
```
tino097, line 597: The OAuth `gmail` engine name collides with the existing local IMAP `gmail` `DatasourceRegistry` entry in `utils/datasources.py`. `restore_namespaced_env()` picks that entry's field schema, so the real OAuth fields (access_token etc.) aren't recognized as known secrets and fall through to coarse redaction.

tino097 (commented):
Inline notes from a review pass on the OAuth Proxy + Data Vault change set.

PR #401, opened 2026-08-26
Scope scratchpad subprocess env to the current data vault

tino097:
Linear: https://linear.app/mindsdb/issue/ENG-392/harden-env-variables-leaking-into-scratchpads

scope scratchpad subprocess env to the current conversation's datasource credentials instead of the full parent environment
stop leaking an unconfigured coding model or provider into scratchpads
let a caller override a pad's env before its next restart
fix the datasource connection test after the env scoping change
guard against one bad vault connection breaking env derivation for the rest

### DIFF
--- a/docs/superpowers/plans/2026-08-26-scratchpad-ds-env-scoping.md
+++ b/docs/superpowers/plans/2026-08-26-scratchpad-ds-env-scoping.md
@@ -1,0 +1,780 @@
+# Scratchpad DS_* Env Scoping Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop a scratchpad subprocess from inheriting `DS_*` datasource credentials it has no business seeing, by building its env from the current data vault instead of trusting whatever is sitting in the parent process's `os.environ`.
+
+**Architecture:** `LocalScratchpadRuntime.start()` gains an explicit `scratchpad_ds_env: dict[str, str] | None` — when set, it strips every inherited `DS_`-prefixed key from the copied parent env and overlays exactly this dict. `ScratchpadManager` computes that dict fresh, at the moment it creates a new pad, from a `data_vault` it's now handed. `ChatSession` passes its already-existing `config.data_vault` through. `None` anywhere in this chain (no vault) means today's behavior, unchanged.
+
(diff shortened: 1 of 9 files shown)


docs/superpowers/plans/2026-08-26-scratchpad-ds-env-scoping.md
```diff
@@ -0,0 +1,780 @@
+# Scratchpad DS_* Env Scoping Implementation Plan
```
pnewsam, line 1: Did you mean to include these?  Maybe better to leave out if they will be redundant with the actual code.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (9 files, +1216/−4) at head `fe1b01b7`: the DS_* strip-and-overlay in `LocalScratchpadRuntime.start()`, the vault-derived overlay in `ScratchpadManager`, the connection-test override path, and the session wiring. The change is fail-closed, keeps the no-vault path byte-for-byte unchanged, and its one load-bearing external assumption checks out — `cowork-server/harness.py:1048` really does pass a (filtered) `data_vault` into `ChatSessionConfig`, so the fix reaches the sidecar path it targets. No actionable defects; one optional note on the committed docs (inline).

### Validation

- Traced the overlay chain: `start()` strips every `DS_*` then overlays the explicit dict; `reset()` and `resume()` both route through `start()`, so an override applied via `set_scratchpad_ds_env()` takes effect on the next restart — matching what both `verify.py` callers do (`get_or_create(..., ds_env_override=…)` then `pad.reset()`).
- Confirmed fail-closed derivation: `_derive_ds_env()` returns `{}` (not `None`) when `list_connections()` raises, so a vault read error strips all `DS_*` rather than falling back to a full inherited copy; per-connection `try/except` isolates one bad record.
- Verified `manager.py`'s new top-level `import DataVault` introduces no cycle (`data_vault.py` imports only stdlib), and CI is green on this exact SHA — `run-tests` (4m30s), `Analyze (python)`, `build`, `CodeQL`, Snyk all pass. Did not re-run locally; CI covers this head.
- Checked the unconditional `env.pop("ANTON_SCRATCHPAD_MODEL"/"PROVIDER")`: it strengthens, and does not regress, the existing `test_get_llm_not_available_without_model` invariant.
- Confirmed the pip-install subprocess (`local.py:1286`) still copies the full env — but it runs `uv`/`pip`, not agent code, and is explicitly carried as a documented follow-up.

docs/superpowers/plans/2026-08-26-scratchpad-ds-env-scoping.md
tino097: i add them because the ENG-1382 @pnewsam but we can remove them

PR #406, opened 2026-08-26
ENG-1858: Record why the completion verifier produced no verdict on turn_completed

alecantu7:
Linear: [ENG-1858](https://linear.app/mindsdb/issue/ENG-1858/a-turn-can-end-because-the-completion-verifier-failed-and-nothing)

## What

`turn_completed` could say **that** the completion verifier failed — `ended_by=handback_verifier_failure`, or `verification_skipped=true` for a latched/denied turn — but nothing said **why**. This adds two content-free properties that surface the classification the verdict loop already computes.

## Why

Re-measured 2026-08-26 in PostHog 424726, 14 days, real turns only: **343 handbacks / 6,468 turns = 5.3%**, 50 installs, avg 514k tokens (3.1× a completed turn, ~176M total), still 6.3% on the newest build. Concentrated by model (nemotron 83/195 = 42.6%; `qwen3.5-122b` 23/26; `gpt-4o` 19/28) but with 59 properties on the event and zero about the verifier, none of it is groupable. The likely causes — model can't produce the forced tool call (ENG-1095), model narrates past the budget ladder (ENG-1081), provider blip (ENG-673) — produce byte-identical events today.

## How

The verdict loop in `session.py` already assigns `verdict_failure ∈ {truncated, transient, hard, denied}` per attempt for the latch, then drops it at end of scope. This PR:

- Adds `TurnCost.verifier_failure` and `TurnCost.verifier_error_type`, stamped at **every** no-verdict exit — one stamp after the attempt loop (covers denied latch, hard latch, failed re-probe, honest handback) plus one at the latched-skip site (`latched_hard` / `latched_denied`, no call was made so no exception type).
- `verifier_error_type` is `_safe_error_type(exc)` — the existing helper, class name only, never the message — with `StructuredOutputError` split into `:no_call` (model never produced the forced call, the nemotron suspicion) vs `:unusable_call` (call present, schema rejected it). Same class, opposite cures, so the bare type would merge them back into one PostHog row.
- Emits both next to `error_type`; `send_event(**extra)` passes new keys through untouched, so no sink change. Also added to the `turn_cost` log line and to the `analytics.py` key-list comment.

**Deliberately not done:** reusing `error_type` (ENG-1689). Its documented contract is "empty on handback"; overloading it would change its meaning by `ended_by` and silently alter existing queries.

## Before / after

| | before | after |
|---|---|---|
| handback event | `ended_by=handback_verifier_failure`, `error_type=''` | + `verifier_failure=hard`, `verifier_error_type=StructuredOutputError:no_call` |
| latched-skip event | `ended_by=completed`, `verification_skipped=true` | + `verifier_failure=latched_hard`, `verifier_error_type=''` |
| verified event | unchanged | + `verifier_failure=''`, `verifier_error_type=''` |
| user-visible behaviour | — | **unchanged** — instrumentation only |

After a week of traffic on a build carrying this, `GROUP BY planning_model, verifier_failure, verifier_error_type` partitions the population into budget problems / capability problems / provider noise, and the nemotron follow-up gets written from data.

## Tests

`tests/test_turn_cost_terminals.py` — 5 new tests (one per class: truncated, hard/no_call, hard/unusable_call, transient, and the verified-turn empty case) plus 3 existing tests extended (handback, denied, hard-latch/re-probe sequence). 31/31 green locally.

Mutation-verified, each watched failing:
- remove the post-loop stamp → 7 tests fail
- remove the latched-skip stamp → 1 test fails
- remove the `:no_call`/`:unusable_call` suffix → 3 tests fail

Ruff on the touched files: 47 findings before, 47 after (zero new).

## Security check

Done. The two new properties are a fixed-vocabulary string and an exception **class name**; the exception message (which can quote model output derived from the user's conversation) never leaves the process — the existing `_safe_error_type` helper is reused unchanged, and the `StructuredOutputError` suffix is derived from a bool. No new inputs, endpoints, or credentials. No raw model free-text reaches the analytics sink (ticket's explicit constraint).

## Delivery

anton-only. Reaches users via anton `main` → cowork-server's `anton-agent` git pin → the desktop train; web via the scratchpad image rebuild. Nothing to coordinate in other repos.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -534,6 +534,26 @@ def _safe_error_type(exc: BaseException) -> str:
         return "unavailable"
 
 
+def _verifier_error_type(exc: BaseException | None) -> str:
+    """`_safe_error_type` for the verdict call, with the one split it needs.
+
+    A non-truncated `StructuredOutputError` is two different failures that
+    ENG-1095 pulls apart: the model never produced the forced tool call at all
+    (`:no_call` — the shape a model without reliable tool-calling shows), or it
+    produced one the schema rejected (`:unusable_call` — fable's `{}` at 8
+    output tokens). Same class name, opposite cures, so the type alone would
+    merge them back into one PostHog row (ENG-1858). Content-free: the suffix
+    is derived from a boolean, never from the message. Cannot raise: the only
+    attribute read is a plain bool set in `StructuredOutputError.__init__`.
+    """
+    if exc is None:
+        return ""
+    name = _safe_error_type(exc)
+    if isinstance(exc, StructuredOutputError):
+        return name + (":unusable_call" if exc.reached_tool_call else ":no_call")
+    return name
+
+
 def _is_provider_auth_error(exc: BaseException) -> bool:
     """A provider-auth 401 — anton's "Invalid API key — …" copy from
     `openai.py`/`anthropic.py` (ENG-1310): the credential is wrong, not the
@@ -2689,12 +2709,13 @@ class ChatSession:
             _root_cause_fields = {}
         logger.info(
             "turn_cost session=%s turn=%d ended_by=%s verification_skipped=%s "
-            "tokens_total=%d "
+            "verifier_failure=%s verifier_error_type=%s tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
             "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d "
             "by_role=%s %s",
             self._session_id, turn_index, tc.ended_by,
-            str(tc.verification_skipped).lower(), tc.total_tokens,
+            str(tc.verification_skipped).lower(),
+            tc.verifier_failure, tc.verifier_error_type, tc.total_tokens,
             tc.input_tokens, tc.output_tokens, tc.cache_read_tokens,
             tc.cache_creation_tokens, tc.llm_calls, tc.rounds,
             tc.continuations, tc.peak_context_tokens, tc.duration_ms,
@@ -2808,6 +2829,12 @@ class ChatSession:
                 # ENG-1632) must be excludable from the honest-stop
                 # denominator without a per-conversation Langfuse hop.
                 verification_skipped=str(tc.verification_skipped).lower(),
+                # WHY the verifier produced no verdict (ENG-1858): the loop's
+                # own truncated/transient/hard/denied class plus the content-
+                # free exception type. Empty on verified turns. See
+                # `TurnCost.verifier_failure`.
+                verifier_failure=tc.verifier_failure,
+                verifier_error_type=tc.verifier_error_type,
                 tokens_total=str(tc.total_tokens),
                 input_tokens=str(tc.input_tokens),
                 output_tokens=str(tc.output_tokens),
@@ -5240,6 +5267,12 @@ class ChatSession:
                     # emit (late finalizers would see a later turn's state).
                     if self._turn_cost is not None:
                         self._turn_cost.verification_skipped = True
+                        # ...and WHY it was skipped (ENG-1858): no call was
+                        # made this turn, so there is no exception type — the
+                        # latch reason is the whole story.
+                        self._turn_cost.verifier_failure = (
+                            f"latched_{self._verifier_latch_reason or 'hard'}"
+                        )
                     break
                 _verifier_log.info(
                     "completion-verifier re-probing after %d skipped verifications",
@@ -5258,6 +5291,10 @@ class ChatSession:
             # latch: a blown budget is a tail sample of one model's verbosity,
             # an identical hard error twice is a capability problem (ENG-1155).
             verdict_failure: str | None = None
+            # The last attempt's exception, kept only long enough to stamp
+            # its TYPE on the turn books below (ENG-1858). `except ... as exc`
+            # unbinds `exc` at the end of each clause, hence the copy.
+            verdict_exc: BaseException | None = None
             for attempt, budget in enumerate(_VERIFIER_TOKEN_BUDGETS):
                 try:
                     verdict = await self._llm.generate_object_code(
@@ -5276,6 +5313,7 @@ class ChatSession:
                         _VERIFIER_TOKEN_BUDGETS
                     )
                     verdict_failure = "truncated" if exc.truncated else "hard"
+                    verdict_exc = exc
                     _verifier_log.info(
                         "completion-verifier verdict=%s budget=%d output_tokens=%d "
                         "stop_reason=%s retrying=%s",
@@ -5290,6 +5328,7 @@ class ChatSession:
                     # transient tuple). Retrying, diagnosing, or telling the
                     # user buys nothing: handled below by latching silently.
                     verdict_failure = "denied"
+                    verdict_exc = exc
                     _verifier_log.info(
                         "completion-verifier verdict=DENIED budget=%d error=%s",
                         budget, _safe_error_detail(exc),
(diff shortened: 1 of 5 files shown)


alecantu7 (commented):
Self-review (adversarial pass before requesting review). Two things a reviewer should see were weighed, plus what was verified:

- **Mutation-verified** all three stamp/derivation sites — post-loop stamp (7 tests fail without it), latched-skip stamp (1), `:no_call`/`:unusable_call` suffix (3).
- **Full suite locally**: 2,669 passed; the 5 failures in `test_cloud_turn_process.py` were a stale editable-install in the fresh worktree venv (`ModuleNotFoundError: anton` in the spawned subprocess) — pass after `uv sync --reinstall-package anton-agent`, and pass on clean `origin/staging` in a separate worktree. Not related to this diff.
- **Content-safety re-check**: every value emitted is either a literal from a fixed set (`truncated/transient/hard/denied/latched_*`) or `type(exc).__name__`. `str(exc)` / `repr(exc)` / `exc.args` are never touched on this path.

anton/core/session.py
```diff
@@ -5240,6 +5267,12 @@ async def _stream_and_handle_tools(
                     # emit (late finalizers would see a later turn's state).
                     if self._turn_cost is not None:
                         self._turn_cost.verification_skipped = True
+                        # ...and WHY it was skipped (ENG-1858): no call was
+                        # made this turn, so there is no exception type — the
+                        # latch reason is the whole story.
+                        self._turn_cost.verifier_failure = (
+                            f"latched_{self._verifier_latch_reason or 'hard'}"
```
alecantu7, line 5274: **Vocabulary note.** `latched_hard` / `latched_denied` are two values the ticket's `How` did not list (it said `""/truncated/transient/hard/denied`). Added deliberately: a latched skip made **no** verdict call, so reusing bare `hard`/`denied` here would make a skipped turn indistinguishable from a turn whose call actually failed once you filter on `verifier_failure` alone. Ticket description updated to match. Cardinality cost: +2 values.

anton/core/turn_cost.py
```diff
@@ -124,6 +124,25 @@ class TurnCost:
...
+    # state at emit (late finalizers would see a later turn's latch).
+    verifier_failure: str = ""
+    # Exception TYPE of the last verdict attempt — `_safe_error_type`, i.e.
+    # the class name only, never the message (which can quote model output
+    # derived from the user's conversation). `StructuredOutputError` carries
+    # a `:no_call` / `:unusable_call` suffix: the model never produced the
+    # forced call vs produced one the schema rejected — the split ENG-1095
+    # needs. Empty when no exception ended the verdict (verdict produced, or
+    # latched skip with no call made).
+    verifier_error_type: str = ""
```
alecantu7, line 145: **Query trap, documented rather than changed.** `verifier_error_type` is empty on latched skips (no call → no exception), so `GROUP BY verifier_error_type` alone **undercounts** hard failures — the latched turns show as `''`. The intended query pairs it with `verifier_failure`. Considered stamping the *original* failure's type onto later skipped turns and rejected it: that would require carrying exception state across turns on the session, exactly the read-live-latch-state-at-emit pattern the `verification_skipped` comment above warns against.

alecantu7 (commented):
## Adversarial self-review — no finding survived refutation (head `be685ea5`)

Ran the full `/pr-review` pass from a scratch worktree of the PR head against fresh `origin/staging` (`47fc554b`, local == remote). What was hunted and what killed each candidate:

| candidate | how it was refuted |
|---|---|
| Stamp reads live latch state at emit (the #309 late-finalizer trap) | Both stamps write onto `self._turn_cost` **inside the verdict loop**, during the owning turn; `_emit_turn_cost` emits the books it was handed and only the owner clears the slot (`is_owner = self._turn_cost is tc`). `TurnCost` is constructed fresh per turn at both entry sites (L3502, L3982), so no field carries over. |
| A `verification_skipped=True` site left unstamped | All four sites (L5242 latched-skip, L5368 denied, L5383 failed re-probe, L5415 hard-latch) are either the latched-skip stamp or downstream of the single post-loop stamp. Test `test_hard_latch_and_failed_reprobe…` asserts the values across all of them, including the re-probe. |
| Log-format arity broken by the two new `%s` | Exercised at runtime with `log_cli`: line renders `… verification_skipped=false verifier_failure=hard verifier_error_type=StructuredOutputError:no_call …`, no `--- Logging error ---`. |
| New keys filtered by the direct PostHog sink | `_posthog_body` forwards every `params` key except cache-buster/`action`/`timestamp`; `send_event(**extra)` is unfiltered; `root_cause_*` already arrives this way. |
| Another repo reads `turn_completed` properties and enumerates keys | Grepped cowork-server, scratchpad-controller, anton_services, cowork: none read `ended_by`/`verification_skipped`/`error_type`. cowork-server's `turn_completed` is the cloud-turn wire `kind`, unrelated. |
| Content leak into analytics | Emitted values are literals from a fixed set, `type(exc).__name__`, or `f"latched_{reason}"` with `reason ∈ {"", "denied", "hard"}` (all four writers of `_verifier_latch_reason` checked). `str(exc)`/`.args` never touched. `/security-review` could not run (needs cwd inside the repo; this session's cwd resets) — pass done by hand on the 76 logic lines. |
| Staging drift since the PR base | Only #404 (ENG-1796) landed; touches none of these files; `git merge-tree` reports 0 conflicts. |
| Tests pass for the wrong reason | `_verifier_fields` reads the last `send_event` kwargs by key — a missing key is a `KeyError`, not a pass. Each stamp site and the suffix derivation were mutated and watched fail (7 / 1 / 3 tests). |

Executed: `tests/test_turn_cost_terminals.py` + 7 related suites → **186 passed**; full suite earlier 2,669 passed (5 subprocess failures were a stale editable install in the worktree venv, pass after reinstall and on clean staging).

**Pre-existing, not this PR:** the verifier call produces no Langfuse observation even on gateway traffic (ticket comment). Separate ticket.

alecantu7:
Addressed the two low follow-ups from the adversarial review in `2bd95b4e`: `latched_denied` now has an assertion (mutation-verified — collapsing the skip-site stamp to the bare latch reason fails 2 tests), and `StructuredOutputError.reached_tool_call`'s docstring now states the flag is meaningful on non-truncated errors, which is what the `:no_call`/`:unusable_call` split relies on. No logic change.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed all 5 changed files — an instrumentation-only change that stamps two content-free properties (`verifier_failure`, `verifier_error_type`) on every no-verdict exit of the completion-verifier loop. The stamp sites cover the full no-verdict partition, the values are class-name/enum only, and every path is test-pinned. No actionable findings.

### Validation

- Ran `tests/test_turn_cost_terminals.py` at head `2bd95b4e` in a clean worktree: **31 passed** — matches the "31/31 green" claim, including the 5 new tests and 3 extended ones.
- Traced every no-verdict exit in `session.py` (denied latch, hard latch, failed re-probe, first-hard-failure honest handback, exhausted-truncated, transient): all reach the single post-loop stamp at `session.py:5383`; the pre-call latched-skip exit stamps `latched_<reason>` at `session.py:5268`. No no-verdict path is left unstamped, and a produced verdict (COMPLETE/WAITING/STUCK) correctly leaves both fields `""`.
- Confirmed content-free / the ticket's security constraint: `_verifier_error_type` returns `type(exc).__name__` plus a suffix derived from the `reached_tool_call` bool — never `str(exc)`. `reached_tool_call` is a plain bool always set with a `False` default in `StructuredOutputError.__init__`, so the helper cannot raise, as its docstring claims.
- Verified the new properties actually reach PostHog: `turn_completed ∈ _POSTHOG_EVENTS`, so `send_event` routes direct to `/capture/` and `_posthog_body` copies all params into `properties` with no allowlist — new keys pass through untouched, no sink change needed.
- `verdict_exc` copy is correct: Python unbinds `except … as exc` at clause end, so keeping the last attempt's exception requires the in-clause copy; it is overwritten per attempt, yielding the last attempt's type as intended.
- Confirmed the `provider.py` and `analytics.py` hunks are documentation-only, and the two new `turn_cost.py` fields default to `""` with no stale-across-turns risk (`TurnCost` is rebuilt per turn).

Nit (non-blocking): `verdict_failure or "hard"` at `session.py:5384` is a dead fallback — `verdict_failure` is always set whenever `verdict is None`. Harmless defensiveness; no change needed.

PR #410, opened 2026-08-28
ENG-1638: Answer "which model are you" from the served model, never from config

alecantu7:
Linear: [ENG-1638](https://linear.app/mindsdb/issue/ENG-1638) · Surface map: [Where the Model Gets Named](https://claude.ai/code/artifact/2ceffbfd-bb74-4d3e-b3f5-03cbf443fb80) · Sibling PR (independent, either merge order): cowork `alejandrocantu/eng-1638-desktop-model-label-consistency`

## What

The `RUNTIME IDENTITY` system-prompt block was built from **configuration** and then told the model it *"already knows"* its provider and model and must *"NEVER ask"*. Three production failure modes (all in the ticket, traces verified by hand):

1. A configured-but-unapplied local model (`gemma4:e4b`) reported as running while MindsHub served the turn — which is what made ENG-1634's silent gateway fallback invisible.
2. `mindshub_air` presented as a model name, and the model reasoning its way to a fabricated third-party name.
3. On the web pod the block rendered **empty** with the mandate intact → a Grok session answered *"No — I'm Anton, not Grok."*

One block was answering two questions. This PR splits them.

## How

**`anton/core/llm/identity.py` (new)** — the rule, mirroring how every other product surface already names a model (picker, billing, website all show the alias label, never the vendor model):

| Situation | Identity line |
|---|---|
| `mindshub_air` requested | **"MindsHub Air"**, always — plus: the model behind it is *not disclosed*; never name, guess, or **deny** it (craigk's "prove it" case). |
| Anything else, response seen | The **served** model from `LLMResponse.model` — MindsHub echoes the resolved id (verified live today: `mindshub_air` → `gpt-5.6-luna`); Ollama/LM Studio echo their own name. |
| Turn 1, no response yet | The requested id, labelled *"not yet confirmed"*. |
| Nothing known | *"say you cannot verify — do not guess, do not deny."* No mandate without data. |

**Configured LLM block** — `build_runtime_context` moved here (import-light so the pod can use it), re-exported from `anton.chat_session` because cowork-server imports it from there. Still says which provider/SDK code the agent writes should use, now labelled as such. `Workspace:` and `Memory mode:` dropped — neither served either question and the path leaked into traces.

**Plumbing** — `LLMResponse.model` captured at all six construction sites (OpenAI chat complete/stream, Responses complete/stream, Anthropic complete/stream); `LLMClient.last_served_model` records the planning role's; `ChatSession._build_system_prompt` renders the lines per turn; `cloud_turn/session.py` now injects the configured block like desktop does. `turn_cost.py`'s comment claiming the gateway doesn't echo the served model was wrong — corrected (the field there deliberately stays the requested alias for dollar math).

## Deviations from the ticket's Done-when, stated so QA can accept or reject

- **Item 4** said the MindsHub identity is the *catalog `label`* of the requested alias via `list_models`. I did **not** add a `/v1/models` fetch to the prompt path (a network call on every session build / every pod turn, with a new failure mode). Instead: `mindshub_air` → "MindsHub Air" via a one-entry `OPAQUE_ALIAS_LABELS` map, and every other MindsHub alias reports the **served** id from turn 2 (`grok` → `grok-4.6`, which is what the picker's label "Grok 4.6" means anyway). Turn 1 on a non-Air MindsHub alias therefore says `grok (not yet confirmed)` rather than "Grok 4.6". Reading `label` live is the eventual source and is noted in the module docstring.
- **Reach of the served-model line (corrected after adversarial review, commit 4e90b23):** it only appears where the `LLMClient` persists across turns — the **CLI**. cowork-server builds a fresh session + client every turn (`harness.py:_build_chat_session`) and the web pod is one process per turn, so on **desktop and web the line is always the requested alias** — `- Serving model: grok (the model requested for this conversation).` — which is what the picker shows. Honest, but unconfirmed; the wording no longer says "not yet confirmed". Seeding the served id across turns is the host's job and belongs with ENG-2050. Air is unaffected (rule keys on the requested alias).
- The identity lines live in the cache-stable prefix; on the CLI the prefix changes **once** per session (turn 1 requested → turn 2 served). Desktop/web and all Air sessions never change.

## Tests

`tests/test_runtime_identity.py` (new, 33 tests): the rule; sanitisation of `response.model` as untrusted prompt input (control chars stripped, 80-char cap, non-strings dropped); section rendering incl. the empty-context case; client capture on `plan`/`plan_stream` and *not* on the coding role; session rendering across turns through a real `LLMClient`; every provider construction site with SDK-shaped fakes. `test_cloud_turn_session.py` gains the pod-injects-configured-block test; `test_chat_context.py:277` (asserted the old *"NEVER ask the user which"*) updated to the new wording and asserts the mandate is gone.

**Mutation-verified: 8/8** mutations of the fix (Air branch, cannot-verify fallback, client capture, pod injection, each of the four provider capture sites) fail a named test.

Full suite locally: 2694 passed; the only failures are 5 `tests/test_cloud_turn_process.py` subprocess tests that fail **identically on pristine `origin/staging`** in this environment (stale-editable-install artefact, see ENG memory) — CI runs them on Linux.

## Security check

Done. `response.model` from a BYOK/local server is untrusted text entering the system prompt: interpolated as a plain string via `sanitize_model_name` (printable only, no newlines, ≤80 chars), never as instructions. The workspace path no longer reaches the prompt or traces. No credential, authz, or deserialisation surface touched. No gateway internals beyond the model id are exposed — and for Air, not even that.

## Delivery

anton `staging` → weekly `main` train → cowork-server bump. Nothing in cowork-server or scratchpad-controller needs to change: `response.model` is a response-side field, and `build_runtime_context` keeps its import path.

### DIFF
--- a/anton/core/llm/identity.py
+++ b/anton/core/llm/identity.py
@@ -60,16 +60,27 @@ OPAQUE_ALIAS_LABELS: dict[str, str] = {MINDSHUB_AIR_ALIAS: "MindsHub Air"}
 _MAX_MODEL_NAME_LEN = 80
 
 
+#: The deprecated pin prefix cowork-server still resolves (``latest:sonnet``).
+#: A stale ``latest:mindshub_air`` pin is overlaid onto settings verbatim, so
+#: without stripping it the Air rule would miss and the served vendor id would
+#: leak — the exact thing this module exists to prevent.
+_LEGACY_PIN_PREFIX = "latest:"
+
+
 def sanitize_model_name(value: object) -> str | None:
     """A model id safe to interpolate into the prompt, or ``None``.
 
     Non-strings (a Mock in tests, ``None`` from an SDK that omitted the field)
     are dropped rather than stringified. Newlines and control characters are
-    removed so the value cannot start a new prompt line; the length is capped.
+    removed so the value cannot start a new prompt line; the length is capped;
+    the deprecated ``latest:`` pin prefix is dropped so an alias compares (and
+    reads) the same however it was stored.
     """
     if not isinstance(value, str):
         return None
     cleaned = "".join(ch for ch in value if ch.isprintable() and ch not in "\r\n").strip()
+    if cleaned.lower().startswith(_LEGACY_PIN_PREFIX):
+        cleaned = cleaned[len(_LEGACY_PIN_PREFIX):].strip()
     if not cleaned:
         return None
     return cleaned[:_MAX_MODEL_NAME_LEN]
--- a/anton/core/llm/identity.py
+++ b/anton/core/llm/identity.py
@@ -1,0 +1,202 @@
+"""What the agent says when asked which model is serving the conversation.
+
+ENG-1638. The system prompt used to hand the model one ``RUNTIME IDENTITY``
+block built from *configuration* and then tell it "you already know what model
+you are running on — NEVER ask". Configuration is the right input for exactly
+one of the two questions that block answered:
+
+- **"What should code you write call?"** — the configured provider/SDK. Correct,
+  and kept (``build_runtime_context``).
+- **"What is answering me right now?"** — NOT configuration. ``mindshub_air`` is
+  a gateway alias, not a model; a mis-applied local endpoint (ENG-1634) leaves
+  the configured name pointing at a model that never ran; and on the web pod the
+  block rendered empty while the mandate stayed, so the model denied being the
+  model it was (a Grok session answering "No — I'm Anton, not Grok").
+
+This module owns the second question. The rule mirrors how every other surface
+in the product already names a model (the composer picker, billing, the
+website): the alias the user picked, never the vendor model underneath it.
+
+- ``mindshub_air`` → **"MindsHub Air"**, always. It is a product tier whose
+  backing model changes without notice (pricing page / FAQ), so the agent must
+  neither name nor deny what is underneath — it says the model is not
+  disclosed. Every other catalog alias is *named after* its model
+  ("Grok 4.6", "GPT 5.6 Luna"), so for those the served model is the honest
+  answer and reveals nothing the picker didn't.
+- Anything else → the model the provider **actually reported serving**
+  (``LLMResponse.model``, echoed by MindsHub, OpenAI, Anthropic, Ollama and
+  LM Studio alike), falling back to the requested id before the first response
+  arrives — labelled as unconfirmed, so a turn-1 answer is still honest.
+- Nothing known → the agent is told to say it cannot verify, and never to guess
+  or deny. The old prompt had no such fallback; that absence is what turned a
+  wrong input into a confident falsehood.
+
+Kept free of heavy imports so the cloud pod (``anton.cloud_turn``) can use it
+without pulling in the CLI's rich/typer stack.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    from anton.config.settings import AntonSettings
+
+#: The MindsHub free-tier alias. Same literal as ``minds_client.MINDS_FREE_TIER_MODEL``;
+#: duplicated rather than imported so this module stays import-light.
+MINDSHUB_AIR_ALIAS = "mindshub_air"
+
+#: Aliases the agent reports by their product label instead of the served model.
+#: Seeded from the catalog's ``label`` for the one alias whose label does NOT
+#: name the model underneath. Every other MindsHub alias is named after its
+#: model, so reporting the served id for those matches the picker already.
+#: Reading ``label`` live from ``/v1/models`` is the eventual source; a static
+#: map is enough while Air is the only opaque tier.
+OPAQUE_ALIAS_LABELS: dict[str, str] = {MINDSHUB_AIR_ALIAS: "MindsHub Air"}
+
+#: ``response.model`` from a BYOK / local server is untrusted text headed into
+#: the system prompt. Cap it and keep it on one line so it can carry a name and
+#: nothing else.
+_MAX_MODEL_NAME_LEN = 80
+
+
+def sanitize_model_name(value: object) -> str | None:
+    """A model id safe to interpolate into the prompt, or ``None``.
+
+    Non-strings (a Mock in tests, ``None`` from an SDK that omitted the field)
+    are dropped rather than stringified. Newlines and control characters are
+    removed so the value cannot start a new prompt line; the length is capped.
+    """
+    if not isinstance(value, str):
+        return None
+    cleaned = "".join(ch for ch in value if ch.isprintable() and ch not in "\r\n").strip()
+    if not cleaned:
+        return None
+    return cleaned[:_MAX_MODEL_NAME_LEN]
+
+
+def serving_model_lines(
+    *, requested: object, served: object
+) -> list[str]:
+    """Prompt lines describing the model serving this conversation.
+
+    ``requested`` is the planning model id anton asked for (``settings.planning_model``);
+    ``served`` is what the provider reported back on the last planning response
+    (``LLMResponse.model``), ``None`` before the first response.
+
+    Returns ``[]`` when neither is known — the caller then emits the
+    cannot-verify fallback instead of a mandate.
+    """
+    req = sanitize_model_name(requested)
+    srv = sanitize_model_name(served)
+
+    if req in OPAQUE_ALIAS_LABELS:
+        label = OPAQUE_ALIAS_LABELS[req]
+        return [
+            f"- Serving model: {label} (MindsDB's hosted model tier).",
+            f"- MindsHub does not disclose which model serves {label}. If asked what "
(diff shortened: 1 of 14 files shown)


anton/core/llm/identity.py
```diff
@@ -0,0 +1,202 @@
...
+    ``served`` is what the provider reported back on the last planning response
+    (``LLMResponse.model``), ``None`` before the first response.
+
+    Returns ``[]`` when neither is known — the caller then emits the
+    cannot-verify fallback instead of a mandate.
+    """
+    req = sanitize_model_name(requested)
+    srv = sanitize_model_name(served)
+
+    if req in OPAQUE_ALIAS_LABELS:
```
alecantu7, line 109: **Finding:** cowork-server still resolves the deprecated `latest:` prefix (`app_settings.py:23` — *"the older latest: prefix still resolves but is deprecated"*), and `_overlay_user_settings` copies the stored string onto `AntonSettings` verbatim. So a stale pin of `latest:mindshub_air` reaches this comparison, misses the opaque map, and the agent would report the served id (`gpt-5.6-luna`) for an Air session — the exact leak this ticket exists to prevent. Fix: normalise a leading `latest:` before the lookup (and in the requested-only fallback line, so the user never sees the prefix). Test added.

alecantu7 (commented):
Adversarial self-review (convention 6) before requesting human review. Checked: every `LLMResponse(` construction site carries `model` (6/6); the re-export path cowork-server imports still resolves; the pod's config test proves the block is injected; mutation run 8/8. Findings:

1. **Legacy `latest:` prefix bypasses the Air rule** — see inline comment on `identity.py:93`. Fixed in the next commit.
2. **Cache prefix churn** — the identity lines live in the cache-stable prefix, so a non-Air session busts the prompt cache exactly once (turn 1 requested → turn 2 served). Accepted deliberately: a moving alias re-resolving mid-session is rare, Air (the default and most common) never changes, and putting it in the volatile tail would weaken an instruction the model must weigh above its own priors. Documented at the call site in `session.py`.
3. **Pod with no `request.model`** — `build_runtime_context` would then print AntonSettings' default id. Not reachable from cowork-server (the producer always sends `resolved_planning_model`), and the identity line uses the same `planning_model` the client actually requests, so the two can't disagree. No change.
4. **Local test-suite failures** — 5 `test_cloud_turn_process.py` subprocess tests fail identically on pristine `origin/staging` in this macOS env; not caused here. CI on Linux is the arbiter.

alecantu7:
Self-review finding 1 (legacy `latest:` prefix bypassing the Air rule) addressed in f9b6bbfd: `sanitize_model_name` strips the prefix; two tests added and mutation-checked. Ready for human review — please don't merge without one.

alecantu7 (commented):
**Adversarial review, full pass (`/pr-review`), validated at `f9b6bbfd6c84be8462b5c7c726c0df78a80b7bbc`.** Fresh clone of the PR head, base verified against GitHub (`staging` = 8791894, unchanged since branch). Ran: 304 tests across the touched suites (all pass; CI `run-tests` also green); workspace-wide grep of every changed symbol across anton, cowork-server and scratchpad-controller; an executable repro of the desktop host shape; a hand security pass on the one input surface (`/security-review` cannot run in this session — the working directory is not a repo).

**Risk: low → merge-safe, with one confirmed finding that is a false claim rather than a wrong behaviour.** Details inline.

**Security (hand pass):** `sanitize_model_name` strips U+2028/U+2029/NEL/CRLF/tab/zero-width/RTL-override (probed each); `{}` in a served name survives only as literal text because the section is a `.format()` *value*; 80-char cap holds; workspace path no longer reaches the prompt; no new logging or secret handling. Residual: a local server can still put words in its own model name (`# SYSTEM: …`) — it lands mid-line inside `- Serving model: <X> (as reported …)`, cannot open a new line, and is the same trust a user already extends to a server they configured. Accepted.

**Cross-repo:** `build_runtime_context` import path preserved (cowork-server `harness.py:741`); `SystemPromptContext` shape unchanged (`probe.py:456` passes `runtime_context=""` and now gets the cannot-verify line — harmless, the probe never asks identity); the extra `LLMResponse` field is additive (cowork-server `producer.py:222` constructs it by kwargs). No version-skew hazard: nothing outside anton calls the new `build()` kwarg.

**Second pass — assume each test passes for the wrong reason:** the cross-turn session test (`test_served_model_flows_from_response_into_next_turn_prompt`) reuses ONE `LLMClient` across turns. That is the CLI's shape and exactly why it hid the finding below.

anton/core/llm/identity.py
```diff
@@ -0,0 +1,213 @@
...
+            f"- MindsHub does not disclose which model serves {label}. If asked what "
+            f"is underneath, say it is not disclosed — never name, guess, or deny an "
+            f"underlying vendor model.",
+        ]
+    if srv:
+        return [f"- Serving model: {srv} (as reported by the provider on its last response)."]
+    if req:
+        return [
+            f"- Serving model: {req} (the model that was requested; the provider has "
+            f"not yet confirmed it)."
```
alecantu7: **Finding — confirmed by execution · medium · introduced by this PR (as a claim, not a regression).**

The PR body and Done-when item 5 say the served model is reported "from the second turn on". That is only true where the `ChatSession`/`LLMClient` persists across turns — the CLI. On the two shipping hosts it never happens:

- **desktop**: cowork-server `stream_response` calls `_build_chat_session` per turn (`harness.py:487`), which builds a fresh client (`harness.py:855`), so `last_served_model` is `None` at every prompt build;
- **web**: one pod process per turn, same result.

Repro (fresh client + fresh session per turn, history carried over, provider returning `model="grok-4.6"`): all three turns render `- Serving model: grok (the model that was requested; the provider has not yet confirmed it).` — the served id never appears.

**Impact:** not a wrong answer — the alias IS what the picker shows, and Air (the default) is unaffected — but (a) the steady-state wording "not yet confirmed" promises a confirmation that will never come on desktop/web, and (b) the PR/ticket overclaim a mechanism whose reach is CLI-only.

**Refutation attempted:** looked for a session/client cache keyed by conversation in cowork-server (none: grep `cache|_clients|lru` in harness.py), and for a per-round prompt rebuild inside a turn that would pick up round-1's served model (none: `_build_system_prompt` runs once per turn at `session.py:3546`; the retry path at 4238 rebuilds but the tool loop reuses `system`).

**Smallest fix (this PR):** reword the requested-only line to something true in steady state — e.g. `- Serving model: grok (the model requested for this conversation).` — and add a host-shaped test (new client per turn) that pins that line; correct the PR body and ticket item 5 to "CLI; desktop/web need the host to seed it". **Real fix (follow-up):** the host records the served model per conversation and hands it to the next turn — ENG-2050's territory (it already wants the server to own the served model), so link it there rather than a new ticket.

alecantu7:
Adversarial-review finding (served-model line never reaches the prompt on desktop/web because the client is rebuilt per turn) addressed in 4e90b23: steady-state wording corrected, host-shaped test added (new client per turn × 3, line pinned), docstring states the reach, PR body corrected. The mechanism itself stays — it is real on the CLI and becomes real on desktop/web once the host seeds the served id (ENG-2050). Ready for human review; please don't merge without one.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

Reviewed the full diff: the new `identity.py` rule, `LLMResponse.model` capture across all six provider construction sites, `LLMClient.last_served_model`, the session/pod wiring, and the prompt-template split. Traced it against ENG-1638's Done-when and against the one contract this repo's CI can't see — cowork-server's cross-repo use of `build_runtime_context`. Merge-ready; no actionable findings.

### Validation

- CI green: `run-tests` (4m31s), `verdict-eval`, `build`, CodeQL, Snyk.
- Cross-repo contract preserved. `cowork-server .../harness.py:1067` builds `SystemPromptContext(runtime_context=build_runtime_context(anton_settings), …)`. The re-export from `anton.chat_session` and the string return keep that working with no change. That site builds a fresh `llm_client` per turn, which confirms the documented "desktop reports the requested alias, not the served id" reach.
- Pod wiring valid: `ChatSessionConfig.system_prompt_context` is a real field (`session.py:861`), staging did not previously pass it in `build_cloud_chat_session` (no duplicate kwarg), and `SystemPromptContext` is importable from `anton.core.session`.
- Sanitization sound: `sanitize_model_name` drops non-strings, strips control chars and newlines so untrusted `response.model` cannot open a new prompt line, caps at 80, and strips the `latest:` pin so a stale pin cannot bypass the Air rule. Unicode line separators are non-`isprintable()` and also removed.
- Logic traced: the opaque-alias rule keys on the requested id, so Air always wins over any served value and the vendor id cannot leak; served wins over requested otherwise; empty yields the cannot-verify fallback with no mandate. Backed by the 33 mutation-verified tests.

One item for QA to consciously accept, not a defect: on desktop/web a non-Air MindsHub alias reports the raw requested id (`grok`) rather than the picker's catalog label ("Grok 4.6"). This is the stated deviation from Done-when item 4, deferred to ENG-2050, and flagged by the author in the two self-review commits.

PR #415, opened 2026-08-31
ENG-1957 FIX: Reusing a vaulted password for a second connection 

ea-rus:
**Problem:**
If you'd already saved a password for one connection and reused it while setting up a second one, saving the second connection would just silently fail — the app treated the reused password as fake and dropped it.

**Fix:**
Now it recognizes when you're referencing a credential you already saved and reuses it automatically, without ever exposing the actual password value anywhere it shouldn't be.

Fixes: https://linear.app/mindsdb/issue/ENG-1957/reusing-an-already-vaulted-password-for-a-second-connection-silently

### DIFF
--- a/anton/tools.py
+++ b/anton/tools.py
@@ -74,18 +74,20 @@ async def handle_connect_datasource(session: ChatSession, tc_input: dict) -> str
     from rich.console import Console
     console = session._console or Console(quiet=True)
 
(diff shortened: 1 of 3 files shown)


ea-rus:

tino097 (approved):

PR #419, opened 2026-08-31
ci(scratchpad): run both pod entrypoints before the image can be pushed (ENG-2126)

lucas-koontz:
Preventive work from the 2026-08-31 outage. Ticket: [ENG-2126](https://linear.app/mindsdb/issue/ENG-2126).

## Why

Nothing had ever executed this image's entrypoints outside a pod. The build proved the image was **assembled**; nothing proved it **ran**. So the first evidence that pods could not serve a turn was prod going quiet, and there was no earlier signal anywhere to read.

To be explicit about scope: **this would not have prevented the 2026-08-31 outage.** That was a bad image reference, so the container never started and no image-level check could fire. It closes the adjacent hole, which the investigation kept walking into: an image that pulls fine, starts fine, and cannot serve a turn.

## What it does

The Dockerfile's last layer runs `docker/image_smoke.py` as UID 1000, below `USER 1000`, so every check sees exactly what a scratchpad pod sees: same venv, same interpreter, same uid.

It **runs** both entrypoints the controller execs, with empty stdin. An import proves the module resolves; only running it proves the whole graph loads and the wire still speaks the contract the controller reads.

| Check | What it catches |
|---|---|
| `python -m anton.cloud_turn` exits 0 with exactly one JSON line on the protocol descriptor, and that line is a terminal event | An import break anywhere in the turn graph. Also the worst case: exit 0 with nothing printed. The controller reads the terminal off the **event**, not the exit code, so that is a turn which hangs until the stall timer fires rather than one that fails |
| stdout carries exactly one line | Anything else writing to FD 1 corrupts the stream the controller parses. This is why the entrypoint isolates FD 1 at all |
| `python -m anton.core.backends.scratchpad_boot` exits 0 | The single-cell entrypoint. It has no main guard (the module body **is** the cell loop), so running it is the only way to load it. EOF ends the loop with no cell executed |
| reported version equals the one the workflow resolved, and is not `2.0.0*` | A wrong version here is invisible in the cluster: well-formed, reported by every pod, reads as a real cohort rather than a null (ENG-1796) |
| uv executable at `/usr/local/bin/uv` | `scratchpad_boot` shells out to `uv pip install` for a package a cell imports. `ANTON_UV_PATH` names that exact path |
| venv writable by uid 1000 | Same install. A chown that stops covering the venv gives an image that serves turns normally right up until a cell imports something missing |

## Why in the build, not in a step below it

A failure blocks the **push**. An image no pod can serve a turn with never reaches ECR to be pinned by mistake.

It also needs no docker daemon on the `mdb-dev` runner. `build-push-ecr` targets the in-cluster buildkit agent, so the image is never in the runner's local daemon and a `docker run` step would have to pull it back first.

Cost: about 1 second.

## Verified both ways

Against a real `docker build`, not by inspection.

- Good image: `image smoke passed`, build succeeds.
- Startup-path dependency removed (`rm -rf .../site-packages/dill*`): build **fails** at that layer with `ModuleNotFoundError: No module named 'dill'` in the log.

Worth recording from that exercise: removing `httpx` did **not** fail the smoke, because `anthropic` and `openai` import it lazily when a client is constructed. That independently re-confirms the httpx theory considered during the investigation was never the cause. A pod missing httpx starts fine.

## Tests

`tests/test_image_smoke.py`, 9 tests. A gate that reports success no matter what is worse than no gate, so the smoke's own failure modes are pinned: silent exit, polluted stdout, non-terminal event, a check that raises (must be reported, not swallowed), one broken check not hiding the others, and the layer's position below `USER 1000`.

That last one matters: above `USER 1000` every check runs as root and passes for an image the pod still cannot use.

## Note

This PR triggers `scratchpad-dev-build.yml`, so CI here is itself the gate running against a real image build.

### DIFF
--- a/.github/workflows/scratchpad-dev-build.yml
+++ b/.github/workflows/scratchpad-dev-build.yml
@@ -6,6 +6,11 @@ name: Scratchpad image - Dev build on PR
 # rebuilds. The scratchpad-controller references the resulting tag via
 # SCRATCHPAD_CONTROLLER__SCRATCHPAD_IMAGE; there is no Helm chart for this image, so this
 # workflow only builds and scans (no deploy job).
+#
+# The image is smoke-tested inside the build, not here: the Dockerfile's last layer runs
+# docker/image_smoke.py as UID 1000, which executes both entrypoints the controller execs.
+# It sits in the build rather than in a step below it so a failure blocks the PUSH — an
+# image no pod can serve a turn with never reaches ECR to be pinned by mistake.
 
 on:
   pull_request:
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,5 +69,11 @@ RUN printf '#!/bin/sh\nexec python -m anton.core.backends.scratchpad_boot\n' \
 RUN useradd -u 1000 -m -s /bin/sh scratchpad
 USER 1000
 
+# Run both entrypoints once, as the pod's uid, before this image can be pushed.
+# The build is otherwise green for an image no pod can serve a turn with: every
+# check up to here proves the image was ASSEMBLED, none proves it RUNS. See
+# docker/image_smoke.py for what each check is guarding against.
+RUN python /app/docker/image_smoke.py
+
 # The controller always execs an explicit command; this default keeps the image runnable standalone.
 CMD ["python", "-m", "anton.cloud_turn"]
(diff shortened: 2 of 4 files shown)


lucas-koontz:

PR #423, opened 2026-09-01
[ENG-2193] Tell the user the work finished when only the completion check failed

tino097:

the verifier failure and latch-state lines now log at WARNING, so a default-level log names the failure class and exception type
a truncation the larger token budget retries stays at INFO, and the per-turn latched-skip, re-probe and successful-verdict lines stay at INFO
the verifier-failure handback now leads with the work having finished and the reply standing, instead of announcing an internal error
the handback forbids naming a path that is not verbatim in a tool result, so the model stops restating the path it meant to write
the progress line reads "Confirming what was completed..." instead of "Something went wrong — checking in with you..."
tests re-anchor the six pinned prompt strings on two shared constants, so the denied-verdict negative assertions cannot pass vacuously after a reword
new coverage for the handback wording and for the per-attempt log levels
pairs with mindsdb/cowork-server#429, which lets LOG_LEVEL in the desktop config file reach logging

### DIFF
--- a/anton/cli.py
+++ b/anton/cli.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import concurrent.futures
 import importlib
+import logging
 import os
 import shutil
 import subprocess
@@ -38,6 +39,12 @@ from anton.commands.datasource import (
     handle_test_datasource
 )
 from anton.minds_client import minds_v1_base, resolve_and_probe, test_llm
+
+# The CLI configures no logging, so without a handler on the package logger
+# Python's lastResort writes WARNING and above straight to stderr — into the
+# middle of the rich Live render. Propagation is untouched, so a host that does
+# configure logging (the cowork sidecar, the cloud pod) still emits them.
+logging.getLogger("anton").addHandler(logging.NullHandler())
 
 
 def _build_scratchpad_manager(
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -5491,7 +5491,10 @@ class ChatSession:
                     )
                     verdict_failure = "truncated" if exc.truncated else "hard"
                     verdict_exc = exc
-                    _verifier_log.info(
+                    # WARNING only when this attempt ends the loop: a truncation
+                    # the larger budget recovers from is not a failure to report.
+                    _verifier_log.log(
+                        _logging.INFO if retrying else _logging.WARNING,
                         "completion-verifier verdict=%s budget=%d output_tokens=%d "
                         "stop_reason=%s retrying=%s",
                         truncation_verdict(exc),
@@ -5506,7 +5509,7 @@ class ChatSession:
                     # user buys nothing: handled below by latching silently.
                     verdict_failure = "denied"
                     verdict_exc = exc
-                    _verifier_log.info(
+                    _verifier_log.warning(
                         "completion-verifier verdict=DENIED budget=%d error=%s",
                         budget, _safe_error_detail(exc),
                     )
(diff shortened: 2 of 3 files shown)


tino097:

tino097:
**Self review**

Status: this covers local commit `179e06f5`, which is not on the branch at the time of writing (pushed head is `b1cdfb23`). If you are reading this before that push, the Codex comment above is already fixed locally.

**What this changes**

1. The verifier failure and latch-state lines log at WARNING, so a default-level log can carry them. The desktop sidecar runs at WARNING and tees stdout and stderr into `cowork-server.log`, so today a customer's log has uvicorn access lines and nothing else.
2. The per-attempt line now carries `error=` (exception type only, via `_safe_error_detail`) and the two terminal lines carry `model=`. Without those, the log names a verdict label but never the exception type or the verdict model, which is the actual cause on a free-tier account.
3. The hand-back prompt no longer reads as a failed turn, and forbids naming a path that is not verbatim in a tool result. The reported bug was the model restating the path it *intended* to write, which does not exist.

**On the Codex finding above: it is correct, and it was the worst thing in the first version.**

Only the tool-round count gates this path, so a turn whose tool calls failed reaches the same hand-back. The first version asserted "nothing about the work itself went wrong and there is no error for the user to fix", which is this ticket's own misreport with the sign flipped. The prompt now claims only what is knowable, that the reply stands and nothing was rejected because no verdict came back, and instruction 1 asks for an accurate account including any tool that returned an error.

**Four judgment calls worth pushing back on**

1. `verdict=DENIED` at WARNING is one line per message per wallet-empty user, fleet-wide: prod and staging run `LOG_LEVEL=WARNING` and Cowork rebuilds the session per message. I moved the paired "latched after a deterministic denial" line back to INFO to halve that. If it is still too noisy, `DENIED` is the next one to drop, though the ticket did ask for it.
2. The CLI `NullHandler` also silences the pre-existing "diagnosis returned no content" WARNING for CLI users. That line was reaching them through the same accidental `lastResort` channel, into the middle of a `rich` Live render, so I treated this as containment rather than a loss. Verified: with no logging configured nothing prints, and with a root handler the line is emitted unchanged.
3. Instruction 2 permits the model to print the artifact store's real internal path (`.anton/artifacts/<slug-id>/...`). Acceptance criterion 2 wants the location to come from the artifact card, which is a UI concern belonging to cowork's Reveal-folder sub-task. Inside this PR, a correct internal path beats an invented one or silence.
4. Prompt wording cannot be verified by tests. The tests pin strings, not model behaviour, and no eval was run against the new prompt. If there is an eval harness for hand-back turns, this is the change that would benefit from it.

**Checked and clean**

- No content leak from the level raise. Every raised line carries counters, budget, `stop_reason`, static text, or `_safe_error_detail`, which emits the exception type plus a status code or pydantic `loc`/`type`, never `str(exc)`.
- `model=` cannot raise: every production path builds a real `LLMClient` (cowork-server `providers.py`, `cloud_turn/session.py`, the CLI), which has the public `coding_model` property.
- Nothing in cowork-server parses these log lines, so added fields are safe.
- No cross-repo contract touched. `ended_by`, `verifier_failure`, `verifier_error_type` and `verification_skipped` are unchanged.
- Nothing imports `anton.cli` from cowork-server, so the `NullHandler` is confined to the CLI process.

**Checks run**

Full suite 2802 passed, 31 skipped at `179e06f5`. Targeted verifier files 71 passed. CLI containment probe as described above.

**Not in this PR, still open on the ticket**: verifier model resolution (the root cause), cross-message latch persistence, the `delegate` tool-history leak, and cowork's Reveal project folder.

PR #424, opened 2026-09-01
Revert "Yolo mode: edit artifact files with a diff" (#422)

torrmal:
Reverts fa8f8533 (#422), which was merged prematurely.

#422's own description says the agent wiring — the `edit_artifact` tool, the progress bridge and provenance — lands in later commits. Those belong in the same PR, so the feature arrives reviewable as a whole rather than as dead code plus a promise.

**Nothing depended on it.** No production module imported `anton.core.yolo` and no tool was registered, so this removes 2872 lines of inert code and changes no behaviour.

```
before revert: 2848 passed, 31 skipped, 2 failed (pre-existing)
after revert:  2761 passed, 31 skipped, 2 failed (pre-existing)
                ^^^^ exactly the 87 yolo tests, nothing else
```

The work returns as one branch carrying both the library and the wiring. Design and rationale stay in [ENG-2186](https://linear.app/mindsdb/issue/ENG-2186/yolo-mode-edit-artifacts-with-diffs-and-let-the-scratchpad-own-the).


torrmal:
Not needed — leaving #422 on staging. It is inert (no production module imports `anton.core.yolo`, no tool registered), so it changes no behaviour and costs nothing sitting there. The agent wiring lands as a follow-up PR instead.

PR #425, opened 2026-09-01
Plug yolo mode into the agent as edit_artifact (ENG-2186)

torrmal:
Follow-up to #422, which added the library and wired it to nothing. This connects it.

## What the agent can now do

> **User:** rename the title to TicTacTris

```
edit_artifact(slug="tictactris", task="rename the title to TicTacTris")

  choosing what to read...
  reading index.html
  plan: Retitled the app to TicTacTris
  wrote index.html

→ Applied: Retitled the app to TicTacTris
  Files: index.html
```

Two model calls. Previously this was the scratchpad writing a Python program to perform string surgery on the file — a second thing that can be wrong, and a subtly wrong one mangles the file rather than leaving it alone.

## The handoff when it does not work

```
Could not apply the change after 3 attempts. Nothing was written and
the artifact is unchanged.

could not find this hunk's lines in the file:
  <title>WRONG</title>

Make this change with the scratchpad instead: read the file, modify it
in Python, and write it back.
```

A **message, not orchestration**. The model already has the scratchpad and knows how to drive it, so the smallest thing that works is to say plainly that nothing was written and what to do next — and to send the diagnosis along with it.

## Streaming

A run makes two model calls, so without progress it is a silent pause. The awkward part is a direction mismatch: `YoloEditor` reports by *calling* a `Progress` object synchronously from inside the work, while a streaming handler reports by *yielding*. `tools/yolo_bridge.py` puts a queue between them and drains it alongside the work, waiting on both at once so a line appears when it is written.

Awaiting the edit and replaying its log afterwards would have produced exactly the pause the progress exists to prevent.

## Also in here

- **`ChatSession.llm_client`** — a property, so a handler does not reach into `_llm`. Everything a handler runs uses the same provider, coding/planning split and tracing as the rest of the turn.
- **`_track_artifact(..., files=...)`** — yolo knows which files it touched. Existing callers pass nothing and keep their behaviour.
- **`rescan_files` after a successful edit** — yolo writes to the folder directly, so `metadata.json` and the README go stale unless the store is told, and nothing complains when they do.
- **`missing_task` and the `yolo_*` outcomes registered in `root_cause`.** All `TIER_SELF`: nothing in the environment is missing, and a diff that would not apply is something the agent recovers from itself. The `yolo_*` ones are built as an f-string so the tree scan cannot see them — listed explicitly so the runtime classifier does not bucket an ordinary fallback as `unclassified`.

## Testing

**11 new tests**, on the seams rather than the engine — the engine has its own 87 from #422.

| | |
|---|---|
| streaming | progress lines arrive, in order, while the work runs |
| provenance | `_artifacts_touched`, `record_turn`, and `files_touched` all correct |
| a new file | `*** Add File:` shows up in `metadata.json` |
| the handoff | nothing written, artifact byte-identical, diagnosis + scratchpad instruction present |
| a failed run | records **no** turn — provenance says which turns changed the artifact, and one that changed nothing did not |
| bad arguments | surface as `missing_slug` / `missing_task`, which tier as self-inflicted |
| a provider crash | propagates rather than returning a silently empty result |

Full suite: **2859 passed**, 31 skipped, 2 failures in `test_chat_scratchpad.py` pre-existing on `staging`.

Phase 2 (the scratchpad's `emit_data` and the `reconcile()` hook) stays open on [ENG-2186](https://linear.app/mindsdb/issue/ENG-2186/yolo-mode-edit-artifacts-with-diffs-and-let-the-scratchpad-own-the).

### DIFF
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -117,6 +117,7 @@ _SENTINEL_REASONS = {
     "scratchpad_missing_name": (TIER_SELF, "missing_name"),
     "missing_name": (TIER_SELF, "missing_argument"),
     "missing_slug": (TIER_SELF, "missing_argument"),
+    "missing_task": (TIER_SELF, "missing_argument"),
     "missing_description": (TIER_SELF, "missing_argument"),
     "invalid_type": (TIER_SELF, "invalid_argument"),
     "invalid_port": (TIER_SELF, "invalid_argument"),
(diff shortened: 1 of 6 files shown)


lucas-koontz:
recheck

PR #426, opened 2026-09-01
[ENG-2193] Count an exhausted verdict ladder toward the completion-verifier latch

tino097:

## Background, if you don't work in this file

After a turn that used tools, anton runs a **completion check**. It asks the cheap coding model one question, "is this task actually finished?", and requires the answer as a structured tool call. If the answer is no, anton keeps working. If the check fails to produce an answer at all, the user gets a message saying an internal error stopped verification.

Three terms used below:

- **ladder** — if the answer gets cut off before the model reaches the tool call, anton retries once with a bigger token budget. 2048 first, then 4096. An *exhausted ladder* means both budgets were spent and no answer came back.
- **latch** — some models fail this check every single time. Rather than interrupt the user on every turn, anton gives up after two failures and stops running the check for a while.
- **re-probe** — the latch is not permanent. Every N turns anton tries the check once more, in case the user switched models or a gateway fix shipped.

## What changed

- **An exhausted ladder now counts toward the latch.** Previously only a provider that rejected the call outright did. This is the bug: the counter was gated on one failure type, so a model that kept getting cut off never latched at all, in any session.
- Renamed `_verifier_hard_failures` to `_verifier_no_verdict_failures`, since it no longer counts only that one type.
- **The latch now records which failures caused it**, so analytics can tell `latched_truncated` from `latched_hard` instead of labelling everything the same. Both new values are documented in `turn_cost.py` and the PostHog register.
- **A latch caused by cut-off answers re-tries after 3 turns instead of 10.** A cut-off answer can succeed on the next call, because output length varies a lot per call. A provider that rejects the call rejects every call, so re-trying it sooner only wastes tokens.
- **That 3-vs-10 choice follows the most recent failure, not the whole history.** Otherwise one cut-off answer during a re-probe would pin the session to the fast, expensive setting forever, against a cause that can never clear.
- **A billing denial keeps naming itself.** It is the only failure the user can fix, by adding credits, so it must not disappear into a generic "mixed" label.
- Rewrote the ladder test, which still asserted the rule this PR deletes, and added end-to-end coverage for the two new analytics values.
- Marked the 3-turn window in its comment as a judgment, not a measurement, so a later change does not cite it as evidence-backed. The queries that would settle it are in the ticket notes.

## Why this matters

The counter was gated on the wrong failure type, so this half of the latch has never fired since it shipped. A model that keeps getting cut off interrupts the user on every single turn of a long CLI session, which is exactly what the latch was built to prevent.

## Notes for the reviewer

**This reverses a decision recorded in a code comment and credited to a reviewer on #299**, that a cut-off answer must never latch because one cut-off is just a sample of a chatty model, not proof about the next call. That argument was about a case the code never actually reached: anthing that reaches the latch check has already failed at *both* budgets, never once.

**An earlier version of this PR justified the cost with a claim that turned out to be false.** It said cut-offs get worse as a session grows, because the text sent to the checker grows with it. That text is hard-capped and levels off after about ten turns. The claim is gone and the 3-turn window is now justified on recoverability instead. The #299 reviewer approved the trade-off while citing the false claim, so that approval is worth re-confirming rather than assuming.

**The real cost, stated plainly.** Unlike a provider rejection, this failure is random. Two unlucky turns in a row can latch a model that mostly works, and turns inside the window run unverified. Three things bound it: the threshold of two, a successful check clearing the latch, and the shorter re-try window.

**Deliberately unchanged:** transient network errors still never latch, and a billing denial still latches immediately on its first occurrence.

**This does not fix the desktop symptom in the ticket.** Cowork builds a fresh session per message and a failed check ends the turn, so at most one failure is ever counted per message and the threshold of two is unreachable there. The only surface whose behaviour changes here is the CLI. That limitation is the reason the ticket has a separate sub-task.

**This is not the fix for the reported customer, and the data says it is a small fix.** Production PostHog (see the ENG-2193 comments from 2026-09-02) showed the reporting customer was on LM Studio local models failing with a bare `BadRequestError`, which this PR does not touch, and that `mindshub_air` had zero verifier hand-backs across 968 turns. The cut-off-answer class this PR newly counts toward the latch had zero occurrences fleet-wide in the last 14 days on instrumented builds (the approving review's PostHog read: `hard` 39, `transient` 12, `denied` 2, `truncated` 0). Read this as a correctness fix for a real logic gap, CLI-only in effect and low urgency, not as progress on the customer's symptom.

## How to test

1. `uv run pytest tests/test_verifier_failsafe.py tests/test_verifier_truncation.py tests/test_turn_cost_terminals.py`
2. To watch the new rule fire, point the coding role at a model that exhausts both budgets and run three tool-using turns in one CLI session. Turn 1 shows the diagnosis, turn 2 goes quiet, turn 3 spends no check at all.
3. To watch the window fix, latch on a provider rejection, let one re-probe get cut off, then confirm the next re-probe is 10 turns later and not 3.

## Ships with

Nothing blocking. mindsdb/anton#427, which tried to cover the per-message case, is closed: production data showed neither model on its list was actually failing. The per-message work has moved to ENG-2244 and ENG-2245.

## Checks

| Check | Observed result |
| -- | -- |
| `uv run pytest` on `87dc40c0` (staging merged in) | 2919 passed, 31 skipped |
| All GitHub checks on the head | 13 passing, including the live `verdict-eval` |


tino097 (commented):
Automated review of `8b9271c3..eaac9449`. Ran `tests/test_verifier_failsafe.py` (24 passed), plus `tests/test_turn_cost_terminals.py` and `tests/test_session_truncation_recovery.py` (49 passed). Also read the paired mindsdb/cowork-server#434 and `anton/cloud_turn/session.py` to trace the host wiring.

Nine inline comments below. Four findings sit on lines outside this diff, so they are here.

### 1. The cloud pod is a per-message host too, and does not get the lowered threshold

`anton/cloud_turn/session.py:606`

`anton.cloud_turn` builds one ChatSession per pod invocation. Its own config says so: `background_memory=False,  # one turn per pod` and `history_store=None,  # disk history OFF` at lines 640 and 637. So the no-verdict counter dies after at most one increment, the default threshold of 2 is unreachable, and every message pays a full-history hand-back. That is the exact condition this PR describes.

The paired fix sets the threshold on cowork-server's in-process `AntonHarness`. But `cowork-server/cowork/build_info.py:68-76` states that `AntonHarness.stream_response` refuses to run in-process when `tenancy_mode == "org"`, so on the cloud deployment the turn executes in a scratchpad-controller pod via `anton.cloud_turn` instead. `verifier_latch_threshold` does not appear anywhere under `anton/cloud_turn/`.

Latent rather than live today: the same docstring notes the web surface has no real traffic yet. Worth saying out loud in the PR either way, because the fix is scoped to the desktop path while the ticket is about the per-message shape.

### 2. The `verifier_failure` value catalogue is not updated

`anton/core/turn_cost.py:144`

`session.py:5471` builds `f"latched_{self._verifier_latch_reason or 'hard'}"`, and `_verifier_latch_reason` is now set to `verdict_failure` (`session.py:5661`), so `latched_truncated` is a new possible value. The doc comment still reads "plus `latched_hard` / `latched_denied` for turns that never made the call". Any query filtering `verifier_failure IN ('latched_hard','latched_denied')` silently drops every truncation-latched skip turn, which is the population this change exists to make visible.

### 3. Two stale comments survive in the touched function

Both are outside the diff hunks but inside `_stream_and_handle_tools`, and both now state the opposite of the code.

`anton/core/session.py:5487-5489`: "Truncation vs hard failure decides whether this counts toward the latch: a blown budget is a tail sample of one model's verbosity, an identical hard error twice is a capability problem." Sits directly on `verdict_failure`. Fifteen lines below, `if verdict_failure in ("hard", "truncated")` counts both.

`anton/core/session.py:5421-5423`: "A verifier that already failed hard twice in a row will keep failing for the rest of the session." Both halves are now wrong. The trigger is N configurable no-verdict failures, and truncation counts.

The PR body says the three comment blocks documenting the old rule were rewritten. These are two more, plus a third flagged inline at `session.py:1149`.

### 4. Repo conventions

Two items from the workspace CLAUDE.md, called out because they are explicit rules rather than taste. Both are arguable given the file's existing style, so take them or leave them.

"No ticket numbers, issue IDs, or ticket names in code, code comments, commit messages, tests, or docs inside a repo." The diff adds new instances: `session.py:449` ("ENG-1079 wants its honest diagnosis"), `session.py:5645`, `session.py:5648`, `tests/test_verifier_failsafe.py:462`. The file is already full of pre-existing IDs, so the established local style contradicts the rule; worth settling which one wins rather than drifting.

"At most 2 lines per comment." The diff adds a 19-line block at `session.py:5641-5659`, a 5-line block at `session.py:445-449`, and a 9-line prose block at `tests/test_verifier_failsafe.py:1028-1036`. Concrete cost rather than style: blocks this long are what let the three surviving contradictory blocks above go unnoticed.

anton/core/session.py
```diff
@@ -5617,33 +5637,33 @@ async def _stream_and_handle_tools(
...
-                        # hard still latches on the second hard failure. That is
-                        # deliberate — only a real verdict proves the model can
-                        # produce one, and a truncation in between is no evidence
-                        # that it can (review: pnewsam on #299).
+                        # The reason carries the class, so the books can tell
+                        # latched_hard from latched_truncated rather than
+                        # mislabelling one as the other.
                         self._verifier_latched = True
-                        self._verifier_latch_reason = "hard"
+                        self._verifier_latch_reason = verdict_failure
```
tino097: `_verifier_latch_reason` is written only here, at the initial latch, and never on a failed re-probe. So the mislabelling this change sets out to fix survives whenever the re-probe fails with a different class than the original latch.

Sequence: a session latches with reason `hard` (kimi-K3 400). Ten turns later the user switches to a verbose model, the re-probe exhausts the ladder, so `verdict_failure == "truncated"` and the failed-re-probe branch at 5627-5639 `break`s without touching `_verifier_latch_reason`. That turn books `verifier_failure="truncated"` (stamped at 5580) while the next nine skipped turns book `latched_hard` (5471). The books now disagree with each other and name a cause that is no longer the cause.

The comment at 5657-5659 ("the books can tell `latched_hard` from `latched_truncated` rather than mislabelling one as the other") is therefore true only for the first latch. `test_latched_truncating_reprobe_stays_latched_without_rediagnosing` walks exactly this path but asserts only `_verifier_latched` and the diagnosis count, never the reason, so it would not catch the drift.

anton/core/session.py
```diff
@@ -5617,33 +5637,33 @@ async def _stream_and_handle_tools(
                         if self._turn_cost is not None:
                             self._turn_cost.verification_skipped = True
                         break
-                    if self._verifier_hard_failures >= 2:
-                        # Second hard failure in a row: the first could have been
-                        # transient, this one establishes the pattern. Latch and
-                        # skip the diagnosis on this turn too — a second
+                    if self._verifier_no_verdict_failures >= self._verifier_latch_threshold:
```
tino097: At the threshold of 1 that cowork-server ships, a single untyped provider blip now ends the turn silently with no hand-back at all.

Anything not in `_TRANSIENT_VERDICT_ERRORS` is classified `hard` by the generic `except Exception` above: a gateway 502/503 surfaced as a plain Exception, a JSON decode error, an SDK wrapper type. At threshold 1 that one occurrence latches on the first message and `break`s before the hand-back, so the user gets the raw streamed reply with zero signal, possibly mid-task.

That also makes the hand-back copy at 5670-5700 ("Your reply above stands, and nothing in your work was rejected, no verdict came back at all") reachable on the cowork path only for typed transients. #423 rewrote that message on this same ticket. Worth stating explicitly that the two halves of ENG-2193 pull against each other here, and that the trade was made knowingly: one improves the message, the other removes it for the host that reported the bug.

anton/core/session.py
```diff
@@ -5602,8 +5622,8 @@ async def _stream_and_handle_tools(
                     if self._turn_cost is not None:
                         self._turn_cost.verification_skipped = True
                     break
-                if verdict_failure == "hard":
-                    self._verifier_hard_failures += 1
+                if verdict_failure in ("hard", "truncated"):
+                    self._verifier_no_verdict_failures += 1
```
tino097, line 5702: Counting truncation toward the latch has a property the PR body's tail-sample argument does not cover: verifier output truncation grows with transcript size, so it is not an i.i.d. model property.

`_build_verify_request(self._history, user_message)` renders a larger transcript as the session goes on, and `exc.truncated` fires when the model narrates past the budget before reaching the forced call. More context to narrate about means more truncation, so truncation is a monotone function of session length.

Consequence in the case the PR argues is safe: a long CLI session with a verbose coding model hits two truncations around turn 30 and latches verification off for the rest of the session in 10-turn windows, and the only thing that could notice is a re-probe that per this diff no longer re-diagnoses and stays latched. The default of 2 softens this but does not remove it. If the trade is accepted, the reasoning belongs in the comment block below, which currently argues only the per-message case.

anton/core/session.py
```diff
@@ -5602,8 +5622,8 @@ async def _stream_and_handle_tools(
                     if self._turn_cost is not None:
                         self._turn_cost.verification_skipped = True
                     break
-                if verdict_failure == "hard":
-                    self._verifier_hard_failures += 1
+                if verdict_failure in ("hard", "truncated"):
```
tino097: Suggestion. `("hard", "truncated")` is now the definition of "counts toward the latch", a concept three comment blocks (391-396, 5437-5442, 5647-5651) and one host-side config knob all reason about. The file already names its two sibling sets at module level with the rationale attached: `_TRANSIENT_VERDICT_ERRORS` (412) and `_DENIED_VERDICT_ERRORS` (437).

A `_LATCHING_VERDICT_FAILURES = ("hard", "truncated")` beside them gives those blocks one anchor to point at instead of three restatements that drift apart, which is how the stale comments in this PR happened in the first place.

anton/core/session.py
```diff
@@ -1133,13 +1147,20 @@ def __init__(self, config: ChatSessionConfig) -> None:
         # latch, each multi-step turn pays a full-history diagnosis call and
         # shows the "checking in" message (ENG-1155). Session-scoped: a hard
         # failure twice in a row latches, a successful verdict clears it.
```
tino097: Stale comment, on the line the diff just renamed. "Session-scoped: a hard failure twice in a row latches, a successful verdict clears it." sits directly above `_verifier_no_verdict_failures`, which was renamed precisely because it no longer counts only hard failures. It also hardcodes "twice" three lines above the configurable threshold assignment at 1159.

anton/core/session.py
```diff
@@ -994,6 +1004,10 @@ class ChatSessionConfig:
     data_vault: DataVault | None = None
     console: Console | None = None
     initial_history: list[dict] | None = None
+    #: Verdict failures before the latch engages; None uses
+    #: `_VERIFIER_LATCH_THRESHOLD`. A host that rebuilds this session per
+    #: message wants 1 — see that constant.
+    verifier_latch_threshold: int | None = None
```
tino097: Design note, not a blocker. This knob requires every per-message host to independently know (a) its own session lifecycle, (b) that this field exists, and (c) that 1 is the right value. cowork-server even needs a `dataclasses.fields()` probe to pass it safely. That is what let `anton/cloud_turn` be missed.

`ChatSessionConfig` already carries lifecycle signals at every per-message call site: `anton/cloud_turn/session.py` sets `background_memory=False` with the comment "one turn per pod" and `history_store=None` with "disk history OFF". A single declared lifecycle, say `single_turn=True`, that the session translates into the threshold (and into the other per-message decisions currently duplicated by hand) fixes every such host at once, and would have surfaced the `cloud_turn` gap at the type level instead of in review.

tests/test_verifier_failsafe.py
```diff
@@ -454,10 +457,10 @@ async def always_400(_schema, *, system, messages, max_tokens):
         await session.close()
 
 
-async def test_truncation_does_not_latch(workspace):
-    """A blown budget is one model being verbose on one transcript — a tail
-    sample, not a capability problem. Only hard failures latch (ENG-1155);
-    truncation keeps its ENG-1081 retry ladder and its honest diagnosis.
+async def test_an_exhausted_ladder_latches_at_the_threshold(workspace):
```
tino097, line 456: This test does not pin the claim its own docstring makes. "The ENG-1081 retry ladder still runs in full first" is the load-bearing distinction between "an exhausted ladder latches" and "a single truncation latches", but the test asserts only `_verifier_latched`, `_verifier_latch_reason` and the diagnosis count.

A regression that incremented `_verifier_no_verdict_failures` on the 2048 truncation and skipped the 4096 retry, halving the verifier's real recovery rate, passes this test unchanged: latch state and diagnosis count are identical either way. Counting `mock_llm.generate_object_code.await_count` would close it.

tests/test_verifier_failsafe.py
```diff
@@ -1010,3 +1023,80 @@ async def always_denied(_schema, *, system, messages, max_tokens):
...
+    mock_llm.generate_object_code = AsyncMock(side_effect=_always_truncated())
+    default = _make_session(workspace, mock_llm)
+    lowered = _make_session(workspace, mock_llm, verifier_latch_threshold=1)
+    try:
+        from anton.core.session import _VERIFIER_LATCH_THRESHOLD
+
+        assert default._verifier_latch_threshold == _VERIFIER_LATCH_THRESHOLD == 2
+        assert lowered._verifier_latch_threshold == 1
+        # A threshold below one would latch before any verdict call ran.
+        assert _make_session(
```
tino097: Two small problems with building the third session inside the `assert`.

It is never closed. `default` and `lowered` are closed in `finally`, this one is not, so it escapes `close()`'s `_reap_tracked_backends` / `_scratchpads.close_all()`. Harmless only because it never runs a turn; the moment this case grows a `turn_stream` it leaks a scratchpad venv into `.pytest-workspace`.

Being inside an `assert` also means the whole "0 clamps to 1" case disappears under `python -O`, leaving the only guard on the `max(1, ...)` clamp at session.py:1162 unexecuted. Assigning it to a local and asserting on that fixes both.

tests/test_verifier_failsafe.py
```diff
@@ -1010,3 +1023,80 @@ async def always_denied(_schema, *, system, messages, max_tokens):
...
+            "no tool call", truncated=True, output_tokens=max_tokens,
+            max_tokens=max_tokens, stop_reason="stop",
+        )
+
+    return truncated
+
+
+@pytest.mark.parametrize("failure, reason", [
+    (_always_truncated(), "truncated"),
+    (AsyncMock(side_effect=RuntimeError("400 tool_choice not supported")), "hard"),
```
tino097: This `AsyncMock` is constructed at collection time and reused across the session, which forces the `isinstance` branch at 1059-1061 whose only job is to undo the asymmetry in the parametrize list. Its `call_count` / `await_count` also accumulate outside any test's control.

Making the second param a plain async callable in the shape of `_always_truncated()` (raising `RuntimeError` instead of `StructuredOutputError`) deletes the branch and the shared state in one move.

tino097:
Thanks — most of this is now in at `bd62a848`. Note first that several comments landed against the revision that lowered a latch threshold to 1. That design is gone: it latched *before* the hand-back, so it produced zero diagnoses per conversation rather than the one the sub-task asks for, which left #423's reworded wording unreachable on desktop. The branch now carries the counter across the session rebuild instead, and `verifier_latch_threshold` no longer exists.

Fixed

`_verifier_latch_reason` going stale on a failed re-probe — correct, and the sequence you traced was exactly right. The class is now recorded on every counted failure, before the latched check, so a re-probe that fails a different way reaches the books. Added `test_a_failed_reprobe_updates_the_latch_attribution`, which walks your sequence and asserts the reason rather than only the latch and the diagnosis count.

Mixed classes reported by arrival order (Codex P2) — the reason accumulates instead of taking the last writer: one class while they agree, `mixed` once they differ. So one exhausted ladder plus one hard rejection is now `latched_mixed`, not whichever came last. `turn_cost.py` enumerates the new value, and `test_mixed_classes_do_not_depend_on_arrival_order` pins it.

`_LATCHING_VERDICT_FAILURES` — added beside `_TRANSIENT_VERDICT_ERRORS` and `_DENIED_VERDICT_ERRORS`, with the rationale on it. You were right about the cause: three hand-restatements is how the stale comments in this PR happened.

The stale comment above the renamed counter — fixed, including the hardcoded "twice".

Truncation growing with transcript size — the argument is in the `_VERIFIER_TOKEN_BUDGETS` block now, as an accepted cost rather than an omission: verifier truncation is a monotone function of session length, not an i.i.d. model property, so two late truncations in a long CLI session can latch verification off in re-probe-sized windows. What bounds it is the threshold of two and a successful verdict clearing the counter.

The ladder test not pinning its own docstring — fixed with `generate_object_code.await_count == 4`, so a regression that counted the 2048 truncation and skipped the 4096 retry now fails.

Answered, not changed

`cloud_turn` (Codex P1) — real gap, and deliberately out of scope. `build_cloud_chat_session` does build a fresh session per request and does not carry the latch, so the web path still re-diagnoses. Carrying it there needs a new `TurnRequestV1` field plus scratchpad-controller's allowlist and a reply-side field its `_STEP_KINDS` whitelist relays — both contracts the working agreement marks as protected. Separate change, not a constant to flip.

A declared lifecycle (`single_turn=True`) over a knob — the knob is gone, so the specific hazard you named is too. The general point stands and is the better shape; it just now applies to `initial_verifier_latch`, which a host has to know to pass. Worth its own change rather than smuggling a lifecycle concept in here.

Stale with the threshold field

The session built inside an `assert`, and the `AsyncMock` in the parametrize list, were both in `test_the_default_threshold_is_unchanged_for_long_sessions`, deleted with the field. Both were fair.

The threshold-of-1 silent-blip case is also gone: the default of two is back, so a single untyped provider error still gets its hand-back.

tino097:
Both halves were right, and the first was a bug I introduced one commit earlier. Fixed at `450c1152`.

The dropped `mixed` reason. Adding it to what the session emits without adding it to what the restore accepts meant a carried `mixed` latch fell to `""` and the stamp's `or 'hard'` fallback then booked it as `latched_hard` — reintroducing, through the restore path, exactly the mislabelling the previous commit set out to remove. My own malformed-input test could not catch it: it asserts the reason is one of `("", "hard", "truncated", "denied")`, so a dropped `mixed` passes.

Rather than just adding the string, the accepted set is now `_LATCH_REASONS` at module scope, next to `_LATCHING_VERDICT_FAILURES`. Emitter and validator drifting apart is the actual defect, and two hand-maintained lists is what let it happen.

The impossible latched state. Also right, and it made the docstring a lie: it promises malformed shapes are ignored while accepting `latched=True` with a sub-threshold count, which would switch verification off for a conversation on storage nobody can account for. The restore now rejects the whole record unless the latch is explicable — either a deterministic denial, which latches on the first occurrence, or a count that reached the threshold.

Worth noting the one legitimate case your comment's rule would have excluded: a denial does latch with zero counted failures, because `"denied"` is not in `_LATCHING_VERDICT_FAILURES` and so never increments the counter. The check allows exactly that and nothing else, and `test_a_denied_latch_carries_no_counted_failures` pins it so a future tightening cannot break the denied path.

`_VERIFIER_LATCH_THRESHOLD` is back as a module constant, not configurable, now that the latch branch and the restore check both compare against it.

Three tests added: `mixed` round-trips, a sub-threshold latch is ignored (parametrized over 0 and 1 failures), and a denied latch with zero failures is still accepted.

Full anton suite: 2907 passed, 31 skipped.

tino097:
@codex review

tino097:
Both fixed at `1a134544`.

Codex, the restored denied latch. This one is a regression against behaviour the existing code states outright, so thank you. The denied branch's own scope note says that for a host which rebuilds per message "the steady-state cost of a persistent denial is one silent verdict call per message, and top-up recovery is simply the next message". Persisting the latch quietly replaced a one-message recovery with up to `_VERIFIER_LATCH_REPROBE_TURNS`.

The fix is not to widen the persisted identity but to stop carrying a denied latch at all. It never needed carrying: that path is already silent, so there is no repeated hand-back to suppress, and the only thing persistence buys is saving one silent verdict call per message — which the code above already accepts as the cost. Everything else about the denial is unchanged: still latches on the first occurrence, still silent, still stamps the books.

Your credential-generation suggestion would be the right shape if a capability latch had the same problem, but it does not: `hard` and `truncated` are statements about the model, and the model name is already in the persisted identity. A provider swap behind an unchanged alias could in principle carry a stale capability latch, which the periodic re-probe resolves; a denial was the case where the window mattered, and it is now gone.

Copilot, `bool(stored.get("latched"))`. Right, and the docstring promised better than the code delivered: the string "false" coerced to True, from a value that arrives out of a host's JSON column. It now takes only a real boolean and ignores the record otherwise, parametrized over "false", "true", 1, 0 and None.

The restore rule also got simpler as a result — a latch is now explicable only by reaching the threshold, with no denied exception to carve out.

Full anton suite: 2912 passed, 31 skipped.

tino097:
@codex review

tino097:
Both Copilot findings are in via the autofix commits (`0e380ec2`, `026c52f2`), and one of them has a consequence worth closing, done at `f758f3c4`.

Treating `denied` as unset when accumulating classes is right — a denial is a wallet or access state, not a capability class, so mixing it into `mixed` would have made the skip log stop describing the denial. But it also means a re-probe can reclassify a denied latch off `"denied"`, and a denied latch never incremented the counter. The state that leaves is `latched=True` with one counted failure — which `_restore_verifier_latch` refuses, because a latch is only explicable at the threshold. So the property could emit a record the restore silently drops.

Unreachable in practice today (a denied latch is not carried, and the sequence needs a long session, which does not persist), but this is the third time the emitter and the validator have disagreed on this branch — first `mixed` being emitted but not accepted, then this. So rather than patch the case, the property now refuses to emit anything the restore would refuse, and the round trip is pinned as an invariant: `test_whatever_is_emitted_survives_being_restored` asserts `restore(emit()) == emit()` over six realistic end-states, so the two cannot drift again without a test failing.

The bool coercion fix is right and needed no follow-up: `isinstance(True, int)` is True, so `int(...)` would have taken `true` as 1 for both counters.

Codex, the cloud path. Correct on the facts, third time raised, and still deliberately out of scope — so I have stopped leaving that only in a PR reply and put it in the `verifier_latch` docstring, where the next person reading the API sees it: only the local path carries this, and the pod needs a `TurnRequestV1` field plus one on the reply stream before it can. Both are contracts this workspace treats as protected, so it is a separate change with the scratchpad-controller in it, not a wiring omission here.

Full anton suite at `f758f3c4`: 2920 passed, 31 skipped.

tino097:
Already in via the autofix at `2171fc0b`, and I verified it does what it claims rather than taking it on faith:

```
initial_verifier_latch is keyword-only: True
  initial_history: kind=POSITIONAL_OR_KEYWORD index=11
  history_store:   kind=POSITIONAL_OR_KEYWORD index=12
  session_id:      kind=POSITIONAL_OR_KEYWORD index=13
```

So the fields that used to follow it keep their positional slots, which is the whole point of the finding. `field` was already imported in this module, and `kw_only` needs 3.10 against anton's 3.11 floor.

Worth adding that within these two repos the break was not reachable — all ~45 `ChatSessionConfig` construction sites use keywords or `**kwargs`, which I checked when the field went in. The reason the fix is still right is that anton is consumed as a library, so a positional caller outside these repos is not something this PR can rule out.

One interaction the finding does not mention, and the one that would have hurt: cowork-server decides whether to pass this field by probing `dataclasses.fields(ChatSessionConfig)`. Had a keyword-only field been absent from that, the guard would have reported the feature unsupported and turned the whole thing into a silent no-op — the same failure mode as the stale pin. It is present, with `kw_only=True` on the field, and the guard predicate returns True. The cowork-server test's stand-in config now mirrors the real keyword-only declaration instead of a positional one, so that stays covered on the side that actually reads it.

Full anton suite at `2171fc0b`: 2920 passed, 31 skipped. cowork-server: 2944 passed, 1 skipped.

tino097:
@codex review

anton/core/session.py
```diff
@@ -382,15 +382,26 @@ def _null_falls_back_to_default(cls, v: object) -> object:
...
+# ladder, not one sample. Against the distribution above — 16 calls spanning
+# 245-1654 output tokens — blowing past 4096 twice is a statement about the
+# model. Typed transients still never latch, see `_TRANSIENT_VERDICT_ERRORS`.
+#
+# Accepted cost, worth knowing: this truncation is NOT an i.i.d. property of the
+# model. `_build_verify_request` renders a transcript that grows with the
+# session, so a verbose model truncates more as a long CLI session goes on, and
+# two late truncations can latch verification off for the rest of it in
+# re-probe-sized windows. The threshold of two, and a re-probe that a successful
+# verdict clears, are what bound that.
```
pnewsam: **Major (intentional — your call):** widening the latch to count an exhausted truncation ladder buys a real, documented cost, and this comment states it plainly: truncation is not i.i.d. because the rendered transcript grows with the session, so a verbose-but-working model truncates *more* as a long CLI session runs on, and two late truncations latch verification off for a re-probe-sized window. The underlying fix is sound and stands alone — a truncating verifier previously never latched even in the long CLI sessions the latch was built for. The only open question is whether the accepted cost is acceptable to you; you explicitly reopened the #299 decision to ask it. Not a defect — flagging it as the decision this PR turns on.

anton/core/session.py
```diff
@@ -1402,6 +1431,19 @@ def last_compaction(self) -> dict | None:
         }
 
 
+    def _note_latch_class(self, failure: str) -> None:
```
pnewsam, line 1459: **Nit:** a `denied` latch that later re-probes into a hard/truncated failure gets relabeled here, so `_verifier_latch_reason` flips off `"denied"` with the counter at 1, and subsequent skip logs/books read `latched_hard` instead of `latched_denied`. Arguably correct — the cause genuinely changed — but it's an attribution edge the tests don't pin. Worth a confirming glance that it's intended.

pnewsam (commented):
## Code Review

**Verdict: COMMENT**
**Model: claude-opus-4-8**

Reviewed the full diff — the latch-rule widening in `session.py`, the books enumeration in `turn_cost.py`, and the reworked tests. The logic is coherent, the `_verifier_hard_failures` → `_verifier_no_verdict_failures` rename is complete across all six source sites and four test sites, and the rewritten comments match the code. No blocking or confirmed defects. The one thing worth a decision is intentional and you already flagged it: this reverses a prior call of yours (#299), so COMMENT rather than APPROVE — the merge turns on your judgment, not on a bug.

### Validation

- Traced the verdict loop end-to-end: a truncation reaches the no-verdict handler only as an *exhausted* ladder (`retrying` is false only at the last budget), so your core justification holds — the old "tail sample" rule never actually gated the latch check.
- Traced all four new/changed test scenarios against the new control flow; each is consistent with the code. Did **not** execute the suite — my local checkout is on `staging`, not the PR head.
- Grepped consumers of `verifier_failure` and the `latched_*` values: only logging/analytics stamping, no strict enum parse the new `truncated`/`latched_truncated`/`latched_mixed` values would break.
- Confirmed the dropped carry-state API left no dangling references.

Two inline notes below.

tino097:
Thanks for looking at this, given it reopens your #299 decision.

On the nit: intended, and you are right that nothing pinned it. Pinned now at `018f8e8b` — `test_a_denied_latch_relabels_when_the_reprobe_fails_differently` walks a wallet-402 latch through the re-probe window into a hard failure and asserts the reason ends as `hard`, plus that the denial latched with the counter still at zero.

Intended because the alternative is worse in the direction that matters. Before this branch the reason was written only at the initial latch and never again, so a session that latched `hard` and later re-probed into a truncation kept booking `latched_hard` for the rest of its life while that turn's own books said `truncated` — the two disagreed and the older one won. Following the current cause fixes that, and the denied case falls out of the same rule: a denial may have been paid since, so a capability failure now is the live cause. The counter staying at zero is right too — the denial never counted, so nothing pretends it did.

On the major, agreed on all of it, including that it is a decision rather than a defect. The two things I would weigh, both bounded:

It needs two exhausted ladders with no successful verdict between them, in one session. A single successful verdict resets the counter, and against the distribution recorded on `_VERIFIER_TOKEN_BUDGETS` — 16 calls spanning 245 to 1654 output tokens, median ~290 — that means a model blowing past 4096 twice with nothing working in between. When it does latch, the re-probe re-tests every `_VERIFIER_LATCH_REPROBE_TURNS` turns and any successful verdict clears it, so the loss is a window rather than the session.

Against that, not counting it leaves a truncating verifier never latching at all, which is the state today: the counter was gated on `verdict_failure == "hard"`, so the model the customer ticket is about could fail every single turn forever without the latch this repo built ever engaging. That is the asymmetry that made me reopen it rather than the per-message host, which is no longer in this PR at all.

The transcript-growth cost is in the comment block rather than only here, so whoever reads that code next inherits the caveat and not just the conclusion. If you would rather not accept it, the fallback that keeps most of the value is counting an exhausted ladder only when the model produced no usable tool call at the LAST budget — narrower, and it would need its own evidence.

Not my call to close: @tino097 owns the trade.

tino097:
Decision on the major: the cost is accepted, with a mitigation aimed at exactly the case you named. `efab6e9f`.

The re-probe window now depends on what latched. Truncation, or a mixed reason that includes it, re-tests after 3 skipped turns instead of 10; a capability rejection or a denial keeps the long window. The reasoning is that your objection is specific to truncation and does not apply to the others: a 400 on forced `tool_choice` makes the same statement on every call, so waiting is free, while transcript-driven truncation can happen to a model that is otherwise working, and that is the case that should not lose verification for long.

So a verbose-but-working model on a long session now loses auto-continue and STUCK detection for at most 3 turns before the verifier is tried again, and one successful verdict clears the latch outright. The price is one verdict call every 3 turns while latched instead of every 10 — paid only by a session that has already produced two exhausted ladders with nothing successful between them.

Two tests: `_reprobe_turns_for` pinned per reason, and an end-to-end one that latches on the second exhausted ladder, confirms no verdict call inside the short window, and confirms the re-probe fires on the turn the window closes.

Full anton suite: 2894 passed, 31 skipped.

If you would still rather truncation never latched, the widening is a self-contained hunk and can come out — but that returns the counter to being gated on `hard`, which is the state where the customer's model never latches at all.

tino097:
@codex review

tino097:
@pnewsam re-confirm request: the premise your sign-off rested on was wrong, so the trade you approved is not quite the trade this PR now makes.

In https://github.com/mindsdb/anton/pull/426#discussion_r3908580770 you flagged the accepted cost as the decision this PR turns on, describing it as "truncation is not i.i.d. because the rendered transcript grows with the session". That was my framing in the comment you were reading, and it does not hold.

`_render_verify_transcript` is hard-bounded: `max_convo=10, max_tool=12, tool_cap=400, text_cap=2000, final_cap=12000`, and `_build_verify_request` passes no overrides. The verify prompt levels off after about ten turns instead of growing with the session. Truncation is also an output overrun measured against `max_tokens`, so even the indirect input-to-narration channel is capped. Both comment blocks that asserted the growth claim are gone.

What this does and does not change:

The widening itself still stands, on the argument you called sound and standalone. `retrying` is false at the last budget, so the only truncation that ever reaches the latch check has already failed at both 2048 and 4096. It is never a single sample.

The accepted cost is different from the one you agreed to. It is not "gets steadily worse as a long session runs on". It is "this failure is random, so two unlucky turns with no success between them can latch a model that mostly works". Bounded by the threshold of two, a successful verdict clearing the latch, and the re-probe.

The 3-turn window is kept but re-justified. Not because truncation worsens over time, but because it can recover: output length varies per call, so a re-probe may land, while a provider that rejects the forced `tool_choice` rejects every call. Your objection was specific to truncation and that reasoning survives, but it is not the reasoning that was in front of you.

One more thing worth your eye, since it cuts against the window. A truncating re-probe runs the whole ladder before giving up, so it costs two calls, not one. At 3 turns that is about 0.67 verdict calls per turn against about 0.1 for a capability latch at 10. My comment claimed "one verdict call per window", which was wrong and is now corrected in the code. So the shorter window is the more expensive one, which is easy to read backwards.

Asking for one of: re-confirm the trade as now stated, or say the window number should change given the real cost. Also fixed from your second note (https://github.com/mindsdb/anton/pull/426#discussion_r3908580778): a denied latch that re-probes into a different failure now keeps naming the denial rather than relabelling, since that is the only class the user can act on, and it is pinned by a test.

tino097:
@codex review

Re-requesting: the previous pass approved `c7be427290`, which is one commit behind. `e83166ae` changed the latch attribution (a denial now keeps naming itself rather than collapsing into `mixed`), corrected the re-probe cost comment, and removed two comments that narrated the change.

alecantu7 (approved):
## Approve — reviewed at `e83166ae`

Re-reviewed against the seven items I raised on the earlier head (`8b9271c3..efab6e9f`, posted to ENG-2193). All seven are closed:

| # | item | resolution |
|---|---|---|
| 1 | the 3-turn window rested on a false premise (the verify prompt does not grow with the session — `_render_verify_transcript` is hard-bounded) | replaced with a sound argument: recovery is plausible after truncation and impossible after a capability failure, so the shorter window buys a real chance of verification returning |
| 2 | `test_verifier_truncation.py:460` still asserted the deleted rule ("a later turn gets a fresh chance — the retry is not latched off") | stale assertions removed |
| 3 | a hard failure could overwrite a `denied` latch reason, erasing the one class with a user remedy | denial now stays the reason, with the rationale written down |
| 4 | skip log cited a single failure as the reason for a latch the threshold says needs two | reworded to state the count accurately |
| 5 | `mixed` was absorbing, so one stray truncation during a hard latch's re-probe permanently dropped the session to the short window | resolved, and better than I proposed — the window keys on `_verifier_last_no_verdict` rather than the accumulated reason, so attribution and window are decoupled |
| 6 | three comment blocks still described the dropped carry-state API | removed |
| 7 | `latched_truncated` / `latched_mixed` reached analytics with no end-to-end test | pinned in `test_turn_cost_terminals.py` and named in the `analytics.py` register |

13/13 checks green, including `verdict-eval`.

### Two non-blocking notes

**The branch is `CONFLICTING` with staging.** Approval doesn't unblock that — it needs a rebase before it can merge. Note that a force-push after rebase will invalidate this approval, since it is recorded against `e83166ae`.

**`_VERIFIER_LATCH_REPROBE_TURNS_TRUNCATED = 3` is still an unmeasured constant.** The justification is now sound reasoning rather than a false premise, which is what I asked for. But nobody has measured that 3 is the right number. The existing `10` was never measured either, so this is not a regression — flagging it only so a later change doesn't cite it as evidence-backed.

### Scope caveat, worth recording on the ticket

This should not be logged as delivering ENG-2193's third acceptance criterion ("the same verifier failure does not re-diagnose on every message"). It cannot:

- **Desktop** rebuilds `ChatSession` per message, so the counter resets before it can reach the threshold of 2.
- **Web** builds one anton session per turn in the scratchpad pod, so the counter can reach at most 1 before the pod exits.

The breaker is therefore unreachable on both product surfaces; the effect is confined to long-lived CLI sessions.

Two supporting measurements, from PostHog 424726 over the last 14 days on instrumented builds (`anton_version >= 2.26.8.31`, the first builds carrying ENG-1858's `verifier_failure`):

- **Zero `truncated` verifier failures** across every endpoint class. Present classes are `hard` (39), `transient` (12) and `denied` (2). So the branch this PR makes reachable currently sees no traffic.
- The population that *is* failing is `endpoint_class=local` — 29 of 29 `hard` / `BadRequestError`, and a provider 400 makes exactly one request with no ladder (verified by repro), so an exhausted ladder cannot arise there either.

None of that argues against merging. The breaker had a branch that could never trip, and closing that is correct on its own merits — the latch was written for exactly the long-session case this restores. It is latent-correctness work plus an attribution fix, not a change to any number we are currently watching. Worth being explicit so it isn't later credited with a fix it did not make.

PR #427, opened 2026-09-02
[ENG-2193] Skip the completion verifier on a model that cannot produce a verdict

tino097:

> [!WARNING]
> **DRAFT, and should not merge as it stands.** The idea may still be right, but the evidence I built it on turned out to be out of date. See Notes for the reviewer.

## Background, if you don't work in this file

After a turn that used tools, anton runs a **completion check**. It asks the cheap coding model one question, "is this task actually finished?", and requires the answer as a structured tool call. If the check fails to produce an answer, the user is told an internal error stopped verification.

That message is what the ticket is about. A customer asked anton to create a file. Anton created it. The reply still ended with an internal-error notice, they read it as "the agent cannot save files", and filed it as a blocker.

Two terms used below:

- **forced tool call** — the check requires the answer in a specific structured form. Some models cannot do this, either because they talk past it or because the provider rejects the request outright.
- **ladder** — if the answer is cut off, anton retries once with a bigger budget, 2048 then 4096.

## What changed

- **Names the models that cannot answer a forced tool call**, with the evidence for each and what would take it off the list.
- **Skips the check entirely on those models**, instead of spending both budgets and then telling the user an internal error happened on a turn that actually worked.
- **Marks the skipped turn as unverified in analytics.** Without this the turn looks identical to one that passed its check, which would quietly inflate our success numbers.
- Added a test that the reply still lands in the transcript exactly once. The skip exits early, and adding the reply twice was a real bug once before.

## Why this matters

On a default free-tier account the check runs on a model that we believed could not answer it. The latch that normally silences a repeated failure needs two failures to trigger, and the desktop app builds a fresh session for every message, so it never gets to two. Every message pays the full cost and shows the error.

## Notes for the reviewer

> [!IMPORTANT]
> **The premise does not hold at the budgets this code actually uses.** Both entries on the list are wrong, and I verified both against the source after review. This is why it is a draft.

**`kimi` was fixed at the gateway.** anton asks for a specific named tool. `mindshub_inference` knows moonshot rejects that form, and rewrites it to a permissive one before sending. Under that rewrite kimi returns correct tool calls in 197 tokens. The rejection this entry cites no longer reaches anton at all, so listing it would switch off a working check for any paid user who picks kimi.

`providers/registry.py:77` sets `supports_named_tool_choice=False` for moonshot, and `InferenceService._effective_tool_request` carries the measurements.

**`mindshub_air`'s evidence is from a budget we no longer use.** The "98.6% of checks returned no answer" figure is from the old 256-token budget. `session.py:378-391` says so in the same paragraph, and gives the current numbers: 245 to 1654 tokens across 16 calls, median about 290, with a 4096 retry behind it. Nearly everything fits. When the same chattiness showed up at two other call sites, the fix was a bigger budget, not a skip.

**Take both off and the list is empty**, so this is a no-op at best, and switches off a working check at worst.

**What is actually still unknown** is which failure really fires on a live free-tier account. anton#423 shipped the WARNING log line that answers it. That measurement should come before this is revived, because it also decides whether the problem is latch-shaped or skip-shaped.

**Smaller things to fix if this is revived:**

- The log line is INFO and should be WARNING. The root logger defaults to WARNING, and the ticket explicitly asks that a customer's default log show the failure.
- The model name is matched by exact string, so a `latest:`-prefixed setting from cowork-server would slip past it. `core/llm/identity.py sanitize_model_name` already handles that prefix.
- Skipping means no verification at all for those accounts, which is the product trade-off the ticket reserved for a decision. This PR decides it by default, and that should be explicit.

## How to test

1. `uv run pytest tests/test_verifier_failsafe.py tests/test_verifier_truncation.py tests/test_turn_cost_terminals.py`
2. Before reviving: reproduce on a free-tier minds-cloud account and read the `completion-verifier verdict=` line to see which failure actually fires at the current budgets.

## Ships with

mindsdb/anton#426, which fixes a separate bug in the same area. Independent, either order.

## Checks

| Check | Observed result |
| -- | -- |
| `uv run pytest` | 2892 passed, 31 skipped |
| All GitHub checks on the head | 13 passing, including the live `verdict-eval` |

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -171,8 +171,11 @@ _pending_lock = threading.Lock()
 #                    spend-ceiling extensions granted this turn, and the
 #                    ceiling one's size), verifier_failure + verifier_error_type
 #                    (WHY the completion verifier produced no verdict — the
-#                    loop's truncated/transient/hard/denied class and the
-#                    content-free exception type; "" on verified turns; ENG-1858),
+#                    loop's truncated/transient/hard/denied class,
+#                    latched_{hard,denied} when an earlier turn latched, or
+#                    skipped_incapable_model when the coding model cannot
+#                    answer a forced tool call, plus the content-free
+#                    exception type; "" on verified turns; ENG-1858),
 #                    harness, surface (desktop / web / cli — WHERE
 #                    the user was, "" when the host did not say; ENG-1945),
 #                    anton_version, conversation_id, turn_index
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -436,6 +436,35 @@ _DENIED_VERDICT_ERRORS: tuple[type[BaseException], ...] = (
     ModelUnavailableError,
 )
 
+# Models that cannot answer a FORCED tool call, so a verdict call on them is
+# spent to produce nothing. Skipping beats discovering it twice per session:
+# the latch below needs two failures, and a host that rebuilds ChatSession per
+# message never reaches two, so every message paid the ladder AND an "internal
+# error" hand-back on a turn whose work had succeeded.
+#
+#   mindshub_air  narrates past the budget instead of reaching the call — 98.6%
+#                 of prod verdicts returned no tool call (ENG-1081). A MindsHub
+#                 alias, so the name does not collide with anyone's own model.
+#   kimi          rejects forced `tool_choice` outright with a 400 (ENG-1095).
+#                 A property of the model family, not of our gateway, so it
+#                 holds wherever the model is served.
+#
+# Matched on the requested alias, which is what `coding_model` holds — the
+# gateway resolves `mindshub_air` to a real model server-side and only echoes
+# that back on the response.
+#
+# Deliberately NOT provider-gated. The provider string has two vocabularies
+# depending on which host wrote the settings (ENG-1695), and the endpoint
+# classifier reads the PLANNING role, so neither answers the coding-role
+# question this needs. Both entries above justify themselves without it.
+#
+# Being out of date is bounded in both directions: an entry that the gateway
+# has since fixed costs that model its verification until the entry is removed,
+# and a broken model missing from the list falls through to the latch exactly
+# as it does today. This is an optimisation over the latch, not a replacement.
+_VERDICT_INCAPABLE_MODELS: frozenset[str] = frozenset({"mindshub_air", "kimi"})
+
+
 # Turns a latched session skips before spending one verdict call to see whether
 # the cause has gone away (user switched model, gateway fix shipped). Without a
 # re-probe the latch is permanent for the session and "reset on a successful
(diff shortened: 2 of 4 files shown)


tino097:
Converting to draft. The premise does not hold at the budgets this code uses, and the evidence I built it on was superseded.

kimi: anton sends a NAMED tool choice (core/llm/client.py:312 -> core/llm/openai.py:296 _translate_tool_choice). mindshub_inference sets supports_named_tool_choice=False for moonshot (providers/registry.py:77) and rewrites named to "auto", under which kimi returns correct tool calls at 197 output tokens (InferenceService._effective_tool_request, the ENG-1095 note). So the 400 this entry cites no longer reaches anton, and listing kimi would disable verification for a paid user whose verdict call works.

mindshub_air: the 98.6% no-tool-call figure is from the retired 256-token budget. session.py:378-391 says so directly, and gives the distribution at the current budget as 245-1654 output tokens over 16 identical calls, median ~290, with a 4096 retry behind it. The repo's answer to the same narration at the two other generate_object_code sites was to raise the budget, not to skip (memory/consolidator.py:177, memory/cortex.py:695).

With both entries removed the set is empty, so this is a no-op at best and switches off a working check at worst.

What is actually still unestablished is the failure class on a real free-tier account, which is the open investigation this should have waited on. anton#423 shipped the WARNING line that instruments it; the next step is a repro on free tier to see which branch fires, not a list built from the old measurement.

tino097:
Closing. Not needed as it stands, and the problem it targets already has two layers handling it.

**Both entries on the list are invalid at the budgets this code uses.** Three independent passes agree: the reviewer subagent, my own check against the gateway source, and Copilot on this PR.

- `kimi` was fixed at the gateway. anton sends a named tool choice, mindshub_inference sets `supports_named_tool_choice=False` for moonshot and rewrites it to `"auto"`, under which kimi returns correct tool calls in 197 output tokens. The 400 this entry cited no longer reaches anton.
- `mindshub_air`'s evidence is from the retired 256-token budget. At the current ladder the measured spread is 245-1654 output tokens over 16 calls, median ~290, with a 4096 retry behind it. Nearly everything fits.

Remove both and the list is empty, so the skip path can never fire.

**The wider reason it is not needed.** This was an optimisation, not the only thing standing between the user and the bug. A repeatedly-failing check already goes quiet on its own through the latch, and a chatty model already gets two chances through the 2048 -> 4096 ladder. This would only have made that quieter sooner and cheaper, and only for a model that genuinely cannot answer a forced tool call. We have not found one.

**What would bring it back.** A free-tier repro that reads the `completion-verifier verdict=` WARNING shipped in anton#423, showing a model that fails the check deterministically at 2048 and 4096. If that turns up, this is small to write again: the mechanism here was sound and tested, only the data was wrong.

If instead that measurement shows the failure is intermittent, then this shape was wrong from the start and the answer is on the latch side, not here.

Branch `tino097/eng-2193-skip-verifier-on-incapable-model` is kept. Full reasoning is in the ticket notes under followups item 0.

PR #428, opened 2026-09-02
Migrate to httpx2 

ZoranPandovski:
`anthropic` and `openai` moved to `httpx2`, but we still pinned `httpx<1` and imported it at module scope, so cold installs resolved both libraries and the `AsyncClient` we handed the SDKs was a different library than the one they used.
Nothing closed the provider connection pools either, so `httpcore2` printed an asyncgen traceback at exit after a normal run.

- `httpx>=0.27,<1` -> `httpx2>=2.7,<3`; `anthropic>=1.0`, `openai>=3.0`; lock regenerated.
- `import httpx` -> `import httpx2 as httpx` in session, web tools, the OpenAI
  provider, the CLI search probes, and tests.
- Added `aclose()` on the providers and `LLMClient`, called from
  `ChatSession.close()` and from the CLI's `asyncio.run` wrapper. Streams close
  in a `finally`.

Also , Release e2e now builds the tag, installs the wheel into a clean venv, and runs the scenarios through it — then `publish` uploads those exact bytes. A source-tree test cannot catch a packaging bug like this, and a PyPI upload cannot be withdrawn..

Fixes [ENG-2234](https://linear.app/mindsdb/issue/ENG-2234/anton-migrate-to-httpx2-with-the-sdks-and-close-provider-http-pools), [ENG-2235](https://linear.app/mindsdb/issue/ENG-2235/anton-verify-the-built-wheel-before-publishing-to-pypi-not-beside-it)

### DIFF
--- a/.github/workflows/tests_e2e_release.yml
+++ b/.github/workflows/tests_e2e_release.yml
@@ -97,7 +97,7 @@ jobs:
 
       - name: Run E2E scenarios against the installed wheel (live)
         env:
-          ANTON_E2E_WHEEL_PYTHON: /tmp/wheelenv/bin/python
+          ANTON_E2E_WHEEL_PYTHON: ${{ env.WHEEL_VENV }}/bin/python
           ANTON_OPENAI_API_KEY: ${{ secrets.ANTON_OPENAI_API_KEY }}
           ANTON_PLANNING_PROVIDER: openai
           ANTON_CODING_PROVIDER: openai
@@ -113,4 +113,8 @@ jobs:
           name: verified-dist
           path: dist/
           if-no-files-found: error
-          retention-days: 1
+          # Long enough that a delayed or re-run publish (environment approval,
+          # transient PyPI failure retried the next day) still finds the bytes.
+          # If it has expired anyway, re-run this whole workflow for the tag —
+          # never rebuild in the publish job, that skips verification.
+          retention-days: 14
--- a/.github/workflows/tests_e2e_release.yml
+++ b/.github/workflows/tests_e2e_release.yml
@@ -1,5 +1,9 @@
 name: Release e2e scenarios
 
+# Verifies the artifact, not the checkout: builds the wheel for the tagged
+# commit, installs it into a clean venv, and runs the live E2E scenarios
+# through that install. Publishing is irreversible, so this must gate it.
+
 permissions:
   contents: read
 
@@ -7,15 +11,24 @@ on:
   workflow_call:
     inputs:
       tag:
-        description: "Release tag the caller just created (e.g. v2.0.5)"
+        description: "Release tag to verify (e.g. v2.0.5)"
+        required: true
+        type: string
+      version:
+        description: "Version the release job minted (e.g. 2.0.5)"
         required: true
         type: string
 
 jobs:
-  e2e-live:
+  verify-wheel:
     runs-on: ubuntu-latest
+    env:
+      WHEEL_VENV: /tmp/wheelenv
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.tag }}
+          fetch-depth: 0  # hatch-vcs needs tags to derive version
 
       - name: Install uv
         uses: astral-sh/setup-uv@v5
@@ -23,11 +36,81 @@ jobs:
       - name: Set up Python
         run: uv python install 3.12
 
-      - name: Run E2E tests (live)
+      - name: Build the distributions
         env:
+          # Match the publish job: build exactly the version that was minted.
+          SETUPTOOLS_SCM_PRETEND_VERSION: ${{ inputs.version }}
+        run: uv build
+
+      - name: Install the wheel into a clean venv
+        run: |
+          uv venv --python 3.12 "$WHEEL_VENV"
+          uv pip install --python "$WHEEL_VENV/bin/python" dist/*.whl
+
+      - name: Wheel must not resolve to the source tree
+        # A green run that imported anton/ from the repo would prove nothing.
+        run: |
+          loc=$(cd /tmp && "$WHEEL_VENV/bin/python" -c \
+            "import anton, pathlib; print(pathlib.Path(anton.__file__).resolve())")
+          echo "anton imported from: $loc"
+          case "$loc" in
+            "$WHEEL_VENV"/*) ;;
+            *) echo "::error::anton resolved outside the wheel venv: $loc"; exit 1 ;;
+          esac
+
+      - name: Installed version matches the release
+        env:
+          EXPECTED: ${{ inputs.version }}
+        # Compare as PEP 440 versions: packaging strips leading zeros, so the
+        # CalVer tag 2026.08.31.4 installs as 2026.8.31.4.
+        run: |
+          cd /tmp && "$WHEEL_VENV/bin/python" - <<'PY'
+          import os, sys
+          from packaging.version import Version
+          import anton
+          got, expected = anton.__version__, os.environ["EXPECTED"]
+          print(f"installed={got} expected={expected}")
+          if Version(got) != Version(expected):
+              sys.exit(f"::error::wheel reports {got}, release minted {expected}")
+          PY
+
+      - name: Console script entry point works
+        # `anton` is what users type; the E2E harness uses `python -m anton`.
+        run: cd /tmp && "$WHEEL_VENV/bin/anton" --help > /dev/null
+
+      - name: Sdist builds, installs and runs
+        # uv build produces a wheel and an sdist, and both get published, so
+        # both must be verified.
+        run: |
+          uv venv --python 3.12 /tmp/sdistenv
+          uv pip install --no-cache --python /tmp/sdistenv/bin/python dist/*.tar.gz
+          cd /tmp && /tmp/sdistenv/bin/anton --help > /dev/null
+
+      - name: Installer path resolves and runs on a cold cache
+        # install.sh installs from the git tag, not the wheel, and a warm local
+        # env can hide a dependency that is no longer pulled in transitively.
+        run: |
+          uv venv --python 3.12 /tmp/gitenv
+          uv pip install --no-cache --python /tmp/gitenv/bin/python \
+            "git+https://github.com/${{ github.repository }}.git@${{ inputs.tag }}"
+          cd /tmp && /tmp/gitenv/bin/anton --help > /dev/null
+
+      - name: Run E2E scenarios against the installed wheel (live)
+        env:
+          ANTON_E2E_WHEEL_PYTHON: /tmp/wheelenv/bin/python
           ANTON_OPENAI_API_KEY: ${{ secrets.ANTON_OPENAI_API_KEY }}
           ANTON_PLANNING_PROVIDER: openai
           ANTON_CODING_PROVIDER: openai
           ANTON_PLANNING_MODEL: gpt-4.1-mini
           ANTON_CODING_MODEL: gpt-4.1-mini
         run: uv run --group dev pytest tests/e2e/ --live -v
+
+      # Hand the publish job these exact bytes. Last step, so a failed
+      # verification never leaves a publishable artifact behind.
+      - name: Upload the verified distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: verified-dist
+          path: dist/
+          if-no-files-found: error
+          retention-days: 1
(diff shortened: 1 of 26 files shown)


ZoranPandovski:

anton/core/llm/provider.py
```diff
@@ -9,6 +9,34 @@
     from anton.core.interaction.elicit import AskAnswer, AskRequest
 
 
+# Providers hold an HTTP pool. Unclosed, httpx2 prints a traceback when the
+# event loop finalizes its async generators, so every entry point closes these.
+# Strong refs on purpose: a provider collected before shutdown leaves its pool
+# to the async-generator finalizer, which is the noise we are preventing.
+_LIVE_PROVIDERS: list = []
```
tino097: Blocking: this converts previously collectable providers into a permanent leak in the one long-lived host.

`register_provider(self)` runs in both providers' `__init__`, and `unregister_provider` only runs from `aclose()`. Nothing outside the CLI drains the list.

cowork-server consumes anton in-process in a long-lived FastAPI process and calls neither `ChatSession.close()` nor `close_live_providers()` anywhere:

- `cowork/services/providers.py:1092` `build_llm_client` constructs 2-3 providers per turn, via `anton_harness._build_chat_session`.
- `cowork/handlers/responses.py:482` constructs a bare `OpenAIProvider` per hosted-org request in `_router_binding`.

So every turn and every gate-model request appends providers plus their httpx2 connection pools to a module-level list that is never drained: unbounded RSS and FD growth in the shared multi-tenant process, where before this change they were garbage collected. `unregister_provider`'s docstring ("so long-running hosts do not accumulate them") describes the intent, but no long-running host is wired to it.

The comment above argues the strong refs are needed because a provider collected before shutdown leaves its pool to the asyncgen finalizer. That is the pre-existing behaviour rather than a new leak, and it only produces exit-time noise in a process that is exiting anyway. A `weakref.WeakSet` gives the shutdown drain without the retention. If the strong refs are load-bearing, this wants a paired cowork-server change that closes both call sites.

anton/cli.py
```diff
@@ -40,6 +41,16 @@
...
 
 
+def _run_and_close(coro):
+    """Run a coroutine, then release provider HTTP pools before the loop dies."""
+    async def _wrapped():
+        try:
+            return await coro
+        finally:
+            await close_live_providers()
+    return asyncio.run(_wrapped())
```
tino097, line 59: Blocking: the primary interactive path never reaches this drain, so the traceback is still reproducible.

`_run_and_close` wraps `_onboard` and the four datasource commands, but `anton` chat goes `cli.py:482` -> `chat.run_chat` -> a bare `asyncio.run(_chat_loop(...))` at `anton/chat.py:1321`. Two ordinary paths still exit with open pools:

1. A provider or model switch mid-session: `chat.py:2069`/`2091` -> `handle_setup_models` -> `chat_session.rebuild_session`, which sets `state["llm_client"] = LLMClient.from_settings(settings)` (`anton/chat_session.py:68`) and returns a new ChatSession (`:79`). Nothing closes the outgoing session, so its providers stay pinned in `_LIVE_PROVIDERS` with live pools for the rest of the process, and `await session.close()` later only reaches the newest client.
2. `await session.close()` (`chat.py:2110`) sits after `except KeyboardInterrupt: pass` rather than in a `finally`, so any other exception escaping `_chat_loop` skips it entirely.

Either wrap `run_chat`'s `asyncio.run` the same way this helper does, or close the outgoing session inside `rebuild_session` and move `session.close()` into a `finally`.

anton/core/session.py
```diff
@@ -2114,8 +2114,14 @@ def _build_core_tools(self) -> None:
...
-        await self._reap_tracked_backends()
-        await self._scratchpads.close_all()
+        try:
+            await self._reap_tracked_backends()
+            await self._scratchpads.close_all()
+        finally:
+            # Provider clients own an HTTP pool. This runs even if the steps
+            # above raise, or the pool outlives the process.
+            if self._llm is not None:
+                await self._llm.aclose()
```
tino097, line 2124: `config.llm_client` is caller-supplied (`session.py:1147`), so `close()` now tears down transports the session does not own.

No shipping host shares a client across sessions today (cowork-server and the pod both build one per turn, the CLI one per rebuild), so this is a latent trap rather than a live break. The pattern does already exist in the suite: `tests/test_chat_scratchpad.py:93-120` shares one `mock_llm` across two sessions and closes both, and passes only because the client is a mock.

Worth either documenting that `ChatSessionConfig.llm_client` transfers ownership, or gating this on an explicit `owns_llm_client` flag, so a future host that reuses one client does not get its pools closed underneath it.

anton/core/llm/openai.py
```diff
@@ -871,6 +897,7 @@ def __init__(
             if not ssl_verify:
                 azure_kwargs["http_client"] = httpx.AsyncClient(verify=False)
             self._client = AsyncAzureOpenAI(**azure_kwargs)
+            register_provider(self)
```
tino097: `register_provider(self)` is duplicated here and at line 910, once per branch of the Azure/non-Azure `if`. A third client branch added later would silently skip registration and its pool would never be closed, which is the bug the registry exists to prevent. One call after the `if/else` covers every branch by construction.

Minor, same area: `_LIVE_PROVIDERS: list` and `register_provider(provider: object)` accept any object, while `close_live_providers` calls `provider.aclose()` on the members, so nothing states the contract the drain depends on. `list["LLMProvider"]` as a string annotation avoids the forward-reference problem.

anton/core/llm/openai.py
```diff
@@ -807,6 +809,24 @@ def _as_int(value) -> int:
     return total - read - write, read, write
 
 
+
+async def _aclose_stream(stream: object) -> None:
```
tino097, line 816: Two hardening points, neither a bug against the SDKs pinned here:

- The selection is by attribute presence, not awaitability. `getattr(stream, "close", None) or getattr(stream, "aclose", None)` short-circuits on a synchronous `close()` if one ever exists alongside an async `aclose()`; `await closer()` then raises TypeError, `except Exception` swallows it, `aclose()` is never attempted, and the connection leaks with no diagnostic, which is the failure this helper was added to prevent. Preferring `aclose` first, or checking `inspect.isawaitable`, removes it. Fine today: `AsyncStream.close` is async in openai 3.6.0.
- `CancelledError` is a `BaseException` and escapes `except Exception`. On a Ctrl+C mid-stream the `finally` at 1346 (and 1691) awaits `stream.close()`, which is itself cancelled immediately, so the socket is not released on exactly the abort path. Not a regression, since nothing closed it before, but the fix stops short there.

tests/e2e/harness.py
```diff
@@ -106,6 +108,35 @@ def timeout(self, base: float) -> float:
...
 
 
+# httpcore2 raises "generator didn't stop after athrow()" while finalizing a
+# response stream at interpreter exit (upstream, pydantic/httpx2). It is not
+# anton output, so strip it before assertions rather than weakening them.
+_UPSTREAM_ASYNCGEN_NOISE_START = "an error occurred during closing of asynchronous generator"
+_UPSTREAM_ASYNCGEN_NOISE_END = "generator didn't stop after athrow()"
+
+
+def _strip_upstream_asyncgen_noise(text: str) -> tuple[str, bool]:
```
tino097, line 121: This can delete a real traceback. Once `skipping` is True every line is dropped until `generator didn't stop after athrow()`, so if that end marker never appears the entire remainder of stderr goes with it.

Two concrete ways that happens: an upstream wording change after an httpcore2 bump, and the `subprocess.TimeoutExpired` path at line 176 where stderr is cut off mid-block. `assert_not_output(result, "Traceback (most recent call last)")` then passes on a run that genuinely crashed, because the traceback was stripped before the assertion saw it.

Bounding the skip fixes it: require both markers to be present before removing anything, or cap the skipped block at a line count.

tests/e2e/harness.py
```diff
@@ -138,13 +179,16 @@ def run_anton(
...
+    stderr, upstream_noise = _strip_upstream_asyncgen_noise(stderr)
+
     return RunResult(
         returncode=returncode,
         stdout=stdout,
         stderr=stderr,
         duration=time.monotonic() - t0,
         timed_out=timed_out,
         cmd=cmd,
+        upstream_asyncgen_noise=upstream_noise,
```
tino097, line 204: `upstream_asyncgen_noise` is written here and declared at line 33, and never read anywhere. Nothing asserts on it.

That leaves the change without a regression guard: if anton starts emitting an asyncgen finalization error again, including from the stream-closing code added in this PR, every e2e scenario stays green and the flag records it into a field no assertion consults. Asserting `not result.upstream_asyncgen_noise` in the live scenarios is the test this change wants, and it is the only place the original symptom would be caught.

.github/workflows/tests_e2e_release.yml
```diff
@@ -1,33 +1,116 @@
...
+
+      # Hand the publish job these exact bytes. Last step, so a failed
+      # verification never leaves a publishable artifact behind.
+      - name: Upload the verified distributions
+        uses: actions/upload-artifact@v4
+        with:
+          name: verified-dist
+          path: dist/
+          if-no-files-found: error
+          retention-days: 1
```
tino097: One day of retention on the only publishable bytes makes a delayed publish unrecoverable.

`publish` runs in `environment: pypi` (`release.yml:60`). If that environment gains a required reviewer or a wait timer, or the job is re-run the next day after a transient upload failure, `download-artifact` finds no `verified-dist` and the step fails. The tag is already pushed by then, and `install.sh:80` installs from `git+https://github.com/mindsdb/anton.git@${LATEST_TAG}`, so users can install a version that does not exist on PyPI. Recovery means re-running the whole verify job, live provider e2e included.

A longer retention, or a documented recovery path in the workflow comment, would cover it.

.github/workflows/tests_e2e_release.yml
```diff
@@ -1,33 +1,116 @@
...
+        # env can hide a dependency that is no longer pulled in transitively.
+        run: |
+          uv venv --python 3.12 /tmp/gitenv
+          uv pip install --no-cache --python /tmp/gitenv/bin/python \
+            "git+https://github.com/${{ github.repository }}.git@${{ inputs.tag }}"
+          cd /tmp && /tmp/gitenv/bin/anton --help > /dev/null
+
+      - name: Run E2E scenarios against the installed wheel (live)
+        env:
+          ANTON_E2E_WHEEL_PYTHON: /tmp/wheelenv/bin/python
```
tino097: This hardcodes the path the job already defines as `WHEEL_VENV` (line 26); `/tmp/sdistenv` and `/tmp/gitenv` are inline literals too. Change `WHEEL_VENV` to avoid a collision or move off /tmp and the wheel install and the resolve check follow it while this step keeps pointing at the old path, handing `subprocess` a nonexistent interpreter. The failure would read as a broken e2e rather than a stale path. `${{ env.WHEEL_VENV }}/bin/python` keeps them together.

tino097 (changes requested):
Reviewed at 6536339. The httpx2 move itself looks right: the import swap is complete, `aclosing` on the responses path is the correct shape, the anthropic provider correctly gets no `_aclose_stream` because it streams under `async with self._client.messages.stream(...)`, and both bumped majors still expose an awaitable `close()` (openai 3.6.0 `_base_client.py:1575`, anthropic 1.3.0 `_client.py:826`). Verifying the built wheel instead of the source tree is a real improvement.

Two things I would want resolved before this merges, both inline:

1. `_LIVE_PROVIDERS` holds strong refs and is drained only by the CLI, so it converts collectable providers into a permanent leak in cowork-server, which never closes a session.
2. The main `anton` chat path never reaches the drain, so the exit traceback this PR exists to remove is still reproducible after a mid-session provider switch.

Two more that need no code change here:

- Nothing tests the close path. `git grep` finds zero references to `aclose`, `register_provider` or `close_live_providers` under `tests/`, so a green `run-tests` says nothing about the fix, and the bare `except Exception: pass` in both `LLMClient.aclose` and `close_live_providers` means a future SDK rename fails silently and the noise returns with no signal. A stub-client test asserting the registry drains and `close()` was awaited would pin both.
- `publish-staging.yml` is untouched: it still builds from a clean checkout and uploads rc's with no wheel verification, and per its own header comment cowork-server staging pins those rc versions. The argument for gating prod applies there too, so if that is deliberately out of scope it is worth saying so.

One question on the new gating: the tag is minted by `auto-release` before `verify-wheel`, so a provider outage or a flaky live scenario now leaves a pushed tag with nothing on PyPI. The comment removed from the old publish step notes that a re-run on an already-tagged head leaves two CalVer tags on one commit and `git describe` resolves the older one. What is the intended recovery there?

tino097 (approved):
LGTM

PR #431, opened 2026-09-02
ENG-2243: stamp a per-attempt turn id so the tool→turn join is exact

alecantu7:
Closes [ENG-2243](https://linear.app/mindsdb/issue/ENG-2243).

## What

`turn_completed` and `tool_completed` both carry `(conversation_id, turn_index)`, and ENG-1486 stamped that pair specifically so a tool row could be joined to its parent turn. **The pair is not unique.** This adds `TurnCost.attempt_id`, stamped on both events as `turn_attempt_id`, which is unique per turn *execution*.

## Why

`turn_index` is a **position in the history**, not a turn identifier — `_turn_count` is seeded by counting the user messages the session was handed (`session.py:1203`), and cowork-server rebuilds the ChatSession on every turn. A retried or cancelled attempt therefore arrives with the same history and stamps the same `turn_index`.

Prod PostHog 424726, 2026-08-28 → 09-01 UTC, AsyncMock rows excluded:

| surface | event rows | distinct `(cid, turn_index)` | inflation | worst key |
| -- | --: | --: | --: | --: |
| **desktop** | 923 | 789 | **14.5%** | **16 rows** |
| web | 314 | 314 | 0.0% | 1 |
| cli | 16 | 13 | (n=16) | 2 |

```
tool rows in window                                  9,911
  ...whose (cid, turn_index) matches >1 turn row     1,832   (18.5%)
```

**Web at 0% is the control** — the pod runs one turn per process (`cloud_turn/__main__.py`) and is handed its history explicitly, so nothing is rebuilt mid-conversation. The colliding keys carry exactly the outcome sets a retry produces (`cancelled, completed`; `error, completed, spend_ceiling`), each row with a distinct `tokens_total`, spanning up to 34 hours — **distinct real attempts, not a double-emit.** That distinction matters: the fix for a double-emit would have been dedupe, which would have been wrong.

## ⚠️ This implements the ticket's recommended option — object here if you disagree

ENG-2243 lists three directions and flags the choice as needing sign-off. This PR takes **option 1: add a field, leave `turn_index` alone.** Rejected alternatives and why:

- **Make `turn_index` monotonic** — changes the meaning of a shipped field. The Langfuse trace name is composed `{harness}:turn-{turn_id}` from it, and `core/artifacts/store.py:314` indexes on it. Also, cowork-server has its own unrelated `turn_index` (0-based, assistant-only, for `DELETE /conversations/{id}/turns/{turn_index}`) — there are already two vocabularies, and adding a third meaning to ours is the wrong direction.
- **Document only** — leaves the join permanently ambiguous.

## How

- `TurnCost.attempt_id`, a **`default_factory`** rather than an argument at each call site. There are two construction sites today (`session.py:3632` non-streaming, `:4145` streaming) and a third that forgot would silently reintroduce the collision. It also makes it a stamped-at-books-open fact like `turn_index`, so a late finalizer reports the id of the turn whose books it holds — the `#309 review` failure mode.
- **Random, not a counter.** Any session-local counter resets on the same rebuild that causes the bug, so nothing derived from session state can be unique across attempts. 16 hex chars = 64 bits, same width and reasoning as the `aid` install fingerprint.
- `_emit_tool_completed` reads the id from the **same books** as `turn_index`, so a tool row and its turn row cannot disagree. Empty when no books are open — a tool that ran outside a turn is not an attempt and must not get an id that looks like one.

### Writer inventory (convention 9)

The behaviour is gated on where the id is stamped, so: `TurnCost(` is constructed in **two** places, both in `anton/core/session.py`, both covered by the factory. Grepped `cowork-server`, `cowork`, `scratchpad-controller` and `cowork_evals` — none construct `TurnCost`; cowork-server's nine `turn_index` references are all its own unrelated conversation-delete field plus one hardcoded `turn_index=1` in the hermes artifact path.

## Verification

- `tests/test_turn_attempt_id.py` (new, **10 tests**). The load-bearing one drives the **actual production shape**: two sessions seeded with the same `initial_history`, asserting `turn_index` repeats (correct — it's a position) while the attempt id does not. `test_two_turns_in_ONE_session_get_different_ids` is the strictest: every other test rebuilds the session per turn, so a per-session id would have satisfied them all while leaving every CLI turn sharing one id.
- **Mutation-verified, 8 mutations, all caught:**

  | mutation | result |
  | -- | -- |
  | drop `turn_attempt_id` from `turn_completed` | 3 failed |
  | drop it from `tool_completed` | 2 failed |
  | constant instead of `default_factory` | 2 failed |
  | **derive the id from `turn_index`** (the plausible wrong fix) | 2 failed |
  | **a stable per-SESSION id** (passed the whole suite before `7c136000`) | 1 failed |
  | `(self._turn_cost or tc).attempt_id` at the turn emit — read live state instead of the handed books | 1 failed |
  | revert to `uuid.uuid4().hex[:16]` | 1 failed, naming hex position 12 |
  | drop `attempt=` from the `turn_cost` log line | 1 failed (**passed 41 tests before this round**) |

- **A warning worth more than the tests, from `c0cf6b9e`: `git merge-tree` undercounts.** It reported 4 conflict regions against #432; there were 5 edits. The fifth was an exhaustive key-set assertion living in the *other* branch's new test file — a file `merge-tree` sees as a clean add, so no conflict count can reveal it. I made the same 4-region estimate off the same tool on #432, so it caught me twice. Anyone widening `tool_completed`'s payload should grep both branches for the key-set assertions rather than trusting a conflict count.

- `test_payload_is_exactly_the_seven_keys_all_strings` → `..._eight_keys...` → `..._ten_keys...` after #432 merged, updated **deliberately**, not loosened — "no surprise keys" is the property that keeps tool arguments and result content out. Same call ENG-1945 made going six → seven.
- One of my own tests was wrong first time: it used `pytest.raises`, but anton's turn loop *swallows* the exception and records the terminal on the event, so it passed vacuously. Rewritten to assert on the emitted row, with a message that fails loudly if the harness stops reaching the error path.
- Full suite on `f3bfa6ac`: **3,007 passed, 31 skipped, 0 failed** — including `test_cloud_turn_process`, which usually fails locally on macOS.

## Expected side effects that are NOT regressions

- **A new high-cardinality PostHog property** — one distinct value per turn, higher than `conversation_id`'s one-per-conversation. Fine as a join key and as `COUNT(DISTINCT …)`; **do not use it as a breakdown dimension.** Flagging it because PostHog auto-creates property definitions on first ingest.
- **Absent on pre-fix builds.** Readers must treat missing as "unknown", never as a value. Documented in the register in `anton/analytics.py`.
- **`turn_index` still repeats, on purpose.** That is not a bug left unfixed; it is what the trace name and artifact index want.

## Not in this PR

- **The skills half of the ticket's `Done when`** — "`turn-cost` and `harness-measure` state whether each figure counts attempts or turns". Those skills do not live in this repo, so it cannot ship here. Saying so rather than silently dropping it: it still needs doing, wherever they live.
- **The Langfuse hop stays ambiguous.** The ticket's consequence 3 is "several traces named `anton:turn-6` in one session", and the register advertises a PostHog→Langfuse forensics hop. An analyst who finds the attempt by `turn_attempt_id` still gets N identically-named traces and no way to pick one. `TraceContext.metadata: dict[str, str]` already forwards to `Langfuse-Metadata`, so it is reachable — but it needs the books opened before the trace context is built, which is a real ordering change. Scoped out deliberately, not overlooked.
- **Backfill.** Historical rows keep the ambiguity.
- **Whether `core/artifacts/store.py`'s turn index has a live collision** from the same root cause. Suspected, unverified, scoped out on the ticket.

## Security check

Performed on the diff. The new field is a random 64-bit hex string from `secrets.token_hex(8)` — no machine identity (unlike `aid`, which hashes the MAC), no user content, no paths, no credentials.

**Corrected in review:** this shipped as `uuid.uuid4().hex[:16]` and this section claimed 64 bits. That is 60 — hex position 12 is uuid4's version nibble, the literal `4` in every id ever generated (measured 2000/2000) — and the comparison to `aid` was backwards, since `aid`'s primary path is `sha256(...).hexdigest()[:16]` and uuid4 is only its fallback. Now `secrets.token_hex(8)`, which is 64 real bits. Collision risk was never the issue either way (the birthday bound at 60 bits is ~1e9 attempts); the claim being wrong was. It rides two events that already carry `conversation_id` and `turn_index`, so no new destination and no widened identity surface: the join it enables is between two events that already shared a key. Nothing new is logged.

## Four `Done when` items with no test, true by construction

Named here rather than covered by four more tests, so review can accept or reject that deliberately:

| `Done when` | Why it holds without a test |
| -- | -- |
| a **cancelled** turn gets its own id | `_emit_turn_cost` fires from the turn's `finally`, so every exit path — completion, hand-back, generator close, cancel, error — closes books that were stamped at open. The cancel terminal is covered for `ended_by` in `test_turn_cost_terminals.py`; the id cannot differ by terminal. |
| a **spend-ceiling** turn gets its own id | Same `finally`. The ceiling is a terminal mark, not a separate emit path. |
| the **Langfuse trace name** is unchanged | `turn_index` is untouched by this PR, and the trace name is built from it. Grep confirms no other reader of `attempt_id`. |
| the **artifact-store index** is unchanged | Same reason — `core/artifacts/store.py` reads `turn_index`, which this PR does not modify. |

Only the error terminal has an explicit id test (`test_the_attempt_id_survives_a_non_completed_terminal`), and that is enough given the shared `finally`.

### DIFF
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -2919,13 +2919,20 @@ class ChatSession:
         except Exception:  # pragma: no cover - defensive
             _root_cause_fields = {}
         logger.info(
-            "turn_cost session=%s turn=%d ended_by=%s verification_skipped=%s "
+            # `attempt` alongside `turn`: `turn_index` is a history position,
+            # so two attempts of one turn used to produce two indistinguishable
+            # log lines (ENG-2243). This line is also the ONLY forensics
+            # surface that survives the collector allowlist (see the note
+            # below) and, per ENG-2193, the only one a desktop customer has —
+            # `cowork-server.log`, not PostHog. Without it the new analytics
+            # property has no fallback at all.
+            "turn_cost session=%s turn=%d attempt=%s ended_by=%s verification_skipped=%s "
             "grace_granted=%s grace_tokens=%d "
             "verifier_failure=%s verifier_error_type=%s tokens_total=%d "
             "input=%d output=%d cache_read=%d cache_creation=%d "
             "llm_calls=%d rounds=%d continuations=%d peak_context=%d duration_ms=%d "
             "by_role=%s %s",
-            self._session_id, turn_index, tc.ended_by,
+            self._session_id, turn_index, tc.attempt_id, tc.ended_by,
             str(tc.verification_skipped).lower(),
             tc.grace_granted or "-", tc.grace_tokens,
             tc.verifier_failure, tc.verifier_error_type, tc.total_tokens,
@@ -3183,17 +3190,27 @@ class ChatSession:
         is reported as ``"unknown"`` rather than coerced either way.
 
         Payload is deliberately name + verdict + duration + exception CLASS
-        + surface (a closed enum, ENG-1945) plus the two join keys, and
-        nothing else. Arguments, result content and ``str(exc)`` routinely
+        + surface (a closed enum, ENG-1945) + the root-cause tier and class
+        (both closed vocabularies, ENG-2247) plus the THREE join keys
+        (``conversation_id``, ``turn_index``, ``turn_attempt_id``), and
+        nothing else — ten keys, and the exact-keys assertion in
+        ``test_tool_completed.py`` is what keeps that number honest. This
+        paragraph is the privacy-audit enumeration, so keep it in step with
+        the payload: arguments, result content and ``str(exc)`` routinely
         carry file paths, user data and credentials-adjacent strings — none
         of them may ever appear here.
 
-        ``conversation_id`` / ``turn_index`` mirror ``turn_completed``'s
-        values exactly (same names, same derivation), so a tool failure spotted
-        in PostHog joins to its parent turn row there and, via
-        ``conversation_id`` → Langfuse ``sessionId``, to the gateway trace of
-        the turn it happened in. Both already ride ``turn_completed`` through
-        this same sink — no new privacy surface.
+        ``conversation_id`` / ``turn_index`` / ``turn_attempt_id`` mirror
+        ``turn_completed``'s values exactly (same names, same derivation), so a
+        tool failure spotted in PostHog joins to its parent turn row there and,
+        via ``conversation_id`` → Langfuse ``sessionId``, to the gateway trace
+        of the turn it happened in. All three already ride ``turn_completed``
+        through this same sink — no new privacy surface.
+
+        Prefer ``turn_attempt_id`` for that join: ``turn_index`` is a history
+        position and repeats across every retry of the same turn, so
+        ``(conversation_id, turn_index)`` matched more than one turn row for
+        18.5% of these rows before ENG-2243.
         """
         try:
             # Same settings resolution as `_emit_turn_cost` above: the
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -3126,6 +3126,16 @@ class ChatSession:
                 # forensics.
                 conversation_id=str(self._session_id or ""),
                 turn_index=str(turn_index),
+                # Unique per turn EXECUTION, where `turn_index` is a history
+                # position that repeats across retries (ENG-2243). This is the
+                # key `tool_completed` joins on; `turn_index` stays the field
+                # the Langfuse trace name and the artifact index are built
+                # from, so both readings remain available:
+                #   COUNT(DISTINCT turn_attempt_id) -> attempts
+                #   COUNT(DISTINCT conversation_id, turn_index) -> turns
+                # Absent on a pre-ENG-2243 build: read that as "unknown",
+                # never as a value.
+                turn_attempt_id=tc.attempt_id,
             )
         except Exception:
             # Reporting must never affect the turn that just ran.
@@ -3211,6 +3221,12 @@ class ChatSession:
             turn_index = (
                 getattr(_tc, "turn_index", 0) or (self._turn_count + 1)
             )
+            # Read from the SAME books as `turn_index` above, so a tool row and
+            # its parent turn row can never disagree about which attempt they
+            # belong to (ENG-2243). Empty outside a turn — the tool ran with no
+            # books open, which is not an attempt and must not be given an id
+            # that looks like one.
+            turn_attempt_id = str(getattr(_tc, "attempt_id", "") or "")
             send_event(
                 settings,
                 "tool_completed",
@@ -3236,6 +3252,12 @@ class ChatSession:
                 # and on an unmigrated handler; see `_tool_failure_cause`.
                 root_cause_tier=_rc_tier,
                 root_cause_class=_rc_class,
+                # Makes the tool -> turn join exact (ENG-2243). Before this,
+                # `(conversation_id, turn_index)` matched every retry of the
+                # same turn, so 18.5% of these rows joined to more than one
+                # turn row and "which tools ran in the attempt that hit the
+                # spend ceiling" had no answer.
+                turn_attempt_id=turn_attempt_id,
             )
         except Exception:
             # Analytics must never affect the tool call that just ran.
--- a/anton/core/turn_cost.py
+++ b/anton/core/turn_cost.py
@@ -37,6 +37,7 @@ Stated gaps (by design, documented on ENG-1288):
 from __future__ import annotations
 
 import time
+import uuid
 from dataclasses import dataclass, field
 
 from anton.core.llm.provider import Usage
@@ -168,6 +169,28 @@ class TurnCost:
     # there gave the abandoned turn a LATER, unrelated turn's index — the
     # cost→Langfuse hop then pointed at the wrong turn (#309 review follow-up).
     turn_index: int = 0
+    # Unique per turn EXECUTION. `turn_index` is only a POSITION in the history
+    # — `_turn_count` is seeded by counting the user messages the session was
+    # handed — and cowork-server rebuilds the session every turn, so a retried
+    # or cancelled attempt arrives with the same history and stamps the same
+    # `turn_index` (ENG-2243). Measured on prod 2026-08-28..09-01: that collided
+    # on 14.5% of desktop turn keys (worst: 16 rows on one key, spanning 34
+    # hours) and left 18.5% of `tool_completed` rows joining to more than one
+    # `turn_completed` row — the join ENG-1486 stamped the pair to enable.
+    #
+    # Random, not a counter, deliberately: ANY session-local counter resets on
+    # that same rebuild, so nothing derived from session state can be unique
+    # across attempts. 16 hex chars = 64 bits, the same width and reasoning as
+    # the `aid` install fingerprint (see `anton/analytics.py`) — collision-free
+    # at our volume, short enough to read.
+    #
+    # A `default_factory` rather than a value passed in at each construction
+    # site: there are two today (the streaming and non-streaming turns in
+    # `session.py`) and a third that forgot would silently reintroduce the
+    # collision this field exists to remove. Also makes it a stamped-at-open
+    # fact like `turn_index` above, so a late finalizer reports the id of the
+    # turn whose books it holds rather than whichever turn owns the slot now.
+    attempt_id: str = field(default_factory=lambda: uuid.uuid4().hex[:16])
     # When the turn ended. None until resolved. A late finalizer cannot know the
     # real end, so it falls back to `last_activity_monotonic` (the last LLM
     # call) — understating but bounded, rather than measuring up to whenever
--- a/anton/core/turn_cost.py
+++ b/anton/core/turn_cost.py
@@ -36,8 +36,8 @@ Stated gaps (by design, documented on ENG-1288):
 
 from __future__ import annotations
 
+import secrets
 import time
-import uuid
 from dataclasses import dataclass, field
 
 from anton.core.llm.provider import Usage
@@ -180,9 +180,17 @@ class TurnCost:
     #
     # Random, not a counter, deliberately: ANY session-local counter resets on
     # that same rebuild, so nothing derived from session state can be unique
-    # across attempts. 16 hex chars = 64 bits, the same width and reasoning as
-    # the `aid` install fingerprint (see `anton/analytics.py`) — collision-free
-    # at our volume, short enough to read.
+    # across attempts. `token_hex(8)` is 16 hex chars of 64 real bits —
+    # collision-free at our volume (the birthday bound is ~5e9 attempts), short
+    # enough to read in a log line.
+    #
+    # NOT `uuid.uuid4().hex[:16]`, which this field shipped as in review and
+    # which is 60 bits, not 64: hex position 12 is uuid4's version nibble, so
+    # it is the literal `4` in every id ever generated (measured 2000/2000).
+    # The width claim was wrong and so was the comparison to the `aid` install
+    # fingerprint — `aid`'s primary path is `sha256(...).hexdigest()[:16]`
+    # (`analytics.py:205`), genuinely 64 bits; its uuid4 form is only the
+    # fallback when no stable node id is available.
     #
     # A `default_factory` rather than a value passed in at each construction
     # site: there are two today (the streaming and non-streaming turns in
@@ -190,7 +198,7 @@ class TurnCost:
     # collision this field exists to remove. Also makes it a stamped-at-open
     # fact like `turn_index` above, so a late finalizer reports the id of the
     # turn whose books it holds rather than whichever turn owns the slot now.
-    attempt_id: str = field(default_factory=lambda: uuid.uuid4().hex[:16])
+    attempt_id: str = field(default_factory=lambda: secrets.token_hex(8))
     # When the turn ended. None until resolved. A late finalizer cannot know the
     # real end, so it falls back to `last_activity_monotonic` (the last LLM
     # call) — understating but bounded, rather than measuring up to whenever
--- a/anton/core/turn_cost.py
+++ b/anton/core/turn_cost.py
@@ -188,9 +188,11 @@ class TurnCost:
     # which is 60 bits, not 64: hex position 12 is uuid4's version nibble, so
     # it is the literal `4` in every id ever generated (measured 2000/2000).
     # The width claim was wrong and so was the comparison to the `aid` install
-    # fingerprint — `aid`'s primary path is `sha256(...).hexdigest()[:16]`
-    # (`analytics.py:205`), genuinely 64 bits; its uuid4 form is only the
-    # fallback when no stable node id is available.
+    # fingerprint: `get_installation_id()` in `anton/analytics.py` takes the
+    # `sha256(str(node)).hexdigest()[:16]` branch whenever the machine has a
+    # real MAC, which is genuinely 64 bits; its `uuid4().hex[:16]` branch is
+    # the fallback for a host with no MAC to read (Docker with stripped
+    # networking), and carries the same 60-bit shortfall described above.
     #
     # A `default_factory` rather than a value passed in at each construction
     # site: there are two today (the streaming and non-streaming turns in
--- a/tests/test_turn_attempt_id.py
+++ b/tests/test_turn_attempt_id.py
@@ -159,11 +159,23 @@ async def test_the_attempt_id_survives_a_non_completed_terminal(workspace):
 
 
 async def test_tool_row_and_turn_row_agree_on_the_attempt_id(workspace):
-    """Both read the same books, so the join cannot be ambiguous.
+    """Both rows carry the same attempt, so the join cannot be ambiguous.
 
-    Reading live session state in `_emit_tool_completed` instead would let a
-    late finalizer pair a tool row with a different attempt — the same class of
-    bug `turn_index`'s stamped-at-open comment records (#309 review).
+    How each side gets there differs, and the earlier version of this docstring
+    had it backwards (#431 review). `_emit_tool_completed` DOES read live
+    session state — `_tc = getattr(self, "_turn_cost", None)` at
+    `session.py:3229`. It is safe because a tool row is always emitted
+    synchronously inside its own live turn and the owning turn nulls the books
+    at close, so the read is never late; not because it reads stamped books.
+
+    The stamped-at-open guarantee is load-bearing on the TURN side, where the
+    emit genuinely can be late — see
+    `test_a_late_finalizer_reports_its_own_attempt_not_the_live_turns`.
+
+    The distinction matters for whoever changes this next: if the tool emit
+    were ever moved off the synchronous path (queued or batched), this test
+    would still pass while the join silently started pairing tool rows with a
+    neighbouring attempt.
     """
     from unittest.mock import AsyncMock
 
@@ -224,6 +236,27 @@ def test_every_turncost_gets_an_id_without_the_caller_passing_one():
     assert all(len(i) == 16 and all(c in "0123456789abcdef" for c in i) for i in ids)
 
 
+def test_all_sixteen_characters_carry_entropy():
+    """The width claim in `turn_cost.py` has to be true, not just plausible.
+
+    The field first shipped as `uuid.uuid4().hex[:16]`, whose comment claimed
+    64 bits. It is 60: hex position 12 is uuid4's version nibble, so it is the
+    literal `4` in every id ever generated. The test above could not catch that
+    — 200 uuid4-derived ids are all unique and all lowercase hex, and a fixed
+    nibble costs 4 bits without failing either assertion.
+
+    So assert the property the comment claims: every position varies. 200
+    samples over 16 possible values makes a false failure ~(15/16)**200 ≈ 3e-6
+    per position, and a positionally-fixed nibble fails deterministically.
+    """
+    ids = [TurnCost().attempt_id for _ in range(200)]
+    fixed = [pos for pos in range(16) if len({i[pos] for i in ids}) == 1]
+    assert not fixed, (
+        f"hex positions {fixed} are constant across 200 ids — the field is "
+        "narrower than the 64 bits its comment claims"
+    )
+
+
 def test_the_id_is_stamped_at_books_open_not_read_at_emit():
     """Stable for the life of the books, like `turn_index`.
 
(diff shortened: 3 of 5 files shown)


alecantu7:
## Self-review (convention 6)

Four findings. One fixed in `02e2c772`, three considered and declined with reasons.

### F1 — the non-streaming `turn()` path was untested — **FIXED** in `02e2c772`

`turn()` opens its own books at `session.py:3632` and calls `_emit_turn_cost` separately from `turn_stream`. Every test in the first commit drove `turn_stream` only. The `default_factory` makes the id correct there by construction — but "correct by construction" on one path is exactly how the other path ends up missing a field, which is the shape of the original defect.

Added `test_the_non_streaming_turn_path_stamps_an_id_too`: two runs on the same seeded history, asserting `turn_index` repeats and the attempt id does not. Mutation-verified — replacing the factory with a constant fails it.

### F2 — `tc.attempt_id` is a new AttributeError surface inside a silent `except` — declined

`_emit_turn_cost` is wrapped in `except Exception: pass`, so **any** attribute that does not resolve silently drops the whole `turn_completed` event. Adding a field adds a way to lose it.

Checked before declining:
- `TurnCost` is never pickled, dilled, or reconstructed from a dict — grepped, no hits. So there is no older-shape instance that could arrive without the field.
- No test or caller feeds a `TurnCost` stand-in to the emit; the `SimpleNamespace` uses in `test_turn_cost.py` are `usage` objects, not books.
- Full suite green.

Declined because the surrounding code reads ~20 attributes off `tc` directly (`tc.ended_by`, `tc.error_type`, `tc.verifier_failure`, …). Being defensive on only the newest field would be inconsistent and would *hide* a genuine contract violation rather than surface it. The contract is "`tc` is a real `TurnCost`", and it holds.

### F3 — new high-cardinality PostHog property — declined, documented instead

One distinct value per turn, higher cardinality than `conversation_id`'s one-per-conversation. There is no cheaper option that still makes the join exact — that is what "unique per attempt" means. Recorded in the PR body and the `analytics.py` register as: fine as a join key and in `COUNT(DISTINCT …)`, **not** a breakdown dimension.

### F4 — the ticket's skills half cannot ship here — declined, called out

ENG-2243's `Done when` includes "`turn-cost` and `harness-measure` state whether each figure counts attempts or turns". Those skills are not in this repo (`.claude/skills/` does not exist here), so the work cannot land in this PR. It still needs doing wherever they live — flagged rather than silently dropped, since a `Done when` item quietly unmet is how a ticket closes on four of five criteria.

---

### One thing a reviewer should push on

The **⚠️ option choice** in the PR body. I implemented ENG-2243's recommendation (add a field, leave `turn_index` alone) without waiting for sign-off, because it is the only one of the three that does not move the Langfuse trace name or the artifact-store index. If you would rather `turn_index` became monotonic, this PR is the wrong shape and should be closed rather than amended — that option needs a writer inventory across anton + cowork-server + the pod.

### Verification on the final head

- Full suite, fresh `uv sync --frozen --all-groups` venv: **2,898 passed, 31 skipped, 0 failed** (includes `test_cloud_turn_process`, which usually fails locally on macOS).
- 5 mutations total, all caught: drop the stamp from either event, constant instead of factory, derive the id from `turn_index` (the plausible wrong fix), and F1's factory mutation.

anton/core/session.py
```diff
@@ -3110,6 +3110,16 @@ def _emit_turn_cost(
...
+                # Unique per turn EXECUTION, where `turn_index` is a history
+                # position that repeats across retries (ENG-2243). This is the
+                # key `tool_completed` joins on; `turn_index` stays the field
+                # the Langfuse trace name and the artifact index are built
+                # from, so both readings remain available:
+                #   COUNT(DISTINCT turn_attempt_id) -> attempts
+                #   COUNT(DISTINCT conversation_id, turn_index) -> turns
+                # Absent on a pre-ENG-2243 build: read that as "unknown",
+                # never as a value.
+                turn_attempt_id=tc.attempt_id,
```
tino097, line 3145: Reading the passed-in books here rather than `self._turn_cost` is the load-bearing choice, and nothing tests it.

`session.py:4606` calls `_emit_turn_cost(expected=_turn_cost_books, exc=_turn_exc)` on the abandoned-generator path, with a comment saying the guard exists "so a late finalizer can't close a newer turn's" books. That is exactly the case where `tc is not self._turn_cost`, and it is the case `attempt_id`'s stamped-at-open design exists for.

`test_the_id_is_stamped_at_books_open_not_read_at_emit` doesn't cover it — it constructs a `TurnCost`, mutates it, and asserts the field is stable. It never invokes `_emit_turn_cost`. So changing this line to `self._turn_cost.attempt_id` would pass the whole new module while stamping a newer turn's id on an abandoned turn's row, reintroducing the `#309` mis-attribution. `turn_index` has a guard for that class in `test_turn_cost_terminals.py`; `attempt_id` doesn't.

A test that opens books, opens a second set, then emits the first would pin it.

anton/core/session.py
```diff
@@ -3110,6 +3110,16 @@ def _emit_turn_cost(
                 # forensics.
                 conversation_id=str(self._session_id or ""),
                 turn_index=str(turn_index),
+                # Unique per turn EXECUTION, where `turn_index` is a history
```
tino097, line 3136: Separate point on the same emit: the attempt id doesn't reach the `turn_cost` structured log line ~80 lines above, which still logs `session=%s turn=%d ended_by=%s ...` — the ambiguous pair and nothing else.

Two attempts of one turn therefore produce two indistinguishable log lines, which is the ambiguity this PR removes on the analytics side left in place on the only local forensics surface. The comment further down this method explicitly relies on that line surviving when properties don't reach PostHog (ENG-1355), so the new property has no fallback at all.

Worth doing because of what ENG-2193 just established: for a desktop customer the diagnostic channel is `cowork-server.log`, not PostHog. One more format-string argument.

anton/core/session.py
```diff
@@ -3195,6 +3205,12 @@ def _emit_tool_completed(
             turn_index = (
                 getattr(_tc, "turn_index", 0) or (self._turn_count + 1)
             )
+            # Read from the SAME books as `turn_index` above, so a tool row and
+            # its parent turn row can never disagree about which attempt they
+            # belong to (ENG-2243). Empty outside a turn — the tool ran with no
+            # books open, which is not an attempt and must not be given an id
+            # that looks like one.
+            turn_attempt_id = str(getattr(_tc, "attempt_id", "") or "")
```
tino097, line 3246: This reads live session state (`self._turn_cost`), while the turn side at 3122 reads the books handed to it. `test_tool_row_and_turn_row_agree_on_the_attempt_id`'s docstring says the opposite — see the note on that test.

Behaviour is right: a tool row is always emitted synchronously inside its own live turn, and the owner nulls the books at close (`session.py:2724`), so this can't pick up a neighbouring attempt. But the safety comes from the read never being late, not from reading stamped books, and that distinction is what the docstring gets backwards.

anton/core/session.py
```diff
@@ -3220,6 +3236,12 @@ def _emit_tool_completed(
                 # and on an unmigrated handler; see `_tool_failure_cause`.
                 root_cause_tier=_rc_tier,
                 root_cause_class=_rc_class,
+                # Makes the tool -> turn join exact (ENG-2243). Before this,
+                # `(conversation_id, turn_index)` matched every retry of the
+                # same turn, so 18.5% of these rows joined to more than one
+                # turn row and "which tools ran in the attempt that hit the
+                # spend ceiling" had no answer.
+                turn_attempt_id=turn_attempt_id,
```
tino097, line 3277: `_emit_tool_completed`'s own docstring (~3170) still reads "name + verdict + duration + exception CLASS + surface (a closed enum, ENG-1945) plus the two join keys, and nothing else". The payload is ten keys now — two arrived with #432's merge, this one makes three — and the join keys are three, not two. The following sentence naming `conversation_id` / `turn_index` as what mirrors `turn_completed` is also short by this field.

That paragraph is the privacy-audit enumeration, so a short list is the wrong kind of stale. The module register in `analytics.py` and the test module docstring were both widened; this one is the straggler.

Smaller thing in the register: it says `turn_attempt_id` is "Empty when the tool ran with no turn books open", but both emit sites (3986, 5362) are inside the tool loop, so that state is unreachable. Meanwhile the divergence that *is* reachable goes undocumented — with books closed, `turn_index` would fall back to `_turn_count + 1` while this would be `""`, so the two join keys disagree about whether a parent exists.

anton/core/turn_cost.py
```diff
@@ -168,6 +169,28 @@ class TurnCost:
...
+    # the `aid` install fingerprint (see `anton/analytics.py`) — collision-free
+    # at our volume, short enough to read.
+    #
+    # A `default_factory` rather than a value passed in at each construction
+    # site: there are two today (the streaming and non-streaming turns in
+    # `session.py`) and a third that forgot would silently reintroduce the
+    # collision this field exists to remove. Also makes it a stamped-at-open
+    # fact like `turn_index` above, so a late finalizer reports the id of the
+    # turn whose books it holds rather than whichever turn owns the slot now.
+    attempt_id: str = field(default_factory=lambda: uuid.uuid4().hex[:16])
```
tino097: "16 hex chars = 64 bits" is 60. `uuid4().hex[:16]` spans hex positions 0-15, and position 12 is the version nibble — always the literal `4` — so it is 15 random nibbles. Every id in production will have `4` as its 13th character.

The `aid` comparison is backwards too: `aid`'s primary path is `hashlib.sha256(...).hexdigest()[:16]`, which is genuinely 64 bits. `uuid4().hex[:16]` only appears there as a fallback. So "same width and reasoning as the `aid` install fingerprint" is wrong on the reasoning, not just the arithmetic — same claim in the PR body's security check.

Collision risk is unaffected: the birthday bound at 60 bits is ~1e9 attempts. `secrets.token_hex(8)` is a real 64 bits, cheaper than building a UUID object, and makes the comment true as written. `test_every_turncost_gets_an_id_without_the_caller_passing_one` checks length and charset only, so it can't catch the fixed nibble either way.

tests/test_turn_attempt_id.py
```diff
@@ -0,0 +1,289 @@
...
+    assert rows[0]["turn_attempt_id"]
+
+
+# ── The join it exists to fix ────────────────────────────────────────
+
+
+async def test_tool_row_and_turn_row_agree_on_the_attempt_id(workspace):
+    """Both read the same books, so the join cannot be ambiguous.
+
+    Reading live session state in `_emit_tool_completed` instead would let a
```
tino097: This inverts what the code does. `_emit_tool_completed` **does** read live session state — `_tc = getattr(self, "_turn_cost", None)` at `session.py:3213`. What it reads is a books object whose `attempt_id` was stamped at open, so the value is right, but the mechanism named here as the rejected alternative is the mechanism in use.

It matters because of where the real hazard lives. The late-finalizer case is on the **turn** side (`_emit_turn_cost(expected=...)` at `session.py:4606`), and that is the one with no test. The tool side is safe only because its read is never late — if someone later moved this emit off the synchronous path (a queued or batched emit), this docstring would read as a guarantee that was never implemented.

Suggest: say the tool row reads the live books and is safe because the emit is synchronous within the turn, and put the stamped-at-open claim where it is load-bearing.

tino097 (commented):
Re-reviewed at `c0cf6b9e`, base `c79131de` (#432's merge). CI green, `run-tests` 5m2s.

The diagnosis and the mechanism both check out. Two `TurnCost(` construction sites, both covered by the `default_factory`. `_emit_turn_cost` fires from the turn's `finally` on every exit — completion, hand-backs, generator close, errors — so all three observed collision shapes get an id by construction, even though only the error terminal has a test. `self._turn_cost = None` on owner close, so no tool row can inherit a stale attempt. `turn_index` untouched, so the trace name and the artifact index are unaffected. The writer inventory holds for anton.

`test_two_turns_in_ONE_session_get_different_ids` is the strongest test in the set — every other one rebuilds the session per turn, so a per-session id would have satisfied them all while leaving every CLI turn sharing one id.

Six inline comments. The first is the one I would act on: the late-finalizer invariant the stamped-at-open design exists for is real, reachable at `session.py:4606`, and untested.

Three more that don't anchor to a diff line:

**The `$insert_id` rationale is now stale.** `_posthog_body`'s docstring says there is deliberately no `$insert_id` because "the natural key would be `(conversation_id, turn_index)`, but an abandoned turn's books and a later retry can legitimately share both". This PR creates the key that would work. Not arguing to add it — `TurnCost.emitted` already stops the same books emitting twice, so the value is marginal — but the reasoning as written now points at a solved problem.

**The Langfuse hop stays ambiguous.** ENG-2243 lists "several traces named `anton:turn-6` in one session" as consequence 3, and the register advertises the PostHog-to-Langfuse forensics hop. An analyst who finds the attempt by `turn_attempt_id` still gets N identically-named traces and no way to pick one. `TraceContext.metadata: dict[str, str]` already forwards to `Langfuse-Metadata`, so it is reachable, though it needs the books opened before the trace context is built. Scope question rather than a defect — worth a "Not in this PR" line either way, the way the skills half already got one.

**The Verification section is stale by three commits.** It says 5 tests (there are 7) and 4 mutations, but `7c136000` records a fifth — a stable per-session `attempt_id` passing the entire suite — which is the most interesting one in the set. And `c0cf6b9e`'s finding isn't there at all: `git merge-tree` reported 4 conflict regions and there were 5 edits, because an exhaustive key-set assertion living in the *other* branch's new test is invisible to a conflict count. That is a genuinely useful warning for the next person widening this event, and right now it exists only in a commit message. I gave a 4-region estimate off the same tool on #432, so it caught me too.

Four of the ticket's `Done when` assertions aren't present — a cancelled turn and a spend-ceiling turn each getting their own id, the Langfuse trace name unchanged, the artifact-store index unchanged. All four are true by construction for the reasons above, so a line in the body saying so reads better than four more tests.

anton/core/session.py
alecantu7: Confirmed and fixed — this was the one real gap.

Reproduced your mutation exactly: `turn_attempt_id=(self._turn_cost or tc).attempt_id` passed all 9 tests in the module and all of `test_turn_cost_terminals.py`. So the invariant the stamped-at-open design exists for had no coverage at the emit, only at the dataclass.

`test_a_late_finalizer_reports_its_own_attempt_not_the_live_turns` — opens abandoned books, installs a newer set as the live turn, emits the abandoned ones, asserts the row carries the abandoned id, not the live one, and that the slot is still owned. It is now the only test that kills that mutation.

Fixed in `fc70969c`.

anton/core/session.py
alecantu7: Taken, and stronger than it looks once ENG-2193 is in the frame — a desktop customer's diagnostic channel is `cowork-server.log`, so the log line is the only surface they have and the new property had no fallback at all.

`turn_cost session=%s turn=%d attempt=%s ended_by=%s ...`, with a comment saying why it sits next to `turn`.

Worth noting how untested that line was: dropping the argument passed **41 tests**. So `test_the_structured_log_line_carries_the_attempt_too` asserts on the *abandoned* books, which pins the source and not just the presence of the key.

Fixed in `fc70969c`.

anton/core/session.py
alecantu7: Correct on all three counts — behaviour right, docstring backwards. Fixed at the docstring (see the reply on `test_turn_attempt_id.py:164`), and I took your suggested framing almost verbatim: the tool row reads the live books and is safe because the emit is synchronous within its own live turn and the owner nulls the books at close; the stamped-at-open claim now lives on the turn side, where it is load-bearing.

Your point about a future queued or batched emit is the reason this mattered rather than being a wording nit.

anton/core/session.py
alecantu7: Both fixed.

The docstring now enumerates ten keys and names all three join keys explicitly, with a line saying that paragraph *is* the privacy-audit enumeration so it has to be kept in step with the payload. Agreed that a short list is the wrong kind of stale there.

The register wording: replaced the unreachable claim with the divergence that is reachable — with books closed `turn_index` falls back to `_turn_count + 1` while this would be `""`, so the two keys would disagree about whether a parent turn row exists. Plus a line to prefer this key and read an empty value as a pre-ENG-2243 build, never as "no parent".

Fixed in `fc70969c`.

anton/core/turn_cost.py
alecantu7: You are right, and I verified it before changing anything:

```
hex[12] over 2000 uuid4s: {'4': 2000}
entropy in hex[:16]: 60 bits
```

60, not 64. And the `aid` comparison was backwards — `analytics.py:205` is `sha256(...).hexdigest()[:16]`, the uuid4 form at `:201` is only the fallback. So the comment was wrong on the reasoning as well as the arithmetic, and so was the PR body's security check. Both corrected.

Now `secrets.token_hex(8)`. Also added `test_all_sixteen_characters_carry_entropy`, which asserts every hex position varies across 200 ids — reverting to uuid4 fails it and names position 12. The old test could not see a fixed nibble: 200 uuid4-derived ids are all unique and all lowercase hex, so a 4-bit loss failed neither assertion.

Fixed in `fc70969c`.

tests/test_turn_attempt_id.py
alecantu7: Agreed — this inverted the code and I have rewritten it along the lines you suggested.

It now says the tool row reads the live books (`session.py:3229`) and is safe because the emit is never late, not because it reads stamped books; points at the new turn-side test for the stamped-at-open guarantee; and states the consequence you identified — if the tool emit were ever moved off the synchronous path, the old docstring would have read as a guarantee that was never implemented, and this test would have kept passing while the join silently paired tool rows with a neighbouring attempt.

Fixed in `fc70969c`.

alecantu7:
Thanks — all six inline comments actioned in `fc70969c`, plus the three body-level points. Rebased onto `f3bfa6ac` first (six commits since `c79131de`, no conflicts), so the head you reviewed at `c0cf6b9e` has moved.

**The one real defect was yours to find:** `_emit_turn_cost` reading the handed books had no test. Your suggested mutation, `(self._turn_cost or tc).attempt_id`, passed all 9 tests in the module and all of `test_turn_cost_terminals.py`. It now fails exactly one.

**And I was wrong about the width.** Verified before changing anything: `hex[12]` is `4` in 2000/2000 uuid4s, so 60 bits, and `aid`'s primary path really is sha256 with uuid4 only as its fallback. Both the comment and the PR body's security check were wrong. Now `secrets.token_hex(8)`, with a test that asserts every hex position varies — the old one checked length and charset, so it could not see a 4-bit loss.

On the three that did not anchor to a diff line:

- **`$insert_id` rationale** — corrected in place. You are right that this PR creates the key that would work; the standing reason is the second half of that comment, that `TurnCost.emitted` already stops a double emit one layer earlier and for every sink at once. So it is kept out on redundancy, not on the absence of a key. Not adding it.
- **The Langfuse hop** — now an explicit "Not in this PR" entry, with the mechanism (`TraceContext.metadata` → `Langfuse-Metadata`) and why it is not a one-liner: it needs the books opened before the trace context is built, which is a real ordering change. Scoped out deliberately rather than overlooked, same as the skills half.
- **Verification section** — rebuilt. 10 tests (was 5), 8 mutations (was 4), including `7c136000`'s per-session-id mutation you flagged as the most interesting, and `c0cf6b9e`'s `git merge-tree` finding is now in the body rather than only a commit message: it reported 4 conflict regions where there were 5 edits, because an exhaustive key-set assertion in the *other* branch's new test file is invisible to a conflict count. Written as a warning for the next person widening this payload, and it caught me twice off the same tool.

The four unasserted `Done when` items are now a table in the body saying they are true by construction and why — the shared `finally`, and `turn_index` being untouched — so they can be accepted or rejected deliberately rather than filed as a miss. Agreed a line reads better than four more tests.

Full suite on `f3bfa6ac`: **3,007 passed, 31 skipped.** Head is `fc70969c`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: claude-opus-4-8**

I reviewed the full diff (5 files, +535/−23): the new `TurnCost.attempt_id`, its stamping on `turn_completed`/`tool_completed` and the `turn_cost` log line, and the tests. The change is additive analytics — a random per-execution id on two events that already carry `conversation_id`/`turn_index` — and both emit paths are inside the existing `try/except` that never affects the turn. No actionable findings.

### Validation

- Grepped every `TurnCost(` construction site: all use keyword args or none, so inserting `attempt_id` mid-dataclass shifts no positional call. This was the main correctness risk and it's absent.
- Counted the reworked `turn_cost` log line: 21 `%`-specifiers, 21 arguments, `attempt=%s`/`tc.attempt_id` inserted in matching positions. Balanced.
- Confirmed `secrets.token_hex(8)` yields 16 hex chars / 64 bits, and `default_factory` runs per-instance — every construction gets a distinct id.
- Traced the reviewer's late-finalizer finding to `test_a_late_finalizer_reports_its_own_attempt_not_the_live_turns`; it asserts on the abandoned books and kills the `(self._turn_cost or tc).attempt_id` mutation. The other five review comments are resolved in `fc70969c`.
- CI: all 14 checks green, including `run-tests` (5m39s) and CodeQL/Snyk.

The `turn_index`/`turn_attempt_id` divergence when no books are open (real index vs empty id) and the new high-cardinality PostHog property are both documented as intentional — not defects.

alecantu7:
Pushed `a456dbee` — comments and docstrings only, no behaviour change. Flagging it because it lands after @pnewsam's approval on `fc70969c`.

**Why it was worth a re-push.** Line numbers in comments mislead the next reader, and every `file.py:NNN` pointer in the files this PR touches was already wrong — four of them introduced by `fc70969c` itself. I quoted the numbers from the `c0cf6b9e` review, the rebase onto `f3bfa6ac` moved them, and my own log-line comment moved them again. Same day, twice.

```
session.py:3229  -> _emit_tool_completed was at 3237
session.py:4643  -> _turn_stream_inner's finally was at 4660
session.py:2856  -> _emit_turn_cost's handed-books read was at 2854
session.py:3632  -> turn() was at 3846            (predates fc70969c)
analytics.py:205 -> an unrelated comment line
```

`session.py` itself carries **zero** `file.py:NNN` references, so symbol names were already the house style and these were the outlier. All five now name the function and quote the expression.

**Two things the audit caught that a number swap would have hidden.**

1. **The function I cited does not exist.** I wrote `_aid()`; it is `get_installation_id()`. Found by grepping every symbol the new comments name — same class of error as the stale numbers, and invisible to a line-number fix.

2. **`get_installation_id`'s docstring has the identical 60-bit defect.** Its Returns clause promises "64 bits of entropy" for both branches, but the no-MAC branch persists `uuid4().hex[:16]` — 60 bits, for exactly the reason @tino097 identified in `attempt_id`. Docstring corrected. **The derivation is deliberately unchanged:** switching it would give every already-fingerprinted Docker install a new `aid` and break the continuity of that identity, which is a migration, not a cleanup.

   This also corrects my own reply on that thread. I said sha256 is `aid`'s primary path "and uuid4 only the fallback", which is true but not for the reason I implied — sha256 is taken whenever `uuid.getnode()` returns a real MAC, and the uuid4 branch is the no-MAC case (Docker with stripped networking), not a later fallback in the same chain.

Full suite on `f3bfa6ac`: **3,007 passed, 31 skipped.** Diff from `fc70969c` is +25/−12 across three files, all comments/docstrings — `git diff fc70969c a456dbee` if you want to confirm nothing executable moved. Sorry for the extra round; re-approval needed.

PR #432, opened 2026-09-02
ENG-2247: stamp the root-cause tier and class on tool_completed

alecantu7:
Closes [ENG-2247](https://linear.app/mindsdb/issue/ENG-2247).

## What

`tool_completed` gains `root_cause_tier` and `root_cause_class`, drawn from `anton/core/root_cause.py`'s existing vocabulary — so "why do tool calls fail" becomes a grouped query instead of one Langfuse trace at a time.

## Why

`error_type` only fills in when a handler **raises**. `scratchpad` — 79% of tool volume, failing **9.35%** of 13,356 calls — doesn't raise: a cell that errors is a normal outcome, so the handler returns `ToolOutcome(ok=False, reason="NameError: …")`. Result, 30 days prod:

| `error_type` | failures | share |
| -- | --: | --: |
| **(empty)** | **1,122** | **83.4%** |
| `RuntimeError` | 116 | 8.6% |
| `ValueError` | 107 | 8.0% |

Shipping `reason` instead is not an option — it's prose carrying file paths and user input, ruled out by ENG-1486's security check. But anton **already classifies every failure per call** into a content-free vocabulary, and throws the answer away.

Measured, 14 days, mock-excluded, carrying builds only:

| | |
| -- | --: |
| classified failures | 2,216 |
| **self-inflicted** | **1,032 (47%)** — every class name discarded |
| unclassified | 679 (31%) |
| external wall | 438 (20%) |
| transient | 68 (3%) |
| `classify_errors` | **0** — the classifier has never raised |
| reason coverage (turns that had failures) | **97.5%**, p50 = 1.0 |
| failures that gain a real class | **69.4%** |

## The constraint that shaped the implementation

The class is dropped by an early return in `RootCauseLedger.add`:

```python
        if not rc.trip_eligible:
            return                      # ~80% of failures stop here
        self.exact[rc.key] += 1
        self.classes[rc.cls] += 1       # only external_wall reaches this
```

**That early return is load-bearing.** Its comment: *"kept out of the rungs a trip could ever read. This is where 'structurally cannot trip' actually happens."* It's how ENG-1531's breaker is stopped from firing on the agent's own bugs — the ENG-836 ping-pong `self_inflicted` exists to exclude.

**So this does not widen the ledger.** Widening `self.classes` would arm the breaker on `NameError`, and it's the obvious-looking fix. Reading the classifier directly bypasses the ledger entirely, so nothing a trip rung reads can change. That's checked by the full suite, not just by argument.

Second reason the per-call route wins: the ledger is cumulative for its lifetime (`event_fields`' own docstring: *"The names read as per-turn and are not"*). One `tool_completed` row is one call, with no such caveat.

## The ordering problem the ticket described — it dissolved

The ticket warned the emit runs *before* the classification (3874 before 3908; 5249 before 5314) and proposed hoisting or reordering. Neither is needed:

`classify(reason, result_text)` consults `result_text` **only** to build the `identifier` of a reason-less failure — the class in that branch is the constant `"unclassified"` either way. So the class is identical with or without it, and both call sites already have `reason` in scope at the emit. No hoist, no signature change, no statements moved in the dispatch loop.

## Design decisions

- **Two fields, not one.** The ticket asked for the class; the tier is what makes it actionable — `NameError` and `ModuleNotFoundError` are both `ok=false` with an empty `error_type`, and without the tier they're the same row. 4 values, deterministic from the class. Flagging it as beyond the ticket's letter.
- **A cause only where `ok is False`.** `ok is None` is an unmigrated handler (ENG-2248); the event already says `ok="unknown"`, and attaching a cause would invent a verdict from prose the model can influence.
- **`root_cause_*` naming** deliberately matches turn_completed's tally, signalling one vocabulary rather than a parallel taxonomy.

## Verification

- **10 tests** in `tests/test_tool_completed.py` (9 new, plus the widened exact-keys test) and a `_closed_vocabulary()` helper. The four strongest arrived in later commits, so they are easy to miss skimming the first diff:
  - `test_the_emitted_class_is_always_from_the_closed_vocabulary` — the closed set is **derived from `root_cause.py`'s tables**, not copied, so extending the vocabulary keeps it green while a value from outside fails it. Driven with adversarial reasons: a novel exception type, a bare sentence, injected-looking text, a status code, a 4k string.
  - `test_the_tool_row_and_the_turn_tally_agree_on_the_tier` — also asserts `root_cause_wall == 0` and `top_class == ""`, which pins that this change does **not** arm ENG-1531's breaker.
  - `test_nonstreaming_turn_path_also_emits` and `test_the_nonstreaming_turn_path_stamps_the_cause_too` — **complementary, neither subsumes the other**; each catches a mutation the other misses (table below, and in the second test's docstring).
  - `test_the_cause_is_the_shared_classifier_verbatim` pins the emitted values against `classify()` **itself** rather than hardcoded strings, so extending the vocabulary can't silently desync the event.
- **9 mutations, all caught.** The last four came from `/pr-review` and a concurrent review session, and include the two that matter most:

  | mutation | result |
  | -- | -- |
  | drop both kwargs from the emit | 7 failed |
  | stamp a cause on success | 1 failed |
  | **leak the prose `reason` instead of the class** | **4 failed** |
  | invent a parallel taxonomy (constant class) | 4 failed |
  | stamp a cause for unmigrated handlers | 1 failed |
  | **widen `RootCauseLedger.add` to count non-trip-eligible classes** — the obvious-looking fix, which arms ENG-1531's breaker on `NameError` | **1 failed** |
  | **emit `rc.key` (`class:identifier`), leaking absolute paths and internal hostnames** | **4 failed** |
  | drop the kwargs from the **non-streaming** emit only | 1 failed *(survived all 2,927 before the guard was added)* |
  | `root_cause_class` sourced from `error_type` (wrong variable) | 1 failed |

- **End-to-end wire probe**, urlopen patched, a real turn driven through `turn_stream` with a failing `PermissionError` outcome whose reason contains `/Users/someone/.aws/credentials`:

  ```
  tool_completed POSTs captured: 1
    name=scratchpad ok=false error_type=''
    root_cause_tier ='external_wall'
    root_cause_class='permission_denied'
  secret path anywhere on the wire: False
  ```

  This is the ENG-1355/ENG-1495 failure class — a property that looks right in code and reaches nothing, or reaches too much. Both directions checked.

- `test_payload_is_exactly_the_seven_keys_all_strings` → `..._nine_keys...`, widened **deliberately**: "no surprise keys" is the property that keeps arguments and result prose out.
- Full suite, fresh `uv sync`: **2,925 passed, 31 skipped, 0 failed**, on `origin/staging` @ `ca249551`.

## Merge order — reviewer please note

[#431](https://github.com/mindsdb/anton/pull/431) (ENG-2243) adds `turn_attempt_id` to the **same** `send_event` call and moves the same exact-keys assertion. Branched from `staging` rather than stacked, per convention 3.

**Whichever merges second needs a rebase — measured, not estimated.** `git merge-tree` of the two heads gives **4 conflict regions across 2 files** (`analytics.py` and `turn_cost.py` auto-merge):

| file | regions | what |
| -- | --: | -- |
| `anton/core/session.py` | 1 | both PRs append kwargs to the same `send_event(...)`; keep both blocks |
| `tests/test_tool_completed.py` | 3 | the module docstring's payload list, the test **name** (`nine_keys` vs `eight_keys`), and the key-set literal |

Resolution: keep both kwarg blocks, rename to `..._ten_keys_all_strings`, and union the set to `name, ok, duration_ms, error_type, surface, conversation_id, turn_index, turn_attempt_id, root_cause_tier, root_cause_class`. Mechanical — no semantic interaction, the fields are independent — and the exhaustive exact-keys assertion catches a botched union immediately. ~10 minutes.

Merging **#431 first** is marginally cheaper: it has the smaller test-file footprint, so #432 absorbs the union once.

## Expected side effects that are NOT regressions

- **Two new PostHog properties.** Both closed-vocabulary and low-cardinality (4 tiers; the class set is bounded by `root_cause.py`). Safe as breakdown dimensions, unlike `turn_attempt_id`.
- **Empty on success, and on `ok="unknown"` rows.** Read them as a pair with `ok`; absent ≠ "no cause", it means no verdict.
- **The `external_wall` / `transient` shares carry a known skew, and this PR is what makes it matter.** `root_cause.py`'s status branch mis-tiers in both directions, and its own comment defers the fix on the grounds that it is *"tolerable while nothing consumes the tier"* — **this PR is that consumer.** Measured through `_tool_failure_cause`:

  | reason | (tier, class) |
  | -- | -- |
  | `FileNotFoundError: [Errno 2] ... '/data/500/report.csv'` | `transient` / `http_500` |
  | `FileNotFoundError: [Errno 2] ... '/data/report.csv'` | `external_wall` / `missing_file` |
  | `RuntimeError: record 404 not found` | `external_wall` / `missing_file` |

  So a real missing-file wall publishes as a retryable blip whenever a path or port carries a bare 3-digit 4xx/5xx token, and an agent bug publishes as an environment wall. Not fixed here — it needs an HTTP-shape precondition on the status branch, which is ENG-1531's call and touches shared classification every consumer reads. Flagged by @tino097; the direction is the safe one (it inflates walls rather than hiding them), but read the two shares with this in mind.

- **This field does NOT cover the ENG-2193 verifier population.** `generate_object_code(_VerifierVerdict)` is a forced-schema call, not a tool call, so it never enters tool dispatch and never produces a `tool_completed` row. The free-tier verifier failure ENG-2193 is about will not appear under any `root_cause_class`. Stated because the two tickets landed the same week and the field name invites the assumption.

- **~31% will read `unclassified`.** Not missing input — coverage is 97.5%. It's exception types absent from the vocabulary hitting `classify()`'s residual branch (`RuntimeError` alone was 116 of 30 days' raise-shaped failures and is in none of the tables). Cheap follow-up once the residual is visible per tool; deliberately not in this PR.

## Security check

Performed on the diff, and it is the central design constraint rather than a formality. Only the closed-vocabulary `tier`/`cls` tokens are emitted — never `reason`, never `result_text`, both of which are in scope at the emit site and are exactly what a future contributor will reach for. Mutation M3 exists to fail loudly if anyone does. The end-to-end probe confirms a credential-bearing reason produces `permission_denied` on the wire and nothing else. No new destination: the properties ride an event already going to the PostHog direct sink.

### DIFF
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -195,16 +195,36 @@ _pending_lock = threading.Lock()
 #                    root_cause_tier + root_cause_class (WHY it failed, in
 #                    `anton/core/root_cause.py`'s vocabulary — `self_inflicted`
 #                    / `transient` / `external_wall` / `unclassified`, and the
-#                    class within it, which for a self-inflicted failure IS the
-#                    exception name). `error_type` above only fills in on a
+#                    class within it, which is USUALLY the exception name but
+#                    is a sentinel class where the handler declared the failure
+#                    itself: `empty_code`, `missing_argument`,
+#                    `invalid_argument`, `unknown_resource`, `missing_name`.
+#                    Do NOT join this against Python exception names — nine
+#                    `_SENTINEL_REASONS` map to `self_inflicted` with non-
+#                    exception classes, and `scratchpad_empty_code` is one of
+#                    scratchpad's own, so that join silently drops every
+#                    argument-validation failure. `timeout` is likewise a
+#                    literal, not a type name. `error_type` above only fills in on a
 #                    RAISE; the dominant tool returns a verdict instead, so 83%
 #                    of failures had no cause before ENG-2247. Both are "" on
 #                    success and on an unmigrated handler (`ok="unknown"`).
 #                    Read them as a PAIR with `ok` — a cause only exists where
 #                    the handler said `ok=false`. Same vocabulary as
-#                    turn_completed's `root_cause_*` tally, so the per-call and
-#                    per-turn views cannot disagree; unlike that tally, these
-#                    are per CALL and carry no cumulative-ledger caveat.
+#                    turn_completed's `root_cause_*` tally, so where both
+#                    populate they agree; unlike that tally, these are per CALL
+#                    and carry no cumulative-ledger caveat.
+#                    They do NOT reconcile row-for-row, and the gap is
+#                    systematic rather than rare. An `ok="unknown"` call (any
+#                    handler returning a plain string — most of them, see
+#                    ENG-2248) is COUNTED by the turn tally as `unclassified`
+#                    and left BLANK here, because inventing a verdict from
+#                    prose the model can influence is the ENG-1276 defect one
+#                    level up. So `sum(tool rows by tier)` runs systematically
+#                    short of `turn_completed.root_cause_failures`; that is the
+#                    design, not drift. (The inverse also exists but is
+#                    unreachable today: a multimodal `ok=false` result would
+#                    populate here and not in the tally — see
+#                    `_tool_failure_cause`.)
 #                    conversation_id + turn_index mirror turn_completed's
 #                    values, so a tool row joins to its parent turn row and,
 #                    via Langfuse sessionId, to the gateway trace.
--- a/anton/analytics.py
+++ b/anton/analytics.py
@@ -192,6 +192,19 @@ _pending_lock = threading.Lock()
 #                    One event per executed tool call; tool arguments and
 #                    result content are deliberately absent (ENG-1486).
 #                    surface mirrors turn_completed's (ENG-1945).
+#                    root_cause_tier + root_cause_class (WHY it failed, in
+#                    `anton/core/root_cause.py`'s vocabulary — `self_inflicted`
+#                    / `transient` / `external_wall` / `unclassified`, and the
+#                    class within it, which for a self-inflicted failure IS the
+#                    exception name). `error_type` above only fills in on a
+#                    RAISE; the dominant tool returns a verdict instead, so 83%
+#                    of failures had no cause before ENG-2247. Both are "" on
+#                    success and on an unmigrated handler (`ok="unknown"`).
+#                    Read them as a PAIR with `ok` — a cause only exists where
+#                    the handler said `ok=false`. Same vocabulary as
+#                    turn_completed's `root_cause_*` tally, so the per-call and
+#                    per-turn views cannot disagree; unlike that tally, these
+#                    are per CALL and carry no cumulative-ledger caveat.
 #                    conversation_id + turn_index mirror turn_completed's
 #                    values, so a tool row joins to its parent turn row and,
 #                    via Langfuse sessionId, to the gateway trace.
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -3180,6 +3180,10 @@ class ChatSession:
                 settings = AntonSettings()
             from anton.analytics import send_event
 
+            # Derived here, not handed in — see the docstring. `reason` is
+            # consumed and dropped; only the two closed-vocabulary tokens go on.
+            _rc_tier, _rc_class = _tool_failure_cause(ok, reason)
+
             # send_event takes STRING values only — every extra is a wire
             # parameter (tests/test_ask_user.py:496 exists because a first
             # draft assumed a (name, props) shape).
@@ -3187,9 +3191,6 @@ class ChatSession:
             # the live TurnCost carries the authoritative index (an explicit
             # turn_id on the cloud path, the local counter otherwise), so the
             # values here and on the turn's own row can never disagree.
-            # Derived here, not handed in — see the docstring. `reason` is
-            # consumed and dropped; only the two closed-vocabulary tokens go on.
-            _rc_tier, _rc_class = _tool_failure_cause(ok, reason)
             _tc = getattr(self, "_turn_cost", None)
             turn_index = (
                 getattr(_tc, "turn_index", 0) or (self._turn_count + 1)
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -603,9 +603,20 @@ def _tool_failure_cause(ok: "bool | None", reason: str) -> "tuple[str, str]":
     """`(tier, class)` for one failed tool call, or `("", "")` (ENG-2247).
 
     Reuses `root_cause.classify` — the SAME vocabulary the turn-level
-    `root_cause_*` tally is built from — so a tool row and its turn cannot
-    disagree about what a failure was. A parallel taxonomy here would put two
-    spellings of one failure in two fields.
+    `root_cause_*` tally is built from — so for a string tool result the row
+    and its turn cannot disagree about what a failure was. A parallel taxonomy
+    here would put two spellings of one failure in two fields.
+
+    **Scoped to string results deliberately, and it is not a full agreement
+    guarantee.** This runs unconditionally at the emit, but
+    `_record_root_cause` is reached only from the string branch — the
+    multimodal arms `continue` past it. So a handler returning
+    `content=[...]` with `ok=False` puts a cause on the tool row that the turn
+    tally never counts (measured: row `self_inflicted`/`NameError`, tally
+    `root_cause_failures=0`). Unreachable today — no handler returns
+    multimodal content at all — but `ToolOutcome.content` permits it, so
+    whoever migrates the first one must add `_record_root_cause` to both list
+    arms, alongside the nudge the #308-review NOTE already asks for there.
 
     **Deliberately does NOT touch `RootCauseLedger`.** That ledger drops every
     non-trip-eligible class one line into `add()` (`if not rc.trip_eligible:
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -599,6 +599,44 @@ def _safe_error_detail(exc: BaseException) -> str:
     return name
 
 
+def _tool_failure_cause(ok: "bool | None", reason: str) -> "tuple[str, str]":
+    """`(tier, class)` for one failed tool call, or `("", "")` (ENG-2247).
+
+    Reuses `root_cause.classify` — the SAME vocabulary the turn-level
+    `root_cause_*` tally is built from — so a tool row and its turn cannot
+    disagree about what a failure was. A parallel taxonomy here would put two
+    spellings of one failure in two fields.
+
+    **Deliberately does NOT touch `RootCauseLedger`.** That ledger drops every
+    non-trip-eligible class one line into `add()` (`if not rc.trip_eligible:
+    return`), and that early return is how "structurally cannot trip" is
+    implemented for ENG-1531's breaker — widening it to keep `self_inflicted`
+    names would arm the breaker on the agent's own `NameError`s, the exact
+    ENG-836 ping-pong the tier exists to exclude. Reading the classifier
+    directly bypasses the ledger, so nothing a trip rung reads can change.
+
+    `result_text` is deliberately not taken. `classify` consults it only to
+    build the `identifier` of a reason-less failure; the CLASS in that branch is
+    the constant `"unclassified"` either way. So the class is identical with or
+    without it — which is what lets this run at the emit site, where the result
+    has not been assigned yet, instead of reordering the dispatch loop.
+
+    Only a handler's explicit `ok is False` produces a cause. `ok is None` is an
+    unmigrated handler (ENG-2248): the event already reports `ok="unknown"`, and
+    attaching a cause to a call we cannot say failed would invent a verdict.
+
+    Never raises: it runs on the reporting path, where an escape would turn a
+    handled tool failure into a dead turn.
+    """
+    if ok is not False:
+        return "", ""
+    try:
+        rc = classify_root_cause(reason or "", "")
+        return rc.tier, rc.cls
+    except Exception:  # pragma: no cover - defensive; classify is total
+        return "", ""
+
+
 def _safe_error_type(exc: BaseException) -> str:
     """The exception's class name, and nothing else, for analytics.
 
@@ -3063,6 +3101,8 @@ class ChatSession:
         ok: bool | None,
         duration_ms: float,
         error_type: str,
+        root_cause_tier: str = "",
+        root_cause_class: str = "",
     ) -> None:
         """Per-tool-call analytics event (ENG-1486).
 
@@ -3130,6 +3170,14 @@ class ChatSession:
                 surface=str(getattr(self, "_surface", None) or ""),
                 conversation_id=str(self._session_id or ""),
                 turn_index=str(turn_index),
+                # WHY the call failed, in `root_cause.py`'s vocabulary
+                # (ENG-2247). `error_type` above only fills in when the failure
+                # was a RAISE; `scratchpad` — 79% of tool volume, 9.35% of it
+                # failing — returns a verdict instead, so 83% of failures
+                # reached analytics with no cause at all. Both empty on success
+                # and on an unmigrated handler; see `_tool_failure_cause`.
+                root_cause_tier=root_cause_tier,
+                root_cause_class=root_cause_class,
             )
         except Exception:
             # Analytics must never affect the tool call that just ran.
@@ -3871,11 +3919,16 @@ class ChatSession:
                 # elicitation is structurally unavailable on this path (no
                 # emitter to render a question), so there is no wait to
                 # subtract.
+                _rc_tier, _rc_class = _tool_failure_cause(
+                    outcome.ok, outcome.reason
+                )
                 self._emit_tool_completed(
                     name=tc.name,
                     ok=outcome.ok,
                     duration_ms=(_time.monotonic() - _tool_t0) * 1000.0,
                     error_type=_tool_error_type,
+                    root_cause_tier=_rc_tier,
+                    root_cause_class=_rc_class,
                 )
                 result = outcome.content
 
@@ -5246,6 +5299,9 @@ class ChatSession:
                     # tool performance. Fixing it needs a session-scoped
                     # accumulator behind `prompt_or_cancel`, which is a change
                     # to the CLI prompt path rather than to this event.
+                    _rc_tier, _rc_class = _tool_failure_cause(
+                        tool_ok, tool_reason
+                    )
                     self._emit_tool_completed(
                         name=tc.name,
                         ok=tool_ok,
@@ -5255,6 +5311,8 @@ class ChatSession:
                             0.0,
                         ) * 1000.0,
                         error_type=_tool_error_type,
+                        root_cause_tier=_rc_tier,
+                        root_cause_class=_rc_class,
                     )
 
                     if isinstance(result_text, list):
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -599,7 +599,7 @@ def _safe_error_detail(exc: BaseException) -> str:
     return name
 
 
-def _tool_failure_cause(ok: "bool | None", reason: str) -> "tuple[str, str]":
+def _tool_failure_cause(ok: bool | None, reason: str) -> tuple[str, str]:
     """`(tier, class)` for one failed tool call, or `("", "")` (ENG-2247).
 
     Reuses `root_cause.classify` — the SAME vocabulary the turn-level
@@ -3121,10 +3121,28 @@ class ChatSession:
         ok: bool | None,
         duration_ms: float,
         error_type: str,
-        root_cause_tier: str = "",
-        root_cause_class: str = "",
+        reason: str = "",
     ) -> None:
         """Per-tool-call analytics event (ENG-1486).
+
+        ``reason`` is the handler's own ``ToolOutcome.reason``, taken so the
+        root-cause class can be derived HERE rather than at each call site
+        (ENG-2247 review). Deriving it here is what makes it impossible to
+        forget: there is no cause argument a new dispatch path can omit, and a
+        caller that passes no ``reason`` at all degrades to ``unclassified``
+        — "we do not know why" — instead of to the empty string, which reads
+        as "this call did not fail". Defaulting the CAUSE was the earlier
+        shape and it failed in the wrong direction; making it required instead
+        was worse still, since a missing argument raises ``TypeError`` at the
+        CALL, outside this method's guard, and kills the turn over telemetry.
+
+        **``reason`` is prose and must never be emitted.** It is a traceback
+        line or a handler message and carries file paths and user input; that
+        is the whole reason ``error_type`` was narrowed to a class name. It is
+        consumed by ``_tool_failure_cause`` two lines down and never touched
+        again — do not add it, ``rc.identifier`` or ``rc.key`` to the payload.
+        ``test_no_prose_from_the_reason_reaches_the_payload`` and the
+        exact-keys assertion both fail if you do.
 
         The verdict the dispatch loop already computes for the UI's
         ``tool_done`` marker, sent where it can be aggregated: without this,
@@ -3169,6 +3187,9 @@ class ChatSession:
             # the live TurnCost carries the authoritative index (an explicit
             # turn_id on the cloud path, the local counter otherwise), so the
             # values here and on the turn's own row can never disagree.
+            # Derived here, not handed in — see the docstring. `reason` is
+            # consumed and dropped; only the two closed-vocabulary tokens go on.
+            _rc_tier, _rc_class = _tool_failure_cause(ok, reason)
             _tc = getattr(self, "_turn_cost", None)
             turn_index = (
                 getattr(_tc, "turn_index", 0) or (self._turn_count + 1)
@@ -3196,8 +3217,8 @@ class ChatSession:
                 # failing — returns a verdict instead, so 83% of failures
                 # reached analytics with no cause at all. Both empty on success
                 # and on an unmigrated handler; see `_tool_failure_cause`.
-                root_cause_tier=root_cause_tier,
-                root_cause_class=root_cause_class,
+                root_cause_tier=_rc_tier,
+                root_cause_class=_rc_class,
             )
         except Exception:
             # Analytics must never affect the tool call that just ran.
@@ -3939,16 +3960,12 @@ class ChatSession:
                 # elicitation is structurally unavailable on this path (no
                 # emitter to render a question), so there is no wait to
                 # subtract.
-                _rc_tier, _rc_class = _tool_failure_cause(
-                    outcome.ok, outcome.reason
-                )
                 self._emit_tool_completed(
                     name=tc.name,
                     ok=outcome.ok,
                     duration_ms=(_time.monotonic() - _tool_t0) * 1000.0,
                     error_type=_tool_error_type,
-                    root_cause_tier=_rc_tier,
-                    root_cause_class=_rc_class,
+                    reason=outcome.reason,
                 )
                 result = outcome.content
 
@@ -5319,9 +5336,6 @@ class ChatSession:
                     # tool performance. Fixing it needs a session-scoped
                     # accumulator behind `prompt_or_cancel`, which is a change
                     # to the CLI prompt path rather than to this event.
-                    _rc_tier, _rc_class = _tool_failure_cause(
-                        tool_ok, tool_reason
-                    )
                     self._emit_tool_completed(
                         name=tc.name,
                         ok=tool_ok,
@@ -5331,8 +5345,7 @@ class ChatSession:
                             0.0,
                         ) * 1000.0,
                         error_type=_tool_error_type,
-                        root_cause_tier=_rc_tier,
-                        root_cause_class=_rc_class,
+                        reason=tool_reason,
                     )
 
                     if isinstance(result_text, list):
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -636,6 +636,15 @@ def _tool_failure_cause(ok: "bool | None", reason: str) -> "tuple[str, str]":
     unmigrated handler (ENG-2248): the event already reports `ok="unknown"`, and
     attaching a cause to a call we cannot say failed would invent a verdict.
 
+    **Emit `tier` and `cls` only — never `identifier`, and never `key`.**
+    `key` is `class:identifier` and is the tempting "more useful" field, but the
+    identifier is extracted from the reason and carries absolute paths and
+    internal hostnames verbatim: `permission_denied:/root/.ssh/id_rsa`,
+    `connection_refused:db.internal.corp:5432`. It is the nearer trap than
+    `reason`, because it sits on the object this function already destructures.
+    `test_no_prose_from_the_reason_reaches_the_payload` fails if either is
+    forwarded — verified by mutation, not by hope.
+
     Never raises: it runs on the reporting path, where an escape would turn a
     handled tool failure into a dead turn.
     """
--- a/tests/test_tool_completed.py
+++ b/tests/test_tool_completed.py
@@ -521,3 +521,74 @@ async def test_no_prose_from_the_reason_reaches_the_payload(workspace):
     assert secret not in blob
     assert ".aws" not in blob
     assert "someone" not in blob
+
+
+def _closed_vocabulary() -> set:
+    """Every class `classify` can return, derived from the module, not copied.
+
+    Derived so that EXTENDING the vocabulary keeps the guard passing, while a
+    value from outside the tables — the unbounded-cardinality failure — fails it.
+    """
+    import anton.core.root_cause as R
+
+    closed = set(R._SELF_INFLICTED) | set(R._TRANSIENT) | set(R._WALL_TYPES.values())
+    closed |= {cls for _, cls in R._SENTINEL_REASONS.values()}
+    closed |= set(R._STATUS_WALLS.values())
+    closed |= {f"http_{code}" for code in R._STATUS_TRANSIENT}
+    return closed | {"unclassified"}
+
+
+def test_the_emitted_class_is_always_from_the_closed_vocabulary():
+    """Bounded cardinality is a property of the FIELD, not of today's inputs.
+
+    A PostHog property with unbounded values is expensive and useless as a
+    breakdown. `reason` is attacker-and-model-influenced prose, so the guard
+    that matters is that nothing derived from it can widen the value space.
+    Adversarial reasons included: a novel exception type, a bare sentence, an
+    injected-looking string, and text carrying a status code.
+    """
+    from anton.core.session import _tool_failure_cause
+
+    closed = _closed_vocabulary()
+    tiers = {"self_inflicted", "transient", "external_wall", "unclassified"}
+    reasons = [
+        "NameError: x", "ModuleNotFoundError: No module named 'z'",
+        "TimeoutError: slow", "PermissionError: nope", "scratchpad_empty_code",
+        "RuntimeError: boom", "SomeVendorSpecificError: novel type",
+        "just a sentence with no exception in it", "",
+        "HTTPError: 503 upstream", "KeyError: 404",
+        "Error: {'sql': 'DROP TABLE users'} failed",
+        "x" * 4000,
+    ]
+    for reason in reasons:
+        tier, cls = _tool_failure_cause(False, reason)
+        assert cls in closed, f"{reason[:40]!r} minted a novel class {cls!r}"
+        assert tier in tiers, f"{reason[:40]!r} minted a novel tier {tier!r}"
+
+
+async def test_the_tool_row_and_the_turn_tally_agree_on_the_tier(workspace):
+    """The per-call field and the per-turn counter must describe one reality.
+
+    They come from the same classifier but by different routes — this one reads
+    it directly, the tally goes through `RootCauseLedger`. A divergence would
+    mean two published numbers disagreeing about the same failure.
+    """
+    session = _session(workspace, [ToolOutcome(
+        content="[error]\nNameError: name 'wb' is not defined",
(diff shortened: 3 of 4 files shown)


alecantu7:
## Self-review (convention 6)

Three findings. Two fixed in `d201b6d7`; one recorded as a known asymmetry.

### F1 — two `Done when` items were asserted in prose, not tested — **FIXED**

The ticket calls for *"cardinality stays bounded… a novel value is an alarm"* and *"the class on a tool row agrees with the turn-level tally"*. The first commit did neither.

**`test_the_emitted_class_is_always_from_the_closed_vocabulary`** — the closed set is **derived from `root_cause.py`'s tables**, not copied, so extending the vocabulary keeps it green while a value from outside fails it. Driven with adversarial reasons: a novel exception type, a bare sentence, injected-looking text, a bare status code, and a 4k string. The real set is 49 values.

This matters more than a normal cardinality nit: `reason` is prose the model can influence, so the guard is that nothing derived from it can widen the value space of a PostHog property.

**`test_the_tool_row_and_the_turn_tally_agree_on_the_tier`** — the two views read one classifier by different routes (this reads it directly; the tally goes through `RootCauseLedger`). It also asserts `root_cause_wall == 0` and `top_class == ""` for a self-inflicted failure, which pins that this change does **not** arm ENG-1531's breaker.

That second assertion earns its keep. Mutation-verified: **widening `RootCauseLedger.add` to count non-trip-eligible classes — the obvious-looking fix — now fails this test.** That's the one way to get this ticket badly wrong, and it's now caught by CI rather than by whoever remembers the comment.

### F2 — multimodal tool results get a cause the ledger never counts — **known, currently unreachable**

On the non-streaming path, `_record_root_cause` is called only in the string branch; the `isinstance(result, list)` branch skips it. My `_tool_failure_cause` runs *before* that split, so a multimodal failure would get a class on the event while contributing nothing to the turn tally — a divergence from what F1's agreement test claims.

**Unreachable today.** The code carries an explicit note: *"NOTE for whoever migrates the first multimodal handler to declare `ok=False`"* — no handler does. So there is no multimodal `ok=False` to diverge.

Not fixed, deliberately: making the event *less* complete to match a gap in the ledger is the wrong direction, and closing the ledger's gap is ENG-1276/ENG-2248 territory rather than this PR's. Recorded so whoever migrates that first multimodal handler knows to check both sides.

### F3 — `_tool_failure_cause` runs outside `_emit_tool_completed`'s `try` — **considered, correct as-is**

The call sits in the dispatch loop, not inside the emit's swallow-everything guard, so a raise there would kill a real turn rather than lose a metric. It is fully wrapped in its own `try/except Exception` returning `("", "")`, matching `_safe_error_type`'s posture in the same file. `classify` is total by construction (every branch returns a `RootCause`) and has raised **0 times in 14 days of production** (`root_cause_classify_errors` = 0).

`BaseException` still escapes — same as `_safe_error_type` — which is correct: a `KeyboardInterrupt` should not be swallowed by a telemetry helper.

---

### Verification on the final head

- Full suite, fresh `uv sync`: **2,927 passed, 31 skipped, 0 failed** (`origin/staging` @ `ca249551`). CI green, 0 failures.
- **7 mutations total, all caught**: drop both kwargs · stamp a cause on success · leak the prose `reason` · invent a parallel taxonomy · stamp a cause for unmigrated handlers · **widen `add()` to arm the breaker** · mint an unbounded class from prose.
- End-to-end wire probe (in the PR body): a real turn with a `PermissionError` reason containing `/Users/someone/.aws/credentials` produces `root_cause_class='permission_denied'` on the wire, with the path absent from the payload.

### For the reviewer

The judgement call worth pushing on is **two fields rather than one**. The ticket asked for the class; I added the tier because `NameError` and `ModuleNotFoundError` are otherwise the same row — both `ok=false`, both empty `error_type` — and the tier is the half that says whose fault it is. It's 4 values and deterministic from the class. If you'd rather ship the class alone, dropping the tier is a two-line change.

anton/core/session.py
```diff
@@ -3871,11 +3919,16 @@ async def turn(self, user_input: str | list[dict]) -> str:
...
                 # subtract.
+                _rc_tier, _rc_class = _tool_failure_cause(
+                    outcome.ok, outcome.reason
+                )
                 self._emit_tool_completed(
                     name=tc.name,
                     ok=outcome.ok,
                     duration_ms=(_time.monotonic() - _tool_t0) * 1000.0,
                     error_type=_tool_error_type,
+                    root_cause_tier=_rc_tier,
```
alecantu7: **Finding 1 — low severity, confirmed by execution: this call site's new fields are entirely unasserted.**

`_emit_tool_completed` defaults both kwargs to `""` (`session.py:3104-3105`), and the non-streaming seam guard `test_nonstreaming_turn_path_also_emits` (`tests/test_tool_completed.py:339`) asserts only `ok`, `error_type` and `name`. So deleting `root_cause_tier=` and `root_cause_class=` from *this* emit changes no payload key and trips no assertion.

I ran exactly that mutation against the whole suite:

```
2927 passed, 31 skipped, 54 warnings in 197.74s
```

Control — the same deletion at the streaming emit (`session.py:5314`): `5 failed, 19 passed`. Clean asymmetry.

**This narrows a body claim.** The mutation table's "drop both kwargs from the emit → 7 failed" is true of the streaming site only; as written it reads as covering the emit.

**No prod impact today** — `ChatSession.turn()` has no production caller (none in anton outside tests, cowork-server only `tests/test_history_compaction_plateau.py:103`, scratchpad-controller none). Which is precisely the point of that guard, per its own docstring: *"the seam guard for the first one that does."* It now guards two of four fields.

**Fix** — three lines in the existing test (`RuntimeError` is in none of `_SELF_INFLICTED`/`_TRANSIENT`/`_WALL_TYPES`, so it lands in the residual branch, which also documents the 31% bucket):

```python
    assert events[0]["root_cause_tier"] == "unclassified"
    assert events[0]["root_cause_class"] == "unclassified"
```

anton/core/session.py
```diff
@@ -599,6 +599,44 @@ def _safe_error_detail(exc: BaseException) -> str:
     return name
 
 
+def _tool_failure_cause(ok: "bool | None", reason: str) -> "tuple[str, str]":
+    """`(tier, class)` for one failed tool call, or `("", "")` (ENG-2247).
+
+    Reuses `root_cause.classify` — the SAME vocabulary the turn-level
+    `root_cause_*` tally is built from — so a tool row and its turn cannot
+    disagree about what a failure was. A parallel taxonomy here would put two
```
alecantu7: **Finding 2 — low severity (unreachable today), confirmed by execution: this "cannot disagree" claim is already false for a multimodal failure.**

`_tool_failure_cause` runs **unconditionally** at the emit, before the list/str split. `_record_root_cause` is reached **only in the string branch** — the multimodal branch `continue`s past it (`session.py:5360`, streaming) or falls in the `isinstance(result, list)` arm (`3937`, non-streaming).

Repro with a handler returning `content=[{"type": "text", ...}], ok=False`:

```
tool_completed root_cause_tier : 'self_inflicted'
tool_completed root_cause_class: 'NameError'
turn_completed root_cause_failures      : 0
turn_completed root_cause_self_inflicted: 0
```

Two published numbers describing the same failure, disagreeing — the thing this sentence rules out.

**Unreachable in production:** `grep 'content=\['` across non-test `anton/` returns no matches — no handler returns multimodal content at all, matching the in-code NOTE at `3947`/`5330`. `ToolOutcome.content` stays typed `str | list[dict]`, so it remains permitted. The ledger's blindness here is **pre-existing**; what's new is the tool row populating where the ledger is blind, under an absolute claim that goes false the day that handler lands.

**Fix:** scope the sentence to string results, or add `_record_root_cause` to the two list branches — the same spot the existing `#308 review` NOTE already tells that migrator to add the nudge. The docstring correction is the proportionate change.

alecantu7 (commented):
Adversarial self-review at `d201b6d7`. The central claim holds under exhaustive checking: all 12 `RootCause(...)` returns in `classify()` yield a `cls` that is a table lookup or a literal, the content-bearing `identifier` is never emitted, and the measured vocabulary is **50 classes / 4 tiers / zero values containing a non-alphanumeric character** — so no path, host or prose fragment can reach the wire through this field by construction. The `Done when` bullet about agreeing with the turn-level tally *is* covered (`test_the_tool_row_and_the_turn_tally_agree_on_the_tier`), and the "class is identical with or without `result_text`" argument verifies on every reachable string-result path.

Two low-severity findings inline, neither blocking. Plus three notes:

- **The #431 rebase is bigger than the body says.** A test merge of #431 into this branch gives `CONFLICT (content)` in **both** `anton/core/session.py` and `tests/test_tool_completed.py` — both PRs rename the same test (`_seven_keys_` → `_eight_keys_` vs `_nine_keys_`) and edit the same set literal and emit region. The merge-order note is right; "one added kwarg line" understates it. Combined set is ten keys.
- **`_tool_failure_cause`'s `except` swallows silently** (`session.py:636`, with `# pragma: no cover`) where `_record_root_cause` a few lines up deliberately calls `note_error()` and carries *no* pragma, on the recorded `#348 review` reasoning that a broken classifier must stay visible. **I consider this refuted** — for every reachable failure `_record_root_cause` classifies the same `reason` moments later and its guard would fire `root_cause_classify_errors` on `turn_completed`, so the signal survives on the turn event. Noting only that the pragma diverges from a decision recorded directly above it.
- **Body staleness:** says "6 new tests"; there are 8 plus `_closed_vocabulary()`. The self-review commit added the two strongest (closed-vocabulary, turn-tally agreement) and they are the ones that answer a reviewer's first questions.

Evidence: baseline `tests/test_tool_completed.py` 24 passed; full suite under mutation **2,927 passed / 31 skipped**; streaming control mutation **5 failed**. All mutations reverted.

tino097 (commented):
Reviewed at `5e64844d` against base `ca249551`. The core design holds up — I checked the load-bearing claims rather than taking them from the description:

- `classify()` reads `result_text` only in the empty-reason branch, which returns the constant `"unclassified"` class. Verified against all 12 `RootCause(...)` returns, so `(tier, cls)` really is identical with and without it and the "ordering problem dissolved" argument is sound. The hoist the ticket recommended genuinely is not needed.
- `RootCauseLedger` untouched. Nothing a trip rung reads can change.
- Both emit sites pass exactly the variables `_record_root_cause` later receives, and both reset per call at 5017-5020 — no stale carry-over across the dispatch loop.
- Only `tier` / `cls` leave the function; never `identifier`, `key`, or `reason`.
- Merge-order note confirmed: `git merge-tree` against #431 conflicts in exactly `session.py` and `tests/test_tool_completed.py`; `analytics.py` and `turn_cost.py` auto-merge. The described resolution is right.
- CI green on this head including `run-tests`. I did not run the suite locally.

Seven inline comments. The first (`_closed_vocabulary()` omitting `"timeout"`) is the only one I would call blocking — the cardinality guard is the argument the whole PR rests on, and it does not cover a class that scratchpad cell timeouts produce in production today. The rest are a mis-scoped invariant in the property doc, one deferred-safety interaction worth naming, and two robustness suggestions the author may reasonably decline.

### Cross-ticket note (ENG-2193)

No conflict: this branches from `ca249551`, which *is* #426's merge commit, so both merged ENG-2193 anton PRs are in the base and nothing here touches the verifier loop, the latch, or the verdict-error sets.

Worth naming as a follow-up rather than a change here — ENG-2193 established that for a desktop customer the diagnostic channel is `cowork-server.log`, not PostHog, and that INFO is below the shipped default level. There is no log line at all on the tool-dispatch failure path (the only one in `session.py` is `_apply_error_tracking`'s INFO for unmigrated handlers), so a customer's log is as empty about tool failures after this as before. The classification now sits in a local at both emit sites; a content-free `logger.warning("tool-failed name=%s tier=%s class=%s")` would put it in the support channel for free.

Separately, the free-tier verifier failure ENG-2193 is about (`generate_object_code(_VerifierVerdict)` on `mindshub_air`) is not a tool call and never enters tool dispatch, so `root_cause_class` will not cover that population. Worth a line in the PR body so nobody reads the new field as answering it.

Two nits not worth their own threads: `_tool_failure_cause`'s annotations are string-quoted in a module with `from __future__ import annotations` on line 1, where every sibling uses bare ones. And ENG-2247's "demonstrated by a query on this ticket" acceptance item is still open — understandably, since it cannot run before ship.

tests/test_tool_completed.py
```diff
@@ -417,3 +432,224 @@ class _PosthogSettings:
...
+
+def _closed_vocabulary() -> set:
+    """Every class `classify` can return, derived from the module, not copied.
+
+    Derived so that EXTENDING the vocabulary keeps the guard passing, while a
+    value from outside the tables — the unbounded-cardinality failure — fails it.
+    """
+    import anton.core.root_cause as R
+
+    closed = set(R._SELF_INFLICTED) | set(R._TRANSIENT) | set(R._WALL_TYPES.values())
```
tino097: The docstring says "Every class `classify` can return, derived from the module, not copied." It isn't. `classify()` returns four class values as bare literals, and **`"timeout"` is in none of the five tables enumerated here**. (`permission_denied` and `connection_refused` are two of the other three, and they pass only because they happen to also be `_WALL_TYPES` values.)

Reachable in production, not hypothetical. `backends/local.py:1089` builds `f"Cell timed out after {total_timeout:.0f}s total"`; that lands in `cell.error`, becomes `tool_reason` via `session.py:5101-5103`, and classifies as:

```
transient  'timeout'   <- Cell timed out after 300s total without producing any output ...
transient  'timeout'   <- RuntimeError: upstream timeout
'timeout' in _closed_vocabulary(): False
```

Not a cardinality risk — `"timeout"` is a bounded literal, so nothing widens in PostHog. The cost is the guard's completeness claim: add a timeout-worded reason to the adversarial list at line 574 and `test_the_emitted_class_is_always_from_the_closed_vocabulary` fails on a legitimate enumerated value, reporting it as a novel class. The natural reaction to that failure is to loosen the set, which is the one move the PR body says must not happen.

Suggest lifting `classify()`'s three tail literals into module constants so the derivation stays honest, then adding a cell-timeout reason to the list. Same shape at line 573, where the four tier strings are hardcoded rather than read from `TIER_SELF` / `TIER_TRANSIENT` / `TIER_WALL` / `TIER_UNCLASSIFIED`.

anton/core/session.py
```diff
@@ -599,6 +599,64 @@ def _safe_error_detail(exc: BaseException) -> str:
...
+    `test_no_prose_from_the_reason_reaches_the_payload` fails if either is
+    forwarded — verified by mutation, not by hope.
+
+    Never raises: it runs on the reporting path, where an escape would turn a
+    handled tool failure into a dead turn.
+    """
+    if ok is not False:
+        return "", ""
+    try:
+        rc = classify_root_cause(reason or "", "")
```
tino097, line 654: `root_cause.py`'s comment block on the status branch says its known both-directions mis-tiering is "tolerable while nothing consumes the tier" and defers the fix to ENG-1531. **This PR is that consumer** — `root_cause_tier` becomes the property the ticket exists to make people group and act on.

Measured through `_tool_failure_cause` on this head:

| reason | (tier, cls) |
| -- | -- |
| `FileNotFoundError: [Errno 2] ... '/data/500/report.csv'` | `transient` / `http_500` |
| `FileNotFoundError: [Errno 2] ... '/data/report.csv'` | `external_wall` / `missing_file` |
| `RuntimeError: record 404 not found` | `external_wall` / `missing_file` |

A real missing-file wall publishes as a retryable blip whenever a path or port carries a bare 3-digit 4xx/5xx token, and an agent bug publishes as an environment wall. Both directions skew the `external_wall` / `transient` shares people will read off the new dimension.

Not asking to fix it here — that is ENG-1531's precondition, and the direction is the safe one. But it belongs under "Expected side effects that are NOT regressions", because the skew stops being a measurement-only caveat the moment this ships.

anton/core/session.py
```diff
@@ -599,6 +599,64 @@ def _safe_error_detail(exc: BaseException) -> str:
...
+        return rc.tier, rc.cls
+    except Exception:  # pragma: no cover - defensive; classify is total
```
tino097, line 656: The sibling guards in `_record_root_cause` call `self._root_causes.note_error()` on exactly this failure, and the comment there says why: "a systematically broken classifier would be indistinguishable from a turn where nothing failed." That same comment records that `pragma: no cover` was **deliberately omitted** there, so a removed `note_error()` cannot hide from the coverage gate.

This guard does neither, and takes the pragma. A classifier exception produces `ok=false` with `root_cause_tier=''`, while the property doc at `analytics.py:201-203` lists exactly two reasons for an empty cause (success, `ok="unknown"`) — nothing tells an analyst the third state exists.

On the string branch `_record_root_cause` still fires a few lines later and `classify_errors` catches it, so the signal survives there. On the multimodal arm it does not, since that arm `continue`s past `_record_root_cause`. Unreachable today (`classify_errors = 0` in prod over 14 days), but the asymmetry cuts against the convention this file just spent a paragraph establishing.

anton/core/session.py
```diff
@@ -3063,6 +3121,8 @@ def _emit_tool_completed(
         ok: bool | None,
         duration_ms: float,
         error_type: str,
+        root_cause_tier: str = "",
```
tino097: Both new kwargs are optional, which is the hole the PR body's own mutation table records: *"drop the kwargs from the non-streaming emit only — 1 failed (survived all 2,927 before the guard was added)."* Omitting them yields a well-formed payload whose keys are present and empty, and empty is a legal value meaning "no failure."

The response was a test for that one call site, which fixes the instance rather than the shape. The next dispatch path — or the #431 rebase, which touches this same `send_event` call — can drop them again with every payload-shape assertion still green.

Consider moving the derivation inside instead: pass `reason` to `_emit_tool_completed` and call `_tool_failure_cause` here, where `ok` already lives. One place, no default to forget, and it removes the duplicated two-line preamble at 3942 and 5322.

anton/core/session.py
```diff
@@ -5246,6 +5319,9 @@ async def _stream_and_handle_tools(
                     # tool performance. Fixing it needs a session-scoped
                     # accumulator behind `prompt_or_cancel`, which is a change
                     # to the CLI prompt path rather than to this event.
+                    _rc_tier, _rc_class = _tool_failure_cause(
```
tino097: This runs above the `isinstance(result_text, list)` split, but `_record_root_cause` (line 5392) sits in the string arm only — the list arm `continue`s past it. So a multimodal `ok=False` outcome now puts a cause on the tool row that the turn tally never counts:

`ToolOutcome(content=[{'type':'text','text':'boom'}], ok=False, reason="NameError: name 'wb' is not defined")` emits `tool_completed{root_cause_tier='self_inflicted', root_cause_class='NameError'}` alongside `turn_completed{root_cause_failures=0}`.

The docstring at 603 names this and calls it unreachable today, which is correct — no handler returns multimodal content. Flagging it only because `analytics.py:206` still tells the reader the two views "cannot disagree", and this is the second way they can. Deriving one `RootCause` before the emit and handing it to both consumers closes this and the `ok=None` gap in one move.

anton/analytics.py
```diff
@@ -192,6 +192,19 @@
 #                    One event per executed tool call; tool arguments and
 #                    result content are deliberately absent (ENG-1486).
 #                    surface mirrors turn_completed's (ENG-1945).
+#                    root_cause_tier + root_cause_class (WHY it failed, in
+#                    `anton/core/root_cause.py`'s vocabulary — `self_inflicted`
+#                    / `transient` / `external_wall` / `unclassified`, and the
+#                    class within it, which for a self-inflicted failure IS the
```
tino097: "which for a self-inflicted failure IS the exception name" is false for the whole sentinel population. `_SENTINEL_REASONS` maps nine reasons to `TIER_SELF` with non-exception classes: `empty_code`, `missing_name`, `missing_argument`, `invalid_argument`, `unknown_resource`.

Live, not theoretical — `scratchpad_empty_code` and `scratchpad_missing_name` are scratchpad's own sentinels, and `test_the_cause_is_the_shared_classifier_verbatim` drives `scratchpad_empty_code`, which yields class `empty_code`.

An analyst who reads this and joins `root_cause_class` against Python exception names, or filters `self_inflicted` expecting `NameError`-shaped values, silently drops every argument-validation failure — one of the two populations the ticket set out to make visible. Suggest "usually the exception name, or a sentinel class where the handler declared the failure itself."

anton/analytics.py
```diff
@@ -192,6 +192,19 @@
...
+#                    exception name). `error_type` above only fills in on a
+#                    RAISE; the dominant tool returns a verdict instead, so 83%
+#                    of failures had no cause before ENG-2247. Both are "" on
+#                    success and on an unmigrated handler (`ok="unknown"`).
+#                    Read them as a PAIR with `ok` — a cause only exists where
+#                    the handler said `ok=false`. Same vocabulary as
+#                    turn_completed's `root_cause_*` tally, so the per-call and
+#                    per-turn views cannot disagree; unlike that tally, these
```
tino097: "the per-call and per-turn views cannot disagree" overclaims. The `ok=None` population is counted by the turn tally and left blank on the tool row:

`ToolOutcome(content="[error] Task failed: something broke", ok=None)` emits `tool_completed{ok='unknown', root_cause_tier='', root_cause_class=''}` while the same turn carries `turn_completed{root_cause_failures=1, root_cause_unclassified=1}`.

Both sides are deliberate — `_record_root_cause:1550-1573` books them because dropping them "was the defect" for `reason_coverage`, and `_tool_failure_cause` refuses to invent a verdict from prose. Neither call is wrong. But any handler returning a plain string gets `ok=None` (`tools/registry.py:144`), so this is most handlers, and an analyst reconciling `sum(tool rows by tier)` against `turn_completed.root_cause_failures` gets a systematic shortfall that this sentence says cannot exist.

Suggest scoping the claim: the two agree on `ok=false` string results; the row is silent where the tally counts an unmigrated handler.

tests/test_tool_completed.py
alecantu7: Fixed in `a7650751`, and it was worse than described — thank you for pushing on this one.

You said three tail literals; there are **four** (`classify` returns bare literals at 264, 348, 355, 357, 360, 363), and the two that looked covered were covered *by coincidence*: `permission_denied` and `connection_refused` pass only because they also happen to be `_WALL_TYPES` values. So the derivation was accidental across the board, not just missing `timeout`.

Confirmed your reachability claim end to end before changing anything:

```
(transient,timeout)  in closed set: False   <- Cell timed out after 300s total...
(transient,timeout)  in closed set: False   <- Cell exceeded inactivity limit
```

Rather than lift the literals into constants and leave the test to reassemble, I moved the enumeration into the module that owns it: `root_cause.py` now names them (`CLS_TIMEOUT`, `CLS_UNCLASSIFIED`, `CLS_PERMISSION_DENIED`, `CLS_CONNECTION_REFUSED`) and exports **`ALL_CLASSES` (50)** and **`ALL_TIERS`** as the authoritative sets, built from the tables plus those constants. `_closed_vocabulary()` returns `set(ALL_CLASSES)` and the tier assertion reads `ALL_TIERS` instead of four hardcoded strings, per your second note.

Adding a class now means adding it there and the guard follows. Mutation-verified in both directions: reverting the helper to the old table-reassembly fails the guard, so it now protects its own derivation.

anton/core/session.py
alecantu7: Taken, but not the way I first tried — and the detour is worth recording because the obvious middle option is actively unsafe.

I initially made both kwargs **required**, reasoning it kills the forgetting hole while keeping prose out of the emitter. Measured before committing to it:

```
TypeError: ChatSession._emit_tool_completed() missing 2 required positional arguments
turn RAISED: TypeError ...
```

The `except Exception: pass` guard is *inside* `_emit_tool_completed`, so a missing argument fails at the **call** and propagates out of `turn()`. That converts a telemetry slip into a dead turn, against this method's own invariant — "Analytics must never affect the tool call that just ran." Strictly worse than the gap it fixes.

So the three shapes, measured rather than argued:

| shape | forgetting yields | turn-safe |
| -- | -- | --: |
| cause kwargs, defaulted `""` | "did not fail" — wrong | yes |
| cause kwargs, required | TypeError out of the turn | **no** |
| **derive here from `reason`** | `unclassified` — honest | yes |

Your version wins on a property neither of us named: it degrades to *"we do not know why"* rather than *"it did not fail"*. For a field whose purpose is measuring failures, failing toward success is the worse direction — and that is the exact mutation this PR already caught once.

Done in `a7650751`: `_emit_tool_completed(..., reason: str = "")` derives via `_tool_failure_cause(ok, reason)` inside the `try`, both preambles deleted. `reason` is consumed and dropped; wire payload byte-identical (probe below). Pinned by `test_a_caller_that_omits_reason_degrades_to_unclassified`, which carries that table in its docstring.

On the prose objection I raised against your suggestion: the payload keys are unchanged and asserted exhaustively, so the risk is proximity rather than leakage. The `reason` parameter now carries the warning in its own docstring, naming `identifier` and `key` too.

```
payload keys: [conversation_id, duration_ms, error_type, name, ok,
               root_cause_class, root_cause_tier, surface, turn_index]
id_rsa on wire: False
```

anton/analytics.py
alecantu7: Fixed in `a7650751`. You are right and the consequence is exactly the one you name — an analyst filtering `self_inflicted` for exception-shaped values drops every argument-validation failure, which is one of the two populations the ticket exists to expose.

The register now reads "usually the exception name but a sentinel class where the handler declared the failure itself", names all five (`empty_code`, `missing_argument`, `invalid_argument`, `unknown_resource`, `missing_name`), and says explicitly **do not join this against Python exception names**. I added `timeout` to that warning too — same shape, also not a type name, and reachable on every scratchpad cell timeout (your other thread).

anton/analytics.py
alecantu7: Fixed in `a7650751`. Both directions of the overclaim, and I took your framing that the gap is systematic rather than rare — `tools/registry.py:144` means any handler returning a plain string lands `ok=None`, so this is most handlers.

The register now says the two agree **where both populate**, then states the shortfall as design: an `ok="unknown"` call is counted by the tally as `unclassified` and left blank on the row, because inventing a verdict from prose the model can influence is the ENG-1276 defect one level up. So `sum(tool rows by tier)` runs systematically short of `turn_completed.root_cause_failures`, and that is expected, not drift.

Also noted the inverse you found on the other thread (multimodal `ok=false` populating here and not in the tally) as unreachable-today, so both asymmetries are written down in one place.

anton/core/session.py
alecantu7: Agreed, and taken as documentation rather than a code change — added under *Expected side effects that are NOT regressions* in the PR body.

Your framing is the part worth keeping: `root_cause.py` defers the mis-tiering on the grounds that it is "tolerable while nothing consumes the tier", and **this PR is that consumer**. So the caveat stops being measurement-only the moment this ships, exactly as you say. The body now carries your three measured rows and points at ENG-1531 as the precondition, with the note that a real missing-file wall reads `transient` whenever a path or port carries a bare 3-digit token, and an agent bug reads `external_wall` — both directions skewing the shares people will read off the new dimension.

Not fixing it here for the reason you give: it needs the HTTP-shape precondition on the status branch, which is ENG-1531's call and a change to shared classification logic that half a dozen consumers read.

anton/core/session.py
alecantu7: Declining this one, but recording why rather than leaving it silent — and your own analysis is most of the reason.

You refuted it for the reachable case yourself: on the string branch `_record_root_cause` classifies the same `reason` moments later and `root_cause_classify_errors` fires on `turn_completed`, so a systematically broken classifier stays visible. Prod bears that out — `classify_errors = 0` across 14 days.

The residual you identify is real: on the multimodal arm the tally is never reached, so a classifier exception there would be invisible. But that arm is unreachable today (no handler returns `content=[...]`), and adding `note_error()` here would need a `RootCauseLedger` reference inside a method that deliberately holds none — the emitter's only job is to ship fields.

The asymmetry you name is fair and I would rather it were consistent. I have left it as-is because closing it properly means `_record_root_cause` in both list arms, which is the same change the `#308 review` NOTE already asks the first multimodal migrator to make. That is where it belongs, not here. If you would rather I drop the `pragma` so the guard at least cannot hide from the coverage gate, say so — that half is free.

anton/core/session.py
alecantu7: Confirmed, and the second half of your point is the actionable one — `analytics.py:206` did still claim the two views "cannot disagree", and that is now corrected (see the thread there).

The divergence itself stays documented rather than fixed: `_tool_failure_cause`'s docstring scopes the claim to string results and names what the first multimodal migrator owes, which is `_record_root_cause` in both list arms alongside the nudge the `#308 review` NOTE already asks for there.

On your suggestion to derive one `RootCause` before the emit and hand it to both consumers — that would close this and the `ok=None` gap together, and it is the right eventual shape. I have not done it here because `_record_root_cause` also books the `tool_ok is None` population that `_tool_failure_cause` deliberately refuses (the `reason_coverage` denominator argument at `_record_root_cause:1550-1573`), so unifying them means deciding whether the tool row starts reporting a cause for unmigrated handlers. That is ENG-2248's question, and it is on that ticket.

tests/test_tool_completed.py
tino097: Confirmed the fix at `a722bf21` — no bare-literal `cls` left in `classify`, all four `CLS_*` constants are in `ALL_CLASSES`, and moving the enumeration next to the tables rather than into the test is the better shape. You're right that it was accidental across the board, not just missing `timeout`.

One leftover, comment-accuracy rather than a broken guard. The four reasons added here are commented as covering "the four classes `classify` returns as literals". Two don't reach those branches:

```
transient      timeout        <- Cell timed out after 300s total ...
transient      timeout        <- Cell exceeded inactivity limit
unclassified   unclassified   <- OSError: permission denied opening the workspace
unclassified   unclassified   <- OSError: connection refused by the backend
```

`_WALL_TYPES["OSError"]` is consulted before the lowercase-phrase branches, and the OSError special case falls through to `unclassified` when the text carries none of "No space / Disk quota / Too many open files / Cannot allocate". So `CLS_PERMISSION_DENIED` and `CLS_CONNECTION_REFUSED` are still unexercised by any input.

The test passes either way, since both classes are in `ALL_CLASSES`. Dropping the `OSError: ` prefix reaches them:

```
external_wall  permission_denied   <- permission denied on /etc/shadow
external_wall  connection_refused  <- connection refused by db.internal:5432
```

Related caveat worth a word somewhere: `ALL_CLASSES` makes the *set* honest, but this guard still only catches an unregistered new literal if some input reaches its branch. `root_cause.py`'s new comment says "Adding a class means adding it here, and the guard follows automatically" — true of the set, optimistic about the guard.

Not blocking; approving.

tino097 (approved):
Approving at `a722bf21`. Re-checked all seven threads against the new head; CI green including `run-tests` (4m27s).

Five fixed, one documented, one declined on grounds I agree with:

- **Vocabulary gap** — fixed at the root rather than in the test. `classify()`'s four tail literals are named constants and `root_cause.py` now exports `ALL_CLASSES` / `ALL_TIERS` as the authoritative enumeration, with the guard reading the export. Verified: no bare-literal `cls` remains, all four constants are in the set. Wider than I called it — `permission_denied` and `connection_refused` were passing only by coincidence.
- **Forgettable cause kwargs** — my suggestion to make them required was wrong, and the measurement showing why is the useful part: `TypeError` lands at the call site, outside `_emit_tool_completed`'s guard, so it propagates out of `turn()` and kills a turn over telemetry. Deriving from `reason` inside the emit is better than either shape, and degrading a forgotten `reason` to `unclassified` rather than `""` fails in the honest direction.
- **Property register** — both overclaims corrected, and the `ok=None` shortfall is now stated as design with its cause, which is what an analyst reconciling the two views actually needs.
- **Tier skew** — right call to take it as documentation. ENG-1531 owns the precondition and it touches shared classification every consumer reads.
- **Silent classifier swallow** — fair decline; the reachable case is covered by `classify_errors` on the turn row, as you say.

One non-blocking note left on the vocabulary thread: two of the four new adversarial reasons carry an `OSError: ` prefix, so `_WALL_TYPES` intercepts before the phrase branches and they land in `unclassified` — the two literal branches they were added for are still unexercised. Comment accuracy, not a broken guard.

Merge-order note from the first pass still stands: `git merge-tree` against #431 conflicts in `session.py` and `tests/test_tool_completed.py` only, and #431 first is the cheaper order.

tests/test_tool_completed.py
alecantu7: Fixed in `a908c0fe` — confirmed your reading before touching it, and it reproduces exactly:

```
unclassified  unclassified  <- OSError: permission denied opening the workspace
unclassified  unclassified  <- OSError: connection refused by the backend
```

`_WALL_TYPES["OSError"]` is consulted first and its special case falls through to `unclassified` unless the text says "No space / Disk quota / Too many open files / Cannot allocate". Prefix dropped; all four now land where the comment claims:

```
transient      timeout
transient      timeout
external_wall  permission_denied
external_wall  connection_refused
```

The comment now says *why* there is no prefix, so the next person does not helpfully add one back.

**Your related caveat was the more valuable half, and I have taken it into the module.** You are right that `ALL_CLASSES` makes the set honest while my comment — "Adding a class means adding it here, and the guard follows automatically" — was optimistic about the guard. A closed-set guard only catches an unregistered value if some input reaches the branch returning it, which is precisely what this commit fixes. `root_cause.py` now says so explicitly, and cites this exact miss as the worked example:

> Adding a class means adding it here. That keeps the SET honest; it does not by itself keep a closed-set guard honest, because such a guard only catches an unregistered value if some input actually reaches the branch returning it. Two of ENG-2247's own adversarial inputs missed their branches (an `OSError: ` prefix sent them down `_WALL_TYPES` instead) and the guard still passed. So a new literal needs BOTH an entry here and an input that reaches it.

Thanks for the review — the vocabulary finding was the one that mattered, and the `TypeError` measurement only happened because you pushed on the kwarg shape. Suite green at the new head, 2,929 passed.

PR #433, opened 2026-09-03
ENG-1361: Card a provider failure instead of telling the user to rephrase

alecantu7:
Closes ENG-1361.

## The bug

A provider failure **at request time** — a 5xx, an unreachable host, an unconfirmed 429 — takes anton's count-based retry path. That path had **no terminal capable of producing an error card**. It asked the model to explain the outage (a call that needs the very provider that is failing) and, when that call failed too, emitted:

> An unexpected error occurred: Could not reach the model provider — check your connection or try again in a moment.. Please try again or rephrase your request.

Rephrasing cannot reach an unreachable provider. Users follow the advice anyway — the reported incident shows the user retyping the same request, then leaving.

**127 turns / 55 users in 30 days** (PostHog `turn_completed`, project 424726), on the current builds, desktop **and** web. Reproduced byte-for-byte against `origin/staging` before writing the fix.

## What changed

**1. Convert at exhaustion.** A transient that outlasts the COUNT budget now raises `ProviderOverloadedError`, exactly as the TIME budget already did — the same product event, reported the same way. It is mapped on *both* transports (`provider_overloaded_info()` for desktop, `remote_turn_error()`'s type-name branch for web) — as are four other curated types; what makes it the right pick here is that it is the one whose card matches this failure.

**Zero server/UI changes is true on DESKTOP only.** On web, two pre-existing transport gaps mean the card lands degraded, and this PR routes a new population into them: (a) `_scrub` serialises as `"Type: message"` and `remote_turn_error` keys on the type name, so `code` never survives — an unconfirmed 429 renders the plain overloaded card with no time-gated Retry; (b) the remote payload carries only `{code, error}` (no `reconnectable`), and the renderer treats absent as BYOK by design, making **"Set up MindsHub" the primary action for org-mode users who are managed by construction** — the ENG-514 mis-nudge, into a Settings section org mode hides. Both are tracked as **ENG-2297** — out of scope here (cross-repo), and the ticket prefers sending `reconnectable` on the remote path over the one-line renderer patch, since that fixes the payload asymmetry rather than one symptom of it.

Three details that are easy to get wrong, each covered by a test:
- Raised **before** the summarize prompt is appended, or an orphan `SYSTEM: The task has failed N times` poisons the next turn.
- **Keeps anton's typed message** (ENG-673's "returned 500") and adds only the attempt count. My first pass replaced it with generic copy and silently broke an existing e2e assertion — caught by the suite.
- **Preserves `code="rate_limited"`** for an unconfirmed 429, since telling a rate-limited user the provider is "having an incident" is the exact mis-report ENG-1537 fixed.

**2. Invert the guard.** The summarize path's re-raise allowlist becomes membership of `CURATED_PROVIDER_ERRORS`, declared in `provider.py` beside the classes themselves. The allowlist shape exposed every new type by default; the new parametrised test shows it was short **five** members. Staging's guard was a 3-tuple plus `ProviderAuthError` via a separate branch (4 effective); the new set is 9, so 6 added, of which exactly 5 change behaviour: the two the ticket named — **`TransientProviderError` and `ProviderOverloadedError`, which are the two with the real behaviour change** — plus `ContextOverflowError`, `StructuredOutputError` and `ContentValidationError`. Of those last three, `StructuredOutputError` and `ProviderOverloadedError` are defensive: traced to their raise sites, they cannot reach that `except` in production. Forgetting now means an exception *propagates* (a generic card — safe) rather than becoming prose that blames the user's wording.

Deliberately a tuple, not a marker base class: a mixin would change the MRO of nine exception classes across a version-skew boundary cowork-server already handles defensively. A module-walk test fails on the *next* class, at authoring time.

**Disclosure:** four of the nine curated types have **no card on either transport** (`ContextOverflowError`, `StructuredOutputError`, `TransientProviderError`, `EndpointConfigurationError`) — cowork-server replaces the message with a flat "An unexpected error occurred." and the client renders a buttonless alert. Two are newly moved in here, so for those the turn trades anton's own diagnosis for less text. Still the right trade (the prose asserted a remedy that could not work), but a trade. Adding those cards is the follow-up.

**3. The copy.** The rephrase advice is gone from the generic fallback **and** from the SYSTEM summarize prompt, which was teaching the model to give the same advice on every successful summarization.

**4. Keep it measurable.** `retry_terminal_reason`, `provider_failure_kind`, `provider_http_status`. Without these, change 1 would make the request-time and mid-stream terminals byte-identical in analytics — `code` splits them but is a card contract that never reaches PostHog, and the retry count was a local variable reaching nothing at all.

| Terminal | `ended_by` | `error_type` | `provider_failure_kind` | `retry_terminal_reason` |
|---|---|---|---|---|
| Mid-stream, incident budget out | `error` | `ProviderOverloadedError` | `overload_signal` | `provider_recovery_timeout` |
| Velocity 429, wait allowance out | `error` | `ProviderOverloadedError` | `rate_limit` | `rate_limit_wait_limit` |
| `Retry-After` beyond cap | `error` | `ProviderOverloadedError` | `rate_limit` | `rate_limit_wait_too_long` |
| Request-time, attempts out | `error` | `ProviderOverloadedError` | `http_5xx` / `connection_failure` | `request_attempt_limit` |
| Non-transient, attempts out | `retry_exhausted` | `BadRequestError` | — (see note) | `request_attempt_limit` |

`provider_http_status` is **independent** of `provider_failure_kind`: the kind classifies *provider* failures and is empty on the last row, but the status is a plain fact about the exception and is still recorded when it has one (an SDK `BadRequestError` books `400`). Self-review finding — the two fields answer different questions, and suppressing a real status to keep the pair symmetrical would discard the only signal those turns carry.

Named for *termination*, not exhaustion: `rate_limit_wait_too_long` is a terminal where nothing ran out. `provider_http_status` is omitted rather than sent blank when absent.

## Behaviour change to know about

`ended_by=retry_exhausted` **does not disappear** — non-transient failures (87 turns / 30 users) still exhaust the attempt budget and still summarize, which is right: the model can genuinely explain a 400, and there is no card to route it to.

**In the CLI**, a request-time 5xx now reaches the setup/retry prompt instead of printing prose — the same treatment a mid-stream overload already gets. Two e2e scenarios asserted the old prose behaviour and are updated, with the reasoning in-file. Their non-zero exit under scripted stdin is *not* new: verified on pristine `staging` that a 401 already aborts a scripted run identically.

## Testing

`tests/test_curated_errors.py` — 51 tests. **Every one mutation-verified**: nine targeted mutants (conversion disabled, old allowlist restored, rephrase copy restored, stamping disabled) each killed by exactly the tests that should catch them, and nothing else.

Full suite: **2970 passed, 31 skipped, 0 failed**.

**Also expected side effects, not regressions:** the new terminal drops tool-history rows, so the card's Retry replays into a conversation missing the tool work — the identical trade already governs Path A and is documented at three sites. And when a *curated* exception ends the summarize call, `ended_by` flips to `error` and `error_type` records the summarize exception rather than the originating one; `retry_terminal_reason` still stamps, so those rows stay identifiable.

## Security

Reviewed. Every `TransientProviderError` message is anton-authored template text plus a validated integer status and a hard-coded provider label — audited all construction sites — so interpolating it into the card forwards no provider body to the client. Change 2 strictly **reduces** exposure: types that previously rendered `str(exc)` as assistant text now propagate and get the server's redacted stand-in instead. No credentials, input validation, path or authz surface is touched.

## Delivery

cowork-server pins anton to git **`main`**, so this reaches users only once it lands there — merging to `staging` ships nothing on its own. `remote_turn_error` already maps the type name, but the scratchpad pod runs a **sha-pinned** image (`values.yaml` / `values-prod.yaml`), so web needs anton on `main` **and a manual `SCRATCHPAD_IMAGE` bump** — it is not automatic.

## Not in scope

- Bare `ConnectionError` from the status mapper's catch-all (28 turns / 4 users), incl. the BYOK out-of-credits 402 — a **builtin**, so it cannot join a curated set without swallowing unrelated socket errors. Needs a type: ENG-1283.
- A BYOK provider's own quota routing to the MindsHub top-up card: ENG-2263.

### DIFF
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -638,12 +638,20 @@ class TransientProviderError(ConnectionError):
     def __init__(
         self, message: str, *, provider: str = "", code: str | None = None,
         retry_after: float | None = None, session_backoff: bool = True,
-        model: str = "",
+        model: str = "", status_code: int | None = None,
     ) -> None:
         super().__init__(message)
         self.provider = provider
         self.code = code
         self.retry_after = retry_after
+        # The HTTP status this was classified FROM, when there was one
+        # (ENG-1361). `code` cannot carry it: a downstream conversion to
+        # ProviderOverloadedError needs `code` for the card vocabulary, so the
+        # originating status would otherwise be lost before it reaches
+        # telemetry. None for a mid-stream failure (the status was 200) and for
+        # a connection error (no response at all) — those are genuinely absent,
+        # not unknown.
+        self.status_code = status_code
         # The model that was in flight when this failed — so a downstream
         # ProviderOverloadedError names the ACTUAL model (planning OR coding),
         # not whatever the session defaults to (ENG-673).
@@ -928,11 +936,13 @@ def classify_transient(
         return TransientProviderError(
             f"{provider or 'The model provider'} is momentarily overloaded.",
             provider=provider, code=etype, session_backoff=session_backoff, model=model,
+            status_code=status_code,
         )
     if isinstance(status_code, int) and 500 <= status_code < 600:
         return TransientProviderError(
             f"{provider or 'The model provider'} returned {status_code}.",
             provider=provider, code=f"http_{status_code}", session_backoff=False, model=model,
+            status_code=status_code,
         )
     if status_code == 429 and not b.get("detail"):
         # Plain rate-limit ("slow down"), NOT an out-of-quota 429. Quota 429s are
@@ -968,7 +978,7 @@ def classify_transient(
             provider=provider, code="rate_limited",
             session_backoff=velocity_confirmed,
             retry_after=retry_after if velocity_confirmed else None,
-            model=model,
+            model=model, status_code=status_code,
         )
     return None
 
@@ -1123,6 +1133,100 @@ class EndpointConfigurationError(ConnectionError):
     the interactive CLI reads the type to default such a failure to ``setup``
     rather than ``retry`` (ENG-1145 review).
     """
+
+
+# --------------------------------------------------------------------------- #
+# Curated failures + the analytics vocabulary for them (ENG-1361)
+# --------------------------------------------------------------------------- #
+
+# Every CURATED failure: a typed exception carrying user-ready copy that a host
+# maps to an actionable card. These must FAIL a turn, never be wrapped into
+# assistant prose — the card can only fire when the exception propagates, and
+# the generic "please try again or rephrase your request" fallback is actively
+# wrong for all of them (rephrasing cannot fix an outage, a dead credential, or
+# an empty wallet).
+#
+# Listed HERE, beside the class definitions, rather than at the consumer in
+# `session.py`: the previous allowlist lived next to the `except` that used it,
+# and drifted three times — each omission found by a user hitting it in
+# production rather than by the suite (ENG-1361, ENG-1310). Adding a class now
+# means ignoring the comment directly above it, and `test_curated_errors.py`
+# fails on any exception class defined in this module that is neither listed
+# here nor deliberately excluded.
+#
+# NOT a marker base class on purpose: a mixin would change the MRO of every one
+# of these across a version-skew boundary that cowork-server already handles
+# defensively (it imports each type lazily precisely because anton's version
+# floats underneath it). High risk, and the module-walk test gives the same
+# guarantee without touching the hierarchy.
+#
+# Builtin `ConnectionError` is deliberately ABSENT even though the status
+# mapper's terminal catch-all raises one: it is the base class of half this
+# tuple and of unrelated socket failures, so listing it would silently curate
+# everything. Giving that catch-all its own type is ENG-1283.
+CURATED_PROVIDER_ERRORS: tuple[type[BaseException], ...] = (
+    ContextOverflowError,
+    TokenLimitExceeded,
+    ProviderAuthError,
+    StructuredOutputError,
+    TransientProviderError,
+    ProviderOverloadedError,
+    ModelUnavailableError,
+    ContentValidationError,
+    EndpointConfigurationError,
+)
+
+
+# The analytics vocabulary for WHY the provider failed, kept deliberately small
+# and closed (ENG-1361). `code` on the exception cannot serve this purpose: it
+# selects the client's error card (`provider_overloaded` / `rate_limited` are
+# matched literally in ChatView.jsx), so it is a wire contract with the
+# renderer, not a free label. And the raw codes mix cardinalities — `http_503`
+# next to `connection_error` — which would spread one failure mode across many
+# analytics rows. The HTTP status, when there is one, rides in its own field.
+PROVIDER_FAILURE_KINDS: frozenset[str] = frozenset({
+    "overload_signal",     # the provider SAID it was overloaded / erroring
+    "rate_limit",          # velocity 429 — waiting is the remedy
+    "http_5xx",            # request-time 5xx status
+    "connection_failure",  # never reached it, or the connection dropped
+    "bad_response",        # a 200 whose body was unusable
+})
+
+# Codes that mean "we got a 200 and the body was unusable": an unclassifiable
+# mid-stream error event, a stream that stopped early, or one that never
+# started. Distinct from `overload_signal` because the provider told us
+# nothing about why — claiming overload here would over-report incidents.
+_BAD_RESPONSE_CODES = frozenset({"stream_error", "truncated_stream", "empty_response"})
+
+
+def provider_failure_kind(code: str | None) -> str:
+    """Map a `TransientProviderError.code` to `PROVIDER_FAILURE_KINDS`.
+
+    Returns "" for anything unrecognised rather than guessing — an empty value
+    in analytics is a prompt to extend the vocabulary, whereas a wrong one is
+    invisible. Every code anton currently mints is covered; the exhaustiveness
+    check lives in `test_curated_errors.py`.
+    """
+    if not code:
+        return ""
+    if code in _TRANSIENT_ERROR_TYPES:
+        return "overload_signal"
+    if code in _BAD_RESPONSE_CODES:
+        return "bad_response"
+    if code == "rate_limited":
+        return "rate_limit"
+    if code == "connection_error":
+        return "connection_failure"
+    if code.startswith("http_"):
+        # Only 5xx is minted with this prefix (`classify_transient`), but parse
+        # rather than trust it: a future 4xx would otherwise be mislabelled as a
+        # server fault, which is the one direction that misleads an operator.
+        try:
+            status = int(code[len("http_"):])
+        except ValueError:
+            return ""
+        return "http_5xx" if 500 <= status < 600 else ""
+    return ""
 
 
 @dataclass
--- a/anton/core/session.py
+++ b/anton/core/session.py
@@ -4695,23 +4695,23 @@ class ChatSession:
                             _model = (getattr(_agent_exc, "model", "") or "") or getattr(
                                 self._llm, "planning_model", ""
                             ) or ""
-                            if _agent_exc.code == "rate_limited":
-                                # An UNCONFIRMED velocity 429 reaches the count
-                                # path (session_backoff=False). It must keep the
-                                # rate-limit code and copy: telling a
-                                # rate-limited user the provider is "having an
-                                # incident" is the mis-report ENG-1537 fixed,
-                                # and buying credits cannot lift a per-minute
-                                # ceiling.
-                                raise ProviderOverloadedError(
-                                    "Too many requests too quickly — this didn't clear after "
-                                    f"{_retry_count} attempts. Waiting a moment and continuing "
-                                    "should work; this isn't a credits problem.",
-                                    provider=getattr(_agent_exc, "provider", "") or "",
-                                    model=_model,
-                                    code="rate_limited",
-                                    retry_after=getattr(_agent_exc, "retry_after", None),
-                                ) from _agent_exc
+                            # NO rate-limit special case here, deliberately.
+                            # A 429 reaching the COUNT path is by definition one
+                            # `classify_transient` could NOT confirm was a
+                            # velocity limit (`session_backoff=velocity_confirmed`),
+                            # and its docstring names that exact population: a
+                            # daily quota in a dialect the string-exact billing
+                            # guards miss (Gemini's RESOURCE_EXHAUSTED) "would
+                            # otherwise spend the whole budget waiting out a
+                            # daily quota that resets at midnight — then be told
+                            # it is not a credits problem." Promising "waiting
+                            # should work" and denying a credits problem is
+                            # precisely that mis-report, on precisely that
+                            # population. It also buys nothing: an unconfirmed
+                            # 429 carries `retry_after=None` (same line), so the
+                            # rate_limited card has no interval to time-gate its
+                            # Retry with. The generic branch below is honest —
+                            # anton's own typed message, no claim either way.
                             # KEEP anton's own classification in the copy. The
                             # typed message already says what happened ("The
                             # model provider returned 500.") and ENG-673 put it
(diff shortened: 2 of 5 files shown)


alecantu7:
## Review trail (convention 6)

@pnewsam — before you start: a **deep adversarial review** already ran on this branch, and four commits of fixes came out of it. Posting the trail because the PR page shows none of it, and two findings are deliberately still open — that distinction is the part worth having up front.

Shape of it: 7 dimensions (cross-repo desktop, cross-repo web, the curated set, the retry terminals, analytics, e2e/test quality, security+copy), each with an independent adversarial pass whose job was to kill the findings. **11 survived, 24 were killed.** 76 claims from the PR body were verified, 34 falsified. Reviewed at `72056b6c`; fixes at `0be455a5`, each re-verified by execution rather than by reading the commit.

### Fixed, and how each was confirmed

| # | Finding | Confirmation at `0be455a5` |
| -- | -- | -- |
| **5** | The unconfirmed-429 branch told users "this isn't a credits problem" and "waiting a moment should work", with a Retry that cannot succeed — on precisely the 429s `classify_transient` could NOT confirm were velocity limits. Its docstring names that population and that exact failure. A test asserted `"credits" in body`, pinning it. | Branch **deleted**. The remaining copy now sits inside `if isinstance(_agent_exc, TransientProviderError) and _agent_exc.session_backoff:` — the velocity-**confirmed** path, where it is true and carries a real `retry_after`. The comment documents the split. |
| **4** | The wrap-up call used raw `plan_stream`, the only `plan_stream` site without the compaction wrapper — so curating `ContextOverflowError` turned a self-fixable failure into a card-less "An unexpected error occurred." | Now `self.plan_stream_with_recovery(...)`; a fourth consecutive overflow still propagates. |
| **2** | The comment claimed forgetting a type means it "PROPAGATES (a generic card — safe)". False twice: the runtime guard is still an allowlist, and the generic path replaces anton's message with a flat string in a **buttonless** alert. | Comment rewritten to state both, plus that four members have no card on either transport. |
| **2b** | The module-walk test covered only `provider.py`. Adding `class ProviderRegionBlockedError(ConnectionError)` to `openai.py` left the whole suite green and the class uncurated. | Re-ran that exact repro: **now fails 1 test.** |
| **7** | Three mutants of new code survived all 2963 tests — the SYSTEM prompt's rephrase wording (the half of fix 3 governing the path that *survives* this change), the count terminal's model attribution (the test was vacuous: both candidate sources held the same string, and cowork-server reads it back to decide `reconnectable`), and `classify_transient`'s `status_code` plumbing (sole production writer of the new field). | All three re-run: **1 failed each.** |

Two things the fixes got that the review did not: an unconfirmed 429 carries `retry_after=None`, so the `rate_limited` card had no interval to time-gate with — the branch was not just wrong, it bought nothing. And `test_an_unconfirmed_429_makes_no_claim_about_credits_or_waiting` pins the **absence** of that copy (no `credits`, no `should work`, no `incident`, `code == "provider_overloaded"`) rather than the presence of a replacement, which is the more durable assertion.

Also dropped: `(ENG-1361)` no longer appears in the live SYSTEM prompt, where a paraphrasing model could have echoed an internal ticket id into a customer's chat.

### Deliberately still open — please read these as scoped, not unfinished

**1. The web transport's card offers the wrong primary action, and it cannot be fixed in this repo.** The body's "zero server or UI changes, already mapped on BOTH transports" is true of the type and false of the card contract:

- `code="rate_limited"` does not survive `_scrub`'s `"ExceptionType: message"` wire format, and `remote_turn_error` keys on the scrubbed type name — so a web 429 renders the **provider_overloaded** card, not `RateLimitedCard`, with no `retry_after` to time-gate Retry. cowork-server's own comment at `turn_errors.py:955` already documents that the code cannot survive that wire. Pre-existing; this PR routes a new population into it.
- The remote branch sends only `{code, error}` (`responses.py:1046-1050`); the in-process branch also sends `model`, `provider_label`, `reconnectable` (`:1308-1338`). With `reconnectable` absent the renderer takes the BYOK branch **by documented design** (`turnErrorActions.js:19-30`) and makes **"Set up MindsHub" the primary button** — for a population that is minds-cloud by construction (`producer.py:111`), routing into a Settings section org mode hides (`SettingsView.jsx:1169`). That is the ENG-514 mis-nudge on a surface ENG-514 did not cover.

The fix is a one-liner in the **cowork** renderer (`reconnectable={m.reconnectable ?? orgMode}`), so it is out of scope here by construction. Follow-up ticket being filed; two open questions on it — whether it belongs in `cowork` (the one-liner) or `cowork-server` (stop the transports diverging, which also fixes the *next* field), and whether it is better as a scoped reopen of ENG-514.

**2. Four of the nine curated types have no card on either transport** — `ContextOverflowError`, `StructuredOutputError`, `TransientProviderError`, `EndpointConfigurationError`. This is now **disclosed** in the comment at the `except` rather than resolved. Finding 4's compaction fix defuses most of the practical consequence (an overflow now compacts instead of propagating), but it is a known state, not an oversight. Worth a deliberate yes/no from you on whether `TransientProviderError` in particular belongs in the set, since propagating it raw is the one case that puts a non-card-authored string on a card.

### Verified — no need to re-check these

- `ProviderOverloadedError` maps to the `provider_overloaded` card on **desktop** identically from the new raise site.
- The new terminal genuinely raises **before** the summarize prompt is appended — no orphan `SYSTEM: The task has failed N times` poisoning the next turn.
- **Path A is unchanged** (the ticket's regression guard: a mid-stream drop still cards after the time budget).
- anton's typed message is preserved, so ENG-673's "returned 500" still reaches the user.
- The **security claim holds.** Every `TransientProviderError` construction site was enumerated: anton-authored template text plus a validated integer status and a hard-coded provider label — no SDK body, `.message` or `str(exc)` is interpolated. Change 2 strictly *reduces* exposure.
- `compare_digest` in `artifacts.py:295` is genuinely untouched — both sides of that comparison stay filesystem-derived.
- Org-mode artifact addressing is id-based (`artifact_scope.py:88`), so a display-label change cannot misroute it.

### Numbers, corrected

The body predates the later commits. At `0be455a5`: **3048 passed / 31 skipped**, and `tests/test_curated_errors.py` collects **51** (the body's 42, and the 2961 full-suite figure, were both accurate when written).

Two arithmetic notes for the record. The old allowlist held **three** types (`TokenLimitExceeded`, `ModelUnavailableError`, `EndpointConfigurationError`) plus `ProviderAuthError` via a separate branch = four effectively re-raised; the tuple has nine, so **six added and exactly five newly propagating** — mutant-confirmed. The body names three of those five; `TransientProviderError` and `ProviderOverloadedError` are the undisclosed two, and they are the ones with the real behaviour change. And "the only type already mapped on both transports" — five are; the operative fact (ProviderOverloadedError is) still holds.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (5 files, +997/−28), the surrounding retry/terminal machinery in `session.py`, `classify_transient` and the new `CURATED_PROVIDER_ERRORS` in `provider.py`, the mid-stream construction sites in `openai.py`/`anthropic.py`, and `test_curated_errors.py`. The change converts a request-time transient that outlasts the count budget into the same `ProviderOverloadedError` card the time budget already raises, replaces the summarize-path allowlist with `CURATED_PROVIDER_ERRORS`, drops the misleading rephrase copy, and adds three telemetry fields. No actionable findings — the one inconsistency I chased (mid-stream terminals emitting `provider_http_status="200"`) is disproven by the production path, which carries `status_code=None`. No actionable findings.

### Validation

- Confirmed head SHA `0be455a5`; all 14 CI checks green, including `run-tests` (4m59s) and `verdict-eval`.
- Traced the count-path terminal: `TransientProviderError` (session_backoff=False) → `_retry_count` exhausts → stamps `request_attempt_limit` → raises `ProviderOverloadedError` before the SYSTEM prompt is appended. Non-transients still summarize with `ended_by="retry_exhausted"`. Matches the terminal table and tests.
- Verified `plan_stream_with_recovery` reads `self._history` via the default factory and re-raises `ContextOverflowError` on the 4th overflow — the wrap-up was the only `plan_stream` site on `staging` lacking recovery; the change makes it consistent with every other site in the file.
- Disproved the suspected telemetry bug: `openai.py` documents mid-stream overloads as a "BARE APIError with no status_code", so `classify_transient` gets `None`, `_stamp_retry_terminal` books `None`, and the field is omitted — exactly as the design comments state. The `(200, …)` classify test is a defensive case no live provider hits.
- Confirmed the cancelled-path clearing of the three new fields is ordered before the `elif exc` branch in `_emit_turn_cost` and is covered by `test_a_stop_after_a_retry_terminal_books_no_terminal_reason`.

One informational note (not a finding, no change requested): routing the wrap-up through `plan_stream_with_recovery` also enables `native_web_tools` on that call when the session has them, which the raw `plan_stream` call did not. It's consistent with every other recovery site and harmless for a summarize prompt, so I'd leave it — flagging only so it's a known, deliberate delta.

alecantu7:
## Re-requesting review — head has moved twice since the approval

@pnewsam your approval is recorded on `0be455a5`; head is now **`96b1a6c5`**. GitHub still shows the PR as Approved, which is exactly the condition under which the new commit gets skipped — so, concretely, here is what is *not* covered by that review:

### One substantive change

**`577c9b51` — keep the session's native web tools off the retry wrap-up call.** This is your own informational note, taken up rather than left.

Routing the wrap-up through `plan_stream_with_recovery` attached `self._native_web_tools` unconditionally — the wrapper adds them independent of the `tools` argument, so passing only `system=` still got them. Two reasons that is wrong on this call specifically, neither applying to the agent-loop callers:

- the prompt says *"Stop retrying"* and asks the model to summarize and explain, so handing it a research tool contradicts the instruction; and
- that prompt embeds `Latest error: {exc}`, which can carry file paths and user content — a provider-side search would take a query derived from it to a search backend. That is a new hop for error text, introduced by accident.

`plan_stream_with_recovery` gains `allow_native_web_tools: bool = True`. The default keeps all five existing callers byte-identical; only the wrap-up passes `False`. It needs that method for the **compaction**, not for what rides along with it.

Second-order path checked: on overflow the wrap-up calls `_summarize_history`, which uses `self._llm.summarize(...)` — a separate method that never touches the injection. So no path from this call reaches a provider-side search.

Mutation-verified, one test each: reverting the opt-out; dropping the argument at the call site; flipping the default to `False` (which would silently strip the whole agent loop's tools); and reverting the compaction change back to raw `plan_stream`.

### Two rebases, no content change

`0be455a5` → `96b1a6c5` also replayed the branch onto current `staging`, which has since taken **#432** (ENG-2247) and **#431** (ENG-2243) — both of which touch `session.py`, and #431 also `turn_cost.py` and `analytics.py`.

Verified rather than assumed, because #431 adds a `TurnCost` field *and* new `turn_completed` properties in the same two places this PR does:

- both field sets coexist (`retry_terminal_reason` / `provider_failure_kind` / `provider_http_status` alongside `attempt_id`);
- in the emission, #431's `turn_attempt_id=` lands **after** this PR's conditional `**{...}` spread for `provider_http_status`. Legal, and green — but it is the shape a clean textual merge would hide, so it was worth executing rather than reading;
- #431's `analytics.py` changes are comments plus the new property — no change to how values are serialised, so the conditional-omit behaviour is unaffected.

Suite on the rebased branch: **3034 passed / 31 skipped / 0 failed**. All 13 checks green at `96b1a6c5`, 0 behind `staging`.

### One question from earlier that is still open

Whether **`TransientProviderError`** belongs in `CURATED_PROVIDER_ERRORS` at all. Propagating it raw is the one case that puts a non-card-authored string in front of the user, since it has no card on either transport. It is disclosed in the comment at the `except` rather than resolved, and a deliberate yes/no would close it either way.

Separately: the web-transport gaps flagged in the earlier review are now tracked as **ENG-2297** (wrong primary CTA for managed users, plus `code` not surviving `_scrub`) — out of scope here by construction, since the fix is in `cowork`/`cowork-server`.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Re-review at head `96b1a6c5`. The only delta since my earlier approval (`0be455a5`) is the `allow_native_web_tools` commit, which resolves the one informational note I left last time — the retry-exhausted wrap-up silently acquired the session's native web tools when it moved onto `plan_stream_with_recovery`. That commit gates the injection behind a new `allow_native_web_tools: bool = True` parameter and passes `False` only at the wrap-up site. No actionable findings.

### Validation

- Confirmed head `96b1a6c5`; all CI green — `run-tests` (5m38s), `verdict-eval`, CodeQL, Snyk, build all pass.
- Verified the guard is surgical: `plan_stream_with_recovery` now reads `if allow_native_web_tools and self._native_web_tools`; the wrap-up caller is the sole `False`; the other four recovery callers (agent loop) inherit the `True` default and stay byte-identical.
- Confirmed the guard is applied once at kwargs construction, so tools stay off across the compaction/retry iterations inside the recovery loop — the security intent (no error-derived query reaching a provider-side search) holds for every overflow restart, not just the first stream.
- Checked the injection isn't re-applied elsewhere on this path: the non-streaming `plan_with_recovery` is untouched and unused by the wrap-up; on overflow the wrap-up's `_summarize_history` uses `self._llm.summarize(...)`, which never touches the native-web-tools injection.
- New regression tests are sound: `test_the_wrapup_call_does_not_get_the_sessions_web_tools`, `test_other_recovery_callers_keep_their_web_tools`, and `test_the_wrapup_still_compacts_on_overflow` pin all three properties (opt-out, default preserved, compaction survives).

Everything approved previously — the count-path `ProviderOverloadedError` conversion, the `CURATED_PROVIDER_ERRORS` allowlist, the dropped rephrase copy, and the three telemetry fields — is unchanged.

PR #435, opened 2026-09-03
ENG-2248: declare explicit ok/reason verdicts on six anton tool handlers

alecantu7:
## What

18 of 27 tools never declare success or failure. **3,038 of 18,108 `tool_completed` rows in the last 30 days carry `ok="unknown"` (16.8%)**, and for every one of them the ENG-1276 substring fallback — not the handler — decides whether the agent's error streak climbs.

This migrates the six highest-volume anton handlers to an explicit `ToolOutcome` verdict: `recall_skill` (1,153 calls), `open_artifact` (279), `list_artifacts` (273), `memorize` (129), `read_image` (128), `create_skill_draft` (72). Together **2,034 of the 3,038 unknown rows**.

## This is not a measurement-only change

The ticket originally said it was. That was wrong, and it has been corrected.

`_apply_error_tracking` opens with `if ok is not None:` and uses the handler's verdict **directly** to drive `error_streak[tool_name]`, which fires the resilience nudge at 2 and the circuit breaker at 5. Every `ok=` added here is a behaviour decision. So the migration is tiered on purpose:

| Tier | What | Verdict | Behaviour |
|---|---|---|---|
| 1 | Verified successes | `ok=True` | **Free.** The legacy matcher finds none of its five markers in a success body, so the streak reset either way. Proven, not asserted — see below. |
| 2 | Indisputable failures | `ok=False` | **A real change, accepted.** The call cannot succeed on retry, so it belongs in the streak. |
| 3 | Ambiguous outcomes | left `ok=None` | `recall_skill`'s three `NO MATCH` returns, `open_artifact`'s "no artifact found", `memorize`'s "Memory encoding is disabled" — each with a comment saying why. These keep reporting `unknown`. |

The tier-3 line is the one worth arguing about. `NO MATCH` is not a failure: the response *lists the skills that do exist* and tells the model to proceed without one. Giving it `ok=False` would mean five consecutive "no skill for this" answers trip the breaker on an agent that is working fine.

## `read_image` — the tree's first multimodal verdict

`read_image` returns a `list[dict]` (image block + text block) on success. It now returns `ToolOutcome(content=[...], ok=True)`, and it must stay `ok=True`.

Both tool loops' list arms skip the nudge/breaker **text** and skip `_record_root_cause` — documented in the NOTE at `session.py`'s non-streaming list arm and in `_tool_failure_cause`'s docstring, which calls the shape "unreachable today". A handler returning a list with `ok=False` would climb the streak, trip nothing the model can see, and put a cause on no turn tally. Every failure branch in `read_image` returns a plain string, and `test_read_image_never_pairs_a_list_with_a_failure_verdict` is an AST scan that keeps that shape shut tree-wide rather than relying on today's inputs not reaching it.

## Tests

`tests/test_tool_verdict_migration.py` — 15 tests that state both directions explicitly instead of leaving them to be discovered in production:

- each migrated success produces the **same streak outcome** and the **same result text** as the bare string did (tier 1 is free);
- a migrated failure **now reaches the streak where a bare string did not** (tier 2 is real);
- the **two-failure nudge** and **five-failure breaker** are asserted as *accepted consequences*, named as such in the test names;
- one success clears a streak built from migrated failures;
- the `NO MATCH` family is asserted **still unverdicted on purpose** — so removing that deliberate gap fails a test rather than passing silently;
- `create_skill_draft` with no store is the one tier-2 case here that **is** trip-eligible (`store_unavailable` was already `TIER_WALL` before this PR), stated explicitly.

Seven mutations verified, each killed by the intended test: giving `NO MATCH` a verdict, flipping a tier-1 success to `ok=False`, removing the nudge threshold, `read_image` success → `ok=False` on a list, `missing_file` remapped to `TIER_WALL`, `read_image` not-found reverted to a bare string, `create_skill_draft` success verdict dropped.

Full suite: **2,944 passed, 31 skipped.**

## What is still unknown after this — not complete, and not claimed to be

Projected residual **~1,004 rows (5.5%)**, by owner:

| Owner | Rows | Tools |
|---|---|---|
| **cowork-server** | 438 | `set_status` 216, `lookup_connector` 85, `request_credentials` 58, `report_success` 40, `report_failure` 25, `set_field_status` 11, `label_connection` 2, `request_extra_field` 1 |
| **anton, not yet migrated** | 532 | `scratchpad` 216 (1.5% of 14,315 — mostly verdicted already), `web_fetch` 95, `select_path` 81, `connect_new_datasource` 68, `ask_user` 38, `publish_or_preview` 34 |
| **unlocated** | 34 | `generate_artifact` 18, `generate_prd` 16 — no tool definition found in any repo in the workspace |

Two caveats stated rather than buried:

1. **5.5% is a floor.** The deliberate tier-3 branches inside the six migrated handlers keep emitting `unknown`, and their share of those 1,561 calls is not measurable from PostHog today — `unknown` rows carry no `reason`. The exact figure lands one release after this ships, when the `ok` column reports it directly.
2. `select_path` (81) and `ask_user` (38) are **deliberately deferred**. Both return through the shared `_status()` choke point at `tool_handlers.py:968`; migrating them means deciding a vocabulary for that helper, which is its own change and does not belong wedged into this one.

The cowork-server half is smaller than it looks, and **not** the same fix as this PR. Seven of its eight tools are the connector probe, and every one of those handlers returns the literal string `"ok"` — which contains none of the five legacy markers, so the substring fallback already classifies them correctly. For 293 of the 441 rows there is no behavioural harm, only a cosmetic `unknown` where PostHog could say `true`. The three non-probe tools (`lookup_connector` 87, `request_credentials` 59, `label_connection` 2) do return real error strings, some carrying no marker at all, so those failures read as successes and reset the streak — real, but confined to the error branches of ~148 rows a month.

**Correction to an earlier draft of this body.** It claimed this work also resolved the 490 unjoinable rows tracked in the measurement ticket for `tool_completed` — "same tools, same root cause". That is wrong. Those rows are unjoinable because they carry **no `conversation_id`** (the probe runs outside a turn), which is a different field from `ok`; a verdict migration fixes none of them. Measured:

```
tool                     no_cid   no_turn    rows
set_status                  216         0     216
report_success               40         0      40
report_failure               25         0      25
set_field_status             11         0      11
request_extra_field           1         0       1
lookup_connector              0         0      87
request_credentials           0         0      59
label_connection              0         0       2
TOTAL                       293         0     441
```

The two anomalies share a set of tools, not a fix. Tracked as notes on ENG-2301 rather than a separate ticket.

## Security

No new input reaches a shell, a filesystem path, or a log. The `reason` values are fixed literals from the closed `_SENTINEL_REASONS` vocabulary, never interpolated from tool input. The result text the model reads is byte-for-byte what it was before the migration, asserted per handler.

Linear: ENG-2248

### DIFF
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -142,17 +142,33 @@ _SENTINEL_REASONS = {
     # A refused package spec (flag/URL/path-shaped entry) is the agent's own
     # bad argument, not an environment wall (ENG-1635).
     "package_install_rejected": (TIER_SELF, "invalid_argument"),
+    # `read_image`, ENG-2248: both are the agent's own file choice, and both
+    # messages tell it what to do instead (use a real image / resize).
+    "not_an_image": (TIER_SELF, "invalid_argument"),
+    "image_too_large": (TIER_SELF, "invalid_argument"),
     # The agent named something that does not exist. Ambiguous — it could be a
     # genuinely absent resource — but the agent chose the identifier and can
     # list the real ones, so it resolves to the non-tripping side.
     "artifact_not_found": (TIER_SELF, "unknown_resource"),
     "unknown_datasource": (TIER_SELF, "unknown_resource"),
+    # Same shape for a path: `read_image` reports the file the agent named as
+    # absent. Genuinely absent files exist, but the agent supplied the path and
+    # can list the directory, so it resolves to the non-tripping side.
+    "missing_file": (TIER_SELF, "unknown_resource"),
     # Genuine walls: the environment is missing something the agent cannot add.
     "package_install_failed": (TIER_WALL, "missing_dependency"),
     "store_unavailable": (TIER_WALL, "service_unavailable"),
     # Could be environment or config and the sentinel does not say which, so it
     # stays out of every trip rung until something distinguishes them.
     "launch_failed": (TIER_UNCLASSIFIED, "unclassified"),
+    # `read_image`'s catch-all read failure wraps a bare `except Exception`, so
+    # one sentinel covers a permissions wall, a corrupt file the agent itself
+    # wrote, and a decode bug. Nothing here distinguishes them, so it stays out
+    # of every trip rung (ENG-2248).
+    "read_failed": (TIER_UNCLASSIFIED, "unclassified"),
+    # A bare `except Exception` around a PIL round-trip: a missing Pillow is a
+    # wall, a corrupt BMP is not, and the sentinel cannot tell them apart.
+    "bmp_convert_failed": (TIER_UNCLASSIFIED, "unclassified"),
 }
 
 # Identifier extraction, per class. Kept narrow on purpose: a wrong identifier
--- a/anton/core/root_cause.py
+++ b/anton/core/root_cause.py
@@ -154,7 +154,15 @@ _SENTINEL_REASONS = {
     # Same shape for a path: `read_image` reports the file the agent named as
     # absent. Genuinely absent files exist, but the agent supplied the path and
     # can list the directory, so it resolves to the non-tripping side.
-    "missing_file": (TIER_SELF, "unknown_resource"),
+    #
+    # NOT named `missing_file`, though that is what it describes (#435 review).
+    # `missing_file` is already a CLASS in this module, reached by a different
+    # path — `_WALL_TYPES` maps FileNotFoundError to it, `_STATUS_WALLS` maps
+    # 404 to it, it is the sole member of `_EXACT_ONLY_CLASSES`, and it gets
+    # path-identifier extraction below. A sentinel KEY of the same name that
+    # resolves to a DIFFERENT class is the kind of collision a future editor
+    # reads straight past.
+    "path_not_found": (TIER_SELF, "unknown_resource"),
     # Genuine walls: the environment is missing something the agent cannot add.
     "package_install_failed": (TIER_WALL, "missing_dependency"),
     "store_unavailable": (TIER_WALL, "service_unavailable"),
--- a/anton/core/tools/recall_skill.py
+++ b/anton/core/tools/recall_skill.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from anton.core.tools.registry import ToolOutcome
 from anton.core.tools.tool_defs import ToolDef
 
 if TYPE_CHECKING:
@@ -115,20 +116,50 @@ def _already_in_history(session, label: str) -> bool:
         return False
 
 
-async def handle_recall_skill(session: "ChatSession", tc_input: dict) -> str:
-    """Look up a skill by label and return its declarative procedure."""
+async def handle_recall_skill(
+    session: "ChatSession", tc_input: dict
+) -> "str | ToolOutcome":
+    """Look up a skill by label and return its declarative procedure.
+
+    Verdicts (ENG-2248). `ok` drives the per-tool error streak, so it fires the
+    resilience nudge at 2 consecutive failures and the circuit breaker at 5 —
+    a verdict here is a behaviour decision, not a label. This tool ran 1,150
+    times across 452 installs in 30 days, so a wrong `ok=False` reaches
+    everyone.
+
+    * `ok=True`  — a procedure was returned, INCLUDING the already-recalled
+      stub: that is a success with a deliberately short body, not a failure.
+    * `ok=False` — the call could not be served at all: no label, or no store.
+      Repeating either cannot help, so it SHOULD reach the streak. This is the
+      intended, accepted behaviour change.
+    * `ok=None`  — the NO MATCH family, left unmigrated ON PURPOSE. See the
+      comments at those returns.
+    """
     label_in = (tc_input.get("label") or "").strip()
     if not label_in:
-        return (
-            "ERROR: recall_skill requires a non-empty 'label' parameter. "
-            "Pick one from the procedural memory list in your system prompt."
+        # Tier 2 (ENG-2248): a malformed call. Retrying it unchanged cannot
+        # work, so repetition IS thrash and belongs in the streak.
+        return ToolOutcome(
+            content=(
+                "ERROR: recall_skill requires a non-empty 'label' parameter. "
+                "Pick one from the procedural memory list in your system prompt."
+            ),
+            ok=False,
+            reason="missing_name",
         )
 
     store = getattr(session, "_skill_store", None)
     if store is None:
-        return (
-            "ERROR: no skill store is wired into this session. "
-            "Procedural memory is unavailable right now."
+        # Tier 2: the host wired no store, so no label can ever work.
+        # `store_unavailable` is an existing `_SENTINEL_REASONS` key mapping to
+        # external_wall/service_unavailable — reused, not invented.
+        return ToolOutcome(
+            content=(
+                "ERROR: no skill store is wired into this session. "
+                "Procedural memory is unavailable right now."
+            ),
+            ok=False,
+            reason="store_unavailable",
         )
 
     skill = store.load(label_in)
@@ -137,6 +168,13 @@ async def handle_recall_skill(session: "ChatSession", tc_input: dict) -> str:
         closest = store.closest_match(label_in)
         if closest is None:
             available = [s["label"] for s in store.list_summaries()]
+            # Tier 3 (ENG-2248): deliberately ok=None, NOT a failure. The tool
+            # worked — it looked, found nothing, and told the model to proceed.
+            # This is the most common non-success on the highest-volume tool, so
+            # ok=False here would push normal exploration into the error streak
+            # and trip the breaker on correct behaviour. If it should ever
+            # count, that needs its own ticket and its own before/after on the
+            # nudge rate.
             if not available:
                 return (
                     f"NO MATCH: no skill named '{label_in}', and the procedural "
@@ -148,7 +186,11 @@ async def handle_recall_skill(session: "ChatSession", tc_input: dict) -> str:
             )
         skill = store.load(closest)
         if skill is None:
-            # Race or filesystem flake — be defensive
+            # Race or filesystem flake — be defensive.
+            # Tier 3: left ok=None with the rest of the NO MATCH family.
+            # Arguably a real failure (a load that should have worked), but it
+            # is indistinguishable from a plain miss without a store-level
+            # signal, and guessing wrong costs a breaker trip.
             return (
                 f"NO MATCH: '{label_in}' was not found and the closest "
                 f"candidate '{closest}' could not be loaded."
@@ -171,18 +213,29 @@ async def handle_recall_skill(session: "ChatSession", tc_input: dict) -> str:
         # procedure header — otherwise a stub surviving compaction would
         # satisfy _already_in_history forever and the full contract would
         # never be re-sent.
-        return (
-            f"Skill '{skill.label}' was already recalled in this conversation "
-            "— its full procedure is in your context above, under the "
-            f"'# Skill: {skill.name}' heading, and still applies. Not "
-            "re-sending the body."
+        # Tier 1 (ENG-2248): a SUCCESS with a deliberately short body. The
+        # procedure is already in context and still applies, so the tool did
+        # its job. Behaviourally identical to today — a bare-string return
+        # already resets the streak, since the legacy matcher finds none of its
+        # five markers in this text.
+        return ToolOutcome(
+            content=(
+                f"Skill '{skill.label}' was already recalled in this conversation "
+                "— its full procedure is in your context above, under the "
+                f"'# Skill: {skill.name}' heading, and still applies. Not "
+                "re-sending the body."
+            ),
+            ok=True,
         )
 
     # Increment the recommended counter for the *resolved* label, not the
     # input. If the LLM typo'd 'csv-sumary', we credit 'csv-summary'.
     store.increment_recommended(skill.label, stage=1)
 
-    return _format_skill_response(skill, warning=warning)
+    # Tier 1: the procedure was returned. Also covers the closest-match path,
+    # where `warning` explains the substitution — a substitution is still a
+    # served request.
+    return ToolOutcome(content=_format_skill_response(skill, warning=warning), ok=True)
 
 
 RECALL_SKILL_TOOL = ToolDef(
--- a/anton/core/tools/skill_draft.py
+++ b/anton/core/tools/skill_draft.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from anton.core.tools.skill_format import SKILL_FILE, normalize_name
+from anton.core.tools.registry import ToolOutcome
 from anton.core.tools.tool_defs import ToolDef
 
 if TYPE_CHECKING:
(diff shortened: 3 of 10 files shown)


alecantu7:
Corrected a claim in the PR body before anyone reviews against it.

It said the cowork-server follow-up "also closes ENG-1861's 490 unjoinable rows — same tools, same root cause". **Wrong.** Those rows are unjoinable because they carry no `conversation_id` — the connector probe runs outside a turn — which is a different field from `ok`. A verdict migration fixes none of them. Measured over 30 days:

```
tool                     no_cid   no_turn    rows
set_status                  216         0     216
report_success               40         0      40
report_failure               25         0      25
set_field_status             11         0      11
request_extra_field           1         0       1
lookup_connector              0         0      87
request_credentials           0         0      59
label_connection              0         0       2
TOTAL                       293         0     441
```

I had identified the two anomalies as covering the same set of tools and then carried that over to "the same fix closes both", which does not follow.

While checking, the cowork-server half also turned out smaller than the body implied: seven of its eight tools are the probe, and every one of those handlers returns the literal `"ok"`, which trips none of the five legacy markers — so the fallback already gets them right and the gap there is cosmetic. The real (small) defect is the other way round, in `lookup_connector` / `request_credentials` / `label_connection`, whose error strings carry no marker and so read as successes, resetting the streak. Body updated with both.

No code change — the diff under review is unaffected.

alecantu7:
Rebased onto current `staging` (`1933d819`) — the branch was based on `c79131de`, 8 commits behind.

Two files were touched by both sides: `anton/core/tools/tool_handlers.py` and `tests/test_cloud_turn_session.py`, both from [ENG-1382] (#430). `git merge-tree` reported clean, but I did not stop there — that tool undercounted a semantic interaction earlier in this same ticket family, so I rebased and re-ran everything:

**Full suite on `1933d819`: 3,034 passed, 31 skipped.**

Checked specifically whether #430 introduced anything this PR should have covered: it added credential-scoping logic *inside* an existing handler, no new tool and no new bare-string returns, so the ENG-2248 handler inventory is unaffected.

alecantu7:
All three findings actioned in `22344477`, plus the housekeeping. Verified each in code before acting — one of them was not implementable as written.

**F1 — the seam guard. Taken, but inverted, because the literal version cannot pass.** The ticket asks for a guard that "fails when a handler returns a bare value where a `ToolOutcome` is expected". Measured before building it: 9 handlers in `tool_handlers.py` alone have **49 bare returns**, and three of those are *migrated* handlers whose bare returns are deliberate tier-3 exits (`recall_skill`'s NO MATCH family, `open_artifact`'s "no artifact found", `memorize`'s "encoding is disabled"). That guard fails on day one, on purpose-built code.

So it is an **inventory lock** over `anton/core/tools/` instead — which delivers what the ticket wanted, the nineteenth handler arriving loudly, without requiring the migration to be finished first. It fails when a new handler appears, when a migrated one regresses to all-bare, or when someone migrates without recording the tier decisions.

Worth saying: **my first version of it was decorative and my own mutation caught it.** It pinned only the verdict-*declaring* set — so a new handler with zero verdicts was simply absent from that set and passed. It now pins the full census as well. Both directions verified: a new unverdicted handler fails on the census; `list_artifacts` regressing to bare returns fails on the subset.

**F2 — confirmed, and slightly worse than described.** `missing_file` is not just a class name in the same module; it has special handling in four places — `_WALL_TYPES` maps `FileNotFoundError`/`NotADirectoryError`/`IsADirectoryError` to it, `_STATUS_WALLS` maps 404 to it, it is the sole member of `_EXACT_ONLY_CLASSES`, and it gets path-identifier extraction. Renamed the sentinel to `path_not_found`; tier and class unchanged, with a comment naming all four so nobody reintroduces the collision.

**F3 — confirmed, your mutation reproduced.** `recall_skill` has three NO MATCH exits and the old test only ever reached the first. `test_every_no_match_branch_is_unverdicted_not_just_the_empty_store` now drives all three — empty store, populated-with-no-near-miss, and the near-miss-found-but-load-fails race — with fixture-drift assertions, so a store that stops producing a near-miss fails loudly instead of silently exercising the wrong branch. A verdict at branch 1, 2 or 3 now fails; branches 2 and 3 are caught by nothing else.

**Housekeeping.** You were right about the attachment, and the state history confirms it: that ticket was Done from 09-02 20:52 and flipped to In Progress at 09:56:37 — two seconds after this PR was created. Attachment removed, ticket back to Done, and the body no longer contains the magic word next to the id (my own correction paragraph still had it in a quote, which would have re-triggered it).

Also rebased onto current `staging` (`1933d819`, was behind 8) and re-ran everything: **3,036 passed, 31 skipped.**

Thanks — F3 in particular was a gap I would not have found, and F1 turned into a better test than either of us specified.

alecantu7:
## Self-review (convention 6)

At `22344477`. Three commits. One thing here needs a reviewer's conscious yes or no rather than a skim, so it goes first.

### The seam guard is satisfied in substance, not in letter — please accept or reject that deliberately

ENG-2248's revised `Done when` asks for:

> *"An AST seam guard [that] fails when a handler under `anton/` returns a bare value where a `ToolOutcome` is expected — the durable half; without it the nineteenth arrives silently."*

**Taken literally that guard cannot exist while this migration is deliberately partial.** Measured: `tool_handlers.py` alone has **49 bare returns across 9 handlers**, and three of those handlers are *migrated* ones whose bare returns are the deliberate tier-3 exits — `recall_skill`'s NO MATCH family, `open_artifact`'s "no artifact found", `memorize`'s "encoding is disabled". A guard that failed on a bare return would fail on day one, on purpose-built code.

So it is **inverted into an inventory lock** over `anton/core/tools/` (`test_the_set_of_verdict_declaring_handlers_is_pinned`). It pins two things:

1. **the census** — every module-level `handle_*` function found by an AST scan, and
2. **the verdict-declaring subset** — those with at least one `ToolOutcome` return.

That gets the property the ticket actually wanted — the nineteenth handler arrives **loudly** — without requiring the migration to be finished first. It fails when a new handler arrives with no verdict, when a migrated handler regresses to all-bare returns, or when a handler is migrated without updating the list (a one-line change, and the moment to re-check the tier table on the ticket).

Scoped to module-level functions so `HTMLParser.handle_data` in the web-fetch parser is not mistaken for a tool handler, with an `assert seen` so the scan cannot pass vacuously.

**If you would rather have the literal guard, it means finishing the migration first** — including `select_path`/`ask_user`, which need a vocabulary decision for the shared `_status()` choke point. That is a bigger change than this one, which is why it went this way.

### The first version of that guard was decorative, and a mutation caught it

Worth recording because it is the kind of thing that reads as coverage and is not. The first version pinned only the **declaring** set. A new handler with zero verdicts is simply *absent* from that set, so it sailed straight through — the guard could not detect the exact thing it existed for. Pinning the census as well is what closes it. Both directions are mutation-verified: a new unverdicted handler fails on the census, and `list_artifacts` regressing to bare returns fails on the subset.

### Review findings, all three fixed

**A new handler could still arrive silently** — the guard above.

**`missing_file` was two different things.** The new sentinel key `missing_file` shadowed an existing **class** of the same name in the same module, one that carries special handling (`_EXACT_ONLY_CLASSES` membership, path-identifier extraction) and is reached by a different path (`_WALL_TYPES["FileNotFoundError"]`). So the same user-facing failure booked two different classes depending on which path produced it — and since ENG-2247 landed, both reach `tool_completed`'s `root_cause_class` per call, making it two analytics rows for one failure mode. Renamed to **`path_not_found`**; the tier and class are unchanged, and the class `missing_file` is untouched.

**Only one NO MATCH branch was pinned.** `recall_skill` has three NO MATCH exits and the original test only ever reached the first, an empty store. A reviewer's mutation putting a verdict on the third — the fuzzy-match load race — **survived the entire suite**. `test_every_no_match_branch_is_unverdicted_not_just_the_empty_store` now drives all three, with fixture-drift assertions so a store that stops producing a near-miss fails loudly rather than silently exercising the wrong branch. Verified: a verdict at branch 1, 2 or 3 now fails, and 2 and 3 were caught by nothing else.

### What this PR does, and the behaviour change it carries

Six handlers now return `ToolOutcome` with an explicit `ok` (and a `reason` where one exists), so the dispatch loop's `tool_ok` is a real verdict rather than `None`.

**This is not purely a measurement change, and that is the part to read carefully.** `_apply_error_tracking` opens with `if ok is not None:` and uses the handler's verdict **directly** to drive `error_streak[tool_name]` — which fires the resilience nudge at 2 consecutive failures and the circuit breaker at 5. So `ok=False` is not a label; it arms a breaker. `recall_skill` alone is **1,150 calls across 452 installs in 30 days**, so a mis-tiered return reaches everyone.

Hence the tiering, applied per `return` rather than per handler:

| tier | what | behaviour change | in this pass |
| -- | -- | -- | -- |
| 1 | verified successes → `ok=True` | **none** — byte-identical to today | yes |
| 2 | indisputable failures → `ok=False` (missing argument, store unavailable, unreadable file) | small, correct direction | yes, **accepted deliberately** |
| 3 | ambiguous → left `ok=None` **with a comment saying why** | none | no |

Tier 1 is where most of the value is and it is free: the reason a failure rate is uncomputable today is not the failures, it is that the *denominator* is unknown.

**One tier-2 case IS trip-eligible, stated on purpose.** `store_unavailable` maps to `TIER_WALL`, the only trip-eligible tier, so five in a row can trip the breaker — the correct outcome for a capability that is absent. It was already `TIER_WALL` and already emitted at three sites on `staging`, so reachability is not new. Asserted in `test_create_skill_draft_with_no_store_is_an_accepted_wall`.

### Verification

- **Full suite: 3036 passed / 31 skipped.** Migration test file: 17 tests. Rebased onto `staging` (0 behind).
- Every tier-3 return carries a comment naming why it was left, so the next reader sees a decision rather than an omission.
- Tier-1 free-ness is pinned by running the **same success body** twice — once as `ok=None` (what shipped before) and once as `ok=True` — and requiring an identical streak outcome from a primed non-zero streak, with a precondition asserting the body contains none of the legacy markers. So if a future success body ever contains "failed", that test is the thing that says so.
- The 2-failure nudge and 5-failure breaker are **exercised and explicitly accepted**, not discovered later.
- The five new sentinels reuse **existing** class values (`invalid_argument`, `unknown_resource`, `unclassified`), so ENG-2247's closed vocabulary is unchanged, and the new tiers are 3× `TIER_SELF` and 2× `TIER_UNCLASSIFIED` — **zero `TIER_WALL`**, so none of them can arm ENG-1531's breaker.
- A separate AST scan keeps `ToolOutcome(content=[...], ok=False)` shut tree-wide: both tool loops' list arms skip the nudge/breaker text and `_record_root_cause`, so that shape would climb the streak, trip nothing the model can see, and put a cause on no turn tally. `read_image` returns a list only on success.

### This pass is partial, and must not be closed as finished

Per the ticket's own revised `Done when`:

- `select_path` (81 calls) and `ask_user` (38) are **deferred** — both route through the shared `_status()` choke point at `tool_handlers.py:968`, whose vocabulary contains `cancelled`, `timeout` and `needs_confirmation`. None of those is a tool failure, so mapping the vocabulary wholesale to `ok=False` would inflate the rate with correct behaviour. That is its own decision.
- **Six of the eighteen tools are not anton's** — `set_status`, `lookup_connector`, `request_credentials`, `report_success`, `report_failure`, `set_field_status` all live in `cowork-server`. A PR here cannot fix them; they need their own ticket in that repo. Two more (`generate_artifact`, `generate_prd`) have no located owner at all.
- The residual `ok="unknown"` rate needs re-measuring on production after 7 days, split by owner, with the remainder named rather than claimed as eliminated.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

This migrates six anton tool handlers from bare `str`/`list` returns to explicit `ToolOutcome(ok=…, reason=…)` verdicts that drive the per-tool error streak. I reviewed all 10 changed files and traced the change into the machinery it feeds — the dispatcher, both tool loops, `_apply_error_tracking`, and `_record_root_cause`. The tiering (success→`ok=True`, indisputable failure→`ok=False`, ambiguous→left `ok=None`) is applied consistently and every behaviour claim in the description held up against the code. No actionable findings.

### Validation

- **No broken callers.** `ToolRegistry.dispatch_tool` normalizes every handler return into a `ToolOutcome` (`registry.py:142`), so the return-type change is contained. The only production consumer of `recall_skill` internals (`session.py` `_preload_thalamus_skills`) uses `_format_skill_response`/`store.load` directly, not the handler, so it never sees a `ToolOutcome`.
- **Non-updated tests still hold.** `test_session_skills_init.py` reads `dispatch_tool(...).content`; `test_thalamus.py` asserts on dispatched history; `test_artifact_turn_tracking.py` reads back off the store, not the return; `test_cloud_turn_entrypoint.py` only references `handle_memorize` in a comment.
- **read_image list invariant upheld.** Both tool loops' list arms (`session.py:4056`, `session.py:5438`) reset the streak on a list unless `ok is False`, and skip the nudge/breaker text + `_record_root_cause`. read_image returns a list only on `ok=True`; every failure returns a string. The AST guard `test_read_image_never_pairs_a_list_with_a_failure_verdict` pins that tree-wide.
- **Every `reason` maps to a sentinel.** `missing_name`, `missing_slug`, `invalid_type`, `store_unavailable` already exist in `_SENTINEL_REASONS`; the five new ones (`not_an_image`, `image_too_large`, `path_not_found`, `read_failed`, `bmp_convert_failed`) are added with deliberate tiers, and `classify_root_cause` resolves each (`root_cause.py:307`).
- **Tier-3 is behaviour-preserving.** `open_artifact` "no artifact found", `recall_skill`'s NO MATCH family, and `memorize`'s "disabled" still return bare strings (`ok=None`), so the legacy fallback classifies them exactly as before.
- **Tier-2 change is real and scoped as described** — e.g. `open_artifact`'s missing-slug and `recall_skill`'s empty-label now climb the streak where the marker-free string previously reset it.

One non-blocking observation, not a finding: `memorize` reuses `reason="missing_name"` for "No entries provided" and `invalid_type` for "No valid entries" — a slightly loose fit for those generic sentinel keys, but the tier (`TIER_SELF`) is correct and `reason` is a coarse machine bucket, so classification is unaffected.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**

Reviewed the full diff (6 migrated handlers across `recall_skill.py`, `skill_draft.py`, `tool_handlers.py`; new sentinels in `root_cause.py`; the seam guards) against the mechanism it drives — `_apply_error_tracking`, `_record_root_cause`, the `ToolOutcome` dispatch contract, and the documented list-arm gaps in `session.py`. The migration flips verdicts only; the bytes the model reads are unchanged, and the one genuine behaviour change (tier-2 failures entering the streak) is scoped correctly. No actionable findings.

### Validation
- **The change is verdict-only, not content.** `dispatch_tool` already normalized every bare handler return into `ToolOutcome(content=…, ok=None)`, so downstream string/list handling is unchanged — the diff only moves `ok` from `None` to `True`/`False`. Confirmed per handler that `.content` is byte-identical to the prior return.
- **`read_image`'s list is safe.** `ToolOutcome.content` is typed `str | list[dict]`, and the migrated handler is `ok=True`-only; every failure branch returns string content. That keeps the two documented list-arm gaps (`_tool_failure_cause` docstring, non-streaming NOTE) unreachable. An AST scan of the whole `anton/` tree (`test_read_image_never_pairs_a_list_with_a_failure_verdict`) pins the shape tree-wide.
- **Tier-1 is provably free.** In `_apply_error_tracking`, `ok=True` sets `is_error=False` → streak reset — identical to the legacy no-match path a bare success string already took. `_record_root_cause` early-returns on `tool_ok is not False`.
- **Tier-2 is the accepted, correct change.** `ok=False` increments the per-tool streak (nudge@2/breaker@5). Each `ok=False` checked is an unretryable failure or the model's own bad argument; none fires on normal behaviour. Tier-3 (`recall_skill` NO MATCH family, `open_artifact` not-found, `memorize` disabled) correctly stays `ok=None`, each with a comment.
- **Sentinels resolve and degrade safely.** All 10 `reason` values the handlers emit exist in `_SENTINEL_REASONS` with coherent tiers (`store_unavailable`→TIER_WALL trip-eligible; `read_failed`/`bmp_convert_failed`→TIER_UNCLASSIFIED given their bare `except`; `path_not_found`/`not_an_image`/`image_too_large`→TIER_SELF). `classify()` uses `.get()` with a fallback, so an unknown reason can never `KeyError`.
- **Seam guard is durable and non-vacuous.** `test_the_set_of_verdict_declaring_handlers_is_pinned` inventory-locks both the full handler census and the verdict-declaring subset, so a nineteenth handler or a regression fails loudly — the ticket's durable-fix criterion, correctly inverted for a partial migration.
- **Tests run green (worktree at head `22344477`):** verdict-migration file 17/17 incl. both AST guards; adjacent modified files 51/51; `test_cloud_turn_session.py` 59/61 — the 2 failures are a missing `truststore` dep in my env, unrelated to this diff.

One thing to be aware of, not a defect: `read_image`'s tier-2 failures (file-not-found, too-large) count toward the per-tool streak, so 5 consecutive read_image failures trip the breaker. That's the deliberate tier-2 tradeoff — the ticket distinguishes it from NO MATCH precisely because there is no listing to self-correct from, so repetition is genuine thrash. Correct call, worth knowing.

PR #439, opened 2026-09-03
ENG-1424: resolve one publish identity per anton CLI session

alecantu7:
## What

In a single anton CLI session, `publish_or_preview` resolved a **different** `ANTON_MINDS_API_KEY` than the chat, the LLM client and `/unpublish` did. Dashboards were published to one MindsHub account while the rest of the session operated as another — so a published report was invisible to `/unpublish` and could not be taken down from the session that created it.

CLI-only. The desktop harness rebinds `PUBLISH_TOOL` to its own server-side handler and the cloud pod registers no host publish tools, so neither can reach this code path.

## Why

A user published a customer dashboard containing third-party referral data and then asked the agent to take it down. The agent could not find it, told them "there is no unpublish capability", and tried to neutralise the link by publishing a placeholder over it. `/unpublish` — which works correctly — listed a completely different set of reports. Their conclusion was "the unpublish command is broken." It is not: the session was authenticated as two accounts at once and nothing in the product ever said so.

## How

Four config files load in order — `<cwd>/.env`, `<project>/.anton/.env`, `~/.anton/.env`, `~/.cowork/.env` — later file wins. Above all of them sits `os.environ`, which pydantic-settings ranks higher than any `env_file`.

`cli.py` builds `AntonSettings()` while `os.environ` is still clean, so the file chain applies. `_ensure_workspace` then copies the **project** vault into `os.environ`, where it outranks everything. `handle_publish_or_preview` built a **second** `AntonSettings()` and therefore saw the promoted project key, while everything else kept using the object built at startup.

Two changes:

**A — the publish tool stops re-resolving.** It reads the session's settings (`ChatSessionConfig` already carries them; this is a read change, not new plumbing), falling back to a fresh resolve only for a host that passed none. Same `getattr` + capability-check shape `core/session.py` already uses, because `ChatSessionConfig.settings` is typed `CoreSettings | None` and `CoreSettings` has none of the publish fields.

**B — `/publish` stops manufacturing the divergent state.** Eight sites write `ANTON_MINDS_API_KEY`; six target the global vault (`~/.anton/.env`) and the two exceptions were both in `_handle_publish` — writing the one file that gets promoted first. It now persists to the global vault after success, clears the same file on 401, and no longer persists a key before it has authenticated at all (restoring STRC-987's intent, which an eager pre-validation write had defeated).

`_reconcile_publish_identity` runs once at boot, before any promotion, and collapses a legacy project-vault key into the global one — moving it up when the global vault is empty, dropping it when it is not. This is what stops existing per-project copies from staying promoted forever, and it is also what keeps `anton --folder <dir>` working (`_build_env_files` is evaluated at import against `Path.cwd()`, so a `-f` target's vault never enters the chain). It prints what it did: this can change which account publishes, and a silent identity change is the failure being fixed.

**Product decision baked in:** the CLI has one publish identity. Per-project publishing accounts were never a supported feature — no surface exposed or selected them, nothing refreshed them on rotation, and six of eight writers were already global. If they are wanted later they should be built explicitly (named profiles, visible active account, rotation handling), not derived from `.anton/.env` precedence.

### Not in scope

`_has_api_key` is still a **presence** check — a non-empty but invalid key still boots past onboarding. So the ticket's `Done when` on validating keys is only partly met: the write path no longer persists a key before it has authenticated, but the boot path still trusts any non-empty string. Validating at boot means an auth round-trip on every start, which is its own change.

Reconciliation is built on the same limitation and handles it by never destroying anything: when the project and global vaults hold different keys it keeps the global one and archives the project's to `.anton/.env.superseded` (0600), because nothing in that path can tell which of the two authenticates.

`~/.cowork/.env` still outranks `~/.anton/.env`, so a desktop-configured machine can still shadow the global vault. That is ENG-1295's load-chain change and anton#313 owns it.

## Testing

Full suite **3017 passed / 31 skipped** (base `staging`: 2992/31). Every new guard was mutation-verified — reverting each half of the fix independently fails a test.

Also fixes the suite writing credentials to the developer's real home: the publish tests mocked the project workspace but not the home directory, so a full run left `ANTON_MINDS_API_KEY=goodkey` and `ANTON_FIRST_RUN_DONE=true` in the real `~/.anton/.env` — and passed while doing it. `_no_real_home` in conftest redirects both `Path.home()` and `$HOME` (the latter also covers child processes, which is what the e2e scenarios leaked through). It is session-scoped with the package caches pointed back at the real home: a fresh home per test cost 193s → 425s, all of it e2e subprocesses first-running into a virgin directory.

## Security

A security pass ran over this diff and covered: credential persistence location and lifetime, secrets reaching logs/console/tracebacks, config injection through unvalidated input, and file permissions.

- `Workspace.set_secret` **rejects a value containing a newline**. A pasted "API key" could otherwise append arbitrary further `ANTON_*` settings to the file. Guarded in the shared writer because the unvalidated callers are spread across `cli.py`, `/remote`, `/connect` and `commands/setup.py` — this closes those routes too, not just `/publish`.
- File permissions were reviewed and need nothing here: #430 (ENG-1382) landed `_write_private` on staging while this branch was open, creating the vault 0600 atomically via `os.open(O_CREAT, 0o600)`. An earlier revision of this branch carried its own weaker helper (write-then-chmod, leaving a window at umask mode); it has been dropped in favour of staging's, and the duplicate mode tests dropped in favour of `TestSecretVaultFilePermissions`.
- An unvalidated key no longer reaches disk at all, and a key owned by another config file (`~/.cowork/.env`, written by the desktop app) is no longer copied into `~/.anton/.env` — that copy survived desktop sign-out, which scrubs only `~/.cowork/.env`.
- No key is added to logs, console output or analytics. No new endpoint or auth surface.

Closes ENG-1424.


alecantu7:
## Self-review (convention 6)

At `14545c33`. Three commits, and the middle of them was wrong in a way worth reading before the diff — the migration destroyed credentials before it archived them.

### The reconciliation is deliberately non-destructive, and that is the whole design

`_reconcile_publish_identity` collapses a legacy project-vault publish key into the global vault. The obvious implementation — keep the global key, drop the project copy — is what the first version did, and it is wrong for a reason this ticket documents in its own words.

**Nothing in that path can tell which of the two keys authenticates.** Reading a vault is a presence check, which is exactly the `_has_api_key` weakness ENG-1424 describes: *"it tests whether a key string exists, never whether it authenticates."* Assuming the global vault is current gets it backwards on the machine state the ticket actually measured:

| location | state on the reporter's machine, 2026-08-11 |
| -- | -- |
| `<workspace>/.anton/.env` | valid, account `f1ec7bf69`, 48 published reports |
| `~/.anton/.env` | **401 — invalid (7-char, all letters)** |

Executed against that state, the first version removed the working key, kept the one that 401s, and printed *"Removed a stale publish key from this project; using the one in ~/.anton/.env"* — an assertion it had no way to make, and backwards. The account owning those 48 reports became unreachable from the CLI: a worse version of the bug being fixed. No backup, not recoverable from disk.

So: three cases, none lossy.

* **global vault empty** → promote the project key. Also what makes `anton --folder <dir>` work: `_build_env_files` is evaluated at import against `Path.cwd()`, so a `-f` target's vault is not in the settings chain at all and its key would otherwise be invisible.
* **the two are the same string** → the copy carries no information; drop it.
* **they differ** → keep the global one, because one identity is the point, but archive the project's to `.anton/.env.superseded` first, so a user whose working key was the project one gets it back with `cat`.

The console wording changed with it. It no longer claims the retained key is the good one — it says the project's key was saved *"in case it was the one you wanted"*, because nothing here has authenticated either.

### Why the archive is safe to leave in the vault directory

Worth stating, because putting a live credential back into `.anton/` is exactly the kind of thing that undoes a fix. `load_env()` reads `self._env_file` — `.anton/.env` — exactly, with no glob, so `.env.superseded` is never loaded and can never be promoted into `os.environ`. Its header line is a `#` comment as well, so even a future glob would skip it. It is written through the vault's own `_write_private` (atomic `os.open(O_CREAT, 0o600)`) rather than a `write_text` + `chmod`, which would leave the key at umask mode until the chmod landed. Verified `0600` on disk.

### The rest of the change

**A — the publish tool stops re-resolving.** `handle_publish_or_preview` built a *second* `AntonSettings()`, which re-reads config after `_ensure_workspace` has copied the project vault into `os.environ` — and pydantic-settings ranks `os.environ` above every `env_file`. So publishing saw the promoted project key while the LLM client, `/publish` and `/unpublish` all used the object built at startup. It now reads the session's settings, with a `getattr` + capability-check fallback matching the shape `core/session.py` already uses (`ChatSessionConfig.settings` is typed `CoreSettings | None`, and `CoreSettings` carries none of the publish fields).

**B — `/publish` stops manufacturing the divergent state.** Every `ANTON_MINDS_API_KEY` writer now targets the global vault; the two exceptions were both in `_handle_publish`, writing the one file that gets promoted first. It persists after success, clears on 401, and no longer persists a key before it has authenticated (restoring STRC-987's intent, which an eager pre-validation write had defeated).

**Product decision baked in:** the CLI has one publish identity. Per-project publishing accounts were never a supported feature — no surface exposed or selected them, nothing refreshed them on rotation, and six of eight writers were already global. If they are wanted later they should be built explicitly, not derived from `.anton/.env` precedence.

### Verification

- **The destructive case is now a named regression guard**: `test_the_ticket_s_measured_machine_state_loses_nothing`, alongside `test_a_divergent_project_key_is_archived_not_destroyed` and `test_a_duplicate_project_copy_is_dropped_outright`. The earlier test that encoded the false premise (`..._dropped_when_the_global_vault_wins`) is gone.
- Reverting the archive — straight back to destructive — fails 2 tests. Every new guard was mutation-verified; reverting each half of the fix independently fails a test.
- **CLI-only, verified in both directions.** The desktop harness rebinds `PUBLISH_TOOL` to its own server-side handler (`cowork-server`'s `anton_harness/harness.py`, whose `tools.py` notes it is the only tool overridden), and `anton/cloud_turn/` registers no host publish tool at all, so neither surface can reach this code path.
- Full suite **3017 passed / 31 skipped**.
- The suite also no longer writes credentials to the developer's real home: the publish tests mocked the project workspace but not `Path.home()`, so a full run left `ANTON_MINDS_API_KEY=goodkey` in the real `~/.anton/.env` — and passed while doing it.

### Known scope line, not an oversight

`_has_api_key` remains a **presence** check, so a non-empty but invalid key still boots past onboarding. The ticket's `Done when` on validating keys is therefore only partly met: the **write** path no longer persists a key before it has authenticated, the **boot** path still trusts any non-empty string. Validating at boot means an auth round-trip on every start, which is its own change. Reconciliation is built on that same limitation and handles it by never destroying anything rather than by pretending to know which key is live.

Also out of scope: `~/.cowork/.env` still outranks `~/.anton/.env`, so a desktop-configured machine can still shadow the global vault. That is ENG-1295's load-chain change and anton#313 owns it.

tino097 (commented):
Review of the publish-identity reconciliation. 15 findings, ordered most severe first: five look blocking (a key that vanishes for the session, a 401 handler that wipes an unverified key, `os.environ` mutation dropping an exported key, a path guard that can delete the only copy, and unguarded writes that traceback at boot), then five worth fixing, then minor items. Anchors verified against head `14545c33`.

anton/cli.py
```diff
@@ -454,6 +538,12 @@ def main(
     settings = AntonSettings()
     settings.resolve_workspace(folder)
 
+    # Must run before anything reads a key off `settings`, and before
+    # `_ensure_workspace` promotes the project vault into os.environ.
+    if _reconcile_publish_identity(settings):
```
tino097: **The re-resolve cannot see the key reconciliation just moved, so the key is lost for the whole session.**

`_ENV_FILES` is frozen at import (`anton/config/settings.py:19-22`) and includes `~/.anton/.env` only if that file already existed at that moment.

Machine with `<project>/.anton/.env` holding `ANTON_MINDS_API_KEY` and no `~/.anton/.env` (desktop user configured via `~/.cowork/.env`, or a `-f` target). Boot takes the `not global_key` branch: `global_ws.set_secret` creates `~/.anton/.env` and sets `os.environ`, then `project_ws.remove_secret` (`workspace.py:288`) pops that same `os.environ` entry. The re-resolve on line 544 reads only `[.env, <cwd>/.anton/.env]` — the project copy is gone and `~/.anton/.env` was never in the chain. `_ensure_workspace` only calls `global_ws.apply_env_to_process()` when `~/.anton/anton.md` exists, which it does not.

Result: `settings.minds_api_key` is `None`, `/publish` re-prompts for a key, immediately after the console printed *"Moved this project's saved publish key to ~/.anton/.env"*. Self-heals on the next boot only.

anton/chat.py
```diff
@@ -729,17 +778,25 @@ def _make_candidate(path: Path) -> tuple[str, Path, str, str] | None:
             import urllib.error
             if isinstance(e, urllib.error.HTTPError) and e.code == 401:
                 settings.minds_api_key = None
-                if workspace:
-                    workspace.set_secret("ANTON_MINDS_API_KEY", "")
+                # Clear it where it is WRITTEN. Clearing the project vault while
+                # the key lived in the global one is precisely how a rejected
+                # key survived to lock the next session out (ENG-1424).
+                _persist_key_best_effort(console, "", what="clear the rejected key")
```
tino097: **The 401 handler blanks the global key without checking the rejected key came from there** — it destroys an unrelated working key and leaves the one that actually 401'd in place.

`~/.anton/.env` holds a valid `GOOD_A`; `~/.cowork/.env` holds an expired `EXPIRED_B`. The dotenv chain ranks `~/.cowork/.env` last, so the session resolves `EXPIRED_B`. `/publish` 401s, this line writes `ANTON_MINDS_API_KEY=` into `~/.anton/.env` and wipes `GOOD_A`. `EXPIRED_B` is untouched, so every later session resolves it again and 401s again — the exact lock-out the comment on 781-783 claims to fix, now permanent and with the good key destroyed.

Same shape when the resolved key came from `<cwd>/.env` or an exported env var. Clearing needs to be conditional on the rejected value matching what the global vault holds.

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
...
+    project_key = project_ws.get_secret("ANTON_MINDS_API_KEY")
+    if not project_key:
+        return False
+
+    global_ws = Workspace(Path.home(), settings=settings)
+    global_key = global_ws.get_secret("ANTON_MINDS_API_KEY")
+
+    if not global_key:
+        global_ws.set_secret("ANTON_MINDS_API_KEY", project_key)
+        project_ws.remove_secret("ANTON_MINDS_API_KEY")
```
tino097: **`remove_secret` mutates `os.environ`, so reconciliation silently deletes an operator-exported key from the process.**

`ANTON_MINDS_API_KEY=X anton` in a project whose `.anton/.env` holds `Y`, with `Z` in `~/.anton/.env`. `AntonSettings()` correctly resolves `X` (pydantic ranks `os.environ` above every `env_file`). Reconcile takes a divergent branch and calls `project_ws.remove_secret(...)`, which does `os.environ.pop(key, None)` (`workspace.py:288`) and drops `X`. The re-resolve at line 544 now yields `Z` — for the session, for every scratchpad subprocess, and for the `ANTON_MINDS_API_KEY`→`OPENAI_API_KEY` fallback at `core/backends/local.py:633`.

This also makes the docstring's "before any promotion" claim false: `set_secret`/`remove_secret` are themselves promotions.

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
...
+    """
+    from datetime import datetime
+
+    # Reuse the vault's own atomic 0600 writer rather than re-implementing it:
+    # the archive holds the same secret as the file it came from, and a plain
+    # write_text would leave it at umask mode until a chmod landed.
+    from anton.workspace import Workspace, _write_private
+
+    project_path = Path(settings.workspace_path)
+    if project_path == Path.home():
```
tino097: **The home-vs-project guard is a lexical `Path` comparison, so on a non-canonical `$HOME` it misses and the "duplicate" branch deletes the only copy of the key.**

`settings.workspace_path` comes from `Path.cwd()` / `Path(folder).resolve()`, both canonicalised; `Path.home()` is `$HOME` verbatim.

`$HOME=/home/user` where `/home` is a symlink to `/export/home`. Running `anton` from the home directory: `workspace_path` is `/export/home/user`, `Path.home()` is `/home/user` → not equal, guard falls through. `project_ws` and `global_ws` then address the **same file**, `project_key == global_key` trivially, and the `elif` on line 468 runs `remove_secret` — deleting the user's only publish key with no archive written. Same failure for `anton -f ~`.

`tests/test_publish_identity_reconcile.py:212` sets `$HOME` to a tmpdir, so it never exercises a non-canonical home. Compare resolved paths on both sides.

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
...
+        return False  # same file — nothing to reconcile
+
+    project_ws = Workspace(project_path, settings=settings)
+    project_key = project_ws.get_secret("ANTON_MINDS_API_KEY")
+    if not project_key:
+        return False
+
+    global_ws = Workspace(Path.home(), settings=settings)
+    global_key = global_ws.get_secret("ANTON_MINDS_API_KEY")
+
```
tino097, line 507: **Three unguarded filesystem writes now run inside the Typer callback, so an `OSError` aborts `anton` at boot with a traceback.**

Container run as a uid with no writable home, or a read-only checkout: `global_ws.set_secret` (465), `_write_private(archive, ...)` (474) or `project_ws.remove_secret` (466/469/479) raises `OSError`/`PermissionError`. Nothing in `main` catches it, and this is `@app.callback(invoke_without_command=True)`, so **every** `anton` invocation including subcommands dies before terms consent.

The PR's own `_persist_key_best_effort` (`chat.py:459-476`) exists for precisely this reason — its docstring argues `~/.anton` "is reachable" and unwritable — yet these boot-time writes get no equivalent treatment.

anton/chat.py
```diff
@@ -437,6 +437,45 @@ async def _handle_remote(
...
+    went to ``<cwd>/.anton``, which the CLI had just created itself. Against
+    ``~/.anton`` it is reachable, and an escaping OSError would take out the
+    whole turn: nothing between here and ``_chat_loop``'s ``except
+    KeyboardInterrupt`` catches it, so the user would lose the success message
+    AND the view URL, and ``.published.json`` further down would never record
+    the report id — making the next /publish create a duplicate report instead
+    of updating this one.
+    """
+    try:
+        _global_vault().set_secret("ANTON_MINDS_API_KEY", value)
```
tino097: **Clearing writes an empty `ANTON_MINDS_API_KEY=` line instead of removing the entry**, so the empty value outranks lower-precedence config and is exported to every child process.

After a 401, `~/.anton/.env` gains `ANTON_MINDS_API_KEY=` and `os.environ["ANTON_MINDS_API_KEY"] = ""` (`workspace.py:265`). The chain is `[<cwd>/.env, <project>/.anton/.env, ~/.anton/.env, ~/.cowork/.env]` with later winning, so a valid key in `<cwd>/.env` or the project vault is now shadowed by the empty global entry.

In the scratchpad launcher, `core/backends/local.py:633` tests `"ANTON_MINDS_API_KEY" in env` — membership, not truthiness — so with `coding_provider == "openai-compatible"` the child gets `OPENAI_API_KEY=""` and fails with an auth error instead of falling through.

`remove_secret` is the right operation for the clear path.

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
...
+    if not global_key:
+        global_ws.set_secret("ANTON_MINDS_API_KEY", project_key)
+        project_ws.remove_secret("ANTON_MINDS_API_KEY")
+        message = "  Moved this project's saved publish key to ~/.anton/.env"
+    elif global_key == project_key:
+        project_ws.remove_secret("ANTON_MINDS_API_KEY")
+        message = "  Removed a duplicate publish key from this project (same key as ~/.anton/.env)"
+    else:
+        archive = project_ws.env_path.with_name(".env.superseded")
```
tino097: **The archive writes a live API key to a filename the usual gitignore patterns do not match.**

`.anton/.env.superseded` is matched by neither `.env` (exact name) nor `*.env`. A user whose repo ignores `.anton/.env`, or has a global `.env` rule, gets a live credential staged by `git add -A`.

anton writes no `.gitignore` of its own — `Workspace._initialize_once` creates only `.anton/`, `memory/`, `anton.md`, `.env`, `artifacts/` — so nothing suppresses it. Either name the archive so existing patterns catch it, or write a `.gitignore` alongside it.

tests/test_publish_identity_reconcile.py
```diff
@@ -0,0 +1,215 @@
...
+    called = [
+        n.func.id for n in ast.walk(main_fn)
+        if isinstance(n, ast.Call) and isinstance(n.func, ast.Name)
+    ]
+    assert "_reconcile_publish_identity" in called, (
+        "cli.main no longer reconciles the publish identity at boot — legacy "
+        "project-vault keys will be promoted into os.environ again"
+    )
+    assert "_ensure_workspace" in called
+    assert called.index("_reconcile_publish_identity") < called.index("_ensure_workspace"), (
```
tino097: **This ordering assertion does not test ordering.** `ast.walk` is breadth-first, so `called` is depth order, not source order.

Verified by running the test's own collection logic against `anton/cli.py` at this head: `called` is `['AntonSettings', '_reconcile_publish_identity', 'send_event', 'AntonSettings', '_ensure_terms_consent', 'check_and_update', ...]` — `send_event` sorts before `_ensure_terms_consent` even though the source has it after.

Counterexample the assertion accepts: leave `_ensure_workspace(settings)` inside the `if ctx.invoked_subcommand is None:` block (depth 3, as today) and move the reconcile call after that block at statement level (depth 2). `ast.walk` yields reconcile first, `index(...) <` holds, the test is green — and the stale project key has already been promoted into `os.environ`, which is the whole thing this guard exists to catch.

Compare `lineno`/`col_offset` instead of list position.

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
...
+            "  This project had a different publish key than ~/.anton/.env.\n"
+            "  Publishing now uses ~/.anton/.env; the project's key was saved to\n"
+            f"  {archive} in case it was the one you wanted."
+        )
+
+    # Say so. This can change which account publishes (that is the point of
+    # collapsing to one identity), and a silent identity change is the exact
+    # failure ENG-1424 is about. The wording must not claim the retained key is
+    # the good one — nothing here has authenticated either of them.
+    console.print(f"[anton.muted]{message}[/]")
```
tino097, line 577: **Rich markup without `escape()` — a bracketed path crashes boot.**

`message` interpolates `{archive}` (line 483). In a project folder named e.g. `report[final]`, this prints `/path/report[final]/.anton/.env.superseded` as markup and Rich raises `MarkupError` parsing `[final]` as a style tag, inside a function with no error handling.

Same shape at `anton/chat.py:476`, which prints `{exc}` — an `OSError` whose text carries the offending path.

This file already imports `from rich.markup import escape` and applies it at ~967/~985, with a comment recording this identical crash from the #317 review.

tests/test_publish_identity_reconcile.py
```diff
@@ -0,0 +1,215 @@
...
+    ``_build_env_files`` is evaluated at import time against ``Path.cwd()``, so
+    with ``anton -f <dir>`` the target's ``.anton/.env`` never enters the chain
+    and its key is invisible to the session — while ``_ensure_workspace`` still
+    promotes it into os.environ. Before reconciliation that mismatch made the
+    publish tool refuse ("STOP: No Minds API key configured") for a user who
+    could publish before.
+    """
+    project = tmp_path / "proj"
+    project.mkdir()
+    Workspace(project).set_secret("ANTON_MINDS_API_KEY", "PROJECT_ONLY_KEY")
```
tino097, line 79: **`set_secret` writes `os.environ`, and `monkeypatch` cannot undo a set it never recorded** — the value leaks for the rest of the pytest session.

`Workspace(...).set_secret(...)` does `os.environ[key] = value` (`workspace.py:265`). The module's `_clean_env` fixture uses `monkeypatch.delenv(..., raising=False)`, which records nothing when the var is absent, so it has nothing to restore.

After `test_a_home_workspace_is_never_reconciled_against_itself`, `os.environ` carries `ANTON_MINDS_API_KEY=HOME_KEY` for the remainder of the run (`tests/test_publish_api_key.py:211` similarly leaves `stale-bad-key`, then `""`). Any later test building `AntonSettings()` resolves the leaked value, since pydantic ranks `os.environ` above every `env_file` — results become order-dependent.

Line 152 of this file pops the var by hand, so the symptom was noticed; the cause is that these helpers mutate the process env. A `monkeypatch.setenv` to a sentinel in `_clean_env`, or an autouse env snapshot/restore, fixes it for all of them.

tests/conftest.py
```diff
@@ -216,6 +222,103 @@ def _no_browser_windows(monkeypatch):
         monkeypatch.setattr(webbrowser, name, lambda *a, **kw: True)
 
 
+@pytest.fixture(scope="session")
+def _suite_home(tmp_path_factory):
```
tino097, line 226: **Session-scoped shared home means a publish assertion can pass on another test's residue.**

`test_successful_publish_persists_key_to_global_vault` writes `ANTON_MINDS_API_KEY=goodkey` into the shared `~/.anton/.env`; `test_publish_never_writes_the_real_home` then asserts `"ANTON_MINDS_API_KEY=goodkey" in _global_vault_text(home)` against that same file.

Delete the persist call in `_handle_publish` and the second test still passes on the first test's leftovers. Run either alone, under `-k`, or under `pytest-randomly`, and the pair stops covering what it claims.

The 193s→425s cost argument in the docstring is real, but it applies to the e2e subprocess homes; the vault-assertion tests need their own directory — which is exactly why `tests/test_publish_identity_reconcile.py` had to override `home` with a pristine one.

anton/tools.py
```diff
@@ -537,7 +537,21 @@ async def handle_publish_or_preview(session: ChatSession, tc_input: dict) -> str
...
+    # LLM client, /publish and /unpublish were built with, and the session
+    # publishes as one account while `/unpublish` lists another's reports.
+    #
+    # Same getattr+capability shape as `core/session.py` (`_emit_turn_cost`):
+    # `ChatSessionConfig.settings` is typed `CoreSettings | None`, and
+    # `CoreSettings` carries none of the publish fields, so a host that passed
+    # a bare CoreSettings (or no settings) must still fall back.
+    settings = getattr(session, "_settings", None)
+    if settings is None or not hasattr(settings, "minds_api_key"):
+        settings = AntonSettings()
```
tino097, line 559: **The fallback builds `AntonSettings()` without `resolve_workspace()`, so it resolves a different artifacts root than the session it is guarding against divergence from.**

`AntonSettings()` leaves `artifacts_dir` at the relative default `'artifacts'` (`config/settings.py:127`), so line 555 yields `Path('artifacts')` relative to cwd. The session path resolves `<workspace>/.anton/artifacts` via `resolve_workspace` (`config/settings.py:246-247`).

`resolve_publish_target` is then handed the wrong roots, so `published_dir`/`published_key` differ from what `/publish` and cowork-server compute: the same `.published.json` entry gets written under two keys and the next publish creates a duplicate report instead of updating.

Carry `resolve_workspace()` into the fallback, or refuse to publish when the host supplied no publish-capable settings.

anton/workspace.py
```diff
@@ -229,7 +229,17 @@ def set_secret(self, key: str, value: str) -> None:
...
+
+        The value must be single-line. Every caller writes user-supplied input
+        straight through — a pasted "API key" containing a newline would append
+        arbitrary further ``ANTON_*`` settings to the file, which is a config
+        injection with no legitimate use (no key, URL or model name contains a
+        newline). Rejected rather than escaped, so the caller reports it to the
+        user instead of silently persisting half a credential.
         """
+        if "\n" in value or "\r" in value:
+            raise ValueError(f"{key} must not contain a newline")
```
tino097, line 273: **This `ValueError` is caught at 1 of 8 `set_secret` call sites**, so the docstring's "the caller reports it to the user" does not hold for the other seven.

Only `anton/chat.py:474` (`_persist_key_best_effort`) catches it. A value containing `\r` or `\n` reaching `cli.py:891` (`_setup_minds`, first-run onboarding), `cli.py:1133/1184/1242/1343`, `chat.py:258` (`/connect`) or `chat.py:427` (`/remote`) escapes as an unhandled `ValueError`.

For `_setup_minds` that aborts onboarding with a stack trace after `ANTON_MINDS_URL` was already written but the key was not — a half-written vault, which is the state the guard was added to prevent.

anton/chat.py
```diff
@@ -437,6 +437,45 @@ async def _handle_remote(
...
+    except (OSError, ValueError) as exc:
+        console.print(f"  [anton.warning]Could not {what} in ~/.anton/.env: {exc}[/]")
+
+
 async def _handle_publish(
     console: Console,
     settings,
```
tino097, line 493: *(anchored here; the parameter itself is line 482, outside the diff hunk)*

**Dead parameter and a duplicated helper.**

`_handle_publish`'s `workspace` parameter is no longer read anywhere in the body — both key writes that used it were replaced by `_persist_key_best_effort`, leaving only `settings.workspace_path` at line 602. CLAUDE.md: *"No dead code left behind: commented out blocks, unused imports, parameters nothing passes."*

`_global_vault` (line 445-456) re-imports `from pathlib import Path as _P` although `from pathlib import Path` is already at line 9 and used at line 126 — and it duplicates the `Workspace(Path.home())` construction inlined at line 126 (`_handle_connect`) and line 391 (`_handle_remote`), neither of which was converted to call it.

Also duplicated: `tools.py:552-554` re-implements the `getattr(session, "_settings", None)` + fresh-resolve fallback that already exists 437 lines above at `tools.py:115-121`, and twice more in `core/session.py` (3007, 3192).

anton/cli.py
```diff
@@ -407,6 +407,90 @@ def _get_settings(ctx: typer.Context):
     return ctx.obj["settings"]
 
 
+def _reconcile_publish_identity(settings) -> bool:
+    """Collapse a legacy project-vault MindsHub key into the global vault.
```
tino097, line 435: **Ticket ids in the codebase.** CLAUDE.md, *Never in the codebase*: "No ticket numbers, issue IDs, or ticket names in code, code comments, commit messages, tests, or docs inside a repo. They belong in the branch name, the PR title, and the Linear link in the PR body, and nowhere else."

The diff adds `ENG-1424`, `STRC-987`, `ENG-1295` and `anton#313` across source comments, docstrings and test docstrings — `anton/cli.py:413` and `:488`, `anton/chat.py:450, 492, 528, 780, 793`, `anton/tools.py:540`, plus the module headers of both new test files. All three commit subjects (`f3369dc`, `681a098`, `14545c3`) also end in `(ENG-1424)`.

Separately, the same file's comment rule — *"At most 2 lines per comment … Density over length"* — is strained by the 8-to-12-line inline blocks at `chat.py:526-537` and `780-782`, and by this 35-line docstring.

anton/cli.py
alecantu7: Confirmed by execution and fixed in `e24ca41`. Reproduced exactly as described — the chain froze as `['.env', '<proj>/.anton/.env']`, `remove_secret` popped the env var, and the re-resolve came back `None` right after the console said the key had been moved.

The chain is now refreshed in place after reconciliation: `_ENV_FILES[:] = _build_env_files()`. In place rather than rebinding, because `AntonSettings.model_config` holds a reference to that list object and a rebind would not reach it.

Guarded by `test_the_session_can_use_the_key_reconciliation_just_moved`. Worth noting the first version of that test passed *without* the fix: its own `set_secret` setup left the key in `os.environ`, which outranks every env_file and satisfied the assertion on its own. It now clears the variable after seeding so it mirrors a real boot, where the vault is on disk and has not been promoted yet.

anton/chat.py
alecantu7: Confirmed by execution and fixed in `630f7ec`. With `GOOD_A` in `~/.anton/.env` and `EXPIRED_B` in `~/.cowork/.env`, the 401 wiped `GOOD_A` and left `EXPIRED_B` — permanent lock-out, exactly as you describe.

Three cases now, because the binary version got the just-typed key wrong: typed this session (nothing on disk to clear, in-memory reset re-prompts), came from the global vault (clear it), came from anywhere else (leave it alone and say where to look — `~/.cowork/.env`, a local `.env`, or an exported var).

Guarded by `test_401_does_not_clear_a_key_the_global_vault_did_not_supply`.

anton/cli.py
alecantu7: Confirmed by execution and fixed in `630f7ec`. `ANTON_MINDS_API_KEY=X anton` resolved `X` before reconciliation and `Z_GLOBAL` after — the exported key silently gone, for the session and every scratchpad child.

`os.environ` is now snapshotted at entry and restored at exit. Reconciliation is a disk migration and has no business editing the process environment.

You are also right that this made the docstring false; "before any promotion" is true again now that the function leaves `os.environ` as it found it. Guarded by `test_reconciliation_leaves_an_exported_key_alone`.

anton/cli.py
alecantu7: Confirmed by execution and fixed in `630f7ec`. Built `/home -> /export/home`, ran from the home directory: guard `False`, both vaults addressing one file, duplicate branch taken, only key deleted with no archive.

Both sides are resolved now. Guarded by `test_a_symlinked_home_is_still_recognised_as_home`, which builds the symlink rather than setting $HOME to a tmpdir — your point that the existing test could never exercise this was the reason it slipped through.

anton/cli.py
alecantu7: Fixed in `e24ca41`. You are right that this is worse than the `chat.py` case: it is `@app.callback(invoke_without_command=True)`, so an OSError killed every `anton` invocation including subcommands, before terms consent.

Wrapped in `try/except OSError` with a muted line. Failing to migrate is harmless — it leaves the machine exactly as it was — so this must never be fatal. Smoke-tested against a 0500 $HOME: raises `PermissionError`, `main` catches it, boot continues.

anton/chat.py
alecantu7: Fixed in `e24ca41`. The clear path now calls `remove_secret`, not `set_secret(..., "")`.

Your `local.py:633` point is the sharp end of it — membership, not truthiness — so the child was getting `OPENAI_API_KEY=""` and an auth error rather than falling through.

The assertion needed strengthening to catch this: `"stale-bad-key" not in text` is satisfied by an empty value too, so the mutation passed at first. It now asserts `get_secret(...) is None` and that the variable is absent from `os.environ`.

anton/cli.py
alecantu7: Fixed in `e24ca41`. `.anton/.gitignore` containing `*` is written alongside the archive when one is created — self-contained, and it covers `.env` itself, which had the same exposure.

Chose that over renaming because a name matching `*.env` would still be staged by anyone whose ignore rule is the exact string `.env`.

tests/test_publish_identity_reconcile.py
alecantu7: Confirmed and fixed in `e24ca41`. You are right that `ast.walk` is breadth-first, and the counterexample you give is exactly the arrangement the guard exists to reject — reconcile hoisted to statement level, `_ensure_workspace` left nested, walk yields reconcile first, green.

Keyed on `lineno` now.

anton/cli.py
alecantu7: Fixed in `e24ca41` — `escape()` applied to the archive path, and to the `{exc}` interpolation in `chat.py` you flagged alongside it.

Both files already import `escape` and use it elsewhere for this same crash, which makes it a straightforward miss on my part.

tests/test_publish_identity_reconcile.py
alecantu7: Confirmed and fixed in `e24ca41`. `monkeypatch.delenv(..., raising=False)` recording nothing when the variable is absent is the part I had not thought through.

Fixed centrally rather than per-file: a `_no_leaked_credentials` autouse fixture in `conftest.py` touches `ANTON_MINDS_API_KEY` / `ANTON_OPENAI_API_KEY` / `ANTON_ANTHROPIC_API_KEY` with `setenv` before deleting them, so monkeypatch has them on its undo list whatever a test body does afterwards. The hand-rolled `os.environ.pop` at line 152 is gone.

tests/conftest.py
alecantu7: Confirmed and fixed in `e24ca41`. Verified the failure mode you describe: deleting the persist call leaves `test_publish_never_writes_the_real_home` green on the other test's residue.

The vault-content tests in `test_publish_api_key.py` now take a pristine per-test home, the same override `test_publish_identity_reconcile.py` already needed. The session-shared home stays for what it was actually introduced for — the e2e subprocesses, where the 193s→425s cost lives — which is the distinction you drew.

anton/tools.py
alecantu7: Confirmed and fixed in `e24ca41`. A bare `AntonSettings()` leaves `artifacts_dir` at the relative default, so the fallback resolved against cwd while the session resolved `<workspace>/.anton/artifacts` — and the duplicate-report consequence follows from `published_key` diverging, as you say.

The fallback now calls `resolve_workspace()` anchored on `session._workspace` when the host supplied one.

anton/workspace.py
alecantu7: Fair, and the docstring is corrected in `e24ca41` rather than the behaviour.

Of the ~60 `set_secret` call sites, the large majority write literals — provider names, model ids, booleans — which cannot contain a newline, so the raise is unreachable for them. Around ten pass pasted input, and you are right that only `_persist_key_best_effort` catches it, so a newline pasted at first-run onboarding still surfaces as a traceback over a half-written vault.

Rather than claim coverage the code does not have, the docstring now says exactly that and names the gap. Handling it at each prompt is worth doing and I would rather it were its own change than a further expansion of this one — happy to file it.

anton/chat.py
alecantu7: Both fixed in `e24ca41`.

`workspace` is gone from `_handle_publish` and its call site. That also let me delete an assertion that the project vault was not written: with no workspace parameter, it is structural rather than something to check against a mock.

`_global_vault()` no longer re-imports `Path`, and `_handle_connect` and `_handle_remote` now call it instead of re-inlining `Workspace(Path.home())`.

On the third: `tools.py:115-121` is a different shape — it tolerates a `None` result and is telemetry-only, where the publish path needs settings that carry the publish fields and must resolve the workspace. Collapsing them would make one of the two wrong, so I have left them separate.

anton/cli.py
alecantu7: Checked this one carefully and I do not think it holds — flagging rather than acting on it, happy to be corrected.

I could not find the rule. There is no `CLAUDE.md` or `AGENTS.md` in this repo, `CONTRIBUTING.md` says nothing about ticket ids or comment length, and the workspace-level `MindsDB_Repos/CLAUDE.md` has no *Never in the codebase* section — I grepped all of them for `Never in the codebase`, `ticket numbers`, `2 lines per comment` and `Density over length` and got nothing. If it lives somewhere I have not looked, point me at it and I will strip them.

Established practice runs the other way. **49 files** under `anton/` on `staging` cite `ENG-xxx` in comments and docstrings, including the two files this PR touches most:

```
anton/chat.py:1064   # script content (ENG-824's root-caused crash site ...
anton/chat.py:1428   # ENG-1495: see the note on the same field in chat_session.py ...
anton/cloud_turn/session.py:581   # fails every stat with ESTALE under gVisor (ENG-1817).
anton/cloud_turn/session.py:617   # ENG-1638: the pod used to pass no prompt context ...
```

Removing them from this diff alone would make it inconsistent with the surrounding code without making the codebase consistent, and I would rather not do that unilaterally in a review round. If the convention is real, it is worth a separate sweep and a note in a CLAUDE.md so it is discoverable — I would support that.

On comment density: I have trimmed where I could this round, but where a comment records *why* a specific machine state broke something — the invalid-7-character-key case, the breadth-first `ast.walk` trap — I would rather keep the length than lose the reason, since both are things a reader would otherwise have to rediscover. Genuinely open to cutting specific ones if you name them.

anton/chat.py
```diff
@@ -728,18 +781,48 @@ def _make_candidate(path: Path) -> tuple[str, Path, str, str] | None:
...
+                # can be running a key this vault does not hold: blanking it
+                # unconditionally destroys an unrelated working key AND leaves
+                # the one that really 401'd in place, making the lock-out
+                # permanent instead of fixing it (ENG-1424).
+                if key_entered_this_session:
+                    # Typed at the prompt above and never persisted (that is the
+                    # STRC-987 fix), so there is nothing on disk to clear. The
+                    # in-memory reset alone re-prompts on the next /publish.
+                    console.print("  [anton.error]Invalid API key — run /publish again to enter a new one.[/]")
+                elif _global_vault().get_secret("ANTON_MINDS_API_KEY") == rejected:
```
tino097: **The ownership test compares two different parsers of the same file, so a quoted or comment-suffixed vault line takes the "not ours" branch and the rejected key is never cleared.**

`settings.minds_api_key` arrives via pydantic-settings / python-dotenv. `get_secret` goes through `load_env` (`workspace.py:204`), which only `.strip()`s the value. Verified against the repo venv (pydantic-settings 2.13.1):

| `~/.anton/.env` line | `settings.minds_api_key` | `get_secret(...)` |
| --- | --- | --- |
| `ANTON_MINDS_API_KEY="abc123"` | `abc123` | `"abc123"` |
| `ANTON_MINDS_API_KEY=abc123 # note` | `abc123` | `abc123 # note` |

Both take the `else` branch, print "It did not come from ~/.anton/.env", and leave the rejected key exactly where it is. That is the permanent lock-out this branch exists to remove, back again for those two vault shapes.

Blast radius is limited but real: `set_secret` writes `key=value` unquoted, so a vault anton wrote itself is fine. This bites a hand-edited vault or one written by another tool, and `~/.anton/.env` is a plaintext file users are pointed at.

Comparing parsed values on both sides fixes it, or key the decision off something other than raw file text.

Separately, the else-branch message omits `<project>/.anton/.env`, which is still a live chain member whenever reconciliation was skipped (no project key, or a `-f` target).

anton/cli.py
```diff
@@ -407,6 +407,129 @@ def _get_settings(ctx: typer.Context):
...
+    # (workspace.py), so an operator running `ANTON_MINDS_API_KEY=X anton` would
+    # have X silently popped here — and pydantic ranks os.environ above every
+    # env_file, so X is exactly the key that session was resolving. Snapshot and
+    # restore around the whole migration.
+    exported = os.environ.get("ANTON_MINDS_API_KEY")
+
+    if not global_key:
+        global_ws.set_secret("ANTON_MINDS_API_KEY", project_key)
+        project_ws.remove_secret("ANTON_MINDS_API_KEY")
+        message = "  Moved this project's saved publish key to ~/.anton/.env"
```
tino097: **This line names the wrong publishing account whenever `~/.cowork/.env` holds a key.**

Verified in pydantic-settings 2.13.1 that a later `env_file` entry wins, and `_build_env_files` orders `~/.cowork/.env` after `~/.anton/.env`. So on a desktop-configured machine (`~/.cowork/.env` has a key, no `~/.anton/.env`, project vault has one): `global_key` is empty, this branch runs, prints "Moved this project's saved publish key to ~/.anton/.env", and the re-resolve at line 594 comes back with the cowork key. The archive branch is skipped, so no warning is printed either.

The account genuinely did change, and collapsing to one identity is the point. But before this change the project key won, because `_ensure_workspace` promotes it into `os.environ`, which outranks every file. So the user is told their project key is now the publish identity at the exact moment it stopped being one. The divergent branch below is careful not to claim the retained key is the good one; this branch needs the same care.

`test_the_session_can_use_the_key_reconciliation_just_moved` creates no `~/.cowork/.env`, so it cannot see this.

anton/cli.py
```diff
@@ -407,6 +407,129 @@ def _get_settings(ctx: typer.Context):
...
+        archive = project_ws.env_path.with_name(".env.superseded")
+        # `.env.superseded` is matched by neither a `.env` nor a `*.env` ignore
+        # rule, and anton writes no .gitignore of its own, so a live credential
+        # would be staged by `git add -A`. Ignore the whole vault directory
+        # from inside it — self-contained, and it covers `.env` too.
+        gitignore = project_ws.env_path.with_name(".gitignore")
+        if not gitignore.exists():
+            try:
+                gitignore.write_text("*\n", encoding="utf-8")
+            except OSError:
```
tino097: **Three problems in these five lines, and the third undoes the reason the block exists.**

1. **`*` ignores the whole workspace state directory, not the archive.** `memory_dir` defaults to `.anton` and `resolve_workspace` puts `artifacts_dir` underneath it, so this newly ignores `anton.md`, `memory/rules.md`, `memory/lessons.md`, `memory/topics/`, `episodes/`, and every published dashboard under `artifacts/`. Already-tracked files keep working, which is what makes it hard to notice: the user finds out when a new file they meant to commit is silently skipped by `git add -A`. `.env*` covers the archive and `.env` without any of that, and still beats an ignore rule of the exact string `.env`, which was your reason for not renaming.

2. **`if not gitignore.exists()` is an escape hatch that removes the protection entirely.** A project that already has a `.anton/.gitignore`, for any reason, gets nothing, and the credential lands in an unignored file with no warning.

3. **`except OSError: pass` discards the failure and execution falls through to `_write_private(archive, ...)` at 497**, which writes the live key anyway. The one thing the ignore file exists to prevent is precisely what happens when it cannot be written. Skipping the archive on that failure costs nothing: the key stays in `.anton/.env`, the machine is left as it was, and `main` already degrades gracefully. CLAUDE.md, *Code quality*: errors are handled or propagated, never swallowed.

anton/cli.py
```diff
@@ -407,6 +407,129 @@ def _get_settings(ctx: typer.Context):
...
+    global_ws = Workspace(Path.home(), settings=settings)
+    global_key = global_ws.get_secret("ANTON_MINDS_API_KEY")
+
+    # This is a DISK migration; the process environment is not ours to edit.
+    # `set_secret`/`remove_secret` both write os.environ as a side effect
+    # (workspace.py), so an operator running `ANTON_MINDS_API_KEY=X anton` would
+    # have X silently popped here — and pydantic ranks os.environ above every
+    # env_file, so X is exactly the key that session was resolving. Snapshot and
+    # restore around the whole migration.
+    exported = os.environ.get("ANTON_MINDS_API_KEY")
```
tino097, line 514: **The snapshot is not exception-safe, so the exported-key bug is still reachable on the partial-failure path.**

`ANTON_MINDS_API_KEY=X anton`, project vault holds Y, global vault empty. `global_ws.set_secret` (478) writes `os.environ` = Y, then `project_ws.remove_secret` (479) raises `OSError` on a read-only project dir. `main` catches that at 590 and boots on, so the restore at 509 never runs: the session, every scratchpad child, and the `ANTON_MINDS_API_KEY` -> `OPENAI_API_KEY` fallback all keep Y for good, and `migrated=False` means there is no re-resolve either. Same shape if `_write_private(archive, ...)` at 497 raises after the divergent branch is entered.

`try/finally` around the three branches with the restore in the `finally` closes it. Worth doing precisely because `test_reconciliation_leaves_an_exported_key_alone` guards the happy path only.

tino097 (commented):
Four blocking findings on the fixes in `630f7ec` and `e24ca41`. Two are new defects introduced by the fixes; the 401 one is a fix I had marked as holding, which turns out to hold only for a vault anton wrote itself.

I have non-blocking notes as well (an unrestored `_ENV_FILES` global, `except OSError` missing `UnicodeDecodeError`, one more unescaped interpolation, and four test-isolation items). Holding those back so this round stays on what blocks.

Also withdrawing my ticket-id comment. The rule I cited lives in a `CLAUDE.md` outside this repo, `anton` uses ticket refs in comments as a matter of course, and there are pre-existing ones in these same files. That was the wrong finding to raise.

tino097 (changes requested):

anton/chat.py
alecantu7: Confirmed and fixed in `4bc1d15`. Reproduced both rows of your table against the repo venv — `load_env` returns `"abc123"` and `abc123 # note` where pydantic-settings returns `abc123` — so both took the else branch and the rejected key stayed put.

Added `workspace.vault_key()`, which reads a vault through `dotenv_values` (the same parser pydantic-settings uses), and the 401 ownership test now goes through it. Guarded by `test_401_clears_a_quoted_vault_line`.

Worth flagging that my first attempt at the guard did not catch this: asserting `vault_key()` returns the unquoted value tests the helper, not the branch that uses it, so the mutation still passed. The test now drives `_handle_publish` with a quoted vault and asserts the key is gone afterwards.

Also took the second half — the else-branch message now lists `.anton/.env` in this project alongside `~/.cowork/.env`, a local `.env` and an exported var.

anton/cli.py
alecantu7: Fixed in `4bc1d15`. You are right, and it is the same mistake I had just been careful to avoid one branch below: the divergent branch does not claim the retained key is good, while this one asserted an outcome it had not checked.

The message now appends a note when `~/.cowork/.env` holds a key: that it takes precedence and publishing still uses it. The move genuinely happened, so "moved" stays; what it no longer implies is that the moved key is now the identity.

You are also right about the test — it created no `~/.cowork/.env` and could not see this.

anton/cli.py
alecantu7: All three fixed in `4bc1d15`, and the third was the worst of them.

1. `*` → `.env*`. You are right about the blast radius: `memory_dir` defaults to `.anton` and `artifacts_dir` resolves underneath it, so I was ignoring `anton.md`, `memory/`, `episodes/` and every published dashboard — invisibly, because already-tracked files keep working. `.env*` still beats an ignore rule of the exact string `.env`, which was the reason I did not rename.

2. The `if not gitignore.exists()` escape hatch is gone; it appends the pattern when a `.gitignore` is already there.

3. Swallowing the OSError and then writing the key anyway was exactly backwards. Failing to protect the archive now skips the archive: the key stays in `.anton/.env`, the machine is left as it was, and the user is told. Nothing is lost by not migrating.

Guarded by `test_the_archive_is_skipped_when_it_cannot_be_gitignored` and `test_gitignore_scopes_to_env_and_appends_to_an_existing_file`.

anton/cli.py
alecantu7: Confirmed and fixed in `4bc1d15`. `try/finally` around the three branches, restore in the `finally`.

Your framing is the useful part: `main` catching the OSError and booting on is what makes this durable rather than transient — the exported key would have stayed replaced for the session, every scratchpad child and the `ANTON_MINDS_API_KEY`→`OPENAI_API_KEY` fallback, with `migrated=False` meaning no re-resolve to correct it.

Guarded by `test_exported_key_is_restored_even_when_a_write_fails`, which raises from `remove_secret` mid-migration. Your note that the existing test covered only the happy path was exactly right.

anton/cli.py
```diff
@@ -407,6 +407,129 @@ def _get_settings(ctx: typer.Context):
...
+        # `.env.superseded` is matched by neither a `.env` nor a `*.env` ignore
+        # rule, and anton writes no .gitignore of its own, so a live credential
+        # would be staged by `git add -A`. Ignore the whole vault directory
+        # from inside it — self-contained, and it covers `.env` too.
+        gitignore = project_ws.env_path.with_name(".gitignore")
+        if not gitignore.exists():
+            try:
+                gitignore.write_text("*\n", encoding="utf-8")
+            except OSError:
+                pass
```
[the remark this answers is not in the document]
alecantu7: Fixed in `4bc1d15` — same finding as tino097's on this block. The `if not exists()` guard is gone; the pattern is appended when a `.gitignore` is already present, and it is now scoped to `.env*` rather than `*`.

Also: when the ignore file cannot be written at all, the archive is no longer written either — previously the OSError was swallowed and the key was archived regardless, which defeated the protection entirely.

anton/tools.py
```diff
@@ -537,7 +537,30 @@ async def handle_publish_or_preview(session: ChatSession, tc_input: dict) -> str
...
+        # A bare AntonSettings() leaves `artifacts_dir` at its RELATIVE default,
+        # so `Path(settings.artifacts_dir)` below would resolve against cwd
+        # while the session's resolves to <workspace>/.anton/artifacts. That
+        # feeds `resolve_publish_target` the wrong roots, so `published_key`
+        # differs from what /publish and cowork-server compute and the next
+        # publish creates a duplicate report instead of updating this one.
+        # Anchor it on the session's own workspace when the host gave us one.
+        base = getattr(session, "_workspace", None)
+        settings.resolve_workspace(str(base.base) if base is not None else None)
```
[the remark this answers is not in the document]
alecantu7: Fixed in `4bc1d15`. The check now covers `minds_api_key`, `artifacts_dir`, `publish_url` and `minds_ssl_verify` — every field the handler goes on to read — so a partial object fails over to the fallback here rather than raising `AttributeError` mid-publish.

tino097 (approved):
LGTM

PR #446, opened 2026-09-04
Feature/cleanup

ZoranPandovski:
1. Removed the core/dispatch/ messaging-channel layer — registry, router, repository, policy, local_runtime, session_store. 
2. Replaced ChatSession.turn()'s old ~300-line parallel implementation with a thin wrapper over turn_stream(). It had drifted to lack verification and truncation recovery that turn_stream has, and nothing in-repo called it directly. 
3. Fixed a latent bug in the local-vault connect flow: one of three places that generate a random connection slug skipped the collision check, so two connections could silently overwrite each other. Consolidated into one helper.
4. Removed next_connection_number (three implementations) — leftover from a naming scheme nothing uses anymore.
5/ Removed ~30 unused imports, dead functions, and write-only fields across the package.


anton/core/session.py
```diff
@@ -3906,303 +3894,31 @@ def _open_ds_turn_scope(self) -> None:
...
+        parts: list[str] = []
+        final: str | None = None
+        async for event in self.turn_stream(user_input):
+            if isinstance(event, StreamTextDelta):
+                parts.append(event.text)
+            elif isinstance(event, StreamComplete):
+                response = event.response
+                if (response.content or "").strip() and not response.tool_calls:
+                    final = response.content
+        return final if final is not None else "".join(parts)
```
tino097: **The `"".join(parts)` fallback returns every tool round's preamble, not the reply.**

Cause: `parts` accumulates *all* `StreamTextDelta` text for the whole turn, including the preamble the model writes before each tool call. So whenever the last response has empty content, `final` stays `None` and this line returns the concatenation of every preamble.

Reproduced against this head: a 3-round turn (preamble + tool_call, preamble + tool_call, empty content / `end_turn`) returns `'Let me look at the DB.Now checking the second table.'`. The removed implementation returned `response.content or ""`, i.e. `""`.

At that point history's last entry is the tool_result, so the returned string matches nothing that was persisted. That contradicts the docstring above: a streaming consumer read those preambles as two separate messages interleaved with tool output, never as one joined string.

anton/core/session.py
```diff
@@ -3906,303 +3894,31 @@ def _open_ds_turn_scope(self) -> None:
...
+        text delta without a closing response event.
+        """
+        parts: list[str] = []
+        final: str | None = None
+        async for event in self.turn_stream(user_input):
+            if isinstance(event, StreamTextDelta):
+                parts.append(event.text)
+            elif isinstance(event, StreamComplete):
+                response = event.response
+                if (response.content or "").strip() and not response.tool_calls:
```
tino097: **This filter latches the thalamus's accounting-only `StreamComplete`, so a truncated router answer that was deliberately discarded can be returned to the caller.**

Cause: `turn_stream` yields the gate call's response as a bare `StreamComplete` purely for token accounting (session.py:4183-4184, *"surface the gate call's usage as its own StreamComplete so token accounting counts it"*). It is not a reply. But it has `.content` and no `.tool_calls`, so this condition accepts it as `final`.

The bad case is the exact one the router guards against: `thalamus.py:326-331` routes a gate response with `stop_reason == "max_tokens"` to `ACTION_DELEGATE` *because* a truncated reply must never reach the user. Reproduced against this head: content `'SECRET half-written answer that was cut'` on a max_tokens gate response gets latched here, and when the real turn ends with empty content `turn()` returns that truncated text verbatim.

Streaming consumers were never exposed, because they render `StreamTextDelta` only. That is precisely why yielding the event was safe before this wrapper started reading `.response.content` off it. Skipping the pre-loop gate `StreamComplete`, or filtering on `stop_reason`, would close it.

anton/core/session.py
```diff
@@ -4295,9 +4011,7 @@ async def turn_stream(
     ) -> AsyncIterator[StreamEvent]:
         """Streaming turn. Owns the out-of-band emitter's lifetime.
 
-        The emitter exists only for the duration of one streaming turn, which
-        is what makes questions unavailable on the non-streaming `turn()`
-        path: nothing there would render them.
+        The emitter exists only for the duration of one streaming turn.
         """
         # Before any tool task is spawned, so a connect made mid-turn registers
         # into a container this turn still holds.
```
tino097, line 4016: **Now that `turn()` delegates here it gets a live emitter, which removes the gate that made `ask_user` choice questions unavailable on the non-streaming path, and can hang the turn.**

Cause: `elicit()` short-circuits to `AskAnswer(status="unavailable")` only when `session.emitter is None` (`anton/core/interaction/elicit.py:152-153`). Inside `turn()` that used to hold. It no longer does, so `elicit()` calls `begin()`, emits `StreamAskUser` into the generator `turn()` drains and discards, then awaits `elicitor.ask()` — a question no non-streaming caller can see, let alone answer.

Reproduced against this head: a probe with a host-style elicitor reported `HUNG-waiting-on-elicitor`. `ElicitorProtocol.timeout_s` is typed `int | None` (elicit.py:110), so a host that leaves it `None` hangs the turn forever rather than timing out.

The comment at elicit.py:151-154 still asserts the old invariant ("including the non-streaming `turn()`"), so the next reader will trust a gate that no longer exists. Either keep `turn()` emitter-free for elicitation, or re-key that gate on something the wrapper actually sets.

anton/core/session.py
```diff
@@ -3906,303 +3894,31 @@ def _open_ds_turn_scope(self) -> None:
             )
 
     async def turn(self, user_input: str | list[dict]) -> str:
```
tino097, line 3896: **The wrapper does inherit the behaviours whose seam-guard tests were deleted — but its return value is new code with a non-trivial selection rule, and it has no test beyond the one case where every implementation agrees.**

Deleted: `test_the_non_streaming_turn_records_too`, `test_round_cap_grace_also_applies_to_the_non_streaming_turn`, `test_ceiling_applies_to_the_non_streaming_turn`, `test_the_non_streaming_turn_also_refuses_a_cut_open_call`, `test_nonstreaming_turn_path_also_emits`, `test_the_nonstreaming_turn_path_stamps_the_cause_too`, `test_the_non_streaming_turn_path_stamps_an_id_too`, `test_non_streaming_turn_marks_round_cap`. Added: `test_non_streaming_turn_wraps_turn_stream`, a single-round text turn.

The inheritance argument is right for the *behaviours*: recording, round cap, ceiling, truncation refusal and the `ended_by` stamps now come from `turn_stream`. It does not cover lines 3912-3921. And a one-round text turn is exactly the case where the last-`StreamComplete` rule, the join fallback, and the old `response.content or ""` all return the same string — which is why the two bugs above both reproduce against a fully green suite (3075 passed, 31 skipped). The docstring's claim that the return value is "the same message a streaming consumer would have read last" is asserted nowhere.

anton/core/dispatch/adapter.py
```diff
@@ -119,8 +119,6 @@ class OutboundMessage:
 
     address: PlatformAddress
     text: str
```
tino097, line 121: **Narrows a cross-repo data contract: `OutboundMessage.attachments`, `OutboundMessage.reply_to_message_id` and `ChannelSetup.on_metadata` are all removed.**

Cause for flagging rather than accepting as dead code: "unused inside anton" is not the test for this module. cowork-server imports these exact types in `cowork/channels/plugins/{slack,telegram,discord,whatsapp}.py` and `cowork/channels/runtime.py` — and the docstring this PR *adds* to the module says cowork-server built its registry "against these same data types".

Nothing over there reads the removed fields today, so this does not break at import. The cost is that the contract becomes asymmetric: `runtime.py:573` already iterates `event.message.attachments` on the inbound side, so a channel can receive a file while an agent reply can no longer carry one, and `on_metadata` (display name, member count, group/DM flag) loses its only channel.

Per the root CLAUDE.md, narrowing a cross-repo contract needs an explicit task and a `## Ships with` note. Neither is in this PR.

anton/commands/datasource/connect.py
```diff
@@ -334,7 +334,7 @@ async def _retry_custom_credentials() -> bool:
         return session
 
     assert engine_def is not None
-    # _telemetry("ds_connect_attempt", engine=engine_def.engine)
+    _telemetry("ds_connect_attempt", engine=engine_def.engine)
```
tino097: **This re-enables a double-count that commit 8ee747b3 ("Commentout double connection attempt event") deliberately silenced.**

Cause: `anton/tools.py:125` already emits `send_event(_settings, "ds_connect_attempt", engine=engine)` before it calls `handle_connect_datasource(..., from_tool_call=True)`. Un-commenting this line means every model-driven `/connect` to a known engine reports two attempts for one attempt.

That inflates the denominator of the connect funnel, so any attempt -> success rate computed from it is roughly halved. It is also an unrelated behaviour change in a PR scoped to dead-code removal. If the intent is to fix the earlier workaround properly, the fix belongs at the `tools.py` call site (one emitter, not two) and in its own change.

anton/commands/datasource/verify.py
```diff
@@ -201,43 +204,11 @@ async def handle_test_datasource(
...
-        console.print(
-            f"[anton.success]        ✓ Connection test passed for"
-            f" [bold]{slug}[/bold]![/]"
-        )
-    console.print()
+    # interactive=False: `/test` is a one-shot check, not an edit flow —
+    # fail closed on the first error rather than prompting to re-enter
+    # credentials (that's what `/edit` is for). `retry_fields` is unused
+    # in this mode, so engine_def.fields is just a placeholder.
+    await run_connection_test(
```
tino097, line 221: **Delegating to `run_connection_test` silently changes `/test`'s terminal output in three ways.**

Cause: `run_connection_test` is `/connect`'s helper, so it brings `/connect`'s conversational wording plus its own banner and verdict lines, and those lines do not know the slug.

`/test postgresql-ab12cd34` now prints the banner at 203-205 ("Testing connection **postgresql-ab12cd34**...") immediately followed by the helper's own "(anton) Got it. Testing connection..." (`connect.py:53`) — "Got it." is odd in a one-shot check — and then "Connected successfully!" / "Connection failed." with no slug, where the removed code said "Connection test passed for **postgresql-ab12cd34**!". Testing several connections in a row, a user can no longer tell which verdict belongs to which.

The error surface flips too: the removed code took `next((ln for ln in error_text.splitlines() if ln.strip()))`, the first non-empty line; the shared helper takes `reversed(...)`, the last. A Python traceback now surfaces the final exception line instead of the header. CLAUDE.md: preserve CLI output unless the task says to change it.

tests/conftest.py
```diff
@@ -150,6 +166,40 @@ def make_mock_llm() -> AsyncMock:
...
+        return gen()
+
+    mock.plan_stream = _plan_stream_from_plan
+    # Verifier default: a clean COMPLETE, so a tool-loop test that doesn't
+    # care about verification ends after its scripted rounds instead of
+    # falling into the INCOMPLETE-continuation loop on a bare Mock verdict.
+    # Tests that exercise the verifier set their own `generate_object_code`.
+    from anton.core.session import _VerifierVerdict
+
+    mock.generate_object_code = AsyncMock(
```
tino097: **Baking a COMPLETE verdict into the shared fixture makes the completion verifier a guaranteed no-op for every test that uses `make_mock_llm` without an override.**

The cause is understandable and the comment states it honestly: the migrated tests traverse the verifier for the first time now that they run through `turn_stream`, and a bare `Mock` verdict drops them into the INCOMPLETE-continuation loop. But the fix lands on the default rather than on the tests that needed it.

Consequence: the verifier's latch / re-probe / continuation machinery (session.py:5580-5900) is never entered by any test using this fixture. A regression there — a latch that never releases, a re-probe that never fires, a continuation that double-appends — passes silently everywhere except the few tests in `test_turn_cost_terminals.py` that build their own `_verdict_llm`.

An opt-in shape (`make_mock_llm(verifier="COMPLETE")`, or a separate `mock_llm_verified` fixture) keeps the migration green without disarming the verifier suite-wide.

tests/conftest.py
```diff
@@ -126,6 +126,22 @@
 os.environ["ANTON_POSTHOG_KEY"] = ""
 
 
+async def run_turn(session, user_input) -> str:
+    """Drive one turn through `turn_stream` and return the reply text.
+
+    The migration shim for tests written against the removed non-streaming
+    `ChatSession.turn()`: drains the stream, concatenates the text deltas —
+    the same text `turn()` used to return.
```
tino097: **"the same text `turn()` used to return" is not accurate, and the shim also diverges from the new `turn()` — so the ~25 tests migrated onto it no longer assert either contract.**

Three different rules are in play: the removed `turn()` returned `response.content or ""`; this shim returns every `StreamTextDelta` joined; the new `turn()` returns the last non-tool-call `StreamComplete`'s content, falling back to joined deltas.

They coincide only for a single-round text turn — which happens to be every migrated call site, so the suite stays green while the assertions have quietly stopped covering any return contract at all. Confirmed by probe on this head: the same multi-round turn yields `'Let me look at the DB.Now checking the second table.'` from this shim and `""` from the removed implementation.

If the shim is meant to stand in for `turn()`, calling `session.turn()` directly would make these tests cover the new code (and would catch the two bugs flagged in session.py). If it is meant to be "all streamed text", the docstring should say that instead.

anton/core/tools/tool_defs.py
```diff
@@ -1,4 +1,4 @@
-from anton.core.tools.progress import ToolProgress
+from anton.core.tools.progress import ToolProgress  # noqa: F401 — re-export
```
tino097: **`# noqa: F401` where the language has a way to say this.**

Cause: the import *is* a public re-export, and F401 fires only because the syntax does not say so. `from anton.core.tools.progress import ToolProgress as ToolProgress`, or an `__all__` entry, states the intent directly, is understood by ruff and by every type checker with no suppression, and survives a linter config change that the pragma would not.

Root CLAUDE.md, Code quality: "Fix the cause, never the check. No `# type: ignore`, `# noqa`, `eslint-disable`... to get past a failing check."

tests/test_turn_cost_terminals.py
```diff
@@ -313,27 +313,6 @@ async def test_round_cap_reports_the_cap_not_cap_plus_one(workspace):
...
-
-
-# ---------------------------------------------------------------------------
-# The three hand-back terminals. Every one of these marks survived deletion
-# with the suite green (#309 review mutation run) — and they are precisely the
-# "expensive turn ended badly" values the `group by ended_by` contract needs.
-# ---------------------------------------------------------------------------
-
-
 def _verdict_llm(status: str, *, tool_rounds: int = 1):
```
tino097, line 318: **The deleted test took with it the section header that documented the tests still starting here.**

Cause: `test_non_streaming_turn_marks_round_cap` sat directly under the header, so removing the test removed the comment above it — but that comment described what *follows*, not what was deleted.

It read: "The three hand-back terminals. Every one of these marks survived deletion with the suite green (#309 review mutation run) — and they are precisely the 'expensive turn ended badly' values the `group by ended_by` contract needs." That is about `_verdict_llm` and the STUCK / DENIED / ERROR tests below, all of which remain. The mutation-run evidence for why those three assertions are load-bearing is now lost, so the next cleanup pass has nothing telling it they are not redundant. Worth restoring above this line.

tests/test_thalamus.py
```diff
@@ -373,18 +380,18 @@ async def test_confirmed_router_auth_failure_falls_back_to_planning(self):
...
-        planning.complete.assert_awaited_once()
+        planning.stream.assert_called_once()
 
     async def test_image_turns_skip_thalamus(self):
         llm = make_mock_llm()
         llm.gate = AsyncMock()
         llm.plan = AsyncMock(return_value=_response("Nice chart."))
         session = ChatSession(ChatSessionConfig(llm_client=llm, router_enabled=True))
-        reply = await session.turn(
+        reply = await run_turn(session,
```
tino097: **Trailing space after `run_turn(session,`.**

An artifact of the mechanical `turn(` -> `run_turn(` rewrite, and the only trailing-whitespace line the diff introduces: `grep -n '[[:space:]]$' tests/test_thalamus.py` at this head returns just 394. Harmless at runtime, but it is what a formatter or a future ruff W291 gate flags.

tino097 (commented):
Review of the cleanup at `0416678`. The full suite is green on this head (3075 passed, 31 skipped); the three `turn()` findings below reproduce anyway, which is the substance of the review.

The delegation is the right call and most of the deletions are clean. Three blockers, all in the new `turn()` wrapper's return path, each reproduced with a probe against this head; then two behaviour changes riding along outside the stated dead-code scope; then smaller items. Each inline comment carries the mechanism.

### Blockers

1. `session.py:3921` — the `"".join(parts)` fallback returns concatenated tool preambles instead of the reply.
2. `session.py:3919` — the `StreamComplete` filter latches the thalamus's accounting-only event, so a truncated router answer that the router deliberately sent to DELEGATE can be returned verbatim.
3. `session.py:4017` — `turn()` now holds a live emitter, which removes the gate that made `ask_user` choice questions `unavailable` on the non-streaming path. A probe with a host-style elicitor hung, and `timeout_s: int | None` means a host leaving it `None` hangs forever.

All three share one root cause: the wrapper treats the event stream as a source of reply text, but `turn_stream` emits `StreamComplete` and text deltas for accounting and rendering, not as a "this is the reply" signal. The suite stayed green because the single added test is a one-round text turn, the one case where every candidate implementation returns the same string.

### Stale docs, outside the diff so no inline anchor

- `docs/docs/developer/scratchpad-runtime.md:35` — "`cancel_all_running()` handles Ctrl-C": the method is removed here, and nothing handled Ctrl-C through it even before, so the line was doubly wrong.
- `docs/docs/developer/scratchpad-runtime.md:85` — "A `RemoteLightsailScratchpadRuntime`...": class and factory removed.
- `docs/docs/developer/adding-a-datasource.md:142` — "typos via `DatasourceRegistry.fuzzy_find()`": method removed.
- `anton/README.md:876` — `Lesson` described as a dataclass with `rule`, `triggers`, `detector`: `triggers` removed.
- `anton/README.md:945` and `docs/docs/developer/cerebellum-and-acc.md:124` — `_schedule_acc_flush()` "runs at the same two spots, end of `turn()` and end of `turn_stream()`": one spot now.
- `anton/core/dispatch/adapter.py:27` — dangling `:class:`PermissionPolicy`` Sphinx reference to the `policy.py` this PR deletes, in the only surviving dispatch module.

anton/core/session.py
ZoranPandovski: Confirmed. Not fixed yet, fixing next.

anton/core/session.py
ZoranPandovski: Confirmed. The fix for the item above will cover this too.

anton/core/session.py
ZoranPandovski: Confirmed. Not fixed yet, fixing next.

anton/core/session.py
ZoranPandovski: Agreed. Will add real coverage once the wrapper logic above is fixed.

anton/core/dispatch/adapter.py
ZoranPandovski: Fixed. Restored the three fields, contract is unchanged.

anton/commands/datasource/connect.py
ZoranPandovski: Fixed. Reverted, matches the existing emit in tools.py.

anton/commands/datasource/verify.py
ZoranPandovski: Fixed. Banner and verdict wording restored, only the error-line fix kept.

tests/conftest.py
ZoranPandovski: Agreed. Not fixed yet, planned.

tests/conftest.py
ZoranPandovski: Agreed. Not fixed yet, planned.

anton/core/tools/tool_defs.py
ZoranPandovski: Agreed. Not fixed yet, planned.

tests/test_turn_cost_terminals.py
ZoranPandovski: Agreed. Will restore the comment.

tests/test_thalamus.py
ZoranPandovski: Good catch. Not fixed yet, fixing next.

anton/core/session.py
ZoranPandovski: Fixed. turn() no longer interprets the event stream; it reads history[-1] instead, which every terminal path already guarantees is the right message. Reproduced your exact scenario as a test.

anton/core/session.py
ZoranPandovski: Fixed by the same change. Reading history[-1] means the thalamus's accounting-only StreamComplete is never inspected at all. Added a dedicated test reproducing the truncated-gate-response case.

anton/core/session.py
ZoranPandovski: Fixed. turn_stream() now takes enable_interaction=False, which turn() passes so it never creates a live emitter. Verified with a host elicitor (timeout_s=None) under a 5s wait_for; no hang.

anton/core/session.py
ZoranPandovski: Added three tests covering exactly this: the preamble-concatenation case, the truncated-gate-response case, and the emitter/ask_user case. Each reproduces the bug on the old code and passes on the fix.

tests/conftest.py
ZoranPandovski: Fixed. The docstring now says what run_turn actually returns (all streamed text concatenated) and states plainly it's not equivalent to turn(), with the divergence explained.

tests/conftest.py
ZoranPandovski: Fixed. Removed the blanket default from make_mock_llm. Only 6 tests across 3 files actually needed it; each now sets its own generate_object_code, matching the existing convention. Verified the other 141 make_mock_llm call sites in the suite were not relying on it.

anton/core/tools/tool_defs.py
ZoranPandovski: Fixed. Replaced the noqa with import ToolProgress as ToolProgress, which states the re-export directly.

tests/test_turn_cost_terminals.py
ZoranPandovski: Fixed. Restored the comment above the three hand-back terminal tests.

tests/test_thalamus.py
ZoranPandovski: Fixed. Trailing whitespace removed.

tino097 (approved):
LGTM

PR #456, opened 2026-09-08
ENG-1204 validate generated html/xlsx artifacts for local version

ea-rus:
Update: 
- to check what type of artifact is generated
- run validator based on file type

Html:
- if it is electron distribution:
  - loads the page in cowork's bundled Chromium in headless mode, flags console errors, failed local requests, crashes, empty render.
- otherwise: skip

Excel:
- check if we have libreoffice installed (soffice): 
  - re-save file with libreoffice and look at converted file to find formula errors
- if no office installed - use python
  - check circular formulas with openpyxl (a new dependency)

New env vars:
- ANTON_XLSX_LINT_OFFICE (optional LibreOffice path override, auto-detected via PATH otherwise) 
- ANTON_HTML_LINT_BROWSER (browser binary path, required for HTML checks to run — host must set it, cowork sets it to process.execPath).

Related PRs:
- https://github.com/mindsdb/cowork/pull/954

Fixes: https://linear.app/mindsdb/issue/ENG-1204/anton-presents-generated-files-as-finished-without-ever-opening-them


alecantu7 (commented):
Reviewed all three ENG-1204 PRs together, at `38b36e74` / `62c18813` / `f85ae03b`.

**The mechanism is sound, and one part is better than the ticket specified.** I'd asked for findings to surface through `open()`/`list()`; this puts them inline on the cell's own tool result (`[artifact lint]` appended to `result_text`). That's stronger, and it's the thing the ticket was most worried about — ENG-350's guard shipped into the live tool schema and fired **0 times across 676 cells**. Text already in context can't be skipped. `except Exception → lint_messages = []` also satisfies the best-effort criterion: a crashing checker can't fail the cell.

**Verified by execution:** all 15 circular cells flagged — `solo_rows = [18, 19, 21, 23, 24]` plus E/F pairs on 6/7/8/12/26, which matches the ticket's list exactly; a clean workbook produces nothing; `E5`'s correctly-qualified self-reference is correctly *not* flagged; the status is omitted rather than `None` when there's no verdict; 3159 passed / 37 skipped.

---

### 1. Two tests fail on macOS, pass in CI

    FAILED tests/test_artifact_lint_status.py::test_not_validated_when_a_configured_checker_could_not_run
    FAILED tests/test_tool_handlers_html_lint.py::test_a_configured_browser_that_fails_to_produce_output_sets_not_validated

Three sites hardcode `ANTON_HTML_LINT_BROWSER=/bin/true`. **macOS has no `/bin/true`** — it's `/usr/bin/true`. So the path doesn't exist, the code correctly reads that as "no browser configured", sets no status, and the assertion fails. Swapping the path: **8 passed, 2 skipped**, both green.

CI is Linux so it never sees this, but it goes red on every dev machine here. `shutil.which("true")` fixes it.

### 2. The badge never reaches a web user

The only wiring is `anton_harness/harness.py:607`, reading `session.artifact_lint_status` off the in-process `ChatSession`. Of the three callers of `publish_and_card_turn_artifacts`, only that one passes it — and the other two can't: `grep -rn "lint" anton/cloud_turn/` returns nothing, so the pod never emits the status and there is nothing for `turnqueue/remote_turn.py` to forward.

Worth being explicit that **the lint itself does run in the pod**, so a web agent still sees findings and can act on them. It's only the user-visible half that stops at desktop. Nothing in the PR body says this today, and it reads as full coverage.

### 3. `xlsx_office_check.py` is the thing the ticket ruled out by name

204 lines of LibreOffice-backed recalculation, where Out-of-scope reads: *"Formula evaluation (`formulas` / `pycel` / LibreOffice headless)… A separate decision with a real dependency cost; the structural lint delivers most of the value first."*

In its favour: it's optional, degrades to the structural lint, and the `None`-vs-`[]` distinction is exactly the honesty the ticket asks for. But it's a second checker whose behaviour depends on a host binary being present. Related — **`openpyxl` is now a hard install-wide dependency** (not on `staging` before). The ticket said it was "already in the scratchpad environment"; that's the scratchpad env, not anton's own, so every install now carries it.

Not asking for a revert — asking that it be a decision rather than a merge.

### 4. A proposal on the badges

Right now **no badge means five different things**: checked and clean; no checker for that format (`.pdf`, `.docx`, `.csv`, images); `.html` with no browser configured; a file over the 10 MB ceiling; or web. I confirmed this by running the path over each type — all return `status={}`.

Only one case shows `Not validated`: a *registered* checker that ran and returned `None`. So the badge that reads as "we didn't check this" actually means "we tried and the checker broke", while the genuinely-unchecked artifacts are silent.

**I'd drop `not_validated` and keep only `has_errors`** — one badge, one meaning, no implied claim of verification. Easy to add more once there's data on how often it fires and how often the agent fixes things itself; much harder to remove a badge users have already learned.

The `pop` that clears the status when a later cell fixes the file is a nice property regardless — it makes the agent-corrected-it path visible.

---

**What I could not test:** the HTML checker against a real browser; the packaging seam (whether anton finds the Electron binary inside a shipped `.app` — if that discovery is dev-only, the HTML half silently never runs for users); the LibreOffice oracle (no `soffice` available); false positives against real-world workbooks; and **whether the agent acts on a finding**. That last one is the ticket's own acceptance criterion and isn't answerable from a diff — it needs a live run or an eval.

alecantu7 (approved):
Approving at `ce51060a`. The rollback is clean and complete — cross-repo grep for `artifact_lint_status`, `LINT_STATUS_*`, `is_browser_configured` and `not_validated` returns zero hits, so there's no orphaned constant or dead import left behind. CI green, and 98 passed / 6 skipped across the eight lint and artifact test files.

Worth saying explicitly: **this moves the PR toward the ticket, not away from it.** ENG-1204's Done-when never asked for a user-facing badge — it asks that findings reach the agent and that the agent act on them. Nothing regresses.

**Re-verified the headline criterion at this head, both ways.** Rebuilt the customer's workbook shape (E/F pairs on rows 6/7/8/12/26 plus solos 18/19/21/23/24) and ran the lint:

- with LibreOffice present: **15 findings**, `E5`'s correctly-qualified self-reference correctly not flagged
- with the office oracle forced unavailable (the pod and most user machines): **15 findings**, `E5` still not flagged

So the structural lint carries the promise on a bare host. That was the thing most worth confirming after the message loop changed.

---

### 1. The `/bin/true` test now passes on macOS for the wrong reason

This is my earlier finding #1, and the rollback masked it rather than fixed it.

```python
def test_a_configured_browser_that_fails_to_produce_output_is_silent(...):
    """A browser IS configured here — the run itself just didn't produce a usable result..."""
    monkeypatch.setenv("ANTON_HTML_LINT_BROWSER", "/bin/true")
    assert lint_changed_artifact_files(store, before) == []
```

macOS has no `/bin/true` (it's `/usr/bin/true`), so `_discover_browser()` returns `None` — confirmed by running it. The test therefore exercises the *no browser configured at all* path, which is byte-identical to `test_no_browser_configured_produces_no_messages` directly above it. Two tests, one path, and the docstring describes a scenario that isn't reached.

Previously this failed loudly on every dev machine here. Now the assertion is weak enough (`== []`) that the wrong path satisfies it too — which is the worse failure mode, because nothing signals it.

**Fix:** `shutil.which("true")` instead of the hardcoded path, same as before.

### 2. The LibreOffice oracle suppresses the more actionable message

`_xlsx_linter` runs `check_xlsx_via_office` first and only falls back to `lint_xlsx` when it returns `None`. Both message sets, from the run above:

- **with LibreOffice:** `formula 'SLOPE(E6:J6,...)' recalculated to #VALUE! in LibreOffice.`
- **without:** `formula 'SLOPE(E6:J6,...)' references a same-sheet range that contains its own cell — likely a missing cross-sheet prefix.`

The second names the fix; the first names a symptom. Now that the whole design rests on the agent self-correcting from this text, message quality is functional rather than cosmetic — and the host with the *better* tooling gives the agent the *worse* instruction.

There's a second edge in the same precedence: if the oracle returns `[]` the structural lint never runs at all, so coverage genuinely differs by host rather than the oracle being a strict superset.

**Suggestion:** run both and concatenate, or prefer the structural message when both fire on the same cell.

### 3. The agent now gets exactly one chance, at the moment of the write

`lint_changed_artifact_files` has two call sites — `tool_handlers.py:841` and `session.py:5129`, both scratchpad-exec paths. `handle_open_artifact` (`:593`) and `handle_list_artifacts` (`:554`) contain no lint reference, and `_reconcile_files` — the hook point the ticket names — is never wired to it.

So a finding exists only in the tool-result text of the cell that wrote the file. It isn't persisted, isn't re-raised when the agent opens the artifact later, and there's no badge. If the agent doesn't act in that moment, the signal is gone for good — which is exactly your own two escape hatches (out of retries, or a deliberately invalid artifact).

This is a pre-existing deviation, but removing the status is what makes it load-bearing: `artifact_lint_status` used to outlive the tool result, and now nothing does.

**Smallest fix:** lint in `handle_open_artifact` too. That's literally the ticket's first Done-when, and it's the cheap version of a second chance — no new state, no UI.

### 4. Nothing counts findings, so the "add states later" plan has no input

The agreed follow-up was to revisit once we see how often findings appear and how often the agent fixes them on its own. Nothing measures either — `lint_messages` is only string-concatenated into tool-result content (`tool_handlers.py:861`, `session.py:5141`); no collector event, no counter, no log line. Counting `[artifact lint]` in Langfuse won't substitute: the transcript caps at 12 events / 400 chars.

One counter at each call site (findings emitted, artifact still dirty at end of turn) would close it.

### 5. Body still describes the removed states

Lines 18-19 still list `has_errors` / `not_validated` as outputs. You flagged this yourself — noting it only because the body is what QA and future readers see, and our QA comments get written against the merged diff.

### 6. Done-when gap for QA, not for this PR

*"flags all 15 circular cells in `2026 Forecast (7).xlsx` … (both as fixtures)"* — there are no `.xlsx` fixtures committed under `tests/`. The behaviour is verified (above), but the ticket asked for the customer's actual workbook as a regression fixture. Worth deciding deliberately rather than letting QA file it as a miss.

PR #461, opened 2026-09-09
ENG-2421: carry artifact-delivery guidance in the pod's system prompt

alecantu7:
Cause half of [ENG-2421](https://linear.app/mindsdb/issue/ENG-2421/the-agent-still-hands-users-file-links-that-cannot-open-the-eng-2044): web agents keep telling users *"download it here"* with `sandbox:/mnt/…`, bare `/mnt/…`, or `http://127.0.0.1:PORT` links — 4 web users, 11 dead links in the 09-04 review window, all **after** ENG-2044's card Download shipped.

## Root cause

ENG-1636's "point at the Live Artifacts panel, never hand out paths" guidance lives in cowork-server's `_turn_style_context` — injected by the **in-process harness, which refuses to run org turns**. Web turns execute in the scratchpad pod, whose `SystemPromptContext` carries `runtime_context` only (ENG-1638) and **no delivery guidance at all**. The web population has never received the instruction; the failure pattern matches exactly (every post-deploy complainant on the ticket is a web user).

## What

`build_cloud_chat_session` now passes `CLOUD_ARTIFACT_DELIVERY_GUIDANCE` as the prompt suffix — the web-worded equivalent of the desktop block: files appear in the Live Artifacts panel with a Download control; pod filesystem paths and loopback URLs exist only inside the workspace and can never be opened from the user's browser — never present them as links or text, never invent `sandbox:` URLs; HTML/MD artifacts are auto-shared at turn end with the link on the artifact card, which the model does not know and must not guess; on "I can't find the file", point at the panel again, never repeat a path.

Every sentence is provable from the product (ENG-2044 Download, ENG-1680 autopublish, the renderer's link neutraliser). The constant is pod-owned rather than wired over the Redis seam — no producer→controller→contract change, and the desktop harness keeps its own block (consolidating both into one anton-owned source is a reasonable follow-up once cowork-server naturally migrates, per the ENG-1808 pattern).

## Tests

`test_cloud_turn_session.py` (62/62): pins the suffix identity on the captured `ChatSessionConfig`, its load-bearing sentences individually, **and reachability** — the test assembles the real prompt via `ChatSystemPromptBuilder.build()` and requires the guidance inside it, so a prompt-builder refactor that stops appending `suffix` fails it too (review finding). Mutation-verified both ways: unwiring `suffix=` in session.py and disabling the append in prompt_builder.py each fail the test. Full suite: 3,132 passed, 31 skipped.

## Delivery — the step that gets skipped

**Staging**: since ENG-2129, anton's staging publish builds the pod image and scratchpad-controller's `values-staging.yaml` follows the moving `:staging` tag with `imagePullPolicy: Always` (verified on scratchpad-controller `origin/main`, 2026-09-09 — a temporary sha pin from 09-02 was reverted by `9d6230d` on 09-08). So staging web picks this up automatically after merge. **Production** still pins its worker image by hand — the prod half of ENG-2421 lands only when that pin moves to an image carrying this commit, which goes in the QA comment as Step 0. Desktop is untouched (its guidance ships from cowork-server and is byte-pinned there).

Sibling PR (net half, renderer): mindsdb/cowork#956 — no merge-order constraint; they harden different layers of the same failure.

Security pass: prompt-text only — no new tool surface, no secrets, no user input interpolated into the constant.

### DIFF
--- a/tests/test_cloud_turn_session.py
+++ b/tests/test_cloud_turn_session.py
@@ -79,6 +79,38 @@ def test_scratchpad_uses_local_factory_and_is_workspace_bound(tmp_path, monkeypa
     assert cfg.session_id == "conv_1"
 
 
+def test_cloud_prompt_carries_artifact_delivery_guidance(tmp_path, monkeypatch):
+    """ENG-2421: ENG-1636 taught the DESKTOP harness to point users at the Live
+    Artifacts panel instead of local paths — but that block lives in
+    cowork-server's in-process harness, which refuses to run org turns, so web
+    agents were never told and handed users sandbox:/mnt/... and loopback
+    "download links" for weeks after every card-side fix shipped. The pod's own
+    prompt context must carry the web-worded equivalent."""
+    _, cfg = _build(tmp_path, monkeypatch)
+    suffix = cfg.system_prompt_context.suffix
+    from anton.cloud_turn.session import CLOUD_ARTIFACT_DELIVERY_GUIDANCE
+    assert suffix == CLOUD_ARTIFACT_DELIVERY_GUIDANCE
+    # Reachability, not just wiring (review finding on #461): the config field
+    # being set proves nothing if the prompt builder stops appending `suffix` —
+    # exactly the "fix shipped while the failure continued" class this ticket
+    # documents. Assemble the real prompt and require the guidance inside it.
+    from anton.core.llm.prompt_builder import ChatSystemPromptBuilder
+    prompt = ChatSystemPromptBuilder().build(
+        system_prompt_context=cfg.system_prompt_context,
+        conversation_started=True,
+        proactive_dashboards=False,
+        output_dir=str(tmp_path),
+    )
+    assert CLOUD_ARTIFACT_DELIVERY_GUIDANCE.strip() in prompt
+    # The load-bearing sentences, pinned individually so a rewrite that drops
+    # one is caught even if the identity assertion above is loosened later.
+    assert "Live Artifacts" in suffix
+    assert "do NOT hand them its location on disk" in suffix
+    assert "sandbox:/mnt/data/" in suffix
+    assert "127.0.0.1" in suffix
+    assert "never repeat a path" in suffix
+
+
 def test_cloud_workspace_does_not_create_anton_md(tmp_path, monkeypatch):
     """ENG-1817: `.anton/anton.md` is cowork-server's staged copy of the project
     instructions. If the pod creates a template there, the next staging pass
--- a/tests/test_cloud_turn_session.py
+++ b/tests/test_cloud_turn_session.py
@@ -109,6 +109,11 @@ def test_cloud_prompt_carries_artifact_delivery_guidance(tmp_path, monkeypatch):
     assert "sandbox:/mnt/data/" in suffix
     assert "127.0.0.1" in suffix
     assert "never repeat a path" in suffix
+    # The base ARTIFACTS prompt tells the agent to include the primary file's
+    # path and to prefer launch_backend's (loopback) url — the cloud block must
+    # override both by name, not just coexist (review note on #461).
+    assert "OVERRIDES the ARTIFACTS workflow instruction" in suffix
+    assert "launch_backend" in suffix
 
 
 def test_cloud_workspace_does_not_create_anton_md(tmp_path, monkeypatch):
--- a/tests/test_cloud_turn_session.py
+++ b/tests/test_cloud_turn_session.py
@@ -102,6 +102,12 @@ def test_cloud_prompt_carries_artifact_delivery_guidance(tmp_path, monkeypatch):
         output_dir=str(tmp_path),
     )
     assert CLOUD_ARTIFACT_DELIVERY_GUIDANCE.strip() in prompt
+    # Tripwire: the override sentence targets the base ARTIFACTS step-4 text.
+    # If the durable fix lands (web-specific assembly REMOVES that base
+    # instruction), this assertion fails on purpose — the override sentence in
+    # CLOUD_ARTIFACT_DELIVERY_GUIDANCE is then stale and must be cleaned up
+    # rather than left referencing an instruction that no longer exists.
+    assert "include the primary file's path" in prompt
     # The load-bearing sentences, pinned individually so a rewrite that drops
     # one is caught even if the identity assertion above is loosened later.
     assert "Live Artifacts" in suffix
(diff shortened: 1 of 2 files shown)


alecantu7 (commented):
## Self-review — adversarial pass at `dd423252`

Reviewed my own diff before requesting eyes. **One finding, fixed in this head**, plus a measured sharpening of the delivery note in the description.

### Found: the test proved wiring, not delivery

The first version asserted `cfg.system_prompt_context.suffix == CLOUD_ARTIFACT_DELIVERY_GUIDANCE` — that the field is *set*. That is not what the PR claims. The claim is that the guidance reaches the model.

Mutation-tested both directions:

| mutation | before | now |
|---|---|---|
| remove `suffix=CLOUD_ARTIFACT_DELIVERY_GUIDANCE` in `session.py` | ❌ fails (good) | ❌ fails |
| make `prompt_builder` stop appending `suffix` (`if False:`) — **guidance never reaches the model** | ✅ **all 62 passed** | ❌ fails |

That second row is the finding. A refactor that renames, reorders or drops the `suffix` append would have left this PR's test green while the fix silently stopped working — which is precisely the failure ENG-2421 documents (*"the ENG-2044 card fix shipped and the failure continued unchanged"*), reproduced inside the guard meant to prevent it.

Fixed by assembling the real prompt and requiring the guidance inside it. Verified independently before writing the assertion: the assembled prompt grows by **1068 chars** and contains the constant verbatim.

### Survived refutation — no action

- **`suffix` is genuinely consumed** — `prompt_builder.py:200-202` appends it. Not an unreachable field like ENG-1459's `surface=web`; I checked the consumer rather than trusting the field.
- **Nothing after the guidance contradicts it.** Three blocks append after the suffix (timestamp note, `memory_context`, `workspace_context`). I checked `build_workspace_discovery_context`: it emits entry **names** only (`"Project root: data/, report.xlsx"`), never absolute paths, and excludes hidden entries deliberately — so the model is not shown `/mnt/` paths right after being told never to emit them. That was my main worry and it is unfounded.
- **The base prompt carries no competing delivery language** — 0 occurrences of "Live Artifacts", "/mnt/", "127.0.0.1", "location on disk", "absolute path".
- Uses cowork-server's existing `suffix=` mechanism rather than a new one.
- Full suite: 3,132 passed / 31 skipped.

### Sharpening the delivery note in the description

The description says the pins are bumped by hand. Measured, it is worse than that and QA needs the specifics:

```
prod    pod image -> minds-anton-scratchpad:development-head-303bf7c6…
                     = anton 303bf7c6, 2026-09-02, 234 commits behind staging
stg/dev pod image -> …095c17ae…  = anton 095c17ae, 2026-09-01
```

`095c17ae` is **not an ancestor** of `303bf7c6` — prod and staging pods run **divergent** anton builds. So "validate it on staging" is not a valid path to confidence about prod for this change, and the QA Step 0 has to name the digest under test rather than the environment.

This also bears on the ticket rather than the PR: ENG-2421's *"re-measured on a later window: zero users complaining"* would, run today, measure a web population whose pod carries neither half of the anton fix.

pnewsam (approved):
No blocking code defect found. `pytest tests/test_cloud_turn_session.py -q` passed 62 tests, `git diff --check` is clean, and current CI is green.

This is a suitable interim mitigation, but the follow-up should replace prompt-only delivery with a structured artifact contract:

- Anton returns an artifact identity or handle, not a filesystem or loopback URL.
- Cowork finalizes or publishes that artifact and renders the card and download action.
- Web-specific prompt assembly removes the base instruction to include a primary file path instead of relying on a later contradictory suffix.
- The renderer keeps its local-link neutralizer as the final safety net.

The current test proves that the suffix reaches the model. It cannot prove model compliance, and the assembled cloud prompt still contains the earlier universal instruction to include the primary file path. Validate this on staging and complete ENG-2421's later-window remeasurement before closing the ticket.

The rollout note also needs a correction: the current staging branch automatically builds and consumes `minds-anton-scratchpad:staging` after merge. Production still needs its worker-image pin or release path updated.

alecantu7:
@pnewsam — acted on both notes:

1. **The contradicting base instruction — addressed in 2a5f91ec** (within the suffix mechanism): you're right that the assembled cloud prompt still carried ARTIFACTS workflow step 4 ("include the primary file's path … clickable/openable in a plain CLI") and the fullstack "prefer the `launch_backend` url" — on the pod those are exactly the dead path and the loopback URL from ENG-2421. The guidance block now **overrides both by name** ("This OVERRIDES the ARTIFACTS workflow instruction to include the primary file's path…"), so the model sees a resolved deployment exception rather than two conflicting rules; the test pins the override sentences. Your fuller recommendation — web-specific prompt assembly that *removes* the base instruction, and the structured artifact contract (identity/handle from anton, card+download from cowork, renderer as last net) — is the right durable shape and bigger than this PR; it's being proposed as its own follow-up ticket on ENG-2421 rather than bolted on here.

2. **Rollout note — thanks for the ENG-2129 correction**, which improves the delivery story: staging's pod image now builds automatically on the staging publish, so staging web picks this up post-merge without a manual step. Production still needs its worker-image pin/release path updated, and the ticket's re-measurement stays gated on that — it's Step 0 of the QA plan so the re-measure can't run against a pod that doesn't carry this.

Since 2a5f91ec landed after your approval: the delta is the override sentences in the constant + their test pins (62/62, CI green at head).

tests/test_cloud_turn_session.py
```diff
@@ -79,6 +79,49 @@ def test_scratchpad_uses_local_factory_and_is_workspace_bound(tmp_path, monkeypa
...
+        proactive_dashboards=False,
+        output_dir=str(tmp_path),
+    )
+    assert CLOUD_ARTIFACT_DELIVERY_GUIDANCE.strip() in prompt
+    # Tripwire: the override sentence targets the base ARTIFACTS step-4 text.
+    # If the durable fix lands (web-specific assembly REMOVES that base
+    # instruction), this assertion fails on purpose — the override sentence in
+    # CLOUD_ARTIFACT_DELIVERY_GUIDANCE is then stale and must be cleaned up
+    # rather than left referencing an instruction that no longer exists.
+    assert "include the primary file's path" in prompt
```
pnewsam: [Minor] Make this tripwire inspect the base prompt before the suffix is appended. `CLOUD_ARTIFACT_DELIVERY_GUIDANCE` itself contains this exact phrase, so removing the instruction from `ARTIFACTS_PROMPT` does not fail the assertion. I simulated that removal in an isolated copy and this test still passed. Split the assembled prompt at the suffix, or otherwise assert against the pre-suffix/base section, so the durable web-specific assembly forces cleanup of the stale override.

pnewsam (approved):
Re-review of 78999dae. The cloud guidance now names and overrides both conflicting CLI instructions. This is acceptable as the provisional mitigation. The inline test issue is non-blocking and belongs with the durable follow-up.

Validation: the full 62-test cloud-turn suite passed locally. Current CI is green.

PR #462, opened 2026-09-09
[ENG-1278] Mark the boundary where a forced continuation supersedes its answer

tino097:
https://linear.app/mindsdb/issue/ENG-1278/a-forced-continuation-appends-into-the-same-answer-bubble-the-client

## What changed

- gives the verifier's continuation notice its own `continuation` phase, so a client can find the boundary; `analyzing` cannot carry it, since six unrelated sites emit that phase and only free text tells them apart
- exempts it from the pod's 250ms progress rate limit, which could drop it silently while the replacement text still arrived and appended
- announces `handback` from `_stream_handback_diagnosis`, so a hand-back diagnosis is not mistaken for the replacement answer, and documents the phase in the wire contract cowork-server now depends on for correctness
- asks the continuation for a reply that stands on its own, since "do not repeat work already done" and replace-on-boundary cannot both hold
- gates the hand-back marker on a continuation actually being pending, so a turn that never continued is unchanged on the wire and in the persisted event log
- clears the CLI's own text buffer on the boundary, so the instruction above is true on that surface too rather than printing the draft and its restatement back to back

## Why this matters

The verifier legitimately continues an unfinished task, and the user then reads the answer twice. The client half of the fix needs a signal only anton can send, and the instruction anton injects has to agree with what the client does with the reply.

## Notes for the reviewer

- **The prompt change is a behaviour change, and the deliberate one.** With consumers replacing at the boundary, a model that obeyed the old increment-shaped guard would write only its new findings and the user would see those alone, losing the earlier ones. "Do not repeat" now governs the tool work.
- **The CLI change follows from it, not from the ticket.** The ticket scoped the CLI out, but once the instruction claims the reply replaces the previous one, the surface that appends has to honour it. A hand-back deliberately keeps the buffer, because it explains the answer rather than replacing it.
- **The hand-back marker is gated so turns that never continued are unchanged.** The round cap and the spend ceiling hand back on ordinary turns; emitting there would add a wire event and a persisted event row for a turn with no continuation. Two tests pin the gate in both directions.
- **The truncation-failure notice announces a hand-back too.** It is an explanation reached with a boundary possibly pending, so without it a consumer takes it for the replacement.
- **No new pod event kind, so `scratchpad-controller` needs no change.** Both phases ride the existing `progress` kind, whose payload the controller relays with every key intact. A new kind would have been dropped by `_event_line_to_reply`.
- **Neither phase gets a `PHASE_LABELS` entry in the CLI**, for the reason that dict's own comment gives: the branch they take returns before the label fallback.
- **Merge and deploy this last.** cowork-server must already carry its paired PR, or an older `PHASE_LABELS` leaks the raw phase into the thinking line.

## How to test

1. `uv run pytest tests/test_verifier_failsafe.py tests/test_cloud_turn_entrypoint.py tests/test_chat_ui.py`
2. `uv run pytest`
3. With the paired PRs deployed, force a continuation and confirm one coherent answer live and after reload.

## Ships with

- mindsdb/cowork#957, the renderer half.
- mindsdb/cowork-server#505, which turns these phases into SSE events and owns persistence.
- Merge order: cowork, then cowork-server, then anton.

## Checks

| Check | Observed result |
| --- | --- |
| `uv run pytest` | 3140 passed, 31 skipped |

### DIFF
--- a/anton/chat_ui.py
+++ b/anton/chat_ui.py
@@ -416,10 +416,10 @@ class StreamDisplay:
         if not self._active:
             return
 
-        # `continuation` renders identically here: the CLI keeps one text
-        # buffer, so it has no answer bubble to replace, and the raw phase
-        # would otherwise leak into the status line.
-        if phase in ("analyzing", "continuation"):
+        # `continuation` and `handback` render identically here: the CLI keeps
+        # one text buffer, so it has no answer bubble to replace or preserve,
+        # and the raw phase would otherwise leak into the status line.
+        if phase in ("analyzing", "continuation", "handback"):
             self._line1_fun = random.choice(ANALYZING_MESSAGES)  # noqa: S311
             self._line2_status = "Composing response..."
             self._line3_peek = ""
--- a/anton/chat_ui.py
+++ b/anton/chat_ui.py
@@ -416,9 +416,14 @@ class StreamDisplay:
         if not self._active:
             return
 
-        # `continuation` and `handback` render identically here: the CLI keeps
-        # one text buffer, so it has no answer bubble to replace or preserve,
-        # and the raw phase would otherwise leak into the status line.
+        # A forced continuation is told its reply replaces the previous one, so
+        # the buffer that would otherwise print both has to let the earlier
+        # answer go. `finish` prints whatever remains as the final answer.
+        if phase == "continuation":
+            self._pending = ""
+
+        # These three render identically: the status line says the model is
+        # composing, and the raw phase would otherwise leak into it.
         if phase in ("analyzing", "continuation", "handback"):
             self._line1_fun = random.choice(ANALYZING_MESSAGES)  # noqa: S311
             self._line2_status = "Composing response..."
--- a/anton/chat_ui.py
+++ b/anton/chat_ui.py
@@ -416,7 +416,10 @@ class StreamDisplay:
         if not self._active:
             return
 
-        if phase == "analyzing":
+        # `continuation` renders identically here: the CLI keeps one text
+        # buffer, so it has no answer bubble to replace, and the raw phase
+        # would otherwise leak into the status line.
+        if phase in ("analyzing", "continuation"):
             self._line1_fun = random.choice(ANALYZING_MESSAGES)  # noqa: S311
             self._line2_status = "Composing response..."
             self._line3_peek = ""
--- a/anton/cloud_turn/__main__.py
+++ b/anton/cloud_turn/__main__.py
@@ -261,9 +261,13 @@ async def stream_turn(raw_line: str, emit, session_builder=None) -> None:
                     seen_tool_progress.add(event.id)
                 # Step-creating/closing phases and the first tool_progress per
                 # id must never be dropped (they open/close renderer steps);
-                # the rest is rate-limited.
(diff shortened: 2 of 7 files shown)


alecantu7 (commented):
Reviewed at `315fa8a0` alongside the two siblings, and traced the marker from here all the way to the browser.

**Reusing `progress` phases instead of adding a ninth event type was the right call**, and it's what keeps this at three repos rather than four. I went in expecting the classic version of this bug — instrument the in-process path, ship something the web population never sees — and it isn't here. The controller relays `progress` by kind and passes every other field through untouched, so `phase: "continuation"` reaches cowork-server without a controller change. I checked that rather than assuming it.

The part I'd have gotten wrong myself is the rate-limit exemption. Both phases carry meaning that a dropped frame destroys, and in opposite directions — losing `continuation` reinstates the duplicated answer, losing `handback` lets the explanation pass as the replacement and the real answer disappears from the transcript. Removing them from `always` reds two tests, including `test_the_handback_marker_outlives_a_progress_flood_too`, so the guard is real and not decorative.

Gating the handback marker on `continuation > 0` is also right: the round cap and spend ceiling hand back on perfectly ordinary turns, and emitting there would put a wire event and a persisted row on a turn that never continued. Both directions are pinned.

Two small things, neither blocking:

The `chat_ui.py` change fixes the CLI's version of this, which is real but slightly wider than ENG-1278 — the ticket scopes the CLI's concatenation problem to ENG-901. It's small and clearly correct, so I'd keep it; just worth a line in the body so it isn't a surprise to whoever reads the diff against the ticket.

The contract docstring is now the best description of this event vocabulary anywhere in the three repos — the note about why `continuation` and `handback` are exempt for *different* reasons is the kind of thing that stops the next person "simplifying" the exemption list. Worth knowing that cowork-server's `stream_formatter.py` docstring is the one place that's now out of step: it lists `response.answer_reset` but not `response.answer_restore`. Flagged on #505.

109 tests pass in the touched files. No findings against this one.

alecantu7 (approved):
Approving `315fa8a0`. No findings — I traced the marker from here to the browser rather than reviewing the diff alone.

Reusing `progress` phases instead of adding a ninth event type is what keeps this at three repos, and it's the ticket's own advice. The controller relays `progress` by kind and passes every other field through untouched, so `phase: "continuation"` reaches cowork-server with no controller change — I checked that rather than assuming it, since the missing fourth PR is exactly the shape this class of bug usually takes.

The rate-limit exemption is the part I'd have gotten wrong. Both phases carry meaning a dropped frame destroys, in opposite directions: losing `continuation` reinstates the duplicated answer, losing `handback` lets the explanation pass as the replacement and the real answer vanishes from the transcript. Removing them from `always` reds two tests. Gating the handback marker on `continuation > 0` is right too — the round cap and spend ceiling hand back on ordinary turns, and both directions are pinned.

No merge-order constraint on this one; mindsdb/cowork#957 then mindsdb/cowork-server#505 have theirs.

Minor, non-blocking: the `chat_ui.py` fix is real but slightly wider than ENG-1278, which scopes the CLI's concatenation problem to ENG-901. Worth a line in the body so it isn't a surprise to whoever reads the diff against the ticket.

PR #466, opened 2026-09-10
[ENG-2422] Tell a web turn where credentials go instead of asking for them in chat

tino097:
https://linear.app/mindsdb/issue/ENG-2422/a-web-user-cannot-create-a-connector-credential-at-all-so-the-only

## What changed

- stops asking the user for credentials in chat, and forbids storing a pasted
  one
- tells a cloud turn where credentials go, since nothing on this surface could
- closes the two solicitations the first pass missed, and tests the property
- points the referral at the entry that exists, and states the desktop parity
- keeps this change to the web turn and moves the rule into the pod suffix
- makes the rule outrank what renders after it, not only what came before: the
  remembered-rules snapshot is appended after the suffix, so an org rule asking
  for an API key was never countermanded by a guard claiming precedence over
  earlier text alone
- covers the channels that instruct from outside the prompt: `recall_skill` is
  allowlisted and the skill tree is the tenant's, so a user-authored skill body
  saying to ask for a key arrives as tool output, outside a guard bounded by
  "in this prompt"

The fifth commit narrows the first four; the last two answer review findings. The net diff against `staging` is two
files, `anton/cloud_turn/session.py` and `tests/test_cloud_turn_session.py`:
the shared base prompt and the connector tool descriptions are byte-identical
to base, so the CLI and the desktop are untouched. Reviewing the net diff is
easier than reading the commits in order.

## Notes for the reviewer

- **The whole rule lives in the suffix, on purpose.** With the shared prompt
  reverted, nothing else on this surface forbids asking for a credential, tells
  the user to rotate one that arrives anyway, or forbids storing it.
- **The no-storing clause is load-bearing.** `memorize` and
  `create_skill_draft` are both allowlisted in the pod, memory writes are
  applied org-side and replayed on later turns, and the relay's scrubber
  matches four shapes that a GitHub PAT, a WordPress application password and
  an SMTP password all miss. Without it, "get the secret out of the transcript"
  reads as "store it somewhere else", which would turn one exposed trace into
  permanent exposure.
- **The suffix countermands the shared prompt in words, and cannot do it by
  position.** The base prompt still lists credentials among the things to ask
  the user for. The suffix does not render last: `ChatSystemPromptBuilder.build`
  appends the relevance-filtered memory snapshot after it, and a recalled skill
  body never passes through the prompt at all. So the text claims precedence
  over every channel explicitly rather than relying on placement. A test
  asserts the shared invitation is still present, so if it is ever removed this
  override gets revisited rather than lingering.
- **No cache effect.** The suffix is two module constants joined at import, so
  it is byte-identical every turn. It lengthens the cached block, which is
  written once and then read at about 0.1x, and adds neither invalidator from
  the prompt-cache investigation. Note for anyone reasoning about placement
  here: the ordering above is not a caching argument. anton places no
  `cache_control`, and the gateway wraps the whole system string in one block
  with a single breakpoint at its end, so position inside `system` has no cache
  consequence either way.
- **Known cost: the desktop solicitation stays live.** `anton/tools.py` still
  says to ask for credentials in chat on the console-less path, which is Cowork
  desktop, and ENG-1849 observed five users pasting live credentials there in a
  single week. Deliberately out of scope here; it wants its own change.
- **The referral names the sidebar label, not "Connectors".** `Connectors` also
  titles the Code page whose non-Electron branch saves personal tokens through
  `/connectors/connections/validate-and-save`, a route with no org-mode
  handling, into a vault no pod turn reads. Pointing a web user there would
  reproduce this dead end one step later. A rename in cowork has no test on
  this side to catch it.
- **It names the desktop app deliberately.** An earlier revision forbade that,
  on the grounds that a credential added there is unreachable from a web turn.
  The ticket asks for parity stated plainly and the web UI already points at
  the desktop app in two places, so silence had the agent contradict the
  screen. The caveat the UI omits rides along in the prompt.
- **No `surface` read.** `build_cloud_chat_session` has one production caller,
  the headless entrypoint, so the suffix is web-only by construction.
  Branching on `surface` would break the relay's stated invariant that the
  `trace` block is observability only.
- **Acceptance criteria are partly unmet, by design.** Criteria 3 and 4 are
  met. Criterion 1 is met only for the connectors auth actually supports.
  Criterion 2 needs transcript-side redaction and criterion 5 is now ENG-2661,
  which has to close before the ticket can. This PR should not close it.
- **Found while reviewing, not fixed here:**
  `/connectors/connections/save` and `/validate-and-save` have no `org_mode`
  branch although five siblings in the same module do, and live callers exist
  in the shell-agnostic renderer tree. An org web user can store a credential
  nothing can read. Worth its own ticket.

## How to test

1. Run the suites below; the prompt-level assertions are the behaviour.
2. To see the referral as the model does, build a cloud session and render its
   system prompt: the suffix names `Connect Apps and Data`, names no connector,
   carries the desktop referral with its caveat, and forbids storing a value.
3. Confirm nothing changed off the web path: `git diff staging..HEAD -- anton/core/llm/prompts.py anton/tools.py` is empty.

## Checks

| Check | Observed result |
| --- | --- |
| `uv run --group dev pytest tests/ --ignore=tests/e2e` | 3098 passed, 31 skipped |
| `uv run --group dev pytest tests/e2e/` | 40 passed |

Both re-run at the head commit. Every new assertion was also confirmed to
invert at the base revision, so they fail on the original bug rather than
passing vacuously.


ZoranPandovski (approved):
Please just fix conflicts and minimize the comments

PR #470, opened 2026-09-11
ENG-2423: ground the agent in what Cowork is and where it runs

alecantu7:
Part 2 of 2 for ENG-2423. Sibling: **cowork-server** — the gate half. Neither blocks the other; see *Merge order* below.

## What and why

Anton has no grounded knowledge of its own product, so it answers questions about Cowork from base-model priors — where the AI assistant with a desktop app is ChatGPT.

Three reported cases:

* A user asking *"how do i install this on my pc"* was told to install the **ChatGPT desktop app**, with a `chatgpt.com/download` link. Same user, *"how do i unstall you on my computer"* → the ChatGPT app again.
* A prospect evaluating Cowork as the agent platform for their own SaaS was told *"I can't reliably identify a current public product called MindsHub CoWork"*, then handed `docs.mindsdb.com` — the wrong product line.
* A user could not find the Hermes harness switcher. It is web-only, he was on desktop, and nothing in the system knew that, so anton troubleshot his logs as if Hermes were a third-party Nous Research binary.

## The change

Split by cost, so the always-on prompt does not grow with the facts.

**A `PRODUCT` sub-block in `RUNTIME IDENTITY`** — a handful of lines beside the two ENG-1638 already renders. What the product is, that it is not another vendor's assistant, and **which surface this conversation is on**. ENG-1638 solved this exact class (agent confidently wrong about itself) with a small structured block plus "answer from these lines and nothing else"; this is the same shape for a third question.

The surface comes from the session's already-validated `surface`, so it costs no new plumbing — and an unreported surface **omits** the claim rather than guessing, matching the serving-model rule (an absent fact reads as unknown; a wrong one reads as certain).

**A `cowork-product` built-in skill** for everything that needs detail: download URL, per-OS install and uninstall steps, the installed app name, and the desktop-vs-web feature differences. Always-on cost is one description line; the body loads on `recall_skill`, which the prompt block points at. Same pointer-plus-skill split `BACKEND_GENERATION_PROMPT` already uses.

### Why the skill lives here and not in cowork-server

Three homes exist and only this one works:

| home | read-only | desktop | web | stale per-user copies |
| :-- | :-- | :-- | :-- | :-- |
| `cowork-server` harness `skills/` | yes | ✅ | ❌ in-process path only, org turns never take it | no |
| `cowork-server/cowork/skills_builtin/` | **no — seeded copy** | ✅ | ✅ | **yes** |
| `anton/core/memory/builtin_skills/` | yes | ✅ | ✅ | no |

`skills_builtin/` re-seeds "without touching ones the user edited or deleted" (its own comment), so a corrected fact would never reach anyone who edited or deleted their copy — the bug again, slower. The price of this home is delivery: a fact correction ships on an anton `main` merge plus a pod image rebuild, not same-day.

## Every fact was read out of the code that implements it

Not written from memory. The tests assert the ones that would silently rot:

* Installed app name is **`MindsHub Cowork`** — `productName` in cowork's `package.json`. This is the exact string the ChatGPT answer got wrong.
* Installers ship for macOS, Windows **and Linux** (`build-installers.yml`: `.pkg`, `.exe`, `.deb` amd64 + arm64). The repo docs still say "macOS + Windows"; that is stale.
* The **account-wide** harness toggle is **browser-only** (`SettingsView` gates the `Agent Harness` group on `host.isWeb`), while **Coding Mode is desktop-only** and is where desktop picks a harness per task. The ticket says "web-only"; the fuller truth is that desktop is not without a control, it is a different one — which is what the affected user needed to hear.

## Reachability — checked, because I have shipped an unreachable value before

* **Desktop:** cowork-server sets `skills_extra_roots` but never `builtin_root`, so anton's packaged builtins still resolve.
* **Web pod:** `_stage_skills` sets only `skills_root` — "Builtins resolve via SkillStore regardless" — and the pod passes `surface` into `ChatSessionConfig` (ENG-1459), so the block names the browser there.
* The skill is present in the built wheel (verified by building one and listing it), not just on disk.

## Testing

`3165 passed, 31 skipped` — full suite, no failures.

12 new tests. Each mutation-verified — watched failing against unfixed code before being accepted:

* PRODUCT block never rendered → 4 fail.
* Surface line emitted when the surface is unknown → 2 fail.
* Harness direction flipped in the skill → 1 fail.

Back-compat is explicitly tested: a host that supplies no product block renders exactly as before, with no dangling header.

## Merge order

Independent of the cowork-server PR; neither breaks the other alone. **Anton first is better**: the gate PR delegates product questions to this agent, so landing the grounding first means the delegated answer is already correct.

Delivery note for whoever tracks it: cowork-server pins anton to the git **`main`** branch, so merging here to `staging` does not reach cowork-server — that happens when it lands on anton `main`. Web needs a scratchpad pod image rebuild on top.

## Security check

Performed. No endpoints, authz, secrets, deserialization or persisted state touched. The additions are static prompt text and a packaged markdown file; the only dynamic input is the session's `surface`, which is validated against a closed set upstream (`_validated_surface`) and is not interpolated as anything but a dictionary key.

Closes part of ENG-2423.


alecantu7 (commented):
## Self-review — five passes, four findings, all fixed here

Reviewed this against cowork-server#519 as one change, since the gate and the agent need the same facts. Recording what was checked and what moved, so the trail is visible rather than living in a chat log.

### The design, and why it is split this way

Two layers. `identity.py`'s PRODUCT block ships **unconditionally** in every prompt — the identity assertion and the surface line. The `cowork-product` skill carries the detail and is recalled on demand. That split is deliberate: the part that must never fail costs a fixed number of tokens every turn, the bulky part costs nothing until it is needed.

The surface line is what addresses the fourth user in the ticket — the one who could not find the harness switcher. Confirmed the claim it encodes: `SettingsView.jsx:1793` gates that group on `host.isWeb`, with desktop deliberately using Coding Mode instead, so "browser only" is accurate.

### Findings fixed during review

**1. The always-on line omitted Linux while the skill listed it** (`57d6fb02`). The skill was right: `build-installers.yml` declares `linux_amd64`/`linux_arm64` with `default: true`, and neither `staging-build-deploy.yml` nor `prod-build-deploy.yml` passes those inputs, so `.deb`s ship on every release. The wrong fact was in the higher-authority place — the unconditional block, which a user meets without the model choosing to recall anything. That is this ticket's own failure mode reproduced with our fact instead of a base-model prior.

Three tests now pin it, and the second one matters more than the obvious one:

    trim Linux from the prompt line only   -> both tests FAIL
    trim Linux from BOTH sources           -> agreement test PASSES (blind)
                                              build-pinned test FAILS

So `test_platform_lists_match` alone would let a shared-wrong-list through; `test_both_name_every_platform_that_ships_an_installer` is what actually catches it. Verified by mutation rather than assumed.

**2. Brand vs. legal entity** (`32d0db6f`). The always-on line now says "made by MindsHub" rather than "MindsDB, Inc." I checked whether that creates a new confusion given the app renders `© MINDSDB, INC.` and `MANAGED BY MINDSDB` on screen — it does not: the skill carries the relationship explicitly ("The company behind MindsHub is MindsDB, Inc., which is the name on the © notice… explain the relationship plainly"). Consistent with ENG-2311, which kept the copyright on the legal entity for the same reason.

**3. A gap when the skill does not cover the question** (`32d0db6f`). The rule line now says to say so plainly and point at https://docs.mindshub.ai rather than fill the gap with a guess. Confirmed that URL matches `MINDS_DOCS_URL` in cowork.

### Fact-sheet claims verified against the code

* Browser at `https://cowork.mindshub.ai` — matches `mindsUrls.ts:10`.
* Installers `.pkg` / `.exe` / `.deb` — exactly the three in `build-installers.yml`.
* Harness toggle browser-only — `SettingsView.jsx:1793`.
* Linux builds unsigned, macOS/Windows signed — matches the workflow's own comment.

### One thing left, cosmetic

The comment on the brand decision cites "129 MindsHub to 24 MindsDB". I could not reproduce those figures from an obvious scope — counting all of `src/renderer` gives 985:219. The ratio direction is the same either way so the argument stands, but the number is not re-derivable by the next reader; worth naming the scope or dropping the digits.

Suite green: 3168 passed / 31 skipped. Ready for human review.

anton/core/memory/builtin_skills/cowork-product/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+  A user who wants to update manually can reinstall from the download page.
+- The browser version updates when the page is reloaded; there is nothing for
+  the user to install.
+
+SURFACES — what differs between desktop and browser
+This is the section to check before telling a user where to click. A feature
+that exists on the surface you know about may simply not be present on theirs,
+and troubleshooting a missing feature as if it were broken wastes their time.
+
+- Choosing the agent (Anton or Hermes), account-wide: **browser only**, under
```
tino097: Worth re-checking this section before merge. cowork-server merged #506 ("Remove hermes harness", ENG-2608) to staging on 09-14, deleting the whole `cowork/harnesses/hermes_harness/` tree. The cowork UI on staging still renders the Agent Harness group (`SettingsView.jsx:1794`) and the Hermes composer option (`Composer.jsx:368`), so the UI and the server now disagree.

That leaves three claims in this file stating a removed harness as verbatim fact: this one, line 40, and the Coding Mode line at 78. The file's own rule ("if the app and this file disagree, the app is right") does not resolve a UI-versus-server disagreement.

Not an authoring mistake, the merge came after this branch. But this is the ticket's own fourth user case, and the fact sheet is the highest-authority place to carry a wrong answer.

anton/core/llm/identity.py
```diff
@@ -136,6 +136,81 @@ def serving_model_lines(
...
+#: Deliberately a handful of lines, not a fact sheet. Everything that varies by
+#: platform or needs detail (install steps, editions, which features are
+#: web-only) lives in the ``cowork-product`` built-in skill, which the last line
+#: points at — the same pointer-plus-skill split ``BACKEND_GENERATION_PROMPT``
+#: already uses, so the always-on cost stays fixed as the facts grow.
+_PRODUCT_HEADER = "PRODUCT (what you are — not what model is serving you):"
+
+_PRODUCT_LINES = [
+    # "MindsHub", not "MindsDB, Inc.", deliberately. The user-visible strings in
+    # the app run 129 MindsHub to 24 MindsDB, and the journey that reaches this
```
tino097: Still here from the self-review note. The 129:24 count is not re-derivable from an obvious scope (all of `src/renderer` gives 985:219), and the argument does not need the digits. Naming the scope or dropping the numbers would leave the next reader something they can check.

tino097 (approved):
LGTM

PR #471, opened 2026-09-11
ENG-2420: never let an unreplayable tool-call id reach conversation history

alecantu7:
Fixes [ENG-2420](https://linear.app/mindsdb/issue/ENG-2420). Paired with cowork-server PR (link below) — **independent, no merge order required**.

## The bug

A `ToolCall` built with an empty `id` is appended to the turn's result, written into `session.history`, and serialised back out on the next request as `"call_id": ""`. The provider then rejects the **whole request** — `400 Invalid 'input[N].call_id': empty string` — on that turn and on every later turn of the conversation. From the user's side the thread is simply dead, with an error naming an internal field they cannot act on; scratchpad state, datasources and accumulated context are all stranded, and nothing tells them a new conversation is the only way out.

Nothing self-heals it: a 400 is not transient so it is never retried, and the only trimming paths run on a caught `ContextOverflowError` or on pressure computed from a *successful* response's usage — a 400 produces neither.

## What the investigation added to the ticket

Everything below was confirmed **by execution**, not by reading.

**Three trigger shapes, not one.** The ticket describes only the first; the other two never touch the `:1610` buffer path, so a guard written for that path alone would have missed them.

| shape | result |
|---|---|
| `function_call_arguments.delta`/`.done` with no preceding `output_item.added` | `id=''`, `name=''` |
| `output_item.added` **does** arrive, carrying a blank `call_id` | `id=''`, `name='scratchpad'` |
| `output_item.added` arrives with an item type other than `function_call` | `id=''`, `name=''` |

**Four construction sites, not one** — `_stream_via_responses` (the filed one), `_parse_response_object`, the chat-completions stream (**this is the transport the gateway population runs**), and chat-completions non-streaming (the SDK types `id` as a required `str`, which `""` satisfies). Anthropic has no empty-id *buffer* path — a missing `content_block_start` appends nothing — but writes `block.id` straight through, so the guard belongs on the **value**.

## The fix

1. **`ensure_replayable_tool_call`** guards all six `ToolCall` construction sites and **salvages, never drops**: an unusable id is replaced with a minted one, and an unusable *name* is replaced with `UNNAMED_REPLAYED_TOOL` so the call fails as an ordinary `Tool ... not found` tool_result the model recovers from inside the turn. Dropping was the first design and was worse — only the three non-streaming paths call `raise_on_empty_response`, so a dropped sole call ended a streaming turn silently. Returning unconditionally is also what keeps this layer, the read-time repair and the two belts agreeing about the same two fields; they disagreed in the first revision and a stored `{"id": "call_x", "name": ""}` block slipped through all three.
2. **`repair_replayed_tool_ids`** heals conversations already poisoned on disk, at `ChatSessionConfig.initial_history` — the one ingress desktop, the cloud pod and the CLI all share.
3. **A belt** in the session's tool loop and the scratchpad boot loop filters the response once, before either history or dispatch reads it, so a `tool_use` is never dropped while still being dispatched.

### Why the repair is count-preserving

This is the part that is easy to get wrong, and my first two drafts both did. `last_compaction.covered_through` is a **positional** count into `initial_history` that the host maps onto its own message list (cowork-server's `_persist_history_compaction` does `tail_start + covered - 1`). Dropping a message here would move a saved compaction cutoff onto the wrong message and silently drop or duplicate real history on the next turn — a worse bug than this one, and one that would be blamed on compaction.

The obvious workaround (keep the message, swap in a placeholder) is *also* wrong: it inflates `is_user_turn` and, in exactly the shape being repaired, leaves two consecutive user messages — which `_seed_history` and `_validate_history_for_provider` both treat as a provider hazard. So blocks are re-keyed in place, the tool reply keeps a `tool_result` block, and message count never changes. Each of those three constraints has its own test.

## Seam checks (convention 11)

- **Running what CI runs** — `staging` has moved since this opened, so the branch and the merge are no longer identical. Re-verified on the **merged** tree: merge is clean and the merged suite is green (see the comments below for the run at each head). Re-run it again if `staging` moves further.
- **Whose invariant** — ENG-664's compaction mapping (explicitly preserved + tested) and ENG-1808's row validation (not modified; a sibling function added in the paired PR).
- **Storage on write** — a turn whose rows are unreplayable is now persisted text-only rather than poisoning the conversation.
- **How it ships** — cowork-server pins anton from git `main`, so this reaches users only once it lands on anton `main`, not on `staging`.
- **Helper names** — renamed my own `drop_unreplayable_tool_rows` → `reject_unreplayable_tool_rows` in the paired PR, because it refuses the whole set rather than dropping rows individually.
- **Stale descriptions** — the `_turn_history` module docstring was updated in the paired PR; the ticket description has been rewritten to match what actually shipped.

## Security pass

Done, covering the changed lines. The new log lines carry only the unusable id and the tool **name** — never `tc.input` or tool-result content, either of which can hold user credentials; rejections in the paired PR log the offending field's *type*, never its value. No new endpoints, no auth surface, no deserialization, no path handling. The minted ids are `uuid4().hex` and are per-call handles with no security role.

## Tests

39 new tests in `tests/test_eng2420_empty_tool_call_id.py`. Full suite: **3191 passed, 30 skipped**, no regressions.

Mutation-checked rather than assumed:

| mutation | tests that fail |
|---|---|
| reader guard becomes a pass-through | 8 |
| repair drops messages instead of preserving count | 6 |
| repair rewrites the tool reply to plain text | 2 |

## Self-review trail

An adversarial pass ran **before** this was committed, against the fix design rather than the diff, and rejected two earlier versions of the repair (the two failure modes described above) and one earlier version of the persistence change (which would have imported the pod's size budgets onto desktop). Those findings are already addressed here rather than left as review comments. One of my own new tests was also wrong — it assumed the pod sanitizer would drop an oversize turn when it shrinks the result first — and was corrected.

## Not verified

The ticket's origin evidence (5 occurrences, one user, desktop, 2026-09-01) was **not** independently re-checked: Langfuse was unreachable during this work. Every mechanism claim above is independent of it; the claims about frequency and population are not.

**Residual risk:** a repaired call whose name was also empty replays as `unavailable_tool_call`. anton rebuilds its tool list every round, so replaying a call to a tool not currently offered is already an ordinary shape — but that was not confirmed against a live provider. For an already-poisoned history the alternative is a guaranteed 400, so it is strictly better either way.

### DIFF
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -73,6 +73,22 @@ _INJECTED_HELPER_NAMES = frozenset(
 _INJECTED_HELPERS: dict = {}
 
 
+def _drop_unreplayable_calls(response) -> None:
+    """Strip calls that cannot be replayed in history (ENG-2420)."""
+    from anton.core.llm.provider import usable_call_id
+
+    import logging as _logging
+
+    calls = getattr(response, "tool_calls", None) or []
+    keep = [tc for tc in calls if usable_call_id(tc.id) and tc.name]
+    if len(keep) != len(calls):
+        _logging.getLogger(__name__).error(
+            "ENG-2420: dropped %d unreplayable tool call(s) during scratchpad boot.",
+            len(calls) - len(keep),
+        )
+        response.tool_calls = keep
+
+
 def _inject_helper(name: str, fn) -> None:
     """Expose a helper to scratchpad code, and remember the object we injected.
 
@@ -651,6 +667,14 @@ if _scratchpad_model:
                     max_tokens=max_tokens,
                 )
 
+                if not response.tool_calls:
+                    return response.content
+
+                # Same ENG-2420 belt as the session's tool loop. Blast radius
+                # here is only this boot loop's own `messages` list — nothing
+                # is persisted to a user conversation — but an unreplayable id
+                # still 400s every remaining round of the boot.
+                _drop_unreplayable_calls(response)
                 if not response.tool_calls:
                     return response.content
 
--- a/anton/core/backends/scratchpad_boot.py
+++ b/anton/core/backends/scratchpad_boot.py
@@ -73,20 +73,20 @@ _INJECTED_HELPER_NAMES = frozenset(
 _INJECTED_HELPERS: dict = {}
 
 
-def _drop_unreplayable_calls(response) -> None:
-    """Strip calls that cannot be replayed in history (ENG-2420)."""
-    from anton.core.llm.provider import usable_call_id
-
+def _ensure_replayable_calls(response) -> None:
+    """Make every call on `response` replayable in history (ENG-2420)."""
     import logging as _logging
 
+    from anton.core.llm.provider import ensure_replayable_tool_call, usable_call_id
+
     calls = getattr(response, "tool_calls", None) or []
-    keep = [tc for tc in calls if usable_call_id(tc.id) and tc.name]
-    if len(keep) != len(calls):
+    broken = [tc for tc in calls if not (usable_call_id(tc.id) and usable_call_id(tc.name))]
+    if broken:
         _logging.getLogger(__name__).error(
-            "ENG-2420: dropped %d unreplayable tool call(s) during scratchpad boot.",
-            len(calls) - len(keep),
+            "ENG-2420: %d unreplayable tool call(s) during scratchpad boot.", len(broken),
         )
-        response.tool_calls = keep
+        for tc in broken:
+            ensure_replayable_tool_call(tc)
 
 
 def _inject_helper(name: str, fn) -> None:
@@ -674,9 +674,7 @@ if _scratchpad_model:
                 # here is only this boot loop's own `messages` list — nothing
                 # is persisted to a user conversation — but an unreplayable id
                 # still 400s every remaining round of the boot.
-                _drop_unreplayable_calls(response)
-                if not response.tool_calls:
-                    return response.content
+                _ensure_replayable_calls(response)
 
                 # Build assistant message with text + tool_use blocks
                 assistant_content = []
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -385,6 +385,12 @@ class AnthropicProvider(LLMProvider):
                                 "id": block.id,
                                 "name": block.name,
                                 "json_parts": [],
+                                # Same record-don't-re-derive rule as both
+                                # OpenAI readers. Nothing updates id/name after
+                                # this on the Anthropic path, so the two are
+                                # equivalent here today — the flag keeps them
+                                # equivalent if that ever changes.
+                                "started": False,
                             }
                             # Gated like both OpenAI readers, which this path
                             # did not match. Unguarded, a blank id opened a UI
@@ -393,6 +399,7 @@ class AnthropicProvider(LLMProvider):
                             # call's id, which is now the MINTED one
                             # (review: pnewsam on #471).
                             if block.id and block.name:
+                                blocks[idx]["started"] = True
                                 yield StreamToolUseStart(id=block.id, name=block.name)
                         elif block.type in ("thinking", "redacted_thinking"):
                             # Adaptive thinking (triggered by output_config.effort,
@@ -435,10 +442,10 @@ class AnthropicProvider(LLMProvider):
                                     parse_error=parse_error, repaired=repaired,
                                 )
                             ))
-                            # Same gate as the Start above, and the ORIGINAL id
-                            # rather than the minted one — an End with no Start
-                            # is a worse event stream than neither.
-                            if info["id"] and info["name"]:
+                            # Gated on whether a Start was emitted, and on the
+                            # ORIGINAL id rather than the minted one — an End
+                            # with no Start is a worse event stream than neither.
+                            if info.get("started"):
                                 yield StreamToolUseEnd(id=info["id"])
 
                     elif event.type == "message_delta":
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -31,6 +31,7 @@ from .provider import (
     retry_after_seconds,
     compute_context_pressure,
     origin_is_known_third_party,
+    replayable_tool_call,
     wallet_denial_code,
     raise_on_empty_response,
 )
@@ -280,9 +281,16 @@ class AnthropicProvider(LLMProvider):
             if block.type == "text":
                 content_text += block.text
             elif block.type == "tool_use":
-                tool_calls.append(
+                # Anthropic's reader has no empty-id BUFFER path — a missing
+                # `content_block_start` appends nothing at all — but `block.id`
+                # is written straight through, so a relay that sends a blank id
+                # produces the identical poisoned history (ENG-2420). The guard
+                # belongs on the VALUE, not on the missing-start case.
+                call = replayable_tool_call(
                     ToolCall(id=block.id, name=block.name, input=block.input)
                 )
+                if call is not None:
+                    tool_calls.append(call)
 
         # The SDK hands back an already-parsed `input`, so unlike the streaming
         # path there is no raw JSON to check: `stop_reason` is the only evidence
@@ -416,12 +424,16 @@ class AnthropicProvider(LLMProvider):
                             # the session what the body was missing. See
                             # `safe_parse_tool_input`.
                             parsed_input, parse_error, repaired = safe_parse_tool_input(raw_json)
-                            tool_calls.append(
+                            call = replayable_tool_call(
                                 ToolCall(
                                     id=info["id"], name=info["name"], input=parsed_input,
                                     parse_error=parse_error, repaired=repaired,
                                 )
                             )
+                            if call is not None:
+                                tool_calls.append(call)
+                            # Original id, not a minted one — see the matching
+                            # note in the OpenAI chat-completions reader.
                             yield StreamToolUseEnd(id=info["id"])
 
                     elif event.type == "message_delta":
--- a/anton/core/llm/anthropic.py
+++ b/anton/core/llm/anthropic.py
@@ -31,7 +31,7 @@ from .provider import (
     retry_after_seconds,
     compute_context_pressure,
     origin_is_known_third_party,
-    replayable_tool_call,
+    ensure_replayable_tool_call,
     wallet_denial_code,
     raise_on_empty_response,
 )
@@ -286,11 +286,9 @@ class AnthropicProvider(LLMProvider):
                 # is written straight through, so a relay that sends a blank id
                 # produces the identical poisoned history (ENG-2420). The guard
                 # belongs on the VALUE, not on the missing-start case.
-                call = replayable_tool_call(
+                tool_calls.append(ensure_replayable_tool_call(
                     ToolCall(id=block.id, name=block.name, input=block.input)
-                )
-                if call is not None:
-                    tool_calls.append(call)
+                ))
 
         # The SDK hands back an already-parsed `input`, so unlike the streaming
         # path there is no raw JSON to check: `stop_reason` is the only evidence
@@ -388,7 +386,14 @@ class AnthropicProvider(LLMProvider):
                                 "name": block.name,
                                 "json_parts": [],
                             }
-                            yield StreamToolUseStart(id=block.id, name=block.name)
+                            # Gated like both OpenAI readers, which this path
+                            # did not match. Unguarded, a blank id opened a UI
+                            # step keyed "" that nothing could retire: the
+                            # marker that actually closes a step carries the
+                            # call's id, which is now the MINTED one
+                            # (review: pnewsam on #471).
+                            if block.id and block.name:
+                                yield StreamToolUseStart(id=block.id, name=block.name)
                         elif block.type in ("thinking", "redacted_thinking"):
                             # Adaptive thinking (triggered by output_config.effort,
                             # set above when self._reasoning_effort is configured)
@@ -424,17 +429,17 @@ class AnthropicProvider(LLMProvider):
                             # the session what the body was missing. See
                             # `safe_parse_tool_input`.
                             parsed_input, parse_error, repaired = safe_parse_tool_input(raw_json)
-                            call = replayable_tool_call(
+                            tool_calls.append(ensure_replayable_tool_call(
                                 ToolCall(
                                     id=info["id"], name=info["name"], input=parsed_input,
                                     parse_error=parse_error, repaired=repaired,
                                 )
-                            )
-                            if call is not None:
-                                tool_calls.append(call)
-                            # Original id, not a minted one — see the matching
-                            # note in the OpenAI chat-completions reader.
-                            yield StreamToolUseEnd(id=info["id"])
+                            ))
+                            # Same gate as the Start above, and the ORIGINAL id
+                            # rather than the minted one — an End with no Start
+                            # is a worse event stream than neither.
+                            if info["id"] and info["name"]:
+                                yield StreamToolUseEnd(id=info["id"])
 
                     elif event.type == "message_delta":
                         stop_reason = event.delta.stop_reason
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -1283,8 +1283,18 @@ class OpenAIProvider(LLMProvider):
                                 if tc_delta.function and tc_delta.function.name
                                 else "",
                                 "args_parts": [],
+                                # Whether a Start was actually emitted for this
+                                # call. Recorded rather than re-derived at the
+                                # End: id and name can BOTH be filled in by a
+                                # later chunk (the branch just below), so
+                                # re-testing them there would answer "are they
+                                # set now", not "did we announce this call" —
+                                # and emit an End for a step the consumer never
+                                # opened (ENG-2420).
+                                "started": False,
                             }
                             if tc_state[idx]["id"] and tc_state[idx]["name"]:
+                                tc_state[idx]["started"] = True
                                 yield StreamToolUseStart(
                                     id=tc_state[idx]["id"],
                                     name=tc_state[idx]["name"],
@@ -1398,12 +1408,12 @@ class OpenAIProvider(LLMProvider):
                 id=info["id"], name=info["name"], input=parsed,
                 parse_error=parse_error, repaired=repaired,
             )))
-            # Gated on the same condition as the StreamToolUseStart above, and
-            # on the ORIGINAL id rather than a minted one: a call that never
-            # announced a Start must not emit an End, or the consumer is left
-            # with a step it cannot retire (review: pnewsam on #471, reported
-            # against the Anthropic reader — this is the same shape).
-            if info["id"] and info["name"]:
+            # Gated on whether a Start was actually emitted, and carrying the
+            # ORIGINAL id rather than a minted one: a call that never announced
+            # a Start must not emit an End, or the consumer is left with a step
+            # it cannot retire (review: pnewsam on #471, reported against the
+            # Anthropic reader — this is the same shape).
+            if info["started"]:
                 yield StreamToolUseEnd(id=info["id"])
 
         # Missing finish_reason is ambiguous: it's a genuine truncation only when
@@ -1608,8 +1618,17 @@ class OpenAIProvider(LLMProvider):
                         idx = event.output_index
                         call_id = getattr(item, "call_id", "") or getattr(item, "id", "")
                         name = getattr(item, "name", "") or ""
-                        fc_state[idx] = {"call_id": call_id, "name": name, "args_parts": []}
+                        # `started` for the same reason as the chat-completions
+                        # reader: an `output_item.added` carrying a good
+                        # `call_id` but a blank name emits no Start, and gating
+                        # the End on the id alone then closed a step that was
+                        # never opened (ENG-2420).
+                        fc_state[idx] = {
+                            "call_id": call_id, "name": name,
+                            "args_parts": [], "started": False,
+                        }
                         if call_id and name:
+                            fc_state[idx]["started"] = True
                             yield StreamToolUseStart(id=call_id, name=name)
 
                 # Function-call argument deltas
@@ -1619,7 +1638,10 @@ class OpenAIProvider(LLMProvider):
                     info = fc_state.get(idx)
                     if info is None:
                         # output_item.added didn't surface this call yet — buffer
-                        info = {"call_id": "", "name": "", "args_parts": []}
+                        info = {
+                            "call_id": "", "name": "", "args_parts": [],
+                            "started": False,
+                        }
                         fc_state[idx] = info
                     info["args_parts"].append(delta)
                     if info["call_id"]:
@@ -1648,7 +1670,7 @@ class OpenAIProvider(LLMProvider):
                             parse_error=parse_error, repaired=repaired,
                         )
                     ))
-                    if info["call_id"]:
+                    if info["started"]:
                         yield StreamToolUseEnd(id=info["call_id"])
 
                 # Final completion event carries the resolved Response object
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -42,7 +42,7 @@ from .provider import (
     origin_is_known_third_party,
     wallet_denial_code,
     raise_on_empty_response,
-    replayable_tool_call,
+    ensure_replayable_tool_call,
 )
 
 logger = logging.getLogger(__name__)
@@ -1130,7 +1130,7 @@ class OpenAIProvider(LLMProvider):
                 parsed_input, parse_error, repaired = safe_parse_tool_input(tc.function.arguments or "")
                 # The SDK types `id` as a required `str`, which an empty string
                 # satisfies — a non-conforming server can still send one (ENG-2420).
-                call = replayable_tool_call(
+                tool_calls.append(ensure_replayable_tool_call(
                     ToolCall(
                         id=tc.id,
                         name=tc.function.name,
@@ -1138,9 +1138,7 @@ class OpenAIProvider(LLMProvider):
                         parse_error=parse_error,
                         repaired=repaired,
                     )
-                )
-                if call is not None:
-                    tool_calls.append(call)
+                ))
 
         raise_on_empty_response(
             content=content_text, tool_calls=tool_calls,
@@ -1396,16 +1394,17 @@ class OpenAIProvider(LLMProvider):
             # `info["id"]` is seeded `tc_delta.id or ""` and only updated when a
             # later chunk carries one, so a provider that never sends an id
             # leaves it empty here (ENG-2420).
-            call = replayable_tool_call(ToolCall(
+            tool_calls.append(ensure_replayable_tool_call(ToolCall(
                 id=info["id"], name=info["name"], input=parsed,
                 parse_error=parse_error, repaired=repaired,
-            ))
-            if call is not None:
-                tool_calls.append(call)
-            # Yielded on the ORIGINAL id, not a minted one: no StreamToolUseStart
-            # was emitted for a call that had no id at the time, and an End with
-            # no Start is a worse event stream than no pair at all.
-            yield StreamToolUseEnd(id=info["id"])
+            )))
+            # Gated on the same condition as the StreamToolUseStart above, and
+            # on the ORIGINAL id rather than a minted one: a call that never
+            # announced a Start must not emit an End, or the consumer is left
+            # with a step it cannot retire (review: pnewsam on #471, reported
+            # against the Anthropic reader — this is the same shape).
+            if info["id"] and info["name"]:
+                yield StreamToolUseEnd(id=info["id"])
 
         # Missing finish_reason is ambiguous: it's a genuine truncation only when
         # the stream produced NOTHING (empty + no terminal marker). A stream that
@@ -1643,14 +1642,12 @@ class OpenAIProvider(LLMProvider):
                     # on a truthy id; this append was not, which is how an
                     # id-less call reached history and 400'd every later
                     # request in the conversation (ENG-2420).
-                    call = replayable_tool_call(
+                    tool_calls.append(ensure_replayable_tool_call(
                         ToolCall(
                             id=info["call_id"], name=info["name"], input=parsed,
                             parse_error=parse_error, repaired=repaired,
                         )
-                    )
-                    if call is not None:
-                        tool_calls.append(call)
+                    ))
                     if info["call_id"]:
                         yield StreamToolUseEnd(id=info["call_id"])
 
@@ -1788,12 +1785,10 @@ def _parse_response_object(response, model: str) -> LLMResponse:
             # model deliberately sent with no arguments.
             parsed, parse_error, repaired = safe_parse_tool_input(args_str)
             # Same `or ""` fallback as the streaming reader, same consequence.
-            call = replayable_tool_call(ToolCall(
+            tool_calls.append(ensure_replayable_tool_call(ToolCall(
                 id=call_id, name=name, input=parsed,
                 parse_error=parse_error, repaired=repaired,
-            ))
-            if call is not None:
-                tool_calls.append(call)
+            )))
         # Other item types (web_search_call, reasoning, etc.) are skipped —
         # the model's output_text already incorporates their effects.
 
--- a/anton/core/llm/openai.py
+++ b/anton/core/llm/openai.py
@@ -42,6 +42,7 @@ from .provider import (
     origin_is_known_third_party,
     wallet_denial_code,
     raise_on_empty_response,
+    replayable_tool_call,
 )
 
 logger = logging.getLogger(__name__)
@@ -1127,7 +1128,9 @@ class OpenAIProvider(LLMProvider):
                 # Both flags ride on the ToolCall for the session to act
                 # on. See `safe_parse_tool_input`.
                 parsed_input, parse_error, repaired = safe_parse_tool_input(tc.function.arguments or "")
-                tool_calls.append(
+                # The SDK types `id` as a required `str`, which an empty string
+                # satisfies — a non-conforming server can still send one (ENG-2420).
+                call = replayable_tool_call(
                     ToolCall(
                         id=tc.id,
                         name=tc.function.name,
@@ -1136,6 +1139,8 @@ class OpenAIProvider(LLMProvider):
                         repaired=repaired,
                     )
                 )
+                if call is not None:
+                    tool_calls.append(call)
 
         raise_on_empty_response(
             content=content_text, tool_calls=tool_calls,
@@ -1388,10 +1393,18 @@ class OpenAIProvider(LLMProvider):
             info = tc_state[idx]
             raw_json = "".join(info["args_parts"])
             parsed, parse_error, repaired = safe_parse_tool_input(raw_json)
-            tool_calls.append(ToolCall(
+            # `info["id"]` is seeded `tc_delta.id or ""` and only updated when a
+            # later chunk carries one, so a provider that never sends an id
+            # leaves it empty here (ENG-2420).
+            call = replayable_tool_call(ToolCall(
                 id=info["id"], name=info["name"], input=parsed,
                 parse_error=parse_error, repaired=repaired,
             ))
+            if call is not None:
+                tool_calls.append(call)
+            # Yielded on the ORIGINAL id, not a minted one: no StreamToolUseStart
+            # was emitted for a call that had no id at the time, and an End with
+            # no Start is a worse event stream than no pair at all.
             yield StreamToolUseEnd(id=info["id"])
 
         # Missing finish_reason is ambiguous: it's a genuine truncation only when
@@ -1626,12 +1639,18 @@ class OpenAIProvider(LLMProvider):
                     # a body cut mid-JSON must not raise out of this generator,
                     # and the flags are what let the session refuse the call.
                     parsed, parse_error, repaired = safe_parse_tool_input(raw_json)
-                    tool_calls.append(
+                    # The three yields around this append were already guarded
+                    # on a truthy id; this append was not, which is how an
+                    # id-less call reached history and 400'd every later
+                    # request in the conversation (ENG-2420).
+                    call = replayable_tool_call(
                         ToolCall(
                             id=info["call_id"], name=info["name"], input=parsed,
                             parse_error=parse_error, repaired=repaired,
                         )
                     )
+                    if call is not None:
+                        tool_calls.append(call)
                     if info["call_id"]:
                         yield StreamToolUseEnd(id=info["call_id"])
 
@@ -1768,10 +1787,13 @@ def _parse_response_object(response, model: str) -> LLMResponse:
             # into `{}`: an empty dict cannot be told apart from a call the
             # model deliberately sent with no arguments.
             parsed, parse_error, repaired = safe_parse_tool_input(args_str)
-            tool_calls.append(ToolCall(
+            # Same `or ""` fallback as the streaming reader, same consequence.
+            call = replayable_tool_call(ToolCall(
                 id=call_id, name=name, input=parsed,
                 parse_error=parse_error, repaired=repaired,
             ))
+            if call is not None:
+                tool_calls.append(call)
         # Other item types (web_search_call, reasoning, etc.) are skipped —
         # the model's output_text already incorporates their effects.
 
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -508,6 +508,14 @@ def replayable_tool_call(tc: ToolCall) -> ToolCall | None:
       cannot be replayed costs the whole conversation.
 
     Callers append the RESULT, never `tc`, and treat None as "no such call".
+
+    The minted id is random (`call_anton_*`) because it is minted ONCE, at
+    generation, and then persisted with the rest of this turn's rows. The
+    read-time repair in `repair_replayed_tool_ids` uses a different prefix
+    (`call_repaired_*`) and a DETERMINISTIC, content-derived id, because it
+    re-runs on every load and is never written back — a random id there would
+    change the replayed prefix every turn and miss the prompt cache. The two
+    prefixes also make it obvious in a log which layer fixed a given call.
     """
     import logging as _logging
     import uuid as _uuid
--- a/anton/core/llm/provider.py
+++ b/anton/core/llm/provider.py
@@ -455,6 +455,90 @@ def safe_parse_tool_input(raw_json: str) -> tuple[dict, str | None, bool]:
     return parsed, None, False
 
 
+#: Name given to a replayed tool call the provider never named (ENG-2420).
+#: History items must carry one — the Responses API declares ``name`` required
+#: — and a name outside the CURRENT tool list already replays fine: anton
+#: rebuilds its tool list every round (``_build_tools()`` in the session's tool
+#: loop, so `recall_skill` can add tools mid-turn), which means a call to a
+#: tool that is no longer offered is an ordinary, working shape in history.
+UNNAMED_REPLAYED_TOOL = "unavailable_tool_call"
+
+
+def usable_call_id(value: object) -> bool:
+    """True when `value` can be replayed as a tool-call handle.
+
+    Non-empty *string*, not merely truthy: a dict or list id is valid JSON and
+    passes a truthiness check, then fails much later and much less legibly.
+    Both provider APIs use string ids, so the strictness costs nothing. Same
+    test cowork-server's `sanitize_turn_history_rows` applies to pod rows.
+    """
+    return isinstance(value, str) and bool(value)
+
+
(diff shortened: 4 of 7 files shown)


alecantu7:
## Self-review pass 2 — one finding, fixed in `3d9b37c3`

A second review pass raised that "the repair is persisted, so the cure is permanent". **It isn't**, and checking that turned up a real defect.

`_persist_history()` only writes when `config.history_store` is set — and cowork-server leaves it commented out (`harness.py:1114`, `# history_store=history_store,`). The DB path doesn't cover it either: the harness persists `session.history[seed_len:]`, i.e. *this turn's* messages, so the repaired **seed** never reaches the `messages` table. The repair therefore re-runs on every load.

That makes the `uuid4` mint a real cost rather than a detail:

```
load 1 id: call_repaired_2f9c9217c6e4478d8ed3ab36f2159d21
load 2 id: call_repaired_5fc1513cd6854b259c5b6b6b06b96a88
byte-identical across loads: False
```

A repaired conversation replayed a different `call_id` every turn, so the history prefix was no longer byte-stable and it missed the prompt cache on its **entire prefix, every turn, forever** — for exactly the conversations being healed, and against an explicit design goal of the host that builds that prefix (`_stamp_message`: *"the history prefix stays byte-stable across turns (cache-safe)"*).

Still net-positive versus a dead conversation, but avoidable. The mint is now derived from the block's own name + input — keyed on **content** rather than position, so a prepended compaction summary doesn't change the ids of the tail it kept, with an occurrence counter so a conversation that made the identical call twice still gets two distinct ids.

```
load 1 ids: ['call_repaired_855aaedb6d08a3a834e78545', 'call_repaired_2b1ce7f40a38d6b1d8e45c48']
load 2 ids: ['call_repaired_855aaedb6d08a3a834e78545', 'call_repaired_2b1ce7f40a38d6b1d8e45c48']
byte-identical across loads: True
```

**Why my existing idempotence test missed it:** it re-repairs an *already repaired* list, which is a weaker property than repairing the *same original* twice. The new test does the latter. Mutation-checked — restoring the `uuid4` mint fails it.

**Also addressed from that pass:** `replayable_tool_call`'s docstring now explains the two mint prefixes. The reason is lifecycle, not forensics: the reader mint is random because it's minted once at generation and persisted with the turn; the repair mint must be deterministic because it re-runs on every load.

Suite after the fix: **3172 passed, 30 skipped**.

### Not adopted from that review

It reported anton at `3169 passed / 31 skipped`; I get `3170/30` consistently on the pre-fix commit across repeated runs. One environment-dependent skip, not a regression — flagging rather than chasing.

It also verified the two mint formats against the live provider (`call_anton_*` and `call_repaired_*` both → HTTP 200, empty id → the ticket's exact 400). I have **not** reproduced that myself — Langfuse and the prod gateway are unreachable from where I'm working — so it's recorded in the ticket as verified-in-review rather than as my own result. Note the deterministic mint keeps the same `call_repaired_` prefix and hex-only body, so that verification still applies to the new format.

anton/memory/history_store.py
```diff
@@ -38,6 +39,183 @@ def is_user_turn(message: dict) -> bool:
...
+        return isinstance(block, dict) and not usable_call_id(block.get(field))
+
+    out = [m for m in history]
+    repaired = 0
+
+    for i, msg in enumerate(out):
+        blocks = _blocks(msg)
+        if blocks is None or msg.get("role") != "assistant":
+            continue
+        broken = [b for b in blocks
```
pnewsam, line 158: **Major:** the repair heals a bad id but not a bad name, so one of the two poisoned shapes survives it. `broken` selects on `_bad(b, "id")` alone, so a stored block `{"type": "tool_use", "id": "call_x", "name": "", "input": {}}` is returned untouched.

That shape is exactly what pre-fix anton wrote whenever a provider sent a usable `call_id` with a blank name — reachable at `openai.py:1610` (`name = getattr(item, "name", "") or ""` while `call_id` is good), at `openai.py:1283` (name seeded `""` until a later chunk supplies one) and at `anthropic.py:388`. Verified end to end: `repair_replayed_tool_ids` returns `repaired=0`, and `_translate_messages_to_responses_input` then emits `{'type': 'function_call', 'call_id': 'call_x', 'name': '', 'arguments': '{}'}` onto the wire.

The three layers disagree with each other about this: `replayable_tool_call` treats an empty name as fatal and drops the call, the minting branch just below treats it as repairable and renames to `UNNAMED_REPLAYED_TOOL`, and this predicate does not consider it at all. `test_nothing_unreplayable_survives_onto_the_wire` does assert `name != ""`, but its fixture only ever reaches that state through the id path, so the assertion is vacuous for this shape.

Whether the provider actually 400s on an empty `name` I did not verify against a live endpoint — but the layers should agree either way. Fix: add `or not usable_call_id(b.get("name"))` to the `broken` predicate; the minting branch already does the rename.

anton/core/llm/provider.py
```diff
@@ -455,6 +455,98 @@ def safe_parse_tool_input(raw_json: str) -> tuple[dict, str | None, bool]:
...
+    (`call_repaired_*`) and a DETERMINISTIC, content-derived id, because it
+    re-runs on every load and is never written back — a random id there would
+    change the replayed prefix every turn and miss the prompt cache. The two
+    prefixes also make it obvious in a log which layer fixed a given call.
+    """
+    import logging as _logging
+    import uuid as _uuid
+
+    name = tc.name if isinstance(tc.name, str) else ""
+    if usable_call_id(tc.id) and name:
```
pnewsam: **Major:** this drops a call whose id is perfectly usable when only the name is missing, which is a behaviour change beyond ENG-2420 and a worse failure than the one it replaces. Verified: `replayable_tool_call(ToolCall(id="call_ok", name="", input={"a": 1})) is None`.

Reachable with a good id at three of the six sites, all streaming — `openai.py:1610`, `openai.py:1283`, `anthropic.py:388`. Pre-PR such a call was appended and dispatched, and came back as an ordinary "tool not found" `tool_result` the model recovers from within the turn. Now it vanishes; and since `raise_on_empty_response` is only called on the three non-streaming paths (`anthropic.py:302`, `openai.py:1145`, `openai.py:1808`) — which this PR's own docstring notes — an unnamed call that was the turn's only call now ends the turn silently with no error at all. That is the failure mode the docstring argues against two paragraphs above, for the id case.

The fix is already half-written in this PR: when `usable_call_id(tc.id)` holds but the name does not, set `tc.name = UNNAMED_REPLAYED_TOOL` and return `tc`, matching what the read-time repair does with the same shape.

anton/core/session.py
```diff
@@ -1347,8 +1347,15 @@ def __init__(self, config: ChatSessionConfig) -> None:
...
             ]
+        # Repaired at the ingress every surface shares — desktop, the cloud pod
+        # and the CLI all arrive here — so a conversation ENG-2420 already
+        # poisoned starts replying again on its next turn instead of 400ing
+        # forever. Count-preserving on purpose: `_seed_len` below and
+        # `last_compaction`'s `covered_through` are positional, and the host
+        # maps that count onto its OWN message list.
         self._history: list[dict] = (
-            list(config.initial_history) if config.initial_history else []
+            repair_replayed_tool_ids(list(config.initial_history))[0]
```
pnewsam, line 1357: **Major:** the wiring is largely unpinned — six of the nine call sites can be reverted to their pre-PR form with the full suite (3171 tests) still green:

| site | reverted → |
|---|---|
| `openai.py:1399` chat-completions streaming | fails 1 test |
| `openai.py:1646` Responses streaming | fails 3 tests |
| `openai.py:1791` `_parse_response_object` | fails 1 test |
| `openai.py:1133` chat-completions non-streaming | **green** |
| `anthropic.py:289` non-streaming | **green** |
| `anthropic.py:427` streaming | **green** |
| `session.py:1357` this line — the ingress repair | **green** |
| `session.py:5013` session belt | **green** |
| `scratchpad_boot.py:677` scratchpad belt | **green** |

So the entire repair layer, both belts and the whole Anthropic reader rest on no test at all. `usable_call_id` is unpinned too: replacing it with a bare `bool(value)` is also green, because `test_a_non_string_id_is_not_merely_falsy_checked` passes `name=""` and therefore exits through the name branch without ever exercising the id check.

No test anywhere constructs a `ChatSession` with poisoned `initial_history`, calls `_drop_unreplayable_calls`, or exercises the boot loop's belt — confirmed by grep. The repo already has the pattern for the belts: `tests/test_verifier_truncation.py:320` pins wiring inside `agentic_loop` with `assert "damaged_tool_call_result" in called`. The Anthropic sites need only the analogue of the OpenAI reader tests that already exist here.

anton/memory/history_store.py
```diff
@@ -38,6 +39,183 @@ def is_user_turn(message: dict) -> bool:
...
+                  if isinstance(b, dict) and b.get("type") == "tool_use" and _bad(b, "id")]
+        if not broken:
+            continue
+
+        nxt = out[i + 1] if i + 1 < len(out) else None
+        reply_blocks = _blocks(nxt) if nxt is not None and nxt.get("role") == "user" else None
```
pnewsam: **Minor:** `nxt.get("role")` is evaluated before `_blocks(nxt)`, so a non-dict message directly after an assistant message with a broken `tool_use` raises out of the repair. Reproduced: `AttributeError: 'str' object has no attribute 'get'`, propagating out of `ChatSession.__init__` at `session.py:1357` and killing the session — strictly worse than the 400 being fixed.

Every other message access in this function is isinstance-guarded through `_blocks`/`_bad`, including the loop head three lines up, so this is the one asymmetric spot rather than a house style. I found no host that passes a non-dict (`initial_history` is typed `list[dict]`), so this is defence-in-depth, not a live bug. Fix: `nxt is not None and isinstance(nxt, dict) and nxt.get("role") == "user"`.

anton/core/llm/anthropic.py
```diff
@@ -416,12 +424,16 @@ async def stream(
                             # the session what the body was missing. See
                             # `safe_parse_tool_input`.
                             parsed_input, parse_error, repaired = safe_parse_tool_input(raw_json)
-                            tool_calls.append(
+                            call = replayable_tool_call(
```
pnewsam: **Minor:** on this path the id is minted *after* the stream already announced the original, so the UI step is never retired. `content_block_start` at `anthropic.py:390` yields `StreamToolUseStart(id=block.id, name=block.name)` **unguarded**, unlike both OpenAI readers, which gate on `if call_id and name` (`openai.py:1613`) and `if ... ["id"] and ... ["name"]` (`openai.py:1289`). With `block.id == ""` and a good name, the consumer opens a step keyed `""`, `StreamToolUseEnd(id=info["id"])` closes on `""`, but the dispatched call now carries `call_anton_*` — and the marker that actually retires a step, `StreamTaskProgress(phase="tool_done", id=tc.id)` at `session.py:5323`, carries the minted id.

The comment you added at `openai.py:1405` ("no `StreamToolUseStart` was emitted for a call that had no id") is true for OpenAI and false here. Fix: guard the Anthropic `Start` on `block.id and block.name` to match.

anton/core/session.py
```diff
@@ -4978,6 +5010,7 @@ async def _stream_and_handle_tools(
                         break
 
                 # Build assistant message with content blocks
+                self._drop_unreplayable_calls(llm_response)
```
pnewsam: **Minor:** no "everything was dropped" exit, unlike the scratchpad twin. If this empties `tool_calls` and `llm_response.content` is empty, both `_append_history` calls below are no-ops (empty content returns early at `session.py:1876`), history is unchanged, and the loop re-issues an identical request until `_max_tool_rounds`. `scratchpad_boot.py:678` handles exactly this with an early return. Fix: `if not llm_response.tool_calls: break` after the drop.

anton/memory/history_store.py
```diff
@@ -38,6 +39,183 @@ def is_user_turn(message: dict) -> bool:
...
+        every load and is never written back: cowork-server leaves
+        `history_store` unset, and it persists only `session.history[seed_len:]`
+        — this turn's messages — so the repaired seed never reaches the
+        `messages` table. A random id would therefore differ on every turn,
+        changing the replayed prefix each time and missing the prompt cache on
+        the WHOLE conversation, forever, for exactly the conversations being
+        healed. Byte-stability of the history prefix is an explicit design goal
+        of the host that builds it (see `_stamp_message`, "cache-safe").
+
+        Keyed on content rather than position so a compaction summary being
```
pnewsam, line 97: **Nit:** the stability claim here isn't quite right — the occurrence counter is positional in effect, so content-keying does not survive a compaction that drops a duplicate. Demonstrated: a history with two identical broken calls mints `call_repaired_02534197…` for the second; once a summary replaces the first, the same call mints `call_repaired_d35d0e11…`. The consequence is one cache miss on a prefix the compaction already invalidated, so the code is fine — it's the comment that overstates.

anton/core/backends/scratchpad_boot.py
```diff
@@ -651,6 +667,14 @@ def agentic_loop(
                     max_tokens=max_tokens,
                 )
 
+                if not response.tool_calls:
```
pnewsam, line 670: **Nit:** dead code. `_drop_unreplayable_calls` can only shrink the list, so the identical pre-existing check at line 678 subsumes this one. Delete it.

pnewsam (commented):
**Model: claude-opus-5[1m]**

The three-layer shape is right and the reader coverage matches the ticket's six sites. Two things stop me approving. The layers disagree about whether an empty *name* is unreplayable — the readers drop such a call, the repair renames it, and the repair's selection predicate never looks at it — so a stored `{"id": "call_x", "name": ""}` block passes through untouched and reaches the wire as `name: ''`. And the wiring is largely unpinned: six of the nine call sites can be reverted with the full suite still green, including all three layers past the OpenAI readers. Findings inline.

_Validation: full suite on `3d9b37c3` (3171 passed, 31 skipped); the new file alone (20 passed); a nine-site mutation matrix; direct repros through `uv run python` for each finding below. Reviewed independently twice, second pass tasked with refuting the first — both passes' claims re-verified by hand before posting._

alecantu7:
Thanks — both Major findings reproduced before I touched anything, and the first one changed the shape of the fix. Addressed in `cfb536a5`.

## The two Major behaviour findings were the same disagreement

You're right that the layers disagreed about whether an empty **name** is fatal, and reproducing it made that plain:

```
repair_replayed_tool_ids([... {"id": "call_x", "name": ""} ...]) -> repaired = 0
_translate_messages_to_responses_input(...)  -> {'type': 'function_call', 'call_id': 'call_x', 'name': '', ...}
ensure_replayable_tool_call(ToolCall(id="call_ok", name="", input={"a": 1})) -> None   # dropped a good id
```

Rather than patch each layer to agree with whichever one was right, I made all three **salvage**: mint an unusable id, rename an unusable name, never drop. `replayable_tool_call` is now `ensure_replayable_tool_call` and returns unconditionally.

I checked your premise that dropping was the worse failure before adopting it: `_find_tool` raises `ValueError("Tool X not found")`, which the dispatcher renders as an ordinary `Tool 'x' failed:` tool_result. So a renamed call recovers inside the turn exactly as it did pre-PR. Confirmed.

Salvaging also dissolves two of your other findings **by construction** rather than by another guard:

- **the belts can no longer empty `tool_calls`** — so the "re-issues an identical request until `_max_tool_rounds`" shape you described is unreachable, no `break` needed;
- **the boot loop's duplicate early return** is gone, since nothing shrinks the list any more.

## Wiring — you were right, and it was worse than the table showed

Six of nine unpinned, with the whole repair layer, both belts and all of the Anthropic reader resting on nothing. Rebuilt as a twelve-mutation matrix; every one now fails:

| mutation | result |
|---|---|
| openai chat non-streaming / chat streaming / Responses non-streaming | 1 failed each |
| openai Responses streaming | 3 failed |
| anthropic non-streaming / anthropic streaming | 1 failed each |
| session ingress repair | 1 failed |
| session belt / boot belt | 1 failed each |
| `usable_call_id` → bare truthiness | 2 failed |
| repair ignores bad names again | 2 failed |
| anthropic `Start` unguarded again | 1 failed |

Your point about `usable_call_id` was exact — the old test passed `name=""` and exited through the name branch, so the id check was never reached. It now passes a *good* name.

The new pins include a `ChatSession` constructed with poisoned `initial_history` (nothing did that before), a direct belt test, and AST assertions for both belts using the `test_verifier_truncation.py` idiom you pointed at.

## The Anthropic `Start` finding generalised

Confirmed and fixed. I also applied it to the OpenAI **chat-completions** reader, where the `End` was unguarded in the same way — same shape, same orphaned step, one line away from the code you flagged.

## Minors and nits

- **`nxt.get("role")` before isinstance** — fixed, with a test that a non-dict message after a broken call does not raise.
- **Mint-stability comment overstates** — correct, the occurrence counter is positional in effect. Reworded to claim "stable across repeated loads of the same history", not across every edit of it, and it now says what the compaction case costs.
- **Dead duplicate check in the boot loop** — gone, per above.

## One thing I did not do

You noted you didn't verify against a live endpoint whether a provider actually 400s on an empty `name`. Neither did I — the gateway isn't reachable from where I'm working. It doesn't change the fix: the layers had to agree either way, and salvaging is correct whether or not an empty name is independently fatal.

Suite: **3182 passed, 30 skipped**.

alecantu7:
## Self-audit of the review-response commit — one bug found, fixed in `83e19754`

Checked `cfb536a5` on the principle that a fix is a new change. It had introduced one.

**I wrote a comment that asserted a guarantee the code didn't have** — the same overstatement you caught in the mint-stability note, one commit later.

Fixing your Anthropic `Start` finding, I gated the `End` events on `id and name` and wrote *"a call that never announced a Start must not emit an End"*. That isn't what the code did, in two places:

- **chat-completions** fills in `id` and `name` from *later* chunks (the `else` branch at `openai.py:1293`), so a call whose name arrived in the second chunk emitted no `Start` — and then emitted an `End` anyway, because both fields were set by the time the End was evaluated.
- **the Responses reader** gated its `End` on `call_id` alone, so an `output_item.added` carrying a good `call_id` with a blank name did the same.

Both close a UI step the consumer never opened — the exact defect you reported, surviving in two paths one line away from the one I fixed. Pre-PR the `End` was ungated so these already existed; what my commit added was a comment claiming otherwise.

**Fix:** record the fact instead of re-deriving it. `started` is set when a `Start` is actually yielded, and the `End` is gated on that. Applied to all three streaming readers so they answer the same question the same way. On the Anthropic path nothing updates id/name after the start, so the flag is equivalent today and stays correct if that ever changes.

**Three new tests**, all asserting `starts == ends`: a chat-completions stream whose name arrives in the second chunk, a Responses `output_item.added` with a good id and no name, and a healthy call that must still open and close exactly one step.

**Mutation matrix, including both of my own near-misses:**

| mutation | result |
|---|---|
| chat `End` re-derives `id and name` (what `cfb536a5` shipped) | 1 failed |
| chat `End` ungated (pre-PR) | 1 failed |
| Responses `End` gated on `call_id` only (what `cfb536a5` shipped) | 1 failed |
| anthropic `End` ungated (pre-PR) | 1 failed |
| anthropic `Start` ungated (your original finding) | 1 failed |

Also swept the rest of the commit: no leftover references to the renamed helper, no mangled `append((...))` from the edit, every appended `ToolCall` in both readers routed through the guard, and the boot loop's pre-existing empty-check confirmed still ahead of the belt (which is what made removing my duplicate safe).

Suite: **3185 passed, 30 skipped**.

anton/core/session.py
pnewsam, line 1357: **Major:** the comment above this line says "desktop, the cloud pod and the CLI all arrive here", and the repair's own docstring calls it "the one ingress desktop, the pod and the CLI all share". That holds for CLI *startup* resume (`core/runtime.py:148,197`) but not for the `/resume <sid>` command, which bypasses `initial_history` entirely: `rebuild_session` constructs the `ChatSessionConfig` without it (`chat_session.py:85-110`), then `commands/session.py:124` does `new_session._history = list(history)` straight onto the attribute.

I grepped every ingress to confirm. Three sites pass `initial_history` and are healed — `cloud_turn/session.py:700`, `core/runtime.py:197`, and cowork-server's config. `commands/session.py:124` is the only direct `_history` assignment outside compaction, and it is the one a user reaches by typing `/resume` — which is precisely what someone does after a conversation stops working. So the surface most likely to meet a poisoned history is the one the repair does not cover.

Fix: `repair_replayed_tool_ids(list(history))[0]` at `commands/session.py:124`, and compute `_turn_count` from the repaired list — this line's own `_turn_count` at `session.py:1366` already reads `self._history` rather than `config.initial_history`, for the orphan-`tool_result` case where the two disagree, and the `/resume` path should match.

tests/test_eng2420_empty_tool_call_id.py
```diff
@@ -0,0 +1,577 @@
...
+        )
+        session._ensure_replayable_calls(response)
+        # Salvaged, NOT dropped: emptying the list would leave the round with
+        # no calls and no content, so both `_append_history` calls become
+        # no-ops and the loop re-issues an identical request until the round
+        # cap (review: pnewsam on #471).
+        assert len(response.tool_calls) == 1
+        assert response.tool_calls[0].id and response.tool_calls[0].name
+
+    def test_the_tool_loop_calls_the_belt(self):
```
pnewsam, line 539: **Minor:** this pin is order-insensitive, and order is the whole point of the belt. It collects called attribute names into a set, so it only asserts that `_ensure_replayable_calls` is called *somewhere* in `_stream_and_handle_tools`. Moving the call from `session.py:5019` to after `self._append_history({"role": "assistant", "content": assistant_content})` — which writes the unsanitised ids into history, the exact failure the belt exists to prevent — leaves the full suite green at 3181 passed. Verified as an eleventh mutation.

`test_the_scratchpad_boot_loop_calls_its_belt` at line 556 is weaker again: it parses the whole module, so the call could migrate to any function in the file and still pass.

Fix: assert relative position (the belt call's line precedes the `"type": "tool_use"` construction in the same source), or drive a fake `LLMResponse` carrying a bad id through the loop and assert on the resulting `session.history` block — which would pin behaviour rather than shape.

pnewsam (commented):
**Model: claude-opus-5[1m]**

Re-reviewed at `cfb536a5`. All seven prior findings are fixed, and the salvage-always redesign is a better answer than what I asked for — making `ensure_replayable_tool_call` total removes the drop/keep disagreement between the three layers at the root instead of patching each one. The wiring hole is closed: all nine sites plus the `usable_call_id` isinstance now fail at least one test when reverted, where six of nine were green before.

One new Major, found only because the redesign made me re-read the ingress claim: the repair's "the one ingress desktop, the pod and the CLI all share" is not true of the CLI's `/resume` command, which is the exact action a user takes to get back into a conversation that stopped working. Four lows besides. Not blocking merge — everything here is strictly better than `staging` — but the `/resume` one is worth a line before this lands.

### Prior findings

| finding | status |
|---|---|
| Major — reader dropped a usable-id/empty-name call | **fixed** — renamed to `UNNAMED_REPLAYED_TOOL`, call survives |
| Major — repair healed ids but not names, `name: ''` on the wire | **fixed** — names-first pass; re-ran the wire probe, now `name: 'unavailable_tool_call'` |
| Major — wiring unpinned, 6/9 sites green | **fixed** — 10/10 mutations caught |
| Minor — `nxt.get("role")` AttributeError | **fixed** — `isinstance(nxt, dict)` |
| Minor — Anthropic `StreamToolUseStart` unguarded | **fixed** — Start and End both gated |
| Minor — belt had no all-dropped exit | **fixed** — salvage makes the list length invariant, so the shape is unreachable by construction |
| Nit ×2 — `_mint` docstring, boot dead code | **fixed** |

_Validation: full suite at `cfb536a5` — 3181 passed, 31 skipped. Eleven full-suite mutation runs (the nine sites, `usable_call_id`, plus one ordering mutation). Re-ran the finding-b wire probe and the finding-c non-dict repro by hand. Independent second pass, tasked with refuting the first; every claim below re-verified directly before posting._

anton/core/llm/openai.py
```diff
@@ -1626,12 +1638,16 @@ async def _stream_via_responses(
...
+                    # id-less call reached history and 400'd every later
+                    # request in the conversation (ENG-2420).
+                    tool_calls.append(ensure_replayable_tool_call(
                         ToolCall(
                             id=info["call_id"], name=info["name"], input=parsed,
                             parse_error=parse_error, repaired=repaired,
                         )
-                    )
+                    ))
                     if info["call_id"]:
```
pnewsam, line 1651: **Minor:** this is the one reader whose `StreamToolUseEnd` was not re-gated, so it is now the odd one out. The Start for this path gates on `call_id and name` (line 1613); this End gates on `call_id` alone. A good id with a blank name therefore emits an End with no Start — probed: `Start: []`, `Delta: ['call_ok']`, `End: ['call_ok']`.

That is the same asymmetry this PR just fixed at `openai.py:1406` and `anthropic.py:441`, and the new comment at `anthropic.py:389` claims that path is now "gated like both OpenAI readers" — only one of the two. Consequence is downstream-only (`cloud_turn/__main__.py:250` emits a `tool_end` frame for an id that never opened; the CLI's handler no-ops on an unknown id), and the shape is pre-existing — but the invariant the PR now states makes it inconsistent. Fix: `if info["call_id"] and info["name"]:`.

anton/memory/history_store.py
```diff
@@ -38,6 +39,220 @@ def is_user_turn(message: dict) -> bool:
...
+                    repaired += 1
+            out[i] = {**msg, "content": new_blocks}
+            continue
+
+        minted: list[str] = []
+        for b in new_blocks:
+            if isinstance(b, dict) and b.get("type") == "tool_use" and _bad(b, "id"):
+                b["id"] = _mint(b)
+                if not isinstance(b.get("name"), str) or not b["name"]:
+                    b["name"] = UNNAMED_REPLAYED_TOOL
```
pnewsam, line 198: **Minor:** dead branch. The names-first pass added at lines 136-151 already renames every unusable `name` on every assistant `tool_use` block, so by the time this runs the condition cannot hold. Harmless, but it reads as a second authority on the same rule, which is what the names-first pass was introduced to avoid. Drop it, or say in a comment that it is a redundant assertion rather than a rule.

tests/test_eng2420_empty_tool_call_id.py
```diff
@@ -0,0 +1,577 @@
...
+    def test_a_tool_reply_is_still_not_counted_as_a_user_turn(self):
+        """Rewriting the reply to plain text inflates the session's turn count
+        and leaves two consecutive user messages in exactly this shape."""
+        poisoned = self._poisoned()
+        fixed, _ = repair_replayed_tool_ids(poisoned)
+        assert sum(1 for m in fixed if is_user_turn(m)) == sum(
+            1 for m in poisoned if is_user_turn(m)
+        )
+
+    def test_no_new_consecutive_same_role_pair(self):
```
pnewsam, line 360: **Nit:** near-tautological. The repair never writes `role` and never changes message count, so role adjacency cannot differ unless `test_message_count_is_preserved` has already failed. It costs nothing to keep, but it does not pin the property its name claims — role alternation would be threatened by a message being added or removed, which is the thing the count test covers.

anton/core/session.py
alecantu7: Confirmed and fixed in `c702fe1f`. I grepped the ingresses myself before touching it: `cloud_turn/session.py:700`, `core/runtime.py:197` and cowork-server's config all pass `initial_history` and are healed; `commands/session.py:124` was the only direct `_history` assignment outside compaction.

`restore_session` now does `repair_replayed_tool_ids(list(history))[0]`, and `_turn_count` is counted off the repaired list for the reason you gave — they disagree exactly when an unpairable `tool_result` is rewritten to text.

Two tests: one asserts `/resume` heals a poisoned history end to end (nothing reaches the wire with an empty `call_id` or `name`), the other builds an orphan `tool_result` so the two counts genuinely differ and pins that `_turn_count` matches the repaired list. Mutation-checked: reverting to the raw assignment fails 1, counting off the raw list fails 1.

Your framing is the part worth keeping — the repair claimed an ingress it did not have, and the path it missed is the one a user reaches precisely because their conversation stopped working.

tests/test_eng2420_empty_tool_call_id.py
alecantu7: Right on both counts, fixed in `c702fe1f`. A set of called names only says the belt runs somewhere, and "somewhere" includes after the history write.

Both pins now assert **position** — the belt call precedes the `tool_use` construction in the same source — and the boot pin is scoped to its enclosing function instead of the whole module. I also added the behavioural test you suggested: a poisoned `LLMResponse` driven through the belt, asserting on what a history block built from it would carry.

One thing worth flagging: my first attempt at the boot pin was itself broken. `src.index("_ensure_replayable_calls(response)")` matched the **function definition**, not the call site, so it pinned nothing. Caught it because the assertion raised rather than passed. It now matches the call statement specifically.

Mutation-checked: moving either belt after its history write fails 1 each.

anton/core/llm/openai.py
alecantu7: Already fixed, by `83e19754` — which landed about an hour after this review was written, so it was accurate against `cfb536a5`.

Both Ends now gate on a recorded `started` flag rather than re-deriving the condition:

```
openai.py  chat-completions  ->  if info["started"]:
openai.py  Responses         ->  if info["started"]:
anthropic.py                 ->  if info.get("started"):
```

I went to the flag rather than to `call_id and name` because re-deriving answers "are they set now", not "did we announce this call" — chat-completions fills both in from later chunks, so a name arriving second produced an End with no Start even with the stricter condition. That was a bug my own review-response commit shipped, and I found it auditing that commit. Three tests assert `starts == ends`; the mutation table on that commit includes gating the Responses End on `call_id` alone.

anton/memory/history_store.py
alecantu7: Agreed, dropped in `c702fe1f`. The names-first pass renames every unusable name before the id pass runs, so it was unreachable — and, as you say, a second authority on a rule the names-first pass exists to own.

Mutation-checked in the other direction, to be sure the remaining authority is real: disabling the names-first pass fails 4 tests.

tests/test_eng2420_empty_tool_call_id.py
alecantu7: Fair, dropped in `c702fe1f` rather than strengthened. Roles are never written by the repair and the count is pinned by `test_message_count_is_preserved`, so it could not fail independently of that test — it was reassurance, not a pin.

alecantu7:
## Second review round addressed — `c702fe1f`

Apologies for the delay on this one; the review landed shortly after my last comment and I only picked it up now. Replies are on each thread; summary:

| finding | outcome |
|---|---|
| **Major — `/resume` bypasses the repair** | **fixed.** Confirmed your grep independently: three ingresses pass `initial_history` and are healed, `commands/session.py:124` was the only direct `_history` assignment outside compaction. Now repaired there, with `_turn_count` counted off the repaired list. Two tests. |
| Minor — belt pin is order-insensitive | **fixed.** Both pins assert position; the boot pin is scoped to its function; added the behavioural test you suggested. |
| Minor — Responses `End` gated on `call_id` alone | **already fixed** by `83e19754`, which landed ~1h after your review was written. All three readers now gate on a recorded `started` flag. |
| Minor — dead rename branch | **dropped.** |
| Nit — role-adjacency test | **dropped**, not strengthened — it couldn't fail independently of the count test. |

**Mutation-checked, five this round:** reverting `/resume` → 1 failed; counting turns off the raw list → 1; moving the session belt after the history write → 1; moving the boot belt → 1; disabling the names-first pass → 4.

While fixing your belt-pin finding I broke a test in the same way you'd described: my first boot pin matched the function *definition* (`def _ensure_replayable_calls(response)` contains the call's substring), so it pinned nothing. It raised rather than passed, which is how I caught it. Now matched against the call statement.

## The green on this PR was four days stale, so I re-ran the real thing

Staging has moved **11 commits** since CI last ran here, into `anton/core/session.py` among others. GitHub builds the merge, so I built it locally:

- merge into current `origin/staging`: **clean**
- merged tree, full suite: **3243 passed, 36 skipped** (the branch alone was 3185/30 — staging brought 58 new tests)
- ENG-2420 file on the merged tree: **35 passed**

Worth noting the merged tree needed a dependency the branch didn't (`openpyxl`, new on staging) — a stale local venv reports that as four import errors rather than as a missing dep, which is exactly the failure mode that makes "I ran the suite" untrustworthy after a base moves.

Not merged — leaving that to a human, and flagging that this needs to reach anton `main`, not just `staging`, to heal anyone's conversation.

alecantu7:
## Audit of `c702fe1f` — no code bug, one claim corrected

Checked the newest commit the same way as the last two, since a fix is a new change.

**Code: clean.** The `/resume` diff is four lines of substance; I re-verified the two things that could go wrong in it:

- both new position pins **fail loudly** rather than passing silently if their anchor moves — the session pin raises `ValueError` on a missing anchor (test error, not a green), and the boot pin uses an explicit `assert ... is not None`;
- the behavioural belt test is **not vacuous** — making `_ensure_replayable_calls` a no-op fails 2 tests.

Plus the five-mutation matrix from the commit itself, all caught.

**What the audit did find is a claim, not a defect.** I wrote on the ticket that the repair is *"non-destructive — it never rewrites stored user history"*. That is true here on desktop and web, and **false on the CLI**:

| surface | `history_store` | effect |
|---|---|---|
| desktop / web (cowork-server harness) | unset (`harness.py:1114`) | transient, stored rows untouched |
| CLI startup resume (`core/runtime.py:197-198`) | set | repaired list **written back** next turn |
| CLI `/resume` (`commands/session.py:121`) | set | same |

`_persist_history()` writes `self._history` whenever `history_store` and `session_id` are set, and both CLI paths set them. So on the CLI the first turn after a resume overwrites the session file with minted ids, and an unpairable `tool_result` rewritten to text loses its original block.

Not introduced by this commit — CLI startup resume has done it since `3e881e4e`, and it is arguably the better outcome there since it heals permanently instead of re-repairing every load. The deterministic mint means a re-repair is byte-identical either way. But the claim was over-broad, so I have corrected it on ENG-2420 rather than leave it standing.

**Also re-checked on the merged tree** (staging is 11 commits ahead): merge clean, full suite **3243 passed / 36 skipped**, ENG-2420 file **35 passed**.

anton/memory/history_store.py
```diff
@@ -38,6 +39,218 @@ def is_user_turn(message: dict) -> bool:
...
+        new_blocks = [dict(b) if isinstance(b, dict) else b for b in blocks]
+        if not pairable:
+            # Nothing to pair with, so a minted id would be an orphan
+            # `tool_use` — which is its own 400 on Anthropic. Keep the message
+            # and the block COUNT, drop only the call.
+            for pos, b in enumerate(new_blocks):
+                if isinstance(b, dict) and b.get("type") == "tool_use" and _bad(b, "id"):
+                    new_blocks[pos] = {
+                        "type": "text",
+                        "text": "[a tool call was recorded here without an id and has been removed]",
```
pnewsam: **Minor:** this placeholder throws away the call it replaces, and on the CLI that loss is written to disk.

The tool's name and input are gone:

```
{'type':'tool_use','id':'','name':'scratchpad','input':{'code':'print(df.head())'}}
 -> {'type':'text','text':'[a tool call was recorded here without an id and has been removed]'}
```

On desktop/web this is transient — the harness leaves `history_store` unset. On the CLI both resume paths set it (`runtime.py:197-198`, `commands/session.py:121`) and `_persist_history` (`session.py:1800`) saves the whole `self._history`, so the first turn after a resume overwrites the session file and the original is gone for good. Your own audit flagged the write-back and framed it as "heals permanently" — it also erases.

One line to fix: carry the original into the placeholder, e.g. `[a tool call to 'scratchpad' was recorded without an id and could not be replayed: {"code": "print(df.head())"}]`. The model then keeps the context of what that turn was doing, and nothing is destroyed on write-back. Same point on the `tool_result` placeholder below.

anton/memory/history_store.py
```diff
@@ -38,6 +39,218 @@ def is_user_turn(message: dict) -> bool:
...
+        if blocks is None or msg.get("role") != "user":
+            continue
+        if not any(isinstance(b, dict) and b.get("type") == "tool_result"
+                   and _bad(b, "tool_use_id") for b in blocks):
+            continue
+        new_blocks = []
+        for b in blocks:
+            if isinstance(b, dict) and b.get("type") == "tool_result" and _bad(b, "tool_use_id"):
+                new_blocks.append({"type": "text",
+                                   "text": "[a tool result with no matching call was removed]"})
```
pnewsam: **Minor (same root cause as the `tool_use` placeholder above):** the result body is discarded too.

```
{'type':'tool_result','tool_use_id':'','content':'ROWS: 1,2,3 important output'}
 -> {'type':'text','text':'[a tool result with no matching call was removed]'}
```

This one costs more than the call site does — a tool result is often the only record of what the data actually said, and on the CLI it is overwritten on the next `_persist_history`. Keeping the content after the marker (`[a tool result with no matching call]\n` + the original) preserves it at no structural cost: it is still a text block, so the count-preserving and `is_user_turn` reasoning in the docstring is unchanged.

pnewsam (approved):
**Model: claude-opus-5[1m]**

Approving at `c702fe1f`. Both findings from my last round are genuinely fixed and now pinned by tests that fail when reverted — I re-ran the `/resume` and belt-position mutations myself. Two Minors left, neither unsafe to merge; the description one is worth fixing before it lands, since the body currently says the opposite of the code on the load-bearing decision.

### Prior findings

| finding | status |
|---|---|
| Major — `/resume` bypasses the repair | **fixed** — repaired at `commands/session.py:130`, `_turn_count` counted off the repaired list |
| Minor — belt pins were order-insensitive | **fixed** — both pins assert position, boot pin scoped to its function, behavioural test added |
| Minor — Responses `End` gated on `call_id` alone | **fixed** in `83e19754` — all three readers now gate on a recorded `started` |
| Minor — dead rename branch / Nit — role-adjacency test | **dropped**, agreed |

### What I dropped from my own earlier passes

- **`_discarded_tool_call_events` emitting an `End` for a minted id that never had a `Start`** — refuted. `session.py:209` emits `StreamTaskProgress(phase="tool_done"|"scratchpad_done")` and a `StreamToolResult`, never a `StreamToolUseEnd`. The CLI matches `tool_done` by name, not id (`chat_ui.py:491`), and `cloud_turn` keys off `Start`, so an unmatched id is a no-op in both.
- **Cost or correctness of running the repair on every `ChatSession` construction** — refuted. Three linear passes and one shallow list copy; a healthy history returns the caller's own dicts untouched, uncopied.

### Findings

Two inline. Plus one summary-level:

**Minor: the PR body and the ticket both still describe a design this PR does not ship.** "The fix" section 1 says an unnamed call "is **dropped**, since it cannot be dispatched anyway" — `provider.py:530` renames it to `UNNAMED_REPLAYED_TOOL` and returns it, and its own docstring argues at length why dropping was worse. The body is internally inconsistent as a result: its "Residual risk" paragraph discusses the `unavailable_tool_call` replay that "The fix" says cannot happen. The same sentence is still in ENG-2420 under **Fix → 1**, although the body claims the ticket was rewritten to match what shipped. Also stale: the helper is `ensure_replayable_tool_call`, not `replayable_tool_call`; 35 new tests, not 18; the suite is 3186/31, not 3170/30; and the "branch is at fresh `origin/staging`" seam check is 11 commits out of date. The merge is still clean and green — I checked — but the body is what a future reader will trust about the drop-vs-rename decision, and it currently contradicts it.

**Nit: `history_store.py:228-236` rewrites an unclaimed bad `tool_result` in place**, so if it precedes a good one in the same message the chat-completions translator emits `[… assistant, user, tool]` — a `role:user` between the assistant and its `tool` reply, which OpenAI rejects. I could not find a writer that produces this shape: every anton `tool_result` sources its id from a matching `tool_use.id`, so a bad one implies a bad sibling that pass 2 claims first. The input is already fatal either way. Only worth touching if you want the repair total for arbitrary corrupt input — appending rewritten text blocks at the end of the message would do it.

### Still unverified

Your residual risk stands as you stated it: replaying `unavailable_tool_call` as a `function_call` name absent from the current tool list has not been confirmed against a live provider. The reasoning is sound and the alternative for a poisoned history is a guaranteed 400, so this is an evidence limitation, not an objection.

_Validation: full suite on `c702fe1f` — **3186 passed, 31 skipped**. Merge into current `origin/staging` (11 commits ahead, touching `core/session.py`) is clean and the merged tree is green: **3239 passed, 40 skipped**. New file alone: 35 passed. Enumerated all six `ToolCall` sites (`anthropic.py:290,440`, `openai.py:1134,1407,1668,1810`) — all wrapped, none left. Traced all four history ingresses. Executed the repair against 8 handcrafted histories: count-preserving on every branch, non-mutating (deep-copy compared), idempotent, byte-identical across two loads of the same original, safe on non-dict messages. Confirmed the rename dispatches as an ordinary failed `tool_result` (`registry.py:69` → the loop's `except Exception`). Reviewed independently twice, the second pass tasked with refuting the first — it removed two of my standing findings, both re-verified by hand before posting._

anton/memory/history_store.py
alecantu7: Fixed in `05250030`. You are right that it erases, and right that the CLI write-back makes it permanent — my own audit spotted the write-back and then framed it only as "heals permanently", which was half the story.

The placeholder now carries the call:

```
[a tool call to 'scratchpad' was recorded without an id and could not be replayed: {"code": "print(df.head())"}]
```

Rendering goes through a shared `_render` that handles a string, a list of blocks, or an arbitrary body, never raises and never truncates — the value was already in history at that size, so carrying it costs nothing new. Mutation-checked: discarding the name and input fails 2 tests.

anton/memory/history_store.py
alecantu7: Fixed in the same commit, and I agree this one costs more — the result body is often the only record of what the data said.

```
[a tool result with no matching call]
ROWS: 1,2,3 important output
```

I also took your ordering nit from the review body while here, because it turned out to be the same edit: the rewritten block is now **appended after** its surviving siblings rather than replaced in place, so the chat-completions translator can no longer emit `[assistant, user, tool]`. Block count is unchanged, so the count-preserving and `is_user_turn` reasoning in the docstring still holds. That makes the repair total for the corrupt shape you could not find a writer for, rather than leaving it latent.

Mutation-checked: discarding the body fails 2, dropping `_render`'s list branch fails 1, and rewriting in place instead of appending fails 1.

Two corrections to my own work found while writing those tests — the ordering assertion first checked that no `user` follows the `tool` block, which is wrong (a user message after a tool reply is just the next turn; the rejected shape is a `user` *between* the assistant and its reply), and the list-rendering test asserted only that the strings survived, which passes even with the list branch removed because the JSON fallback contains them too. Both now pin what they claim.

alecantu7:
## All three Minors addressed — `05250030` + body/ticket edits

Thanks — and thanks in particular for dropping two of your own standing findings and saying so. That is more useful to me than the findings would have been.

### The two inline Minors: fixed

Both placeholders now carry what they replace, through a shared `_render` that handles a string, a list of blocks or an arbitrary body without raising or truncating:

```
[a tool call to 'scratchpad' was recorded without an id and could not be replayed: {"code": "print(df.head())"}]

[a tool result with no matching call]
ROWS: 1,2,3 important output
```

You were right that the CLI write-back makes this permanent, and right that my own audit had seen the write-back and framed it only as "heals permanently" — which was half of it.

**I also took the ordering nit from your review body,** because it turned out to be the same edit: the rewritten `tool_result` is now appended **after** its surviving siblings instead of replaced in place, so the chat-completions translator can no longer produce `[assistant, user, tool]`. Block count is unchanged, so the count-preserving and `is_user_turn` reasoning holds. It costs three lines and makes the repair total for a shape neither of us could find a writer for, rather than leaving it latent in a function whose whole job is corrupt input.

Mutation-checked: discarding the call's name+input fails 2, discarding the result body fails 2, dropping `_render`'s list branch fails 1, rewriting in place instead of appending fails 1.

### The summary-level Minor: you were right, and it was the worst of the three

The body described a design this PR does not ship, on exactly the decision most likely to be re-litigated later. Fixed:

- "an **unnamed** call is dropped" → the salvage description, with why dropping was worse and why returning unconditionally is what keeps the three layers agreeing
- `replayable_tool_call` → `ensure_replayable_tool_call`
- 18 tests → 39; suite figure refreshed
- the "branch is at fresh `origin/staging`" seam check now says staging has moved and points at the merged-tree runs

**ENG-2420 carries the same corrections**, including `Fix → 2`, which still claimed the repair covers "the one ingress desktop, the pod and the CLI all share" — the claim your `/resume` finding disproved. It now names `/resume` explicitly and records that unpairable blocks keep their content.

Worth naming the pattern, since it is now three rounds running: the code has held up under your mutations, and what keeps failing is **what I write about it** — a comment asserting a guarantee the code didn't have, a ticket claim that was wrong for one surface, and now a PR body contradicting the design on its load-bearing decision.

### Still unverified, unchanged

The `unavailable_tool_call` replay against a live provider — your framing is exactly right, an evidence limitation rather than an objection. Same for ENG-2420's origin evidence: Langfuse has been unreachable for every attempt, mine and both review passes, and the ticket says so rather than assuming it.

Suite at this head: **3191 passed, 30 skipped**. Not merging — and when this does land it needs to reach anton `main`, not just `staging`, or no user's conversation is healed.

PR #472, opened 2026-09-12
fix(memory): stop rules.md dropping entries on every rewrite

sohanreddyk:
`encode_rule` is read-modify-write: `get_rules()` → append → `save_rules()`, which rewrites the whole file. Two gaps in that round-trip mean the next rule Anton learns silently deletes existing ones.

**Entries above the first `## ` heading are never read.** `get_rules` starts at `current_kind = None` and gates entries on `and current_kind`, so a rule hand-written before any heading never reaches the system prompt — and is dropped from the file by the next `save_rules`.

**Entries under a non-canonical heading are dropped on write.** `get_rules` takes `kind` from whatever heading it finds, so `## Notes` yields `kind="notes"`. `save_rules` buckets into a fixed `{"always", "never", "when"}` dict and skips anything that misses. No error, no log.

Before this change, a `rules.md` containing a `## Notes` section loses that section entirely the first time Anton learns any rule:

```
# Rules

## Always
- Call progress() before long API calls

## Never
- Never use time.sleep() in scratchpad cells

## When
```

`Cortex._compact_file` already guards exactly this, defaulting an unrecognised heading to `_RULE_SECTIONS[0]`, with the comment "Entries above the first heading (a hand-edited file) are kept rather than dropped." This applies the same rule in the storage layer so the two paths can't disagree, and moves the section tuple into `hippocampus.py` so there is one definition.

Entries under an unknown heading are relocated under `## Always` — the same thing compaction already does to them. Preserved and readable, not filed under a custom heading.

Adds three round-trip tests to `tests/test_hippocampus.py`; all three fail before the change. After it, `test_hippocampus.py` + `test_cortex.py` are green (104 passed), including the existing newline-injection tests, as are the consolidator, cerebellum, ACC, desktop-memory, self-awareness, history-compaction and reconsolidator suites (139 passed).

Targeting `staging` since that's where feature PRs are landing — `dev` hasn't moved since May. Happy to retarget.

### DIFF
--- a/anton/core/memory/cortex.py
+++ b/anton/core/memory/cortex.py
@@ -34,7 +34,7 @@ from anton.core.llm.structured import (
 )
 from anton.core.memory.base import HippocampusProtocol
 from anton.core.memory.base import Engram
-from anton.core.memory.hippocampus import Hippocampus
+from anton.core.memory.hippocampus import Hippocampus, RULE_SECTIONS as _RULE_SECTIONS
 
 if TYPE_CHECKING:
     from anton.core.llm.client import LLMClient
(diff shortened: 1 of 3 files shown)


sohanreddyk:

sohanreddyk:
recheck

PR #480, opened 2026-09-16
ENG-2677: stop counting a fetched page or a search hit as a tool failure

alecantu7:
## The bug

`handle_web_fetch_fallback` and `handle_web_search_fallback` both returned a bare `str`, so `ToolOutcome.ok` was `None` and the ENG-1276 substring fallback decided the verdict by scanning the result text for `[error]`, `Task failed:`, `failed`, `timed out`, `Rejected:`.

**On success, that text is the content** — a page of up to 20,000 characters for fetch, up to 20 × 600-character snippets for search. Both are arbitrary web prose. Measured against the real `_apply_error_tracking` with five **successful HTTP 200** fetches:

| # | Page content | Streak |
| -- | -- | -- |
| 1 | "…the merger **failed** to clear regulatory review, revenue rose 12%" | 1 |
| 2 | "…the deployment **timed out** after thirty minutes" | 2 |
| 3 | "…applies when a payment **failed**, for any reason" | 3 |
| 4 | "…Apollo 13's oxygen tank **failed** en route to the Moon" | 4 |
| 5 | "…if the connection **timed out**, check your firewall" | 5 |

Nudge fires at 2; at 5 the agent is told *"The 'web_fetch' tool has failed 5 times in a row. Stop retrying this approach."* — after five successful fetches. `"failed"` is a bare substring match, so **the richer the content, the likelier it is judged a failure.**

## The fix

The verdict information already existed and was being discarded. `_fetch_once` built a `_FetchResult` carrying the real HTTP status; `_fetch_url` unwrapped it to `.text`. Search had no equivalent at all.

- `_fetch_url` now returns the whole `_FetchResult`; the handler declares `ok=True` on a 2xx.
- `_search_exa` / `_search_brave` now return a new `_SearchResult` carrying `served`; the handler declares `ok=True` only when real hits came back.

No text inspection anywhere.

## Why non-success stays `ok=None`

A 404, or "no results", is not a tool failure: the tool worked and the answer was negative. That is the tier-3 category `recall_skill`'s NO MATCH family already occupies, and ENG-2248 is explicit that `ok=False` there *"would push normal exploration into the error streak and trip the breaker on correct behaviour… If it should ever count, that needs its own ticket and its own before/after on the nudge rate."* Deferred accordingly.

## This change is purely subtractive

**No branch in either handler returns `ok=False`**, so it can only ever remove a firing:

| Scenario | Before | After |
| -- | -- | -- |
| 5 good pages, 3 contain "failed" | **nudge fires** ❌ | silent ✓ |
| Search hits whose snippets mention "failed" | **counted as failure** ❌ | `ok=True` ✓ |
| Mixed research, good pages + 404s | streak 1 | streak 0 ✓ |
| 6 × HTTP 403 | silent | silent (unchanged) |
| 5 × timeout | nudge + breaker | nudge + breaker (unchanged) |
| Provider error ("Exa search failed…") | counted | counted (unchanged) |

## Expected side effect — read this before comparing metrics

`_record_root_cause` returns early on `ok=True`, so a successful page mentioning "failed" no longer books a phantom failure. Measured on one such page: `failures=1, tiers={'unclassified': 1}` → `failures=0, tiers={}`.

**`root_cause_failures` and `root_cause_unclassified` will drop for this cohort after release.** That is noise leaving, not tool health improving.

## Tests

- `TestWebFetchVerdict` / `TestWebSearchVerdict` — the handler contracts.
- `TestWebFetchVerdictOnNonHttpFailures` — the exhausted-timeout and unresolvable-host paths, which have no HTTP status.
- `TestWebFetchDoesNotPoisonTheErrorStreak` — pins both states: what the bug looked like (streak 5, nudge, breaker) and what correct looks like (streak 0, silent). These feed `_apply_error_tracking` directly, so they document the consequence rather than guard the handler; the guard is the verdict classes. Stated in the docstring because the first draft claimed otherwise and the mutation run disproved it.
- Existing assertions updated for the new return type via a `_text()` helper; argument-validation and no-provider branches still return `str` and are covered as such.
- Both handlers added to `_VERDICT_DECLARING_HANDLERS`.

**Mutation-verified three ways:** reverting the fetch handler to a bare `str` fails 6 tests, reverting the search handler fails 4 (each including the inventory guard), and marking a served search `served=False` fails exactly 1.

**Full suite, no exclusions: 3,220 passed, 37 skipped.** The first push claimed "3,213 passed" from a run that used `--ignore` on `tests/test_web_tools_live.py` — the one file most exposed to this contract change, which has no skip guard and runs in CI unconditionally. That is why CI went red on the first commit, and the claim is corrected here.

## Scope and security

Anton-internal. Grepped `cowork-server`, `cowork` and `scratchpad-controller`: nothing outside anton calls either handler, `_fetch_url`, or the search helpers. `dispatch_tool` already normalises both `str` and `ToolOutcome` returns by design, `dispatch_tool_stream` is only reachable through it, and the single direct `handler()` call lives in the registry. `cloud_turn/session.py:757` sets `web_fetch_enabled=False`, so the pod path is untouched.

Security: no new input is trusted and no new output surface is created. `ok` is a boolean into a payload that already carried `ok`; the content returned is what was returned before; the SSRF guard and its `blocked` status are unchanged.

## Who this affects

These fallbacks register only when the provider exposes no native web capability — dormant on Anthropic BYOK, OpenAI BYOK and the mdb.ai passthrough. So: desktop users on local models and generic OpenAI-compatible endpoints. 414 `web_fetch` calls / 32 users / 30 days, plus 13 `web_search` calls.

It also makes both tools measurable for the first time: `web_fetch` has reported `ok="unknown"` on 414 of 414 calls, and 92% of those never touch the gateway, so no Langfuse traces exist either. After this, `unknown` means "did not return content", which yields a success rate.

Closes ENG-2677.


alecantu7:
## Self-review trail (CLAUDE.md convention 6)

Everything below was found and fixed **before** requesting review. Recording it so a reviewer sees the hardened version plus what was already caught, and doesn't spend time re-deriving it.

Reviewed at `f03e897d`; head is now `4b7f9c3a`. Base moved `1c02b15a` → `9a0d0252` (ENG-2666, −160 lines in `session.py`); merge is CLEAN and nothing this work depends on changed — the `_apply_error_tracking` body hash is identical across all three commits (`7fe7751140c3`) and `_LEGACY_FAILURE_MARKERS` is unchanged.

### Findings fixed in this PR

**1 · CI was red and I caused it, by running a narrowed suite.** The first push claimed "full suite green" from a run that used `--ignore=tests/test_web_tools_live.py` — the one file most exposed to the contract change, and one with *no skip guard* (`TestWebFetchLive` needs no API key by design, so it runs in CI unconditionally). Convention 11.0 verbatim: I checked the merge was current and never checked the suite was complete. Now re-run with no exclusions.

**2 · Two type errors hidden behind key gates.** Forcing the key-gated Exa/Brave tests to *run* (fake keys, stubbed HTTP) surfaced two failures invisible to CI:
- `_SearchResult` has `.text`, not `.content`, so the `_text()` helper's `hasattr` check missed and returned the object → `TypeError: argument of type '_SearchResult' is not iterable`.
- `_has_https_url_line(out)` received a `ToolOutcome` and called `.splitlines()` on it.

Both introduced by the fix commit. Inspection had already "passed" these files; execution disagreed immediately.

**3 · The tier-3 promise was enforced by wording, not by code or test.** `ok=None` does **not** mean "not counted" — `_apply_error_tracking` short-circuits only on an explicit verdict and otherwise falls through to the substring match. So a 404 avoided the streak purely because `"Fetch returned HTTP 404 for …"` happens to contain no marker. Rewording it to `"Fetch failed with HTTP 404 …"` reinstated the exact false positive this PR removes, with every test still green. `TestTierThreePromiseIsPinnedToTheRealText` now pins **both** halves against the real `_fetch_once` output.

**4 · `web_search` had the identical defect** and the original out-of-scope reason was wrong — "already reports `ok=false` via the exception path" is true of its *failures*, not its *successes*, which return up to 20 × 600-character snippets of exactly the prose the matcher scans. Brought in scope; ticket corrected.

### Two mistakes I made while fixing these, kept visible

- A bulk edit inserted a helper **between `@anthropic_only` and its class**, so the skip decorator landed on the helper and four live tests went from *skipped* to *failed* on a missing key. Restored to the exact prior shape (1 passed, 17 skipped).
- The first version of the tier-3 test **passed under mutation for the wrong reason**: my fixtures raised `ConnectError("getaddrinfo failed")` and `ConnectError("cert verify failed")`, so the assertion matched the *exception text* rather than the template it was meant to measure. Fixtures now use marker-free messages (`"connection refused"`, `"certificate verify error"`). Also, covering only `ConnectError` left `web_tools.py:360` (the non-retryable SSL return, a separate line with its own wording) free to lose its marker unnoticed — it has its own case now.

### Mutation coverage

| Mutation | Result |
|---|---|
| fetch handler → bare `str` | 5 red |
| search handler → bare `str` | 4 red |
| served search → `served=False` | 1 red |
| `200 <= s < 300` → `s >= 200` | 1 red |
| 4xx message reworded to contain "failed" | 8 red |
| `web_tools.py:359` retryable wording loses its marker | 1 red |
| `web_tools.py:360` non-retryable wording loses its marker | 1 red |
| timeout wording loses its marker | 3 red |

Full suite, no exclusions: **3231 passed, 37 skipped.**

### Deliberate, please don't file against it

`TestWebFetchDoesNotPoisonTheErrorStreak` does **not** fail when the handler regresses — it feeds `_apply_error_tracking` directly to document the consequence. Its class docstring names `TestWebFetchVerdict` as the actual guard. The split is intentional.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**  
**Model: unknown**

Reviewed PR #480 at `4b7f9c3a`. The handler contracts, downstream dispatch, failure tracking, telemetry, and current `staging` integration remain consistent. No actionable findings.

### Validation

- Exact-head web and verdict tests: **80 passed**.
- Outcome, root-cause, telemetry, and dispatch tests: **75 passed** with pre-existing AsyncMock warnings.
- An independent second review found no defects.
- The merge against current `staging` is clean.
- All current GitHub CI, build, scope, license, and security checks pass.
- The author's documented self-review findings are fixed in the current candidate.
- I did not rerun the full suite locally. The GitHub full-suite run is green.

**Risk: Low. Confidence: High.** The change is contained. It adds no dependency, persistence, or security-boundary changes. It preserves the intentional `ok=None` behavior for non-success responses.

PR #484, opened 2026-09-17
ENG-2689: Classify an oversized-image refusal as permanent, not transient

alecantu7:
## ENG-2689

A user attached an image the provider permanently refuses. anton read the refusal as a temporary endpoint outage, so it re-sent the identical request four times and then told the user to try again in a moment. The task was left unusable, and sixty seconds after the fourth failure he wrote `scrap the whole thing` and never came back.

Verbatim, identical in all four attempts:

> `400 - {'error': {'message': 'The image you provided requires 32400 patches after processing, exceeding the limit of 30000. Please resize the image and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': 'invalid_value'}}`

## What was actually wrong

The ladder in `_raise_for_status_error` already had the right branch — it just asked the wrong question. It recognised rejections about a content block's **shape** (ENG-1992) and not about its **size**, so this 400 fell through to the generic `ConnectionError`, which asserts the failure is temporary. Downstream, cowork-server's history repair is keyed on the type that branch produces, so the oversized image was never stripped: **every later turn in that conversation re-sent it and failed the same way.** A dead task, not a dead turn.

## Changes

- **`classify_content_rejection()` in `provider.py`**, beside `classify_404` and shared for the same reason — the heuristic must not drift between providers. It covers both families and builds the copy.
- **Called from both mappers.** `anthropic.py` had *no* content branch at all: the ENG-1992 fix landed only in `openai.py`, so a BYOK-Anthropic user hit this bug with none of the handling. That is fixed here rather than left as a second ticket.
- **`ContentTooLargeError(ContentValidationError)`.** Subclassing is load-bearing: every existing caller keeps behaving correctly without knowing the subtype exists. The class *name* is also the only discriminator that survives the remote/pod wire (`_scrub` emits `"<Type>: <message>"`), which is why this is a type and not a `code`.
- **Terminal in `session.py`** — raised after `_seal_dangling_tool_uses`, not with the early-raise types at the top of the handler, because the refusal can land mid-tool-round and an orphan `tool_use` would 400 the *next* turn too, converting a repairable conversation into a differently-broken one. Stamped `retry_terminal_reason="content_rejected"`.
- **The user-facing message carries the provider's own remedy sentence**, credential-scrubbed and capped at 300 chars.

## The conservative bit, on purpose

Misclassifying a 400 here is destructive: downstream it strips **every** image block from the conversation's stored history. So the size family additionally requires the provider to have mentioned an image at all — a 400 about a too-long text field must never cost the user their images. `test_an_oversized_non_image_payload_is_not_a_content_rejection` pins that direction, and removing the guard fails it.

## Verification

Full suite: **3194 passed, 37 skipped**, re-run on the tree rebased onto current `origin/staging` (CI builds the merge, not the branch).

Every new test was mutation-verified — each of these made exactly the intended test fail, and nothing else:

| Mutation | Caught by |
|---|---|
| Remove the session early-raise | all 6 in `test_content_rejection_terminal.py` |
| Remove the size branch from the classifier | 3 mapper tests |
| Remove the `"image" in message` guard | the non-image negative test |
| Move the session raise *before* the seal | the dangling-`tool_use` test |
| Revert the `anthropic.py` wiring | both Anthropic mapper tests |

## Delivery

cowork-server pins anton from the git **`main`** branch, so this reaches users only once it lands on `main` — merging to `staging` alone changes nothing for desktop or web. Companion PRs, both no-ops until then:

- mindsdb/cowork-server — the `content_too_large` code, its card copy, and the widened repair trigger
- mindsdb/cowork — the renderer branch (the ENG-1282 contract requires all three)

**Version skew, verified by execution rather than reasoning** (probed both server versions against the new exception):

| Skew | Desktop | Web |
|---|---|---|
| New server, old anton | unchanged | unchanged |
| Old server, new anton | repaired + no retries, older "already fixed" copy | generic error, no repair — **same as today**, minus the four retries |

No ordering is unsafe; the worst case is strictly better than current behaviour.

## Security

Pass covered the one new path that reaches a user: provider prose is now quoted into a user-visible message. It is run through `scrub_credentials` and capped at 300 chars, so a hostile or verbose BYOK endpoint cannot leak credentials or flood the card, and React renders it as text, not markup. No new endpoints, no authz surface, no deserialization, nothing added to logs.

## Known trade-off, inherited not introduced

The repair strips *all* images in the conversation, not just the oversized one — ENG-1992's deliberate choice, since a provider's request-relative index can't be mapped back to a stored message reliably. For a size failure that is heavier than strictly needed. Called out so it is a decision, not a surprise.


alecantu7:
## Self-review

Three findings from reviewing my own diff. Two were fixed before pushing; one has a follow-up commit (`c65d1218`) so the trail is visible.

**1. A comment asserted something I never checked.** (fixed in `c65d1218`)
`classify_content_rejection`'s docstring justified testing size before shape by claiming "a size refusal can also carry a `param` that satisfies the shape test". I went and checked both live bodies: the OpenAI patch-limit 400 carries `param="input"`, and Anthropic's dimension refusal puts its content path in the *message*, not the param. Neither satisfies the shape predicate. The ordering is still correct — as a tie-break for a body that doesn't exist yet — but the stated reason described one that does. Rewritten to say what is actually true.

**2. The scope I nearly shipped was half the bug.** The ticket described the OpenAI mapper. `anthropic.py` turned out to have no content branch at all — ENG-1992 landed only in `openai.py` — so a BYOK-Anthropic user had this bug completely untreated. Fixing one mapper would have left that, and left the next person to rediscover it. Hence the shared classifier rather than a second inline block.

**3. "Terminal" was nearly terminal in the wrong place.** My first instinct was to add `ContentValidationError` to the early-raise tuple at the top of the handler, next to `TokenLimitExceeded`. That skips `_seal_dangling_tool_uses`, so a refusal landing mid-tool-round would leave an orphan `tool_use` and 400 the *next* turn — trading a repairable conversation for a differently-broken one, on the exact ticket about conversations staying broken. Moved below the seal; `test_a_dangling_tool_use_is_sealed_before_the_turn_fails` fails if it moves back.

### Adjacent — retracted (2026-09-22)

~~The pre-existing early-raise types (`TokenLimitExceeded`, `ModelUnavailableError`, `EndpointConfigurationError`, `ProviderAuthError`) all raise *above* the seal and so have finding 3's problem today. … it is real and worth its own ticket.~~

**This was wrong — please don't file it.** Those four types do raise above `_seal_dangling_tool_uses`, but the state the seal guards against can't occur for them:

- they are raised only by LLM calls;
- no LLM call runs between the `tool_use` append and the `tool_result` append;
- a tool that raises anything (including these, from its own sub-call) is caught by the per-tool `except Exception` and turned into a `tool_result`.

My original probe put the dangling `tool_use` into history by hand, which tests the handler but not whether the state is reachable. Driving the real tool loop instead — each of the four types, both where the tool raises and where the next LLM round raises, then a second turn in the same session — produced **zero** orphans in 8 of 8 scenarios. The zero is not vacuous: the `tool_use` entered history with its `tool_result` paired, and the same detector flags a deliberately built orphan.

The same false rationale appears in this PR's comment above the `ContentValidationError` raise and in `test_a_dangling_tool_use_is_sealed_before_the_turn_fails`. The placement is harmless (merely defensive); only the stated reason is wrong.

alecantu7:
recheck

alecantu7:
## Blocking finding accepted and fixed — `199e51fb`

Reproduced it before touching anything, then again after. You were right, and the reachability gap was bigger than the ticket knew.

### Confirmed independently

```
MRO: BadRequestError -> APIStatusError -> APIError -> OpenAIError
direct mapper     -> ContentTooLargeError | ContentValidationError? True
production order  -> BadRequestError      | ContentValidationError? False
```

### The fix, and what I deliberately did *not* do

The obvious repair — call `_raise_for_status_error` from the `BadRequestError` clause — is wrong. `classify_transient` maps a 400 carrying `type: api_error` to a retryable `TransientProviderError`; I checked:

```
400 + type='invalid_request_error' -> None
400 + type='api_error'             -> TransientProviderError
```

So routing all 400s through the ladder would silently make some terminal 400s retry — a new bug, on a PR about unwanted retries. Instead `_raise_for_bad_request()` names only what it can name (context overflow, then content shape/size) and returns otherwise, so the caller re-raises the SDK error and **every other 400 stays byte-identical**. Six call sites across both providers; the anthropic twin had the same shape, as you said.

### Measured on the incident's own body, end to end

| | provider requests | what the user gets |
|---|---|---|
| before | **4** | turn ends as prose: `An unexpected error occurred: Error code: 400 - {'error': {'message': 'The image you provided requires 32400 patches…` |
| after | **1** | typed `ContentTooLargeError` carrying "Please resize the image and try again." |

Four requests is exactly the four Langfuse traces in the ticket.

**This also corrects the ticket.** ENG-2689's "second defect on current main" claimed users now see *"Server returned 400 — the LLM endpoint may be temporarily unavailable"*. That string lives in the unreachable function, so it was never what anyone got — the raw nested JSON was always the real behaviour. I've corrected the ticket.

**And ENG-1992 was never reachable either.** Its branch sits in the same function. The repair users actually received came from cowork-server independently matching the provider's message text, not from anton's type. Both families now reach the classifier, so ENG-1992's shape case also drops from 4 attempts to 1.

### On the test gap

Your diagnosis of why my tests missed it is exactly right, and I've made that failure mode explicit rather than just patching around it. `tests/test_bad_request_reaches_the_classifier.py` drives the **real provider methods** against the real SDK over a mock transport — both flavors, so all four OpenAI call sites plus both Anthropic ones — with one case running the whole chain through `LLMClient` and `ChatSession`.

Restoring the original bare `raise`:

- new join tests: **6 failed, 4 passed**
- old mapper + session tests: **98 passed**

That 98 is the finding, reproduced as a number.

Full suite 3208 passed, 37 skipped, rebased onto current `staging`.

anton/core/llm/provider.py
```diff
@@ -1156,6 +1162,133 @@ def __init__(self, message: str, *, code: str = "content_validation") -> None:
...
+            f"accept. The provider said: {_quote_provider(raw)} "
+            "That image will be removed automatically so the conversation can "
+            "continue \u2014 re-attach a smaller or lower-resolution copy if you "
+            "still need it."
+        )
+
+    if (
+        ".content[" in par
+        or par.endswith(".content")
+        or any(phrase in low for phrase in _CONTENT_SHAPE_PHRASES)
```
pnewsam: **Major:** This message-only predicate also matches unrelated enum validation errors. For example, an `invalid_request_error` with `param="reasoning_effort"` and `"Supported values are: ..."` becomes `ContentValidationError`. The session propagates that type, and cowork-server responds by removing every stored image from the conversation while leaving the real configuration error unresolved. Require a content-specific parameter or message signal before selecting this destructive repair path, and cover the real provider path with a negative regression test.

anton/core/llm/provider.py
```diff
@@ -1156,6 +1162,133 @@ def __init__(self, message: str, *, code: str = "content_validation") -> None:
...
+
+    raw = message or ""
+    low = raw.lower()
+    par = (param or "").lower()
+
+    if "image" in low and any(phrase in low for phrase in _CONTENT_SIZE_PHRASES):
+        return ContentTooLargeError(
+            "An image in this conversation is too large for the model to "
+            f"accept. The provider said: {_quote_provider(raw)} "
+            "That image will be removed automatically so the conversation can "
```
pnewsam, line 1299: **Minor:** This automatic-removal promise is only true when cowork-server handles the exception. The standalone CLI catches the propagated `ConnectionError` but does not strip image blocks from `ChatSession` history. The next turn sends the same image again. Either use host-neutral copy here or add standalone history repair before the CLI accepts another turn.

pnewsam (dismissed):
## Code Review

**Verdict: REQUEST CHANGES**
**Model: unknown**

The reported oversized-image path now stops after one request and preserves the provider remedy. The shared classifier also turns unrelated request-validation errors into destructive image-history repair. The Major inline finding blocks merge. The CLI copy issue is non-blocking by itself.

### Validation

- Relevant Anton tests — 167 passed.
- Real SDK, `LLMClient`, and `ChatSession` reproduction confirmed the unrelated-enum failure path.
- In-process and remote Anton-to-cowork-server mappings were exercised.
- `git diff --check` — passed.
- GitHub checks were green at `199e51fb`.

alecantu7:
## Both findings fixed — `43e518bb`

Thanks, the Major one is a real data-loss path and I'd reasoned my way to the guard on one branch and not the other.

### Major — confirmed, and it was worse than one repo

Reproduced before changing anything:

```
reasoning_effort   -> ContentValidationError
tool_choice        -> ContentValidationError
ENG-1992 real      -> ContentValidationError
anthropic tag      -> ContentValidationError
```

Then I checked the layer that actually performs the deletion, and **cowork-server's message-phrase fallback has the identical hole independently of this PR** — it is live on `staging` today:

```
cowork-server is_content_validation_error -> True
friendly_turn_error -> ('content_recovery', "An image earlier in this conversation couldn't be sent…  I've fixed it automatically — you can keep going.")
```

So a `reasoning_effort` typo already deletes every image in the conversation and reports success. Fixing only anton would have left that standing, since the fallback runs whenever anton is unimportable or older and the looser of the two decides. Tightened in both, with the same rule: a bare shape phrase needs corroboration — either the `param` points into a content array, or the message names a quoted content-block type. cowork-server#545 carries its half (`befb8dfe`).

After:

```
reasoning_effort   -> None          tool_choice -> None
ENG-1992 real      -> ContentValidationError
anthropic tag      -> ContentValidationError
size (ENG-2689)    -> ContentTooLargeError
```

Unquoted providers now fall through to the old generic handling. Under-matching costs worse copy; over-matching destroys user data, so that is the direction to be wrong in.

Worth naming: I wrote exactly this guard for the SIZE family — it requires the message to mention an image, and the docstring says why — then carried the inherited shape predicate forward without applying the same reasoning. Same hazard, one branch guarded.

### Minor — took the repair option, not the copy option

You offered host-neutral copy or standalone repair. The copy fix would have left ENG-2689's own `Done when` #2 false for CLI users ("the next turn is not blocked by the same content"), so I did the repair: the session strips image blocks from its own history before the terminal raise. Harmless no-op for cowork-server, which rebuilds the session each turn from its own store; for the CLI it is the difference between a dead turn and a dead conversation. Nested `tool_result` images covered too — a tool-returned screenshot is as poisonous and hides a level deeper.

### Tests, on the real provider path as asked

Negative regressions for three unrelated enums × both flavors, plus an end-to-end case asserting a config error leaves the conversation's images intact. Mutation-verified — loosening the predicate fails 7; removing the history repair fails 2; loosening cowork-server's fallback fails 3.

anton **3241 passed**, 37 skipped. cowork-server **4768 passed** (the one failure is the pre-existing `OSError: File name too long` under the long macOS tmpdir, failing identically on clean `origin/staging`).

Re-requesting review.

anton/core/session.py
```diff
@@ -2664,6 +2665,67 @@ async def _backoff_sleep(self, delay: float) -> bool:
...
+        def _strip(content):
+            nonlocal removed
+            if not isinstance(content, list):
+                return content, False
+            out, changed = [], False
+            for block in content:
+                if not isinstance(block, dict):
+                    out.append(block)
+                    continue
+                if block.get("type") == "image":
```
pnewsam: **Minor:** Remove supported `image_url` blocks here as well. `ChatSession.turn_stream()` accepts this public input format, and the OpenAI provider explicitly translates or forwards it. I reproduced an oversized-image 400 through the real SDK, provider, `LLMClient`, and session; a following text-only turn resent the same `image_url`, and history still contained it. Include `image_url` in this predicate and cover the public-ingress continuation path.

anton/core/session.py
```diff
@@ -4248,6 +4310,38 @@ async def _turn_stream_inner(
...
+                    # Raised HERE, after the seal above, rather than with the
+                    # other early-raise types at the top of this handler: the
+                    # failure can land mid-tool-round, and a dangling tool_use
+                    # left in history would 400 the NEXT turn too — turning a
+                    # repairable conversation into a differently-broken one.
+                    if isinstance(_agent_exc, ContentValidationError):
+                        # Remove the poison before giving up, so the NEXT turn
+                        # isn't refused for the same content. Without this the
+                        # turn dies and the conversation dies with it.
+                        _removed = self._strip_image_blocks_from_history()
```
pnewsam, line 4339: **Minor:** Persist the repaired history before propagating this error. The later `_persist_history()` call is skipped when this branch raises, and `close()` does not save. If a saved image becomes invalid under a stricter model, the live session removes it, but `/resume` reloads the original image and repeats the refusal. Persist when `_removed` is nonzero.

pnewsam (commented):
## Code Review

**Verdict: COMMENT**
**Model: unknown**

Both original findings are fixed at `43e518bb`. Two non-blocking gaps remain in the new history repair.

### Validation

- Focused classifier and session suite — 176 passed.
- Independent related regression batch — 199 passed.
- Real SDK, provider, `LLMClient`, and `ChatSession` reproduction covered the `image_url` continuation path.
- GitHub CI — all four workflows passed.
- `git diff --check` — passed.

alecantu7:
## Self-review of the review fix — found another false positive, `106d42c3`

Checked the fix itself rather than only the tests it passes, and the corroboration guard I shipped in `43e518bb` still let through the nastiest member of the family.

**`modalities` legitimately accepts the value `image`.** So its enum error reads `Invalid value: 'text'. Supported values are: 'image', 'audio'` — a real content-block token, in a message with nothing to do with content. It still reached the path that deletes every image in the conversation:

```
modalities naming image    -> ContentValidationError      # after 43e518bb
reasoning_effort           -> None
```

The discriminator both live dialects share: the OpenAI one carries a content-indexed `param`, the Anthropic one carries **no** `param` at all. So a param that is present and is not a content path is positive evidence the rejected thing was not content. The phrase+token route now applies only when the provider named no param:

```
modalities naming image    -> None
reasoning_effort           -> None
ENG-1992 OpenAI real       -> ContentValidationError
ENG-1992 Anthropic real    -> ContentValidationError
size (ENG-2689)            -> ContentTooLargeError
```

cowork-server#545 carries the mirror (`4574c577`); its fallback only sees a repr, so it recovers the param from the string when present and falls back to the token rule when not.

### Also verified the history repair by execution, not by reading

The other half of `43e518bb` mutates `self._history`, which is the riskiest thing in this PR. Measured rather than assumed:

| property | result |
|---|---|
| idempotent on its own output | ✅ second pass removes 0 |
| `tool_use_id` preserved through a `tool_result` rewrite | ✅ |
| string (non-list) content untouched | ✅ |
| image-free messages left as the same object | ✅ |
| repaired history passes `_validate_history_for_provider` | ✅ |

Mutation-verified: dropping the param evidence fails the two `modalities` cases; dropping cowork-server's fails its regression test.

anton **3243 passed**, 37 skipped. cowork-server **4776 passed** (one pre-existing macOS tmpdir failure, identical on clean `origin/staging`).

Worth saying plainly: this is the second time on this PR that I verified components and not the thing they combine into. The reported Major was real, my first fix for it was incomplete, and it took reviewing my own fix as a new change to find the gap.

alecantu7:
## Both non-blocking gaps fixed — `3d41084a`

Confirmed both by execution before changing anything, and both were right.

### `image_url` — and the codebase already knew the answer

`openai.py:524` explicitly forwards an inbound `image_url` from history, and this very file already treats `("image", "image_url")` as the pair at both of its other image sites (`session.py:878` and `:978`). My repair invented a narrower predicate than the convention sitting two screens away, so the OpenAI shape stayed in history and got re-sent — the exact failure the repair exists to stop. Now matches the existing idiom.

### Persistence — the repair didn't outlive the turn

Verified the mechanism: the only `_persist_history()` call is at `:4599`, my raise is at `:4341`, and `close()` (`:2312`) only reaps scratchpads and closes the provider pool. So the repair was memory-only and `/resume` reloaded the original image and repeated the refusal, exactly as you described.

Now saved when anything was removed, wrapped in try/except with a warning: this runs on the error path, where an escape would convert a handled, card-mappable failure into an unexpected one. `test_a_failing_history_store_does_not_change_the_raised_error` pins that — a store raising `OSError` must still surface `ContentTooLargeError`.

### Tests

Three added, each mutation-verified:

| Mutation | Fails |
|---|---|
| strip only the `image` shape | `test_an_openai_shaped_image_url_block_is_removed_too` |
| remove the persistence call | `test_the_repair_is_persisted_so_resume_does_not_resend_it` |
| (guard) store raises on save | error type must be unchanged |

The persistence test asserts on what a resumed session would actually load, not that a method was called.

**3246 passed**, 37 skipped, rebased onto current `staging`.

### One thing to re-check on your side

Your review validated at `43e518bb`. Since then two more commits landed: `106d42c3` fixed a false positive I found in my own guard — `modalities` legitimately accepts the value `'image'`, so its enum error named a content-block token and still reached the destructive repair — and `3d41084a` is this one. Worth a look at `106d42c3` in particular, since it changes the same predicate your Major finding was about.

anton/core/llm/provider.py
alecantu7: Confirmed and fixed — `43e518bb`, then tightened again in `106d42c3`.

Reproduced first: `reasoning_effort`, `tool_choice` and `service_tier` all returned `ContentValidationError`. I then checked the layer that performs the deletion and found cowork-server's message-phrase fallback has the same hole **independently of this PR** — live on `staging` today, so a `reasoning_effort` typo already deletes every image in a conversation and reports success. Fixing only anton would have left that standing, since that fallback runs whenever anton is unimportable or older. Mirror guard in cowork-server#545.

A bare shape phrase now needs corroboration: either the `param` points into a content array, or the message names a quoted content-block type.

`106d42c3` closes one your example didn't cover and my first guard missed: **`modalities` legitimately accepts the value `image`**, so `Supported values are: 'image', 'audio'` names a real content-block token while having nothing to do with content — and still reached the repair. The discriminator both live dialects share is that the OpenAI one carries a content-indexed `param` and the Anthropic one carries none, so a param that is present and is *not* a content path is now positive evidence against.

```
modalities naming image  -> None
reasoning_effort         -> None
ENG-1992 OpenAI real     -> ContentValidationError
ENG-1992 Anthropic real  -> ContentValidationError
size (ENG-2689)          -> ContentTooLargeError
```

Negative regressions run on the real provider path as you asked: four unrelated enums x both flavors, plus an end-to-end case asserting a config error leaves the conversation's images intact. Mutation-verified — loosening the predicate fails 7, dropping the param evidence fails the 2 `modalities` cases.

anton/core/llm/provider.py
alecantu7: Confirmed — and I took the second option rather than the copy one.

Host-neutral copy would have left ENG-2689's own `Done when` #2 false for CLI users ("after such a rejection, the next turn in the same conversation is not blocked by the same content"), so weakening the sentence would have made the message honest about a product gap instead of closing it.

`43e518bb` adds `_strip_image_blocks_from_history()`, called before the terminal raise, so the promise is true for every host rather than only the one that happens to implement it. For cowork-server it is a harmless no-op on a history about to be discarded (it rebuilds the session each turn from its own store); for the CLI's one long-lived session it is the difference between a dead turn and a dead conversation. Nested `tool_result` images are covered too — a tool-returned screenshot is as poisonous and hides a level deeper.

Your two follow-ups on that repair (`image_url`, persistence) are fixed in `3d41084a`.

Verified the mutation by execution rather than by reading, since it rewrites live history: idempotent on its own output, `tool_use_id` preserved through a `tool_result` rewrite, string and image-free content untouched, and the repaired history passes `_validate_history_for_provider`.

anton/core/session.py
alecantu7: Confirmed and fixed — `3d41084a`.

You were right that the codebase already answers this: `openai.py:524` explicitly forwards an inbound `image_url` from history, and this very file treats `("image", "image_url")` as the pair at both of its other image sites (`session.py:878`, `:978`). My repair invented a narrower predicate than the convention two screens away, so the OpenAI shape survived and got re-sent — the exact failure the repair exists to stop. Now matches the existing idiom.

`test_an_openai_shaped_image_url_block_is_removed_too` covers the public-ingress continuation path; reverting to the `image`-only predicate fails it.

anton/core/session.py
alecantu7: Confirmed and fixed — `3d41084a`.

Verified the mechanism you described: the only `_persist_history()` call is at `:4599`, this raise is at `:4341`, and `close()` (`:2312`) only reaps scratchpads and closes the provider pool. So the repair was memory-only and `/resume` reloaded the original image and repeated the refusal.

Now saved whenever anything was removed, wrapped in try/except with a warning — this runs on the error path, where an escape would convert a handled, card-mappable failure into an unexpected one. `test_a_failing_history_store_does_not_change_the_raised_error` pins that: a store raising `OSError` must still surface `ContentTooLargeError`.

`test_the_repair_is_persisted_so_resume_does_not_resend_it` asserts on what a resumed session would actually load back, not merely that a save method was called. Removing the persistence call fails it.

pnewsam (approved):
## Code Review

**Verdict: APPROVE**
**Model: unknown**

All four prior findings are fixed at `3d41084a`. The classifier now rejects unrelated named parameters, and the session repair covers both image shapes and persists before propagating the terminal error. No actionable findings.

### Validation

- Focused classifier, terminal, curated-error, and status-mapper suite — 181 passed.
- Independent provider and session regression suite — 163 passed.
- Real SDK, provider, `LLMClient`, and `ChatSession` probes confirmed one rejection, persisted repair, and a successful image-free continuation for both image shapes.
- GitHub Actions and Snyk checks passed.
- `git diff --check` — passed.

PR #486, opened 2026-09-18
ENG-2863: run the verifier live matrix only when the verifier changed

alecantu7:
> **Blocked on ENG-2892 (2026-09-21).** `verdict-eval` is red on this PR for a reason unrelated to it: the mindshub_inference release this morning made the gateway echo the alias as the response `model`, and the eval's ENG-1687 identity pin raises on every call. The fix is a separate PR against staging; **merge that first, then re-run this PR's checks.** The scope step itself ran correctly in that red run ("full", because this PR changes the eval module).

Linear: [ENG-2863](https://linear.app/mindsdb/issue/ENG-2863/the-verifier-live-eval-spends-its-full-gateway-budget-on-every)

## Why

`verdict-eval` fires real verdict calls at the production gateway on every PR that touches `anton/core/session.py`, a ~6k-line file most PRs touch. In the last 30 days it ran ~200 times; only ~26 of those diffs touched anything the verifier's judgment depends on. Since ENG-2686 the matrix is ~103 calls, so roughly **20,000 gateway calls a month guarded code that did not change**.

## What

**Two tiers, decided per PR at the job level.** `.github/scripts/verifier_eval_scope.py` parses both sides of each trigger file with `ast`, takes the line spans of the verdict class, its constants, the request builder, the transcript renderer and clipper, and the structured-output call path, and intersects the PR's `-U0` hunks with them. A body-only edit inside `_render_verify_transcript` counts; an edit anywhere in `turn_stream`, which merely *calls* the verifier, does not. Any change to the eval module, `structured.py`, or the script itself is always full.

- **full** → the matrix as today (~97 calls after the trim below).
- **guard** → only `test_narrating_model_reaches_a_verdict_at_shipped_budgets`, now parametrized over both matrix aliases: **two calls**, the ENG-1081 truncation guard that is the reason the two plumbing files are trigger paths at all, and enough for `_check_served_model` to keep ENG-1687's repoint detection armed on every guard run. The job summary says so.

Deliberately **not** a `pytest.skip`: the out-of-money branch requires every junit skip to be a gateway denial (ENG-1334; #483 deep review, finding 1).

**One trim:** `one_attempt_give_up` runs at 6 on `mindshub_air`, not 12 — 0 slips in 48+ runs.

## Why spans, not grep

A first cut used `git diff -W` and grepped the function context for the markers. It fired on every edit inside `turn_stream`, because that method contains the verifier call site and its constants, which would have left most `session.py` PRs on the full matrix. Spans from `ast` distinguish *defining* a symbol from *using* it.

## Validation

- `tests/test_verifier_eval_gate.py` pins: every marker resolves to a span in the real code (a rename reds the unit test instead of silently turning the matrix off); a body edit inside a marker counts and one outside does not, including a pure insertion; `-k` selects exactly one existing test; the step contains no `pytest.skip`; `fetch-depth: 0` is set so the base is diffable.
- Real diffs through the script: #483 (verifier) → **full** (`_VerifierVerdict`, `_VERIFIER_JUDGMENT_RUBRIC`, eval module); #480 and #485 → **guard**.
- Synthetic edits against clean HEAD: renderer body → full; `turn_stream` notice text → guard; rubric constant text → full.
- This PR changes the eval module, so its own `verdict-eval` run is **full** — expected, and the last full run this class of PR pays.

Expected effect: ~2,900 gateway calls a month instead of ~20,000 at the same PR volume; no verifier-touching PR runs fewer calls than the incident cases need. Numbers to be confirmed on the ticket after two weeks.

## Security pass

CI configuration, a stdlib-only script that reads git diffs of the repo's own files, and tests. No secrets touched. `BASE_SHA`/`HEAD_SHA` come from the `pull_request` payload and reach the script as `env:` values passed as quoted arguments, never interpolated into shell. **One value is interpolated:** `${{ steps.scope.outputs.pytest_args }}` in the pytest `run:` line. It is safe because the script writes exactly one of two literals there (empty, or `-k narrating_model_reaches_a_verdict`, from a constant), and it must stay unquoted for word-splitting; anyone widening what the script may write to that output should re-read this line first. `GITHUB_OUTPUT` writes are two fixed keys with literal values; `git` is invoked via argv, no shell.

**Review follow-ups folded in before human review** (two self-review rounds, not posted). Round 2: the guard tier reaches both aliases (it called only air, so a haiku-only repoint would have waited for a verifier-touching PR); the YAML fallback and the script's trigger-path entry are pinned by tests; the recount comment's numbers corrected; the script lives under `.github/scripts/` next to its structural twin. Round 1: the forced-tool-call body `_generate_object_with` and `_call_with_auth_confirmation` are now markers (a `tool_choice` mutation inside them read "guard" before); the decision diffs from the **merge base**, not `pull_request.base.sha`, so a verifier commit landing on staging no longer pushes every un-rebased PR back to the full matrix (5 of 5 real guard-eligible PRs flipped before); marker sets are pinned by equality so a deleted entry reds the unit test; a pure-deletion hunk edge is tested; the eval's stale call-count comment is corrected.


.github/workflows/verifier-eval.yml
```diff
@@ -93,6 +99,50 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          if [ -z "$BASE_SHA" ] || [ -z "$HEAD_SHA" ]; then
+            # Not a pull_request payload: fail closed to the full matrix.
+            echo "scope=full" >> "$GITHUB_OUTPUT"; echo "pytest_args=" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          git fetch --quiet origin "$BASE_SHA" "$HEAD_SHA" || true
+          python3 .github/scripts/verifier_eval_scope.py --base "$BASE_SHA" --head "$HEAD_SHA" --github-output "$GITHUB_OUTPUT"
+          if grep -q "^scope=guard" "$GITHUB_OUTPUT"; then
```
ea-rus: This step reads `$GITHUB_OUTPUT` back with `grep` to decide whether to write the summary. That file is meant for passing values to later steps, not for reading them back. Could the summary be written by a separate step with `if: steps.scope.outputs.scope == 'guard'`, or directly by the script?

.github/workflows/verifier-eval.yml
```diff
@@ -101,7 +151,7 @@ jobs:
           VERIFIER_EVAL_REQUIRE_LIVE: ${{ steps.mode.outputs.require_live }}
         run: >-
           uv run --group dev pytest tests/test_verifier_verdict_live.py -v
-          --junitxml=eval-junit.xml
+          --junitxml=eval-junit.xml ${{ steps.scope.outputs.pytest_args }}
```
ea-rus: Could `pytest_args` go through `env:` instead (`PYTEST_ARGS: ${{ steps.scope.outputs.pytest_args }}`), used as unquoted `$PYTEST_ARGS` here? That removes the only `${{ }}` interpolation into shell, the one the description flags as needing care, so widening what the script writes later can't turn into injection.

.github/scripts/verifier_eval_scope.py
```diff
@@ -0,0 +1,201 @@
...
+SPAN_MARKERS: dict[str, tuple[str, ...]] = {
+    "anton/core/session.py": (
+        "_VerifierVerdict",
+        "_VERIFIER_TOKEN_BUDGETS",
+        "_VERIFIER_NO_PREAMBLE",
+        "_VERIFIER_JUDGMENT_RUBRIC",
+        "_build_verify_request",
+        "_render_verify_transcript",
+        "_render_tool_result_content",
+        "_clip_keep_cause",
```
ea-rus: `_IMAGE_PLACEHOLDER` (session.py:968) is missing. `_render_verify_transcript` inlines it into the transcript the verifier judges, so changing `"[image]"` to `""` reads "guard". I checked everything the session.py markers reach transitively, and this is the only name that affects verdict input. Suggest adding it.

.github/scripts/verifier_eval_scope.py
```diff
@@ -0,0 +1,201 @@
...
+        name = None
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            name = node.name
+        elif isinstance(node, ast.Assign):
+            targets = [t.id for t in node.targets if isinstance(t, ast.Name)]
+            name = targets[0] if len(targets) == 1 else None
+        elif isinstance(node, ast.AnnAssign) and isinstance(node.target, ast.Name):
+            name = node.target.id
+        if name in names and name not in spans:
+            spans[name] = (node.lineno, node.end_lineno or node.lineno)
```
ea-rus: `node.lineno` for a function or class is the `def` or `class` line, not the first decorator. Adding `@functools.lru_cache` above `_render_verify_transcript` or `_build_verify_request` reads "guard" (checked against the real session.py). No marker has a decorator today, so this is latent. Starting the span at `min(d.lineno for d in node.decorator_list)` when decorators are present would close it.

.github/scripts/verifier_eval_scope.py
```diff
@@ -0,0 +1,201 @@
...
+    for node in ast.walk(tree):
+        name = None
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            name = node.name
+        elif isinstance(node, ast.Assign):
+            targets = [t.id for t in node.targets if isinstance(t, ast.Name)]
+            name = targets[0] if len(targets) == 1 else None
+        elif isinstance(node, ast.AnnAssign) and isinstance(node.target, ast.Name):
+            name = node.target.id
+        if name in names and name not in spans:
```
ea-rus: `name not in spans` keeps only the first definition. A later `_VERIFIER_JUDGMENT_RUBRIC = "..."` or `_VERIFIER_JUDGMENT_RUBRIC += "..."` further down the module is the one that takes effect, but it reads "guard" (checked against the real session.py; `+=` is also an `AugAssign`, which isn't matched at all). Every marker is assigned once today. Collecting all spans per name, and handling `AugAssign`, would close it.

tests/test_verifier_verdict_live.py
```diff
@@ -1184,7 +1190,9 @@ def acceptable(self) -> tuple[str, ...]:
     # each. An 11-of-12 threshold flaked on its first CI run, and a gate people
     # learn to re-run is not a gate. If haiku stops slipping, delete
     # `skip_models` and its pin and let it gate.
-    runs=12,
+    # Six, not twelve: 0 slips in 48+ runs on air, so twelve bought nothing
+    # over six except gateway calls (ENG-2863).
+    runs=6,
```
ea-rus: The all-runs-must-pass threshold keeps 6 runs flake-safe, but halving runs halves this control's power to catch a low-rate regression, which is what it exists for. At a 1-in-16 STUCK slip rate (haiku's measured rate), 12 runs catch it 54% of the time and 6 runs catch it 32% of the time. "0 slips in 48 runs" speaks to false positives only. Is the saving (6 calls per full run, ~26 full runs a month) worth it? If so, could the comment state the trade-off?

.github/workflows/verifier-eval.yml
alecantu7: Done in 66e0c6d: the summary is its own step, `if: steps.scope.outputs.scope == 'guard'`; the scope step no longer reads $GITHUB_OUTPUT back. Gate test pins both.

.github/workflows/verifier-eval.yml
alecantu7: Done in 66e0c6d: `PYTEST_ARGS: ${{ steps.scope.outputs.pytest_args }}` under `env:`, expanded unquoted as `$PYTEST_ARGS` (comment says why). Gate test asserts no `${{ }}` remains in the run line.

.github/scripts/verifier_eval_scope.py
alecantu7: Agreed and added in 66e0c6d: `_IMAGE_PLACEHOLDER` is a marker; the equality pin on the marker set includes it. Verified it resolves to session.py:968.

.github/scripts/verifier_eval_scope.py
alecantu7: Reproduced (span started at the def line) and fixed in 66e0c6d: a class/function span starts at `min(node.lineno, *decorator linenos)`. New gate test edits a decorator line and expects "full".

.github/scripts/verifier_eval_scope.py
alecantu7: Reproduced both (only the first Assign kept; AugAssign unmatched) and fixed in 66e0c6d: every assignment of a marker yields a span, `ast.AugAssign` is matched, and `touched` checks all spans. New gate test covers a re-assignment and a `+=` on separate lines.

tests/test_verifier_verdict_live.py
alecantu7: You're right that 0/48 bounds false positives only — I had the wrong quantity in mind. Reverted to 12 in 66e0c6d with your detection numbers in the comment (54% vs 32% at 1-in-16). The saving was 6 calls × ~26 full runs a month; not worth it.

ea-rus (approved):

PR #487, opened 2026-09-19
[ENG-2807] Parse the datasource block and provide the typed cloud helper

tino097:
https://linear.app/mindsdb/issue/ENG-2807/integrate-datasource-grants-and-typed-helper-across-cloud-turns

## What changed

- Parse the block and provide the typed Anton datasource helper: the cloud turn contract accepts the optional `datasource` block and rejects a protocol version other than 1, the session hands the scratchpad a `query_minds_data(connection_id, sql, parameters=None)` helper bound to the turn key and the configured https gateway origin, and the legacy static-key injection stays desktop only
- Let the helper report what the gateway said: on an HTTP error the agent sees the gateway's own error code, otherwise `datasource_unavailable`; a result is accepted only when it echoes the requested protocol version, operation, connection and credential version, and the truncation reason passes through
- Address the review findings on the helper: the POST goes over one verified TLS connection that never follows a redirect, so the turn key cannot be replayed to wherever a response points; only the gateway's closed error vocabulary is relayed; column descriptors are bounded like values; the cloud turn's bearer stays out of artifact backends; the legacy static-key path is proven absent from a pod carrying its full configuration

## Notes for the reviewer

- **The strict top-level protocol version check is deliberate.** A later version could carry fields a version 1 parser would silently ignore, so an unknown version fails the turn instead of being parsed as version 1.
- **The turn key rides in the workspace environment overlay,** which artifact backend child processes would also inherit. Nothing leaks today because `launch_backend` is outside the cloud tool allowlist; a change to that allowlist must revisit this.
- **The helper sends `query` only.** `describe` waits for schema discovery in the adapters ticket. Results are bounded (rows, columns, cell size) and returned as data; the prompt tells the agent to treat them as untrusted.
- **https only.** The helper and the session guard refuse a gateway origin that is not a bare https origin, so a plain http local gateway cannot be exercised through the subprocess path; PR environments and staging are https.
- **A datasource block with no gateway origin on the pod fails the turn,** the same loud failure as a missing key prefix on the producer: both are misconfiguration with the flag on, and the controller now replaces pods that predate the origin.
- **The cloud helper does not fill the explainability query trail** the legacy helper wrote; the record's shape (a datasource name versus a connection id) is for the UI ticket to settle.
- **Contracts consumed:** inference #557 (request and response shape of `/v1/datasources/execute`, error body `{code, detail, request_id}`), auth #539 indirectly through the turn key.

## Checks

| Check | Observed result |
|-------|-----------------|
| `uv run pytest tests/test_cloud_turn_contract.py tests/test_cloud_turn_session.py tests/test_cloud_turn_entrypoint.py tests/test_cloud_turn_datasource.py tests/test_scratchpad_child_env.py tests/test_side_effect_result.py tests/test_scratchpad_utf8.py -q` | 171 passed |
| `uv run pytest` | 3270 passed, 51 skipped |
| `git diff --check` | clean |

## Ships with

- mindsdb/scratchpad-controller#54, either order; both before mindsdb/cowork-server#557.


tino097:
@codex review

PR #490, opened 2026-09-21
ENG-2892: treat a gateway alias echo as "served model not disclosed", not a repoint

alecantu7:
Linear: [ENG-2892](https://linear.app/mindsdb/issue/ENG-2892/the-gateway-now-returns-the-alias-as-the-response-model-so-antons)

## Why

The mindshub_inference release to `main` this morning (#535, carrying [ENG-2751](https://linear.app/mindsdb/issue/ENG-2751) / #553 — a decided product change) changed the client-visible `model` field of a chat completion from the resolved id to a configured response model — for the eval's aliases, the alias itself (`haiku` → `haiku`, `mindshub_air` → `mindshub_air`; on the 18th these were `claude-haiku-4-5-20251001` and `gpt-5.6-luna`). Langfuse still records the resolved ids server-side. There is no response header carrying them.

anton's verdict eval pins each matrix slot to the model it was measured on (ENG-1687) and raised `AliasRepointed` on the mismatch — on **every** call. Every verifier-touching PR is red until this lands: last green `verdict-eval` 07:47 UTC today (before the release), first red 15:36 on #486, 24 failed in 59 s, all `AliasRepointed`. Not caused by any anton change.

## What

`_check_served_model` treats `served == alias` as **no information**: it records `"<alias> (alias echoed — served model not disclosed)"` so the served-models report shows the identity half of the matrix is blind for that slot, and it does not raise. A **genuine** repoint — a different real id — still fails, and the pinned id is still accepted verbatim. One gate test pins all three behaviours; no network needed.

Nothing else changes: same cases, same runs, same acceptable sets.

## What this does not restore, and why that is by design

The ENG-1687 identity check is blind while the gateway echoes the alias — and the echo is **not a gateway gap waiting on a header**. It is [ENG-2751](https://linear.app/mindsdb/issue/ENG-2751), decided 2026-09-15 and implemented by mindshub_inference#553: the served model is hidden on purpose, across the whole catalog and on failover routes, and the public docs now say the response `model` "does not name the model that served the request". A customer-visible header would undo that decision, so it is not the route. Consequences: the eval cannot name the model behind a matrix slot — the printed served-models report says `<alias> (alias echoed — served model not disclosed)` for each blind slot — and it also cannot tell when a fallback backend served one of its calls, which the code comments state (the report line does not carry that second caveat). Re-arming ENG-1687 needs a channel customers cannot see — Langfuse still records the resolved id server-side — and that is ENG-2892's remaining item.

**Self-review fold-in (c0b5b140 → head):** the confirmed-pin comparison now runs *before* the echo test, so a one-off run whose alias is itself a real id (`VERIFIER_EVAL_*_MODEL=gpt-5.6-luna`) is recorded verbatim rather than misreported as blind; pinned by a gate test.

## Merge order

**This first**, then [#486](https://github.com/mindsdb/anton/pull/486) (ENG-2863), whose `verdict-eval` is red only because of this; re-run its checks after this lands. Not stacked — both branch from staging.

## Validation

- `tests/test_verifier_eval_gate.py`: new test — echo does not raise and is recorded as not disclosed; a different real id still raises; the pinned id is accepted verbatim. All existing repoint tests unchanged and green.
- This PR's own `verdict-eval` runs the full matrix against today's gateway; it is the live proof.

## Security pass

Test-only change to how one string from a gateway response is classified. The value still goes through `sanitize_model_name` before it reaches any report line.


alecantu7:
**Self-review complete, at `11b2c3d7`. Nothing blocking; ready for human review.**

**Found and fixed** (in `11b2c3d7`):
- **Pin-match order.** The alias-echo test ran before the pin comparison. So when an alias was itself a real id (a one-off `VERIFIER_EVAL_*_MODEL=gpt-5.6-luna` run), a confirmed match was recorded as "not disclosed". Pass/fail was unaffected, but the report was wrong. A confirmed match now wins, and a gate test pins it.
- **Wrong premise in the comments.** They treated the alias echo as a gateway gap to fix with a header. It is [ENG-2751](https://linear.app/mindsdb/issue/ENG-2751), a decided product change (mindshub_inference#553): the served model is hidden on purpose, across the whole catalog and on failover. The code comments and the PR body now say so, and rule out a customer-visible header as the way to re-arm ENG-1687.

**Verified:**
- **Mutations.** Breaking any of the three behaviours fails 2 tests each: pin match wins, echo recorded as blind, genuine repoint still raises.
- **Merged tree.** Staging + this PR passes the gate tests (33). With #486 also merged it's clean at 39.
- **CI.** `verdict-eval` is green on this head, running the full matrix against today's gateway. `run-tests` is still pending.

**Known and left open:**
- ENG-2892's comment and done-when item 4 still describe a header request. They will be updated to the Langfuse server-side route.
- The printed served-models report says "served model not disclosed" but doesn't mention failover. That caveat is in the code comments only.

**Merge order:** this first, then #486, whose `verdict-eval` is red only because of what this fixes.

ea-rus (approved):

